Issue #24, opened 2015-09-18
some more plot types?


jverzani:
Here are some other plot types that might be useful to add to this package. They come from SymPy's plotting module and are supported in PyPlot and others, though not fully:

- [x] 2d contour plot (`contour`?)
- [x] 3d parametric line plot (just a third argument to `plot`?)
- [ ] 3d parametric surface plot (`surfaceplot`?)
- [x] 2d vector (or quiver) plot (`vectorplot`? `quiver`?)

~~plot_implicit for making plots of implicit functions~~

tbreloff:
Thanks for the list. I'll probably take care of a few other things before I get to add these.

tbreloff:
Contours are available: #36 

What do you mean by `plot_implicit`?  Do you have an example viz you can point me to?

jverzani:
The name comes from SymPy (http://docs.sympy.org/dev/modules/plotting.html#sympy.plotting.plot_implicit.plot_implicit). There is also this package for Julia (https://github.com/jverzani/ImplicitEquations.jl). Really this isn't the most important graphic, but it would be nice to be able to push some of the graphing code in ImplicitEquations onto Plots.jl.

The more widely used plot would be an equivalent of PyPlot's `quiver` function.

tbreloff:
I understand now. My gut feeling is that it should live in your ImplicitEquations package. You could consider creating a type which is a subtype of PlotRecipe and which wraps an equation. Then you only need to implement methods to generate the underlying data and Plots will handle the rest. I might not get to this soon, but I'm happy to answer questions.

jverzani:
Okay, thanks for the pointers on how to wrap it.

On Mon, Oct 26, 2015 at 9:37 PM, Tom Breloff <notifications@github.com>
wrote:

> I understand now. My gut feeling is that it should live in your
> ImplicitEquations package. You could consider creating a type which is a
> subtype of PlotRecipe and which wraps an equation. Then you only need to
> implement methods to generate the underlying data and Plots will handle the
> rest. I might not get to this soon, but I'm happy to answer questions.
>
> > On Oct 26, 2015, at 8:38 PM, john verzani <notifications@github.com>
> wrote:
> >
> > The name comes from SymPy (
> http://docs.sympy.org/dev/modules/plotting.html#sympy.plotting.plot_implicit.plot_implicit).
> There is also this package for Julia (
> https://github.com/jverzani/ImplicitEquations.jl). Really this isn't the
[...]

PR #65, opened 2015-11-12
Add link to Immerse.jl example

FavioVazquez:
This adds the Immerse.jl examples from https://github.com/tbreloff/ExamplePlots.jl to the Readme file.

### DIFF
--- a/README.md
+++ b/README.md
@@ -24,7 +24,8 @@ Please add wishlist items, bugs, or any other comments/questions to the issues l
 
 ## Examples for each implemented backend:
 
-- [Gadfly.jl/Immerse.jl](https://github.com/tbreloff/ExamplePlots.jl/tree/master/docs/gadfly_examples.md)
+- [Gadfly.jl](https://github.com/tbreloff/ExamplePlots.jl/tree/master/docs/gadfly_examples.md)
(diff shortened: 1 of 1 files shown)


tbreloff:
Hi thanks for the PR, but I purposely changed it to the way it is now.  Immerse uses Gadfly to do its plotting, so (in theory) those examples should be exactly the same for both... it's just more work for me to maintain both sets of examples.

FavioVazquez:
I see. Well thanks for letting me know that. Great work btw :+1:. I was searching in the documentation and examples and I couldn't find the StreamPlot from Matplotlib http://matplotlib.org/api/pyplot_api.html#matplotlib.pyplot.streamplot, is every kind of plot defined in PyPlot available in Plots.jl? I know this is out of the topic of the PR but just didn't think it's worth to add an Issue with this question. Thanks.

tbreloff:
Quiver/streamplot are on my list, but I need to think through the best way to take in the vector data (U/V). I don't love that PyPlot requires an equally spaced grid. Until I implement it, you can always access the underlying PyPlot figure/axes with the "plt.o" field. Feel free to add specific requirements to #24.

PR #69, opened 2015-11-15
createPlotWindow -> Plot

jverzani:
In UnicodePlots, the function `createPlotWindow` no longer exists. I think just calling `Plot` is sufficient here along with a canvas type, but am certainly no expert in the matter. This pull request does that change. It seems to work on the basic plots I tested. `Pkg.test` had issues, but the error of `QuartzImageIO` not being installed seems unrelated to this change.


tbreloff:
Thanks for the PR.  I had already fixed this in the dev branch, and didn't get around to pushing to master.  I pushed just now... let me know if there are any issues.

jverzani:
All good. Thanks.

PR #85, opened 2015-11-25
don't test for 0.4-

stevengj:
Testing version > 0.4- is wrong because it incorrectly implies that all pre-releases of 0.4 work.  Either test for the specific 0.4 build you need, or test for `VERSION > v"0.4"` (i.e., the release version).

However, since you REQUIRE julia 0.4 anyway, you can omit this test entirely.


tbreloff:
Thanks... your logic makes sense.

PR #86, opened 2015-11-25
WIP: GLVisualize

dlfivefifty:
This is aiming to support GLVisualize.  Right now simple plots like
```julia
glvisualize()
N=20
h = 1./N
x=y=h:h:1
f=(x,y)->exp(-x^2-y^2)
z       = Float64[f(x[k],y[j]) for k=1:length(x),j=1:length(y)]
plot(x,y,z)
```
work.  I'd like to add support for matrix-valued x and y for general grids (GLVisualize supports this, but Plots doesn't).

### DIFF
--- a/src/backends/glvisualize.jl
+++ b/src/backends/glvisualize.jl
@@ -1,0 +1,90 @@
+
+
+# [WEBSITE]
+
+# ---------------------------------------------------------------------------
+
+immutable GLScreenWrapper
+    window
+    render
+end
+
+function _create_plot(pkg::GLVisualizePackage; kw...)
+  d = Dict(kw)
+  # TODO: create the window/canvas/context that is the plot within the backend (call it `o`)
+  # TODO: initialize the plot... title, xlabel, bgcolor, etc
+  w,r=GLVisualize.glscreen()
+  @async r()
+  Plot(GLScreenWrapper(w,r), pkg, 0, d, Dict[])
+end
+
+
+function _add_series(::GLVisualizePackage, plt::Plot; kw...)
+  d = Dict(kw)
+  # TODO: add one series to the underlying package
+  push!(plt.seriesargs, d)
+  x,y,z=map(Float32,d[:x]), map(Float32,d[:y]), map(Float32,d[:surface].surf)
+  GLVisualize.view(GLVisualize.visualize(x*ones(y)', ones(x)*y', z, :surface))
+  plt
+end
+
+function _add_annotations{X,Y,V}(plt::Plot{GLVisualizePackage}, anns::AVec{@compat(Tuple{X,Y,V})})
+  for ann in anns
+    # TODO: add the annotation to the plot
+  end
+end
+
+# ----------------------------------------------------------------
+
+function _before_update_plot(plt::Plot{GLVisualizePackage})
+end
+
+# TODO: override this to update plot items (title, xlabel, etc) after creation
+function _update_plot(plt::Plot{GLVisualizePackage}, d::Dict)
+end
+
+function _update_plot_pos_size(plt::PlottingObject{GLVisualizePackage}, d::Dict)
+end
+
+# ----------------------------------------------------------------
+
+# accessors for x/y data
+
+function Base.getindex(plt::Plot{GLVisualizePackage}, i::Int)
+  series = plt.o.lines[i]
+  series.x, series.y
+end
+
+function Base.setindex!(plt::Plot{GLVisualizePackage}, xy::Tuple, i::Integer)
+  series = plt.o.lines[i]
+  series.x, series.y = xy
+  plt
+end
+
+# ----------------------------------------------------------------
+
+function _create_subplot(subplt::Subplot{GLVisualizePackage})
+  # TODO: build the underlying Subplot object.  this is where you might layout the panes within a GUI window, for example
+end
+
+function _expand_limits(lims, plt::Plot{GLVisualizePackage}, isx::Bool)
+  # TODO: call expand limits for each plot data
+end
+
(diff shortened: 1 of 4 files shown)


dlfivefifty:
@SimonDanisch

SimonDanisch:
Cool, thanks for doing this! I'll take a look at it tomorrow.
Note, that GLVisualize can be a backend for Gadfly (by being a backend for Compose) soon. But I think it'll still be valuable to have GLVisualize as a different backend here, since it supports other plots than Gadfly and Compose is not the best fit for GLVisualize.

tbreloff:
This is exciting... Thanks for working on it! It may be a couple days before I can review completely.

tbreloff:
One quick note. I need to think through how to best support "matrix-valued x and y for general grids".  Since I support so many different input combinations, it can sometimes be tricky to ensure there isn't ambiguity. A m x n matrix might mean a grid/surface, or it might mean n series, each of length m. Context can help, but we need to be sure not to break what already works.

dlfivefifty:
Yeah, I saw that in the code so didn’t try to implement it.   But I think when there are three arguments, x, y and z, the ambiguity is gone.  So it might make sense to create “createKWargsList(plt::PlottingObject, x, y, z;kw…)” that replicates the functionality of “createKWargsList(plt::PlottingObject, x, y;kw…)”, but assumes 3D data.

tbreloff:
You might be right, and worst case we can use the plot type to guide it, so that 3d plots get the logic you want. Thanks again for the WIP

SimonDanisch:
Okay this works for me!
The best I probalby can do to help this right now is to merge [next²](https://github.com/JuliaGL/GLVisualize.jl/tree/sd/next%C2%B2), which will make e.g. annotations a lot easier.

src/plotter.jl
```diff
@@ -38,6 +41,7 @@ backend_name(::UnicodePlotsPackage) = :unicodeplots
 backend_name(::QwtPackage)          = :qwt
 backend_name(::BokehPackage)        = :bokeh
 backend_name(::PlotlyPackage)       = :plotly
+backend_name(::GLVisualizePackage)  = :plotly
```
mmagnuski: shouldn't this be ` = :glvisualize` ?

dlfivefifty: Nice catch, thanks!  It’s fixed now.

src/backends/glvisualize.jl
```diff
@@ -0,0 +1,90 @@
...
+  Plot(GLScreenWrapper(w,r), pkg, 0, d, Dict[])
+end
+
+
+function _add_series(::GLVisualizePackage, plt::Plot; kw...)
+  d = Dict(kw)
+  # TODO: add one series to the underlying package
+  push!(plt.seriesargs, d)
+  x,y,z=map(Float32,d[:x]), map(Float32,d[:y]), map(Float32,d[:surface].surf)
+  GLVisualize.view(GLVisualize.visualize(x*ones(y)', ones(x)*y', z, :surface))
```
tbreloff: I'm assuming that this both updates **and** displays what is on screen?  If so, you probably want to move the call to `GLVisualize.view` to the `display` method below.

`_create_plot` is for initializing the internals, `_add_series` is for updating the internals for a new data series, and the `writemime`/`display` methods below are for taking the internal representation and saving, displaying, or whatever else.  Ideally it's possible to be able to do this:

```
using Plots; glvisualize()
plot(rand(10));

# no windows have been created or displayed

... do other stuff ...

# now display it
gui()
```

I really should learn the GLVisualize internals...

src/backends/glvisualize.jl
```diff
@@ -0,0 +1,90 @@
...
+function _update_plot_pos_size(plt::PlottingObject{GLVisualizePackage}, d::Dict)
+end
+
+# ----------------------------------------------------------------
+
+# accessors for x/y data
+
+function Base.getindex(plt::Plot{GLVisualizePackage}, i::Int)
+  series = plt.o.lines[i]
+  series.x, series.y
```
tbreloff: these methods need to be implemented... they should really have "TODO" comments like the others

src/backends/glvisualize.jl
SimonDanisch: Visualize creates a viewable object. Since it already uploads some data to the GPU, sadly a window (OpenGL context) needs to be present already. This is actually addressed in next², where visualize won't convert anything to OpenGL objects (if you don't pass any yourself), so it will be easier to differentiate between setup and display.

>_add_series is for updating the internals for a new data serie

What does this actually mean? You already plotted e.g. 10 points, and want to update the positions of the points, but leave labels and axis alone?

tbreloff: The distinction between changing internals and displaying may be best seen with something like plotly, or gadfly. In gadfly, you create the gadfly objects/structure in the "internals" methods, and call draw, etc in the display methods. 

For this case, if there's no underlying internals, then you could consider moving most/all of the logic to the display methods, and just loop through plotargs and seriesargs.

tbreloff:
I'm sorry that this fell on the back-burner.  I'm wondering if I should try to merge as it is and we can iterate afterwards.  It would be ideal if you could rebase onto the `dev` branch, and also open the PR against the `dev` branch.  dev and master are usually lined up, but not always, and there's been a bunch of commits since this PR.  Let me know if you can do this, otherwise I can merge and clean up afterwards.

dlfivefifty:
I think this is a good idea.  I don’t really know how to rebase properly, but I suppose its a good time to learn.  Should I rebase to dev and just create a new pull request then?

tbreloff:
That would be great... thanks.

dlfivefifty:
OK There’s a new PR

PR #92, opened 2015-12-18
Add initial support for GLVisualize

dlfivefifty:
This adds basic support for GLVisualize, and supercedes #86


tbreloff:
Thanks!

PR #99, opened 2015-12-29
Add support for levels keyword in PyPlot

dlfivefifty:
The following now works for specifying the levels explicitly:

```julia
using Plots
x=y=0.:.1:1.
z=Float64[exp(x)*cos(y) for x=x,y=y]
plot(x,y,z;t=:contour,levels=[.6,1.5,2.6])
```

This is particularly useful for movies, where you don't want the meaning of the levels to change.

If you are happy with these changes, I can also add bins (i.e., bins/nbins mirrors levels/nlevels), and add support for Gadfly.

### DIFF
--- a/src/args.jl
+++ b/src/args.jl
@@ -2,8 +2,8 @@
 
 const _allAxes = [:auto, :left, :right]
 @compat const _axesAliases = Dict(
-    :a => :auto, 
-    :l => :left, 
+    :a => :auto,
+    :l => :left,
     :r => :right
   )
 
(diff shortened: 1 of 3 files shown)


tbreloff:
Thanks for this. I think I may prefer to have one keyword, levels. Then nlevels can be an alias to it. I can make the change later though. 

Do you know why there are all those end-of-line changes in the commit?

KristofferC:
Some editors automatically strip trailing whitespace on save.

dlfivefifty:
Sorry about that!

KristofferC:
It is a good thing though? You can always add `?w=1` to the URL to remove the whitespace from the diff.

PR #102, opened 2016-01-01
Added basic support for GR.jl

jheinen:


src/backends/gr.jl
```diff
@@ -0,0 +1,232 @@
+
+# https://github.com/jheinen/GR.jl
+
+fig = Dict()
+
+fig[:size] = [500, 500]
+
+@compat const gr_linetype = Dict(
```
Evizero: since min version is 0.4, is compat still serving a purpose here?

jheinen: You are right. I removed the compat macros ...

tbreloff:
Thanks for the effort. I need to review more completely, but a couple things that jump out:

- is the global fig really necessary? Can you just store the reference in the Plot object? (Then say "fig = plt.o" instead of "global fig")
- can you move the display logic to the bottom methods? Most of the methods are for object creation/setup, but you shouldn't open a window or draw to the screen until "display" has been called
- all PRs should be merged into the dev branch. Also this will be much cleaner if you rebase on top of dev instead of merging master into yours

Let me know if you have any questions about any of this. Thanks

PR #107, opened 2016-01-03
Initial support for GR framework (GR.jl)

jheinen:


tbreloff:
Thanks for this!  Due to the embedded merge, and another PR to master, I'll have to do some branch cleanup locally and push new master and dev branches in a bit.  It would be great if you add some TODO items to #108 as well so we all understand the roadmap.

Issue #108, opened 2016-01-03
GR backend


tbreloff:
Current TODO list (assume `y = rand(10)` for the examples):

- [ ] Support the `grid` keyword.  This should display a plot without the grid lines: `plot(y, grid=false)`
- [ ] Support handling of NaN (gap in line segment): `plot(vcat(1:10, NaN, 12:21), vcat(y,NaN,y))`
- [ ] Support removal/setting of xticks/yticks: `plot(y, xticks=nothing, yticks=nothing)`
- [ ] Support xlim/ylim

tbreloff:
I did some very basic initial tests.  Here's some things that I consider bugs (but may be unavoidable, as I don't know GR very well):

- only one plot window can be open at a time
- if you close the window, there is a fatal error and GR is unusable for the rest of the session

In addition, there are tons of errors on the examples... I recommend doing the following:

```
Pkg.clone("https://github.com/tbreloff/ExamplePlots.jl.git")
using ExamplePlots
test_examples(:gr)
```

Note that test_examples also accepts an optional second argument which is the number of a specific example.

tbreloff:
ps - this is the error when you close the window:

```
julia> XIO:  fatal IO error 11 (Resource temporarily unavailable) on X server ":0"
      after 5740 requests (5585 known processed) with 0 events remaining.
```

jheinen:
Thanks for merging the PR and doing initial tests. In terms of those results I would like to mention that:

* GR can handle multiple (and different) output devices (simultaneously). I did not implement this feature in the initial release.
* Closing the window (on Linux) will not re-open the X11 display. For applications that produce "figures" (like Plots), this behavior is inconvenient. I will provide a solution in the next GR release.
* I will try to add the missing features (plot types) ASAP

BTW: It seems to me that some of the "bugs" are simply ignored by other backends, e.g.

    test_examples(:gr,4)

After analyzing the debug output It turns out, that the markersize is given as an array:

    Updating plot items:
    ...
        markersize: 100-element Array{Float64,1}
    ...

Does this make sense? PyPlot seems to ignore this.

tbreloff:
Thanks for the thorough respone @jheinen.  I think this contribution is great, and the time-to-plot seems very quick based on my very brief tests.  

> BTW: It seems to me that some of the "bugs" are simply ignored by other backends

[...]

PR #117, opened 2016-01-28
Improved GR framework support

jheinen:
@tbreloff : Still some TODOs, but probably good enough to merge ...

TODOs:
- linetype: pie
- marker shapes
- multiple figure output
- animation stuff (will be difficult, because GR has builtin functionality for creating videos)

http://pgi-jcns.fz-juelich.de/pub/doc/ExamplePlots.html

### DIFF
--- a/src/backends/gr.jl
+++ b/src/backends/gr.jl
@@ -62,6 +62,8 @@ function gr_display(plt::Plot{GRPackage}, clear=true, update=true,
 
   extrema = zeros(2, 4)
   num_axes = 1
+  cmap = false
+  axes_2d = true
 
   for axis = 1:2
     xmin = ymin = typemax(Float64)
@@ -84,8 +86,15 @@ function gr_display(plt::Plot{GRPackage}, clear=true, update=true,
           else
             xbins = ybins = p[:nbins]
           end
+          cmap = true
           x, y, H = Base.hist2d(E, xbins, ybins)
         else
+          if p[:linetype] in [:contour]
+            cmap = true
+          end
+          if p[:linetype] in [:surface, :wireframe]
+            axes_2d = false
+          end
           x, y = p[:x], p[:y]
         end
         xmin = min(minimum(x), xmin)
@@ -104,6 +113,13 @@ function gr_display(plt::Plot{GRPackage}, clear=true, update=true,
 
   if num_axes == 2
     viewport[2] -= 0.05
+  end
+  if cmap
+    viewport[2] -= 0.1
+  end
+  if !axes_2d
+    viewport[1] += 0.1
+    viewport[2] -= 0.1
   end
   GR.setviewport(viewport[1], viewport[2], viewport[3], viewport[4])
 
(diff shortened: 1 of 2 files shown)


tbreloff:
These plots look really great! Thanks for the effort here... I'll merge when I have a chance.

Evizero:
I agree! Looking forward to messing around with it. The lineplots look even nicer than ggplot2. 

Have you tried using `@manipulate`?

jheinen:
Glad to hear. But our main focus is on performance and interop ... :smile:

KristofferC:
Very nicely looking plots indeed.

tbreloff:
I merged and did some very brief local testing.  I think the biggest todo is to stabilize the plotting workflow as best as possible.  I immediately ran into some errors which in hindsight I already knew about, but it would be good to keep them on the radar if they're fixable:

- if you manually close the plot window, it errors and you can't do any more plotting until you quit/reload.  I close the window habitually, and this would probably mean I can't use it for real work until this is handled cleanly.  I don't want to have to restart my session if I accidentally close a window.
- plotting 0 or 1 points results in errors.  Seems silly, but there's lots of times I initialize a mostly empty plot window expecting to add to it later.
- some errors cause subsequent commands to complain with errors like "GKS: GKS not in proper state. GKS must be either in the state WSAC or SGOP in routine POLYLINE".  It would be best to not allow it to get to this state, and a bonus to call some sort of "resetState" command when creating a new plot, so that past failed plotting commands don't force a session restart
- will it be possible to have more than 1 plot window open at a time?  ideally `plot(...)` will open a new window

All told, I'm very happy that you decided to spend time on this, and if some of these issues are resolved I could see GR becoming a top choice for backend.  Thanks!

jheinen:
Problems 2+3 can be fixed with a simple sanity check in ``gr.jl``.

Problems 1+4 describe the same problem - although GR can handle multiple graphics drivers (of different type) simultaneously, I'm still looking for a solution that fits best into GR's _logical workstation_ concept, which is different from switching to a graphics backend.

I'll fix these problems with the next PR. I'm  just waiting for an inspiring idea ...

jheinen:
> Have you tried using @manipulate?

@Evizero : now I did. 
manipulate

Thanks for the hint.

Evizero:
I just tried it myself. Very nice!

I also tried to switch to the `gr()` backend for some of my own notebooks that I am working on and stumbled on some bug (I assume)

For example `copy` doesn't seem to work as expected

```Julia
using Plots
gr()
plt = scatter(randn(10), randn(10))
copy(plt)
```

@tbreloff @jheinen Since this PR is merged, how do you want to handle `GR` related issue reports? Do you want me to open new issues for each thing or should we create an umbrella issue for now?

tbreloff:
So I'll be honest that I haven't reviewed the GR code in detail yet, which makes it hard for me to meaningfully solve individual issues.  I think this could be up to @jheinen.  One solution is to open separate issues (preferably with a "GR backend: " prefix to the issue's title), then I or @jheinen can close and consolidate to a single umbrella issue (probably #108) when needed.

I do know that there's probably a little overhaul that will be necessary to make sure all workflows are supported (multiple windows, copying plots, animations, embedded gui windows, etc)... but that this can be a gradual process.  I think items that are related in terms of the solution (for example, multiple issues which require independent plot windows) should stay in one issue, and independent issues could live in their own issues.

jheinen:
@tbreloff : Need some help with the show / display logic. ``gr.jl`` has two ``writemime`` methods and both (PNG + SVG) are called when a plot is finalized (results in two files ``gks.png`` and ``gks.svg``). Can I set the ``defaultOutputFormat`` to something other than PNG, e.g. SVG? The PNG ``writemime`` method should only be invoked when the user calls ``savefig`` with a ``.png`` file extension.

Thanks in advance ...

Issue #125, opened 2016-02-02
abline!


diegozea:
Hi! 
It would be great to have an `abline` function, similar to [R](https://stat.ethz.ch/R-manual/R-devel/library/graphics/html/abline.html). 
Best,

KristofferC:
Isn't this implemented with `hline`

tbreloff:
Right now you can do `hline`, `vline`, or add a regression term with `smooth`.  Here's some examples:

![tmp](https://cloud.githubusercontent.com/assets/933338/12753519/7a107596-c995-11e5-9ae9-fdd1326934f4.png)

`abline` could be done manually of course, but I can see the use for it, so I'll leave this open.

pkofod:
Just to be clear, how would you create the line `y=0.5+0.1*x` using the mentioned functions? I mean, as  `abline(a=0.5, b=0.1)` does.

tbreloff:
Right now you would have to add the line manually:

![tmp](https://cloud.githubusercontent.com/assets/933338/12754650/02dc803c-c99a-11e5-8058-af5ee9ef357e.png)

diegozea:
Also you can do
```plot!(0:100, x -> 0.5 + 0.1 * x, linetype=:l)```
but it's annoying when you don't know the limits for x.

**Comment by tbreloff** [comment]
Or better yet: `plot!([0,100], x->0.5+0.1x)` which doesn't plot the [1,2,3,...,99] points.

**Comment by pkofod** [comment]
Okay, so I just misunderstod the "could be done manually of course". So I guess something like

```julia
function abline!(a, b, ...)
    fetch x limits
    plot!([xmin, xmax], x-> b+a*x)
end
``` 
I can do a PR if this would be useful to have.

tbreloff:
Sure if you want to give it a go... this probably belongs in recipes.jl

tbreloff:
Please merge the PR into the dev branch, not master.  It makes my flow a
little easier.  Thanks!

joshday:
Could there be a method for plotting a function which takes the x limits from the current plot?   It's a common use case, and a little more flexible than abline.  `plot!(x -> .5 + .1x)`

pkofod:
@joshday how should we then choose how to sample the domain?

joshday:
I'm not sure I understand the question.  Maybe my comment needs clarification.  This already works:
```julia
scatter(randn(100))
plot!(x -> .5 + .1x, 0, 100)
```

[...]

PR #126, opened 2016-02-02
Add abline! which creates a straight line according to the formula b+…

pkofod:
…a*x.

Implements #125 . The syntax matches with the name such that `abline!(a, b)` adds a straight line `b+a*x` based on only x-min and x-max.

Does it match with your usual style of coding @tbreloff or are there any changes you would like?

### DIFF
--- a/src/Plots.jl
+++ b/src/Plots.jl
@@ -59,6 +59,7 @@ export
   path3d!,
   scatter3d,
   scatter3d!,
+  abline!,
 
   title!,
   xlabel!,
--- a/src/recipes.jl
+++ b/src/recipes.jl
@@ -132,7 +132,7 @@ end
 
 "Adds a+bx... straight line over the current plot"
 function abline!(plt::Plot, a, b; kw...)
-    plot!(plt, [xmin(plt), xmax(plt)], x -> b + a*x; kw...)
+    plot!(plt, [extrema(plt)...], x -> b + a*x; kw...)
 end
 
 abline!(args...; kw...) = abline!(current(), args...; kw...)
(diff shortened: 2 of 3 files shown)


tbreloff:
Thanks... There's a couple things...

- the method should take a "plt::Plot" as the first argument and pass that as the first argument in your inner call to 'plot!'.  Then you can add the convenience "abline!(args...;  kw...) = abline!(current(), args...; kw...)"
- min/max should look at all series, not just the first

Thanks for the PR

pkofod:
Great suggestions. I just amended the changes, but I think you should be able to see the changes.

pkofod:
Do you want `xmin` and `xmax` be exposed to the users?

pkofod:
The failing CI seems to be comming from Gadfly markers, so it must be from the state of dev, not my PR. Added `Base.extrema(plt::Plot)` and changed code accordingly.

tbreloff:
Yes I need to track down the markers behavior... Don't worry about that. 

This looks good... Can you post an example png here for reference? I'll merge today.

pkofod:
I will do that. Do you want xmin, xmax and extrema exported?

pkofod:
Code
```julia
using Plots

# The first series
x1 = [1; 4; 10; 30]
y1 = [0; 10; 3; 5]
# The second series
x2 = [2; 8; 15; 35]
y2 = [3; 5; 4; 8]

plot([x1 x2], [y1 y2])

abline!(0.2, 2)

#or
# plt = plot([x1 x2], [y1 y2])
# abline!(plt, 0.2, 2)

png("abline")
```
abline

tbreloff:
No... there's no need to export extrema, as it's defined in base, and I don't think people should need xmin/xmax directly (and it can always be qualified). I just have a feeling I might want to change those names someday.

pkofod:
Okay! I guess you could always deprecate them, but on the other hand I don't see a lot of requests for them either in other issues :) They'll remain unexported. Then I think we should be good to merge. Do you want me to squash the commits?

tbreloff:
Merged... thanks for the contribution.  I'm happy to get more people involved to push this forward.

PR #113, opened 2016-01-12
added writemime methods for plotly_blink

spencerlyon2:
This should not be merged until after JuliaIO/ImageMagick.jl#22

PR #127, opened 2016-02-03
Sl/plotly blink

spencerlyon2:
ref #113


tbreloff:
Thanks!  I'm working from my Mac today, so hopefully I can get it working!

PR #130, opened 2016-02-03
Add plot!(f::Function) method

joshday:
I think it was this easy.  Tested with gadfly(), pyplot(), plotly().  Thanks for the pointer.

### DIFF
--- a/src/plot.jl
+++ b/src/plot.jl
@@ -413,7 +413,7 @@ function createKWargsList(plt::PlottingObject, x::AVec, y::AVec, surf::Surface; 
 end
 
 function createKWargsList(plt::PlottingObject, f::FuncOrFuncs; kw...)
-  error("Can't pass a Function or Vector{Function} for y without also passing x")
+  createKWargsList(plt, f, xmin(plt), xmax(plt); kw...)
 end
 
 # list of functions


pkofod:
Does it work with a vector of functions? At least the way it is currently, I would expect it to accept a vector of function, as the method is defined for `FuncOrFuncs`.

tbreloff:
Yes it should be this easy :+1:

pkofod:
Great stuff. After you mentioned it in the issue, I can really see great uses for it @joshday

tbreloff:
Here's a fun little demo of the stuff that was just added.  How cool to be able to write `plot!(sin)`!

tmp

KristofferC:
Nice

pkofod:
I know all of the issues are plastered with praise for @tbreloff , but once again: I think you've really started _the_ package to point to for people who want to plot in Julia. Of course a lot of praise should go to the pyplot, gadfly, plotly, etc developers, but Plots.jl really ties it all together really well!

tbreloff:
Thanks @pkofod ... I never get sick of compliments. :+1: 

At some point it would be cool to get the seal of approval from the core Julia developers and get official endorsement (maybe [here](http://julialang.org/downloads/plotting.html)?)... not sure the politics surrounding something like that...

joshday:
@pkofod yes, it works with a vector of functions as well.  Yep, @tbreloff, you got a good thing going here.

PR #133, opened 2016-02-04
Added PGFPlots.jl to list

mykelk:
It would be nice to support a LaTeX-based backend.


pkofod:
It would be nice indeed!

tbreloff:
I agree, but this belongs as an issue... I just started rewriting the docs yesterday, and this whole section is going away.

PR #159, opened 2016-03-09
Fixed some TODOs (Issue $108)

jheinen:
- support the grid keyword
- support handling of NaN (gaps in line segments)
- support xlims/ylims
- removed gr_state macro

The GR.savestate / GR.restorestate problem (caused by exceptions) will be solved
in GR - probably by something like GR.rewind() => to unwind the attribute stack.

### DIFF
--- a/src/backends/gr.jl
+++ b/src/backends/gr.jl
@@ -19,29 +19,12 @@ const gr_markertype = Dict(
   :star4 => -25, :star5 => -26, :star6 => -27, :star7 => -28, :star8 => -29,
   :vline => -30, :hline => -31 )
 
-set_gr_markertype(shape::Shape) = set_gr_markertype(:ellipse)
-set_gr_markertype(shape) = GR.setmarkertype(gr_markertype[shape])
-
-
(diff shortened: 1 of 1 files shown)


tbreloff:
All looks good, except I'm confused... What happens if there's an error outside of GR, but between a state save/restore? Will the unwind happen?

Another quick one... I recently added the option to specify one-sided x/y limits by setting one of the tuple values to +/- Inf. So xlim=(-Inf, 3) will fix the right boundary to 3, but let the left boundary vary with the data. If that's easy to do, now might be a good time to change it.

jheinen:
> All looks good, except I'm confused... What happens if there's an error outside of GR, but between a state save/restore? Will the unwind happen?

Yes. I plan to unwind the attribute stack before creating a new plot.

> Another quick one... I recently added the option to specify one-sided x/y limits by setting one of the tuple values to +/- Inf. So xlim=(-Inf, 3) will fix the right boundary to 3, but let the left boundary vary with the data. If that's easy to do, now might be a good time to change it.

I will implement this with the next patch, too.

tbreloff:
Thanks!

PR #161, opened 2016-03-10
Update GLVisualize

dlfivefifty:
This PR updates the GLVisualize backend to use surface, and the new syntax for GLVisualize.  The following example should work now:

```julia
using Plots
glvisualize()
surface(collect(1:10),collect(1:10),rand(10,10))
```

@SimonDanisch


tbreloff:
Cool thanks!

dlfivefifty:
Yes I’m planning on working on this further at some point: I need it for ApproxFun to do 3D plots, without building in directly support for GLVisualize.  But at least the following works already:

```julia
glvisualize() # or pyplot()
f=Fun((x,y)->exp(-x^2-y^2))
surface(f)
```

dlfivefifty:
The major thing that needs to be fixed is what happens after you close a window, though @SimonDanisch might know best what to do

SimonDanisch:
What's the problem exactly?

tbreloff:
I closed the window and was able to re-run the surface command successfully without error. Guess it works already??

SimonDanisch:
I fixed a few things at some point ;)

PR #169, opened 2016-03-17
Arc and chord diagram

diegozea:
Arc and chord diagrams ;)

Now both support keyword arguments, in particular `zcolor` and `group` and both use the same function to get the lists from an  adjacency matrix. I let you the final tweaks ;)

image

Best

### DIFF
--- a/src/Plots.jl
+++ b/src/Plots.jl
@@ -127,7 +127,8 @@ export
     PlotRecipe,
     # EllipseRecipe,
     spy,
-    arcdiagram
+    arcdiagram,
+    chorddiagram
     # corrplot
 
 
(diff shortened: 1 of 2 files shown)


diegozea:
Related issues: https://github.com/tbreloff/Plots.jl/issues/163 & https://github.com/tbreloff/Plots.jl/issues/147

tbreloff:
Thanks for the contribution @diegozea.  It would be really cool if you could show off a real-world use of these plots!

diegozea:
You're welcome! @tbreloff there is [an example here](http://nbviewer.jupyter.org/github/diegozea/MIToS.jl/blob/7cc27599b9c866d206ce6b04b5b73e2cd6868738/docs/Information.ipynb#Visualize-Mutual-Information) ;)

Issue #156, opened 2016-03-08
3D API


tbreloff:
I have decent support for 3D plotting already, but the internals are a little too hacky/messy for my tastes.  I want to refactor a little:

- [x] Better context-aware handling of input data (for example: matrix gets passed through for a surface plot, but not for a scatter) see: https://github.com/tbreloff/Plots.jl/issues/147#issuecomment-193858073
- [x] Arguments/handling of z-axis settings (subsumes #57)
- [x] `push!`/`append!` and `getindex`/`setindex!` for 3D data

GravityAssisted:
Do you know what all backends support 3D plots ? As far as I know, its only matplotlib, am I correct ?

tbreloff:
Plotly, PlotlyJS, GR, and GLVisualize as well.

pkofod:
(and soon PGFPlots https://github.com/pkofod/pgfplots_temp/blob/a9612f01032cb95922480ab13bdb922882ba27db/PGFPLots.ipynb )

tbreloff:
I just merged (into `dev`) a pretty substantial overhaul to the internals of plot creation.  As part of the overhaul:

- dictionaries that represent keyword arguments now use the typealias: `typealias KW Dict{Symbol, Any}`.  you'll see this all over the place, so take note if you're working on backend code.
- separation of old `createKWargsList` into `process_inputs` and `build_series_args` (also in a new file)... more robust handling of corner cases (I hope!)
- `z` argument is "first class" now, so 3D should be just as natural as 2D in many cases
- replaced keyword splatting with dictionary passing in many places (and many more to come)
- overhaul to the way the `group` arg is processed

I plan to continue code cleanup, but maybe not for a couple weeks.  Please, if you notice anything weird after all these changes, open an issue.

cc:  @spencerlyon2 @pkofod @dlfivefifty

pkofod:
Everything looks OK for the pgfplots branch.

dlfivefifty:
glvisualize() is working.

I'm disappointed that surface(::Matrix,::Matrix,::Matrix) isn't working yet, to support

```julia
using PyPlot

r=linspace(0.,1.,100)
θ=linspace(-π,π,100)
x=Float64[r*cos(θ) for r in r, θ in θ]
y=Float64[r*sin(θ) for r in r, θ in θ]
z=Float64[exp((r*sin(θ))^2+(r*cos(θ))^2) for r in r, θ in θ]

surf(x,y,z)
```

[...]

PR #176, opened 2016-03-29
Implement zlims!

abelsiqueira:
I needed for my animation, so I implemented it.
Not sure if it breaks anything else.

### DIFF
--- a/src/Plots.jl
+++ b/src/Plots.jl
@@ -72,6 +72,7 @@ export
     ylabel!,
     xlims!,
     ylims!,
+    zlims!,
     xticks!,
     yticks!,
     annotate!,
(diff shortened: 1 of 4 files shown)


tbreloff:
Thanks for the PR!  This does break things... you can look at the travis details, or run `Pkg.test("Plots")`.  You may need to add a few packages to get the tests to run.

I think the breakage is because you're always setting the zlim values, even for 2D plots.  The error I saw was a "set_zlim not defined" type of error.  If you're willing to fix so that the tests pass, then great.  Otherwise I'll merge this when I have time to fix it myself.  Thanks.

tbreloff:
Also, this implements part of #156.

tbreloff:
Merged... I'll try to fix this today.

tbreloff:
Should be fixed on dev... Please check if it works for you.

Issue #177, opened 2016-04-05
xlims/ylims not working for logscale in gadfly


mfalt:
It seems like the handling of xlims/ylims is not working in Gadfly when updating a plot, example:
```julia
Plots.gadfly()
plot()
plot!(0.1:0.1:10, 0.1:0.1:10,yscale=:log10, ylims=(0.1,10))
```
which gives a plot from 10^0.1 to 10^10, contrary to how the limits usually work, and contrary to how it works with `pyplot()`. This also make it impossible to have a plot from 10^-2 since negative numbers throw an error.

I'm on julia 0.4.3 with Plots v0.5.3

tbreloff:
I could reproduce on the dev branch.  It's weird that it works properly when creating a new plot, but fails when updating an existing plot.  I'll investigate.

mfalt:
Do you have any idea what might be causing this? I could spend a few hours to try to fix this if I knew where to look, I am quite dependent on this feature.

tbreloff:
> Do you have any idea what might be causing this?

I suspect that Gadfly is applying the transformation function on the `minvalue`/`maxvalue` fields right [here](https://github.com/dcjones/Gadfly.jl/blob/master/src/scale.jl#L268-L283), so to handle that (IMO strange) behavior, Plots would have to do the inverse transformation to the `minvalue`/`maxvalue` parameters [here](https://github.com/tbreloff/Plots.jl/blob/dev/src/backends/gadfly.jl#L383-L384).  Wanna take stab at a PR and check that the changes work?

mfalt:
I think I was able to fix this in #190, your comment was very helpful, even though the problem was not with the push to Scale, it was in Coord.

I left a comment about another bug that I noticed, which is regarding how the limits are handled by Gadfly. If we want them to be able to crop away parts of the plot we should add them in Gadfly.Coord.cartesian when a plot is created, I don't think this is done now.

tbreloff:
I saw the PR... thanks for the contribution!  I'll merge when the tests pass, but just for the future... can you try to branch/merge to the `dev` branch of Plots?  I prefer to accumulate new PRs which change functionality in the "unstable" branch, so that someone that uses master can be somewhat confident that things work.

> If we want them to be able to crop away parts of the plot we should add them in Gadfly.Coord.cartesian when a plot is created

I think I did this correctly at one time... I guess I broke it during a refactor?  If you know how to fix, please do!

mfalt:
[...]

PR #190, opened 2016-04-20
Better handling of xlims in Gadfly

mfalt:
This should solve the problem in issue #177.
You should however be aware that the values ymin/ymax in Gadfly.Scale.ContinuousScale is handled differently than in Gadfly.Coord.Cartesian which is what caused this bug in the first place.

One problem that persists is that while the values in ContinuousScale can extend the range of values shown, it can't decrease it to hide parts of the plot. This might make the axes change the first time an update is made (the behavior is the same without this patch).

### DIFF
--- a/src/backends/gadfly.jl
+++ b/src/backends/gadfly.jl
@@ -355,18 +355,18 @@ function getGadflyScaleFunction(d::KW, isx::Bool)
     hasScaleKey = haskey(d, scalekey)
     if hasScaleKey
         scale = d[scalekey]
-        scale == :ln && return isx ? Gadfly.Scale.x_log : Gadfly.Scale.y_log, hasScaleKey
-        scale == :log2 && return isx ? Gadfly.Scale.x_log2 : Gadfly.Scale.y_log2, hasScaleKey
-        scale == :log10 && return isx ? Gadfly.Scale.x_log10 : Gadfly.Scale.y_log10, hasScaleKey
-        scale == :asinh && return isx ? Gadfly.Scale.x_asinh : Gadfly.Scale.y_asinh, hasScaleKey
-        scale == :sqrt && return isx ? Gadfly.Scale.x_sqrt : Gadfly.Scale.y_sqrt, hasScaleKey
+        scale == :ln && return isx ? Gadfly.Scale.x_log : Gadfly.Scale.y_log, hasScaleKey, exp
+        scale == :log2 && return isx ? Gadfly.Scale.x_log2 : Gadfly.Scale.y_log2, hasScaleKey, exp2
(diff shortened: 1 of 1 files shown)


tbreloff:
it seems like the tests fail with errors now. You can run Pkg.test("Plots") locally to check. The errors I see are of the form:

```
    MethodError: `updateGadflyAxisFlips` has no method matching updateGadflyAxisFlips(::Gadfly.Plot, ::Dict{Symbol,Any}, ::Tuple{Void,Function}, ::Tuple{Void,Function})
    Closest candidates are:
      updateGadflyAxisFlips(::Any, ::Dict{Symbol,Any}, ::Any, ::Any, !Matched::Any, !Matched::Any)
     in updateGadflyGuides at /home/travis/.julia/v0.4/Plots/src/backends/gadfly.jl:451
     in plot! at /home/travis/.julia/v0.4/Plots/src/plot.jl:103
     in plot at /home/travis/.julia/v0.4/Plots/src/plot.jl:56
     in anonymous at /home/travis/.julia/v0.4/FactCheck/src/FactCheck.jl:271
     in do_fact at /home/travis/.julia/v0.4/FactCheck/src/FactCheck.jl:333
     [inlined code] from /home/travis/.julia/v0.4/FactCheck/src/FactCheck.jl:271
     in anonymous at /home/travis/.julia/v0.4/Plots/test/runtests.jl:14
     in facts at /home/travis/.julia/v0.4/FactCheck/src/FactCheck.jl:448
     in include at ./boot.jl:261
     in include_from_node1 at ./loading.jl:320
     in process_options at ./client.jl:280
     in _start at ./client.jl:378
  Error :: (line:-1)
    Expression: plot(Int[1,2,3],rand(3)) --> not(nothing)
```

mfalt:
Whops, I was editing the files in a different directory so I forgot to copy some of the changes. I was able to test it after installing VisualRegressionTests, ExamplePlots (its a bit annoying to have to add them manually), and they complete successfully.

tbreloff:
> after installing VisualRegressionTests, ExamplePlots (its a bit annoying to have to add them manually)

Agreed.  I should certainly publish VisualRegressionTests to METADATA, but I'm not sure that ExamplePlots belongs as a published package.  Does anyone have an opinion about that?

mfalt:
I guess they could still be pulled by the test script if they are not installed, even if they are not in METADATA, this was my main concern, but I'm not sure how its usually done.

tbreloff:
Yes they are cloned in the [travis script](https://github.com/tbreloff/Plots.jl/blob/master/.travis.yml).  For those packages which are used for testing and using a tagged version, I just add them to the `test/REQUIRE` file.

tbreloff:
Merged... Thanks for the contribution!

PR #208, opened 2016-04-28
[WIP] Further changes to PGFPlots backend.

pkofod:
- [x] Make linecolor work as intended for bar plots and histograms
- [ ] build styles and join with commas


tbreloff:
> Make linecolor work as intended for bar plots and histograms

Just to be clear... the convention I'm going for is that linecolor applies to the border for histograms, bars, hexbins, etc, and that fillcolor is the inside.

For shapes, scatters, etc it is markercolor and markerstokecolor.  The distinction is important for something like:

tmp

pkofod:
Yes, that is what I meant by "Make it work" :) As you can see, I currently use `seriescolor`. The problem is `linecolor` sometimes return `:black` (I believe if `linecolor` isn't set), and then `_pgfplots_get_color` will fail, because it assumes a `ColorWrapper`. I can't remember the details, that's why I wrote `#TODO`, to remind me :)

tbreloff:
Cool.  And to expand on colors a little more... you should be prepared to accept a bunch of different color types, but I think they will always be a subtype of ColorScheme.  This includes ColorWrapper, ColorFunction, ColorVector, ColorGradient, ...

Ideally if passed a ColorGradient you would apply accordingly.  So for example, if `markercolor` is `:inferno` then the markers should be colored according to the `marker_z` keyword.  Likewise, if `marker_z` is set but markercolor is not a gradient, you should choose the default gradient instead.

You can see some of this logic with the PyPlot backend... in general you should try to match the functionality/look of PyPlot as much as possible.

pkofod:
I will see what to do about the gradients. I think they might have to be defined and added to the preamble when needed, but I think that is possible in PGFPlots.jl, so that should be doable.

pkofod:
So, the bars are a bit wobbly, but...
wobbly

Should the grid lines be dotted?

tbreloff:
Cool good progress!  I don't care about the gridlines... whatever you think.  Let me know when you want to merge.

pkofod:
As before, random plots (produced when I test various features) are available here: https://github.com/pkofod/pgfplots_temp/blob/master/PGFPLots.ipynb (for the latest version).

I made the changes to the constant dictionaries, and a small change to the way incoming colors are handled. This allows for the `Symbol` version of the colors, but I still have to look into the different types of colors for that part to be great.

I also fixed a tiny bug that caused `:background_color` to be ignored.

If you want, you can merge now, and I will look at the string cat -> `join(vec, ',')` another time. I won't have time to work on in more today.

tbreloff:
Thanks!  Looks great

PR #209, opened 2016-04-28
Added polar plots

jheinen:
@tbreloff : the marker type example is now working, too. There was a problem in the PNG plugin - not related to the Julia wrapper. I started building new deps for GR.jl.

Will add transparency support with a separate PR.


tbreloff:
Awesome!  And glad you tracked down the png issue... let me know when it's fixed.

jheinen:
The binaries are now rebuilt. The GR.jl master branch should work (again). You probably have to re-install or rebuild GR.jl

PR #233, opened 2016-05-08
gr: support for colormaps

jheinen:
- needs further testing
- ``gr.jl`` 3D routines (internally) require 256 contiguous RGB24 color values (with color indexes ranging from 1000 to 1255)

@tbreloff : thanks in advance ...

### DIFF
--- a/src/backends/gr.jl
+++ b/src/backends/gr.jl
@@ -427,6 +427,17 @@ function gr_display(plt::Plot{GRBackend}, clear=true, update=true,
 
   for p in plt.seriesargs
     lt = p[:linetype]
+    if lt in (:hist2d, :hexbin, :contour, :surface, :wireframe, :heatmap)
+      if haskey(d, :color_palette)
+        ci = 1000
+        for cv in d[:color_palette]
+          GR.setcolorrep(ci, cv.r, cv.g, cv.b)
+          ci += 1
+        end
+      else
+        GR.setcolormap(GR.COLORMAP_COOLWARM)
+      end
+    end
     if get(d, :polar, false)
       lt = :polar
     end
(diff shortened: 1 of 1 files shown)


tbreloff:
Thanks Josef.. I might not get to this til tomorrow, but this givesme a
good head start.

jheinen:
I'd also like to add the new background arguments (tomorrow), but I have some questions:

- Are the following assumptions right?
  - ``background_color_inside`` describes the background color of the area within the axes
  - ``background_color_outside`` describes the background color of the area outside the axes
  - ``background_color`` describes the background color of the complete viewport

- Should a backend automatically change foreground colors in a manner, that "dark" backgrounds don't produce "invisible" ticks or axes labels?

- Does the ``aspect_ratio`` keyword change the appearance of the content (only) or the complete viewport? E.g., when using GR's native functions (like ``drawimage``), displaying an image automatically keeps the aspect ratio.

No hurry! I think we have some time to clarify these points until JuliaCon .... but I'd like to get ``gr.jl`` into a stable state (not only a fast replacement for another backend).

tbreloff:
> Are the following assumptions right?

Close.  You can see exactly what happens here:
https://github.com/tbreloff/Plots.jl/blob/dev/src/colors.jl#L363-L411

But in a nutshell, you don't need to support `background_color` or `foreground_color` at all, as those are only used when defining the sub-attributes.  You don't need to do any calculations... Plots does it.   You'll just want to blindly use the `background_color_*` and `foreground_color_*` attributes.

> Does the aspect_ratio keyword change the appearance of the content (only) 

In pyplot it only changes the content (the grid, not the colorbar, for example).  If that's easy then I'd say just do that.  Right now you can do `:none` (don't change anything), `:equal` (ratio == 1), or a number which represents width/height.

tbreloff:
Also so you know... `:color_palette` should probably never be used in backend code.  It is a color list/gradient which helps Plots to pick series colors automatically.  For colormaps, the fields `:markercolor`, `:fillcolor`, etc will have a `Plots.ColorGradient` type.

tbreloff:
And one more thing... the keywords will always be there, so `haskey(d, :xxxx)` will always be true.  (it's not checking what you think it's checking).  You probably want to compare it to a default value (frequently `nothing`) instead.

tbreloff:
I tried: 925243969c79a264f334637eb295290b7422f6cf  but it doesn't seem like it worked as expected.  Maybe I misunderstood when you said:

> 3D routines (internally) require 256 contiguous RGB24 color values (with color indexes ranging from 1000 to 1255)

Any ideas?

PR #244, opened 2016-05-10
Improved scaling of plots with error bars

jheinen:
- added an offset for both axes (in case the plot contains error bars)
- skip empty legends


jheinen:
Thanks!

PR #249, opened 2016-05-11
add support for legend and label in pgfplots

KristofferC:
also adds some extra kwargs for legend positions

image

And one using a new kwarg for pyplot for example `:bottomright`

image

cc @pkofod

### DIFF
--- a/src/args.jl
+++ b/src/args.jl
@@ -780,7 +780,7 @@ function convertLegendValue(val::Symbol)
         :best
     elseif val in (:no, :none)
         :none
-    elseif val in (:right, :left, :top, :bottom, :inside, :best, :legend)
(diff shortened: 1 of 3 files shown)


KristofferC:
Should I throw an error is someone enters a legend kwarg that is not supported by pgfplots, ex `:top`?

pkofod:
No, just allow for it, and place it in supported args. That's what I did for `stepmid` for example. thanks for adding this by the way.

KristofferC:
>  place it in supported args

what do you mean by this? Anything I should change in the PR?

tbreloff:
Merged thanks. Supported args isn't the right place for this (that would check that legends are supported, not `:top`)... you could add a warning and use a default value if you want.

pkofod:
Sorry! I misread, I thought it said "enters a legend kwarg that is not supported by Plots", and I thought Kristoffer meant a kwarg that wasn't supported (currently) by other backends. My bad :)

PR #251, opened 2016-05-11
Added (native) image support

jheinen:
@tbreloff : If the image aspect ratio has to be managed by the backend (wrapper), then the rectangle given by (``xmin, xmax, ymin, ymax``) in ``GR.drawimage()`` has to be adjusted, similar to https://github.com/jheinen/GR.jl/blob/master/src/jlgr.jl#L419-L460


tbreloff:
Cool thanks!

PR #257, opened 2016-05-13
Fix Deprecation Warning

ChrisRackauckas:
Fix for the warning:
sys:1: MatplotlibDeprecationWarning: The set_axis_bgcolor function was deprecated in version 2.0. Use set_facecolor instead.


tbreloff:
This makes every pyplot test fail. (most people are matplotlib version 1.5.1 at most)  I obviously can't merge.  You could either deal with the deprecation warning, or try to come up with some logic to conditionally choose the matplotlib method.  Let me know what you want to do.

ChrisRackauckas:
In that case stick with the warning until some later time.

PR #276, opened 2016-05-25
Fix plotlyjs subplots

spencerlyon2:
Just needed to call `PlotlyJS.relayout!` after `Plots.plotly_finalize`

### DIFF
--- a/src/backends/plotlyjs.jl
+++ b/src/backends/plotlyjs.jl
@@ -216,8 +216,13 @@ function _update_min_padding!(sp::Subplot{PlotlyBackend})
     sp.minpad = plotly_minpad(sp)
 end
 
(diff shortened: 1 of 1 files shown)


tbreloff:
Very good catch.  The real problem is that I don't update the bounding boxes before the `_update_plot` callback, so for it to work you need to call `relayout!` __again__ when finalizing.  The difference with the plotly backend is that plotly does everything after finalize, it's not created, then updated, like in plotlyjs.

I'm going to merge this, but I'll probably change the logic so that we don't need the fix in the future. Thanks for figuring it out!

spencerlyon2:
OK no problem. Feel free to change as needed. Glad we identified the issue

PR #294, opened 2016-06-01
Fixed axes limits problem

jheinen:


src/backends/gr.jl
```diff
@@ -268,16 +268,16 @@ end
...
+gr_y_axislims(sp::Subplot) = axis_limits(sp.attr[:yaxis], false)
+gr_z_axislims(sp::Subplot) = axis_limits(sp.attr[:zaxis], false)
 gr_xy_axislims(sp::Subplot) = gr_x_axislims(sp)..., gr_y_axislims(sp)...
 
 function gr_lims(axis::Axis, adjust::Bool, expand = nothing)
     if expand != nothing
         expand_extrema!(axis, expand)
     end
-    lims = axis_limits(axis, true)
+    lims = axis_limits(axis, false)
```
tbreloff: This will break a recent fix for missing markers in GR when the marker is on the axis limit (which is the reason I added `widen` in the first place).  So I'd prefer to keep this as true (i.e. it should widen) and just remove the `GR.adjustlimits` call.  Thoughts?

jheinen: I'd prefer to fix the marker problem in the right place. Do you have an example?

Right now, markers are drawn if they are in the clip rectangle (closed intervals for X, Y, **including** the endpoints): `xmin <= x <= xmax && ymin <= y <= ymax`

tbreloff: The bug was that the endpoints were not being included... I'll look for the
issue later.

PR #298, opened 2016-06-05
Make animation use ImageMagick.jl if it is installed.

dhoegh:
This will make animations easier to use on windows, as it will be easier to get the required dependencies. 

Btw. awesome work with the package.


tbreloff:
This is great. Thanks! I'll merge later. In the future, please try to work
from the dev branch though.

Issue #210, opened 2016-04-28
Boxplot using only y values


diegozea:
At the moment, you could indicate only Y values for `boxplot`, but the default width looks strange.
Also would be great to support a list of columns names when a wide dataframe is used. I found difficult to plot series of data (Y) since they use the same X value (`1`):

![image](https://cloud.githubusercontent.com/assets/2822757/14890032/593e5f30-0d61-11e6-99be-fa51ad0f28c3.png)

![image](https://cloud.githubusercontent.com/assets/2822757/14890397/ba528926-0d62-11e6-8cfe-5818e3be44a8.png)

Thanks!

tbreloff:
There are a couple issues here:

- Should we set xlims explicitly?  (I lean towards no...)
- How to choose the x values for a boxplot?
- Better handling of arrays of symbols (I agree this is broken right now)

In thinking about this just now, I had the thought that the current method of hoping that a Vector{Any} is good enough to allow dispatch on "processed" data is flawed and ripe for subtle bugs... I should replace the internal logic with a wrapper type:

```julia
immutable InputData{T}
  data::T
end
```

so that it's explicit that an input has been processed and wrapped, and dispatch will never get confused.  I'll create a separate issue for this, and the arrays of symbols issue should be resolved as part of that change.

tbreloff:
Right now I implement the boxplot recipe by explicitly applying the grouping and forcing the xticks to `1:length(shapes)`... this will need to be made more flexible to allow overlaying multiple boxplots.

As a stop-gap solution, you could build the arrays as expected by the current recipe:

![tmp](https://cloud.githubusercontent.com/assets/933338/14891812/43190b22-0d36-11e6-8f54-d332dd07fd20.png)

diegozea:
Boxplot looks broken right now:
![image](https://cloud.githubusercontent.com/assets/2822757/15132478/9761ebec-1630-11e6-97d2-854a71d57c6a.png)

tbreloff:
The weird boxplot drawing issue is fixed. 

I think the solution for the x-axis will be to have some sort of DiscreteAxis type that can map strings, etc to an x/y coordinate. I want to be able to overlay a scatter or violin plot over a boxplot but still allow new series to extend the axis. This can share implementation with the 'setStringVector..." stuff.

diegozea:
In the current master, boxplots are working fine with a categorical variable in x, but it can be used with group.

diegozea:
It can't be used with group.

[...]

PR #306, opened 2016-06-08
Solve a bug with whisker lengths (#210)

diegozea:
Now the whisker goes to the maximum and maximum values of the data inside the limits (Q2 - range*IQR and Q4 + range*IQR). If range is 0, then the whisker goes to the data extrema.
iris


tbreloff:
Thanks!

diegozea:
You're welcome! :)

PR #310, opened 2016-06-09
Add trim kw arg to violin and density

diegozea:
This PR add a `trim` argument to `violin` (default to `true`) and `density` (default to `false`). If `trim` is `true`, the tails are trimmed to the range of the data (`extrema`).

Also set (hard code) `bins=200` for this kind of plots. 

```julia
using RDatasets, Plots
mtcars = dataset("datasets","mtcars")
plot(
violin(mtcars,:Cyl,:MPG),
violin(mtcars,:Cyl,:MPG,trim=false),
density(mtcars,:MPG),
density(mtcars,:MPG,trim=true),
layout = grid(2,2)
)
```

image

### DIFF
--- a/src/recipes.jl
+++ b/src/recipes.jl
@@ -560,13 +560,18 @@ end
 # otherwise, just use a histogram
 if is_installed("KernelDensity")
     @eval import KernelDensity
-    @eval function violin_coords(y, bins = 30)
-        kd = KernelDensity.kde(y, npoints = isa(bins, Integer) ? bins : 30)
+    @eval function violin_coords(y; trim::Bool=false)
+        kd = KernelDensity.kde(y, npoints = 200)
+        if trim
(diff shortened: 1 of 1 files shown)


tbreloff:
Looks nice... thanks

diegozea:
You're welcome!

mauro3:
Thanks for looking into this!  Although this fix did not fix what I'm after :-)  The example on gitter was
`violin(1:10, randn(10,100) )` which should plot 10 violins, the first one with the samples `randn(10,100)[1,:]`, etc.  I.e. a violin plot without using DataFrames.

Tom gave the suggestion to use `violin(1:10, vec(randn(10,100)))` but that did not work, the violins were wrong (apart from the missing cut).  Now this suggestion errors.  Do I need another invokation? Or should I open a separate issue for this?

PR #313, opened 2016-06-10
Improved GR output

jheinen:
- slightly reduced the nominal marker size
- use character height as line spacing criterion when plotting legends
- use GR surface function for filled contours
- avoid calling GR setwindow/adjust* functions with Inf arguments
- corrected bounding box for legend


tbreloff:
Thanks @jheinen! I'll take a closer look at GR today and probably open an issue to track outstanding bugs and missing features.

PR #332, opened 2016-06-13
Fixed problem with log axes

jheinen:


tbreloff:
I made the changes in 266d2ef... figured it was easier than the "git-lympics" required to merge. Thanks for the fix!

PR #334, opened 2016-06-14
conserve user defined GR workstation type

jheinen:


tbreloff:
Thanks.  Can you write one sentence (maybe two) about the GR outputs and environment variable?  I'd like to add something to the docs.

jheinen:
The GKS_WSTYPE environment variable can be used to set the output device. The default workstation type is automatically selected depending on the platform:
 -  211 = X11 (Un*x)
 -  400 = Quartz (macOS only)
 -  42 = GDI (Windows only)

All output devices can also be specified by an alias:

| Type |Description |
|---------------|--------------|
| x11	| X11 (Un*x + macOS) |
| quartz	| Quartz (macOS only) |
| win	| GDI (Windows only) |
| ps, eps | 	PostScript |
| pdf	| Portable Document Format |
| bmp	 | Windows Bitmap (BMP) |
| pgf	 | PGF / TikZ |
| mov	| MPEG-4 video |
| jpeg, jpg | 	JPEG image file |
| png	| Portable Network Graphics file (PNG) |
| tiff, tif	| Tagged Image File Format (TIFF) |
| fig  | Xfig vector graphics file |
| svg	| Scalable Vector Graphics |
| wmf	| Windows Metafile |

The deployment of all these output types and required libraries ist challenging - but we are on a good way to realize a self-contained package for all platforms, especially for Julia.

There are more drivers, e.g. for GUI toolkits (like Qt4, wxWidgets or Gtk) or for OpenGL. The latter is currently rewritten to support Modern OpenGL - a first version will be available in 2-3 months.

There are also plans to develop a platform independent stand-alone Qt 5 backend, which would also solve the "multiple output" problem (figure oriented output with interactive extensions, like zooming and panning). The communication will be realized using an abstract serialization of GR commands, which could also feed a JavaScript based browser component (transpiled from GR code).

tbreloff:
Imagine if I asked for 3-4 sentences :laughing: Thanks!

PR #339, opened 2016-06-17
WIP: Adding support for arrays in violin-plot

mauro3:
Violin plots only work with `DataFrame`, see PR #310 for an example.

I use arrays and I would like to draw violins a locations `x` with distribution from samples in `s`.  Something like this would be the intuitive interface to me:
```
x = [1.0, 7.5, 9.0]
s = rand(100,length(x))
violin(x', s)
```
Which now gives:
vio

Any help appreciated in making the color not cycle.

### DIFF
--- a/src/recipes.jl
+++ b/src/recipes.jl
@@ -69,7 +69,7 @@ function seriestype_supported(pkg::AbstractBackend, st::Symbol)
     if st in supported_types(pkg)
         return :native
     end
-    
+
     haskey(_series_recipe_deps, st) || return :no
 
     supported = true
@@ -341,7 +341,7 @@ end
 # ---------------------------------------------------------------------------
 # bezier curves
 
-# get the value of the curve point at position t 
+# get the value of the curve point at position t
 function bezier_value(pts::AVec, t::Real)
     val = 0.0
     n = length(pts)-1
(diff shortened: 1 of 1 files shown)


src/recipes.jl
```diff
@@ -719,24 +719,39 @@ end
...
+    legend --> false
 
+    if length(d[:x])==1
+        widths, centers = violin_coords(d[:y], trim=trim)
         # normalize
         widths = _box_halfwidth * widths / maximum(widths)
 
         # make the violin
-        xcenter = discrete_value!(d[:subplot][:xaxis], glabel)[1]
+        xcenter = discrete_value!(d[:subplot][:xaxis], d[:x][1])[1]
```
mauro3: Not sure this is right.

tbreloff: Mauro I can't work on this now, but I think we need to talk it through.
Maybe put this on hold until I have time?

mauro3: No probs, take your time.  It works as far as I need it working at the moment and that publication is still a long way off...

mauro3: And thanks!

tbreloff:
I rebuilt the violin recipe using the recently added `Segments` type and also cycling over x values, so the following work as I would expect/want (on dev branch):

screen shot 2016-06-22 at 11 44 30 am

screen shot 2016-06-22 at 11 48 37 am

The examples above pass 2 vectors, where the x-values are cycled over the y values (this is why I take the transpose before converting to a vector).

You can also create 3 series by passing a 1x3 x-matrix and a 100x3 y-matrix:

screen shot 2016-06-22 at 1 09 23 pm

It's important to realize that there may be cases where the first example is what you want... if you want x to be treated as discrete values, map it to strings.  I may someday implement an `xdiscrete` attribute (and same for y, etc), but for now the behavior depends on the input type.

mauro3:
Perfect, thanks!

PR #342, opened 2016-06-17
Added suport for marker borders

jheinen:


tbreloff:
Thanks @jheinen! Example 4 gives an error now:

```
using Plots; gr(); test_examples(:gr,4)
```

I don't have time to fix right now, so I'll wait to merge.

PR #361, opened 2016-06-28
gr: allow simple formulas or LaTeX equations

jheinen:


tbreloff:
Cool!  Thanks @jheinen

PR #368, opened 2016-07-01
Added iTerm2 inline display support for gr

jheinen:
Example:
    using Plots
    gr(display_type=:inline)
    plot(Plots.fakedata(50,5), w=3)


tbreloff:
Exciting!  :+1:

PR #374, opened 2016-07-05
glvisualize backend

SimonDanisch:
Work in progress for the GLVisualize backend

### DIFF
--- a/src/axes.jl
+++ b/src/axes.jl
@@ -23,7 +23,7 @@ function Axis(sp::Subplot, letter::Symbol, args...; kw...)
         lk = Symbol(letter, k)
         lv = _axis_defaults_byletter[lk]
         d[k] = (lv == :match ? v : lv)
-    end     
+    end
 
     # merge!(d, _axis_defaults)
     d[:discrete_values] = []
@@ -309,15 +309,13 @@ function expand_extrema!(sp::Subplot, d::KW)
 
     # expand for bar_width
     if d[:seriestype] == :bar
-        dsym = vert ? :x : :y
+        dsym = (!vert) ? :x : :y
         data = d[dsym]
 
         bw = d[:bar_width]
         if bw == nothing
             bw = d[:bar_width] = mean(diff(data))
         end
-        # @show data bw
-
         axis = sp.attr[Symbol(dsym, :axis)]
         expand_extrema!(axis, maximum(data) + 0.5maximum(bw))
         expand_extrema!(axis, minimum(data) - 0.5minimum(bw))
@@ -489,4 +487,3 @@ function axis_drawing_info(sp::Subplot)
 
     xticks, yticks, spine_segs, grid_segs
 end
-
--- a/src/output.jl
+++ b/src/output.jl
@@ -134,6 +134,7 @@ const _mimeformats = Dict(
 const _best_html_output_type = KW(
     :pyplot => :png,
     :unicodeplots => :txt,
+    :glvisualize => :png
 )
 
 # a backup for html... passes to svg or png depending on the html_output_format arg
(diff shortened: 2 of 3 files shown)


src/backends.jl
```diff
@@ -26,18 +26,18 @@ macro init_backend(s)
     end)
 end
 
-@init_backend Immerse
-@init_backend Gadfly
```
tbreloff: Is there a reason you commented these out?

src/examples.jl
```diff
@@ -252,7 +252,7 @@ PlotExample("DataFrames",
         iris = RDatasets.dataset("datasets", "iris")
         scatter(iris, :SepalLength, :SepalWidth, group=:Species,
             title = "My awesome plot", xlabel = "Length", ylabel = "Width",
-            marker = (0.5, [:+ :h :star7], 12), bg=RGB(.2,.2,.2))
```
tbreloff: Can you revert this?

src/output.jl
```diff
@@ -210,7 +211,7 @@ end
 # IJulia
 # ---------------------------------------------------------
 
-const _ijulia_output = String["text/html"]
```
tbreloff: Please revert this

SimonDanisch: yeah of course, that's why it's a "do not merge" ;) These edits are just there to make my life a bit easier temporarily!

SimonDanisch:
- [ ] correct handling of all the different color attributs
- [ ] text and labels
- [ ] a few 3D plots still don't work perfectly
- [ ] better camera for corner cases (extreme zoom values etc)
- [ ] shape up shape plots
- [ ] fix efficient renderloop (currently inefficient again, since there are weird artifacts otherwise)
- [ ] fix line anti aliasing for all cases
- [ ] make wireframe work for (x,y,z) surface plots

SimonDanisch:
okay this starts looking okay!
Anyone daring to try this out should be on GLWindow, GLAbstraction, GLVisualize, FixedSizeArrays and GeometryTypes master!
New issues:
- [ ] 2D camera can now rotate which is extremely annoying :D Restrict movement!
- [ ] outline of sprites + barplots still not satisfactory
- [ ] 3D axis
- [ ] annotations
- [ ] Reactive message queue fills up with long blocking loops without yield (e.g. @gif)

Here's a little (badly compressed) gif:
surf

It's embarrassing slow though, since GLVisualize and Plots build up the complete plot for every frame which is completely unnecessary! ;) Will need to figure out a API/hijack the existing one, to bypass this!

SimonDanisch:
can someone try this out?! :)

SimonDanisch:
gotta be on:
GLWindow, GLAbstraction, GLVisualize and GeometryTypes master and for FixedSizeArrays on:
https://github.com/SimonDanisch/FixedSizeArrays.jl/pull/157

SimonDanisch:
Should we merge this?
And @tbreloff, could you add me as a collaborator, so that I can develop in a branch? It'll be easier to check out new versions of the backend like this!

tbreloff:
We should, but I'm traveling right now. Did you rebase on a recent dev?

tbreloff:
What a historic day... :tada: :tada: :tada:

src/axes.jl
```diff
@@ -309,15 +309,13 @@ function expand_extrema!(sp::Subplot, d::KW)
 
     # expand for bar_width
     if d[:seriestype] == :bar
-        dsym = vert ? :x : :y
+        dsym = (!vert) ? :x : :y
```
tbreloff: I just reverted this change... any idea why you did this?  The idea is that it's setting a default bar width and expanding the axis extrema.  When the plot is vertical, we want to update the x axis.

SimonDanisch: Well, it was skewed in the wrong direction so I changed this and then it was working...

PR #389, opened 2016-07-12
Legend location in subplot

odow:
@tbreloff This works for me but I have no idea if it impacts other parts of the code base!?!

```
using Plots
pgfplots()
plot(rand(10, 4), layout=4, legend=[:none :topright :bottomleft :bottomright])
```
legendlayout

### DIFF
--- a/src/args.jl
+++ b/src/args.jl
@@ -809,6 +809,7 @@ function convertLegendValue(val::Symbol)
 end
 convertLegendValue(val::Bool) = val ? :best : :none
 convertLegendValue(val::Void) = :none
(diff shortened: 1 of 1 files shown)


tbreloff:
Hi Oscar... This should work without any code changes. Let me take a
look... This method shouldn't be necessary. Thanks.

tbreloff:
Ok I just tested and looked at the code, and I agree we need this, though I'd rather it be:

```
convertLegendValue(v::AbstractArray) = map(convertLegendValue, v)
```

Do you want to make the change? or I can add this and set you as the author

odow:
@tbreloff thanks! Plots is great.

PR #391, opened 2016-07-16
pass axis style to PGFPlots

odow:
This pull request adds the ability to turn off axis ticks when using PGFPlots and fixes a bug that meant the axisstyle commands were never passed to the PGFPlot object.


tbreloff:
Thanks!

PR #425, opened 2016-07-31
Use dirname(@__FILE__) instead of Pkg.dir

tkelman:
This allows installing the package elsewhere.

Add testing against 0.5 to Travis - this runs the most
recent RC now, release once final tags are done


tbreloff:
Thanks for this @tkelman.  Sorry it didn't get into the last tag.

PR #442, opened 2016-08-10
Support *ticks in PGFplots.

pkofod:
ticks

### DIFF
--- a/src/backends/pgfplots.jl
+++ b/src/backends/pgfplots.jl
@@ -222,6 +222,11 @@ function pgf_axis(sp::Subplot, letter)
         kw[Symbol(letter,:max)] = lims[2]
     end
 
+    if !(axis[:ticks] in (nothing, false, :none, :auto))
+       push!(style, "$(letter)"*"tick = {$(join(get_ticks(axis),","))}")
+    end
+
+
     # return the style list and KW args
     style, kw
 end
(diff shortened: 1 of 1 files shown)


tbreloff:
FYI... your commit should have been rebased on dev (`git rebase origin/dev`) and then the PR should have been made against dev.  I should really add docs for contributing...

KristofferC:
This will constantly be a problem for anyone who doesn't use master as "dev". I don't really understand the workflow between having a dev branch separate from master. 

* If it is just for your own playing then isn't it easier to just make a separate branch and then merge it in. 
* If it is to not break stuff for people who are using master? Well, they are using master, deal with it :P
* Is it because the pervious release is to far away so people need to use master for the latest cool shit? Maybe tag more often.

tbreloff:
This case was bad because the PR was branched from over a month ago... didn't have anything to do with dev/master.  However:

> I don't really understand the workflow between having a dev branch separate from master

Master is the "stable development branch", and anything else is possibly not working.  Dev is the "real development branch".  I could reverse these, of course, but then people would check out master and get broken functionality.  It's not practical to tag constantly... it's a waste of community resources.  Maybe if we had a different package system I would tag more often, but it's a time consuming process for several people (namely me and Tony).

> Well, they are using master, deal with it :P

I wish that was practical...  I'd rather tell a few contributors to track dev, instead of telling a ton of users not to check out master.

pkofod:
My bad, such a long time since i PRed against Plots :)

PR #450, opened 2016-08-18
Add eps to savefig

anriseth:
This adds EPS support to `savefig`. 

Maybe I've just misunderstood, but it doesn't seem like this is supported.
Currently, the following does not create an eps image:
```j
plt=plot(cos,0,1)
savefig(plt, "test.eps")
```
Instead it creates a png file "test.eps.png".

### DIFF
--- a/src/output.jl
+++ b/src/output.jl
@@ -36,6 +36,13 @@ function ps(plt::Plot, fn::AbstractString)
 end
 ps(fn::AbstractString) = ps(current(), fn)
 
+function eps(plt::Plot, fn::AbstractString)
+  fn = addExtension(fn, "eps")
+  io = open(fn, "w")
+  writemime(io, MIME("image/eps"), plt)
+  close(io)
+end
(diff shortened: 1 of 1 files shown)


tbreloff:
Thanks for the PR @anriseth ... did you test this locally?  Which backend(s)?

anriseth:
I tested it with PyPlot, patched on top of v0.8.2. I couldn't test it on master because the `@compat`s have been removed [I guess it has something to do with Julia 0.5?].

Haven't tried the other backends.

tbreloff:
Ah ok.  0.8.2 is the last tag for julia 0.4.  Recent commits of Plots are 0.5-only.  I expect this code will work fine, but it's good for you to know that you're submitting the PR to a 0.5-only branch.

Another option is that you re-submit the PR to the `backports` branch (julia 0.4), and I'll cherry-pick it to dev.

anriseth:
I'm okay with leaving it for the julia 0.5 branches, I'll move over soon anyway.

PR #469, opened 2016-09-01
Added support for clims kwarg

jheinen:


tbreloff:
Thanks Josef. I won't be able to merge this right away as I'm traveling
today. Also, we need to solve the line_z and marker_z as well (anything
that will create a colorbar)

PR #471, opened 2016-09-05
Puts some values in the correct spot for PGFPlot

fredrikekre:
Puts `title`, `width` and `height` in the correct structure of `PGFPlots.jl`.

This way the keywords can be changed afterwards, i.e.

Before:
```jl
plt = plot(rand(3),rand(3), title = "Hello.")

p = backend_object(plt)

p.[1].title == "Hello."
    false
```

With this PR:

```jl
plt = plot(rand(3),rand(3), title = "Hello.")

p = backend_object(plt)

p.[1].title == "Hello."
    true
```


tbreloff:
Thanks @fredrikekre

PR #472, opened 2016-09-05
gr: added support for aspect_ratio

jheinen:


tbreloff:
Thanks @jheinen!  I haven't actually tested this, so I'll probably wait to merge into master.

jheinen:
@tbreloff : I compared the results with Matplotlib - setting the ``aspect_ratio`` for ``gr()``  should produce the same results as with ``pyplot()``.

sje30:
Thanks very much. I just pulled the dev branch and it works for me in my limited testing!

PR #474, opened 2016-09-06
forget plot if no label is provided

### LINKED ISSUE #473: Mismatch between linestyle in plot and legend when one label is ="" for Pgfplots
The legend should be orange:

image

KristofferC:
fixes #473 

is this reasonable @pkofod?

image


pkofod:
late to the party, but yes

PR #478, opened 2016-09-09
text rotation, move commands into module, is_marker_supported

SimonDanisch:


src/backends/glvisualize.jl
```diff
@@ -51,7 +51,8 @@ const _glvisualize_seriestype = [
 const _glvisualize_style = [:auto, :solid, :dash, :dot, :dashdot]
 const _glvisualize_marker = _allMarkers
 const _glvisualize_scale = [:identity, :ln, :log2, :log10]
-is_marker_supported(::GLVisualizeBackend, shape::Shape) = true
```
tbreloff: Will `Plots.Shape` be supported eventually?  It's just a wrapper around 2D coordinate arrays.

tbreloff:
It seems that you can't use Plots without GLVisualize, which is a problem (see the travis output)

src/backends/glvisualize.jl
SimonDanisch: Ah yeah, will add it back

SimonDanisch:
Huh, so you don't just include the backend code on demand?

tbreloff:
```julia
julia> module Outer
       module Inner
       println("I get run no matter what!!")
       f(x) = "NO MORE CATS"
       end
       end
I get run no matter what!!
Outer

julia> Outer.Inner.f("What should the next demo include?")
"NO MORE CATS"
```

Hopefully this illustrates why the submodule pattern won't actually help here.  We can't do anything GLVisualize-specific in code that is run on a call to `using Plots`.

SimonDanisch:
Well, you could only `include` it, when doing `glvisualize()`...

SimonDanisch:
Then put the whole thing into a module ;) That seems safer and more encapsulated anyways ;) Or am I missing some reason, why this can't work?

tbreloff:
Submodules feel annoying because, by default, everything gets masked when referring to the outer module.  Here's an example:

```julia
julia> module Outer
       g(x) = x
       module Inner
       importall Outer
       println("I get run no matter what!!")
       g(x::Int) = "Int"
       end
       end
WARNING: replacing module Outer
I get run no matter what!!
Outer

julia> Outer.g(10)
10

julia> Outer.Inner.g(10)
"Int"
```

I would have hoped that an `importall Outer` would mean that we extend `Outer.g`, but instead we mask it with a new method `Inner.g`.  To make this work you have to add `import Outer: g` or qualify with `Outer.g(x::Int) = "Int"`.  This just feels clunky and annoying.

That being said, I'm open to alternative ways to structure and load things, but only if it actually improves something. (I'd rather not focus on it... lets just use the pattern we already have and punt this discussion to another time)

SimonDanisch:
Not sure what's up with `importall`, but `import ..Outer3: g` works as expected ;)

tbreloff:
But that's the point... I don't want to explicitly list every method from Outer.

SimonDanisch:
Hm, well, but it would indeed make the intent clearer, when a backend is extending a Plots.jl function... which should be fairly rare, and shared by all backends?

SimonDanisch:
and if you accidentally not overloade a function, only that backend will fail...And accidentally overloading will be a lot harder :)

tbreloff:
> and if you accidentally not overloade a function, only that backend will fail...And accidentally overloading will be a lot harder :)

I get this reasoning.

So...

- Step 1: get it working with the current pattern so I can merge within breaking non-glvisualize stuff.
- Step 2: create a new branch and try out your proposed refactor.  See if it works and is better in any way

mmagnuski:
:cat2: :cat2: :cat2: :cat2: :cat2: :cat2: 
:cat: No more cats? :cat: 
:cat2: :cat2: :cat2: :cat2: :cat2: :cat2: 
:cry:

tbreloff:
Do you have any progress towards "step 1"?  I want to fix the overwritten `scatter` before tagging a new version of Plots.

SimonDanisch:
Okay, lets try this ;)

tbreloff:
You still have a few references to the `GL` module... do you want to fix those first?

SimonDanisch:
>You still have a few references to the GL module... do you want to fix those first?

These refer to `const GL = Plots`... Just a little trick to make switching back easier ;)

tbreloff:
alrighty!

PR #483, opened 2016-09-13
gif() now accepts an argument specifying number of times to loop

sivark:
Note that GIF playback might have some variability depending on the
player used: https://github.com/ImageMagick/ImageMagick/issues/263


sivark:
Also, gif generation with ffmpeg is unusable for me, so I have not been able to test it.

tbreloff:
Thanks

PR #488, opened 2016-09-15
fix ticks for pgfplots

KristofferC:
https://github.com/tbreloff/Plots.jl/pull/487 to right branch


tbreloff:
Thanks!

pkofod:
btw, check the original PR, I did the exact same thing: https://github.com/tbreloff/Plots.jl/pull/442

pkofod:
Also, I think it worked because I was only addressing adding ticks at custom x and y values, I didn't consider adding text (*ticklabels wasn't there).

PR #491, opened 2016-09-16
stop double adding style options to the axis in pgfplots

KristofferC:
The `append!` already appends the `axisstyle` to `style` so there is no need for the loop after. This caused for me double `xticks` and `xtickslabel` options in the .tex file (which is harmless but unnecessary).

cc @pkofod


pkofod:
I really can't remember the details of this, but it might be something that wasn't removed in the "reorganization".

PR #517, opened 2016-10-02
Fix for dpi in pyplot backend

fedxa:
A simple fix for dpi configuration in pyplot backend.  Easy to check that it should be like this when using, for example GtkAgg backend for matlpotlib.  Very useful on hidpi monitor.

### DIFF
--- a/src/backends/pyplot.jl
+++ b/src/backends/pyplot.jl
@@ -356,7 +356,7 @@ function py_bbox_title(ax)
 end
 
 function py_dpi_scale(plt::Plot{PyPlotBackend}, ptsz)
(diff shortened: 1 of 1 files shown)


tbreloff:
Is it really backwards? Sorry I thought I tested that.

fedxa:
Definitely backwards :)  Was fighting for some time trying to get the labels visible on my laptop.
By the way, is it really a proper behaviour to have the call to :set_size_inches in _before_layout_calcs?  In the way I use it (julia started from emacs ESS, with plot in a separate window) it is a bit annoying -- if I resize the window with a mouse and then replot, I the new plot appears in a part of the window.

PR #530, opened 2016-10-11
gr: changed heatmap code

jheinen:


tbreloff:
Thanks @jheinen I'll check this out

tbreloff:
This is really close to what I was hoping for!  I'm making a few updates... one thing is that I'd like for the "NaN color" to be invisible, not just the background color.  This is important when layering heatmaps on top of something else.  I'll make this change in a bit.  Thanks!

jheinen:
@tbreloff : To support NaNs, we have to draw an image (with support for transparency) instead of a cell array. The cell array indices (1000..1255) have to be replaced with corresponding RGBA values (using zero alpha values for NaNs). I'll make a PR ...

tbreloff:
Please wait... I'm almost done with a few changes

tbreloff:
Got it working! I'll push up in a sec.  Test with:

```julia
using Plots; gr()
z = float((1:10) .* (1:20)')
z[rand(10,20).<.3]=NaN
heatmap(z, clim=(30,100), bg=:orange)
```

tbreloff:
you'll need PlotUtils master and Plots dev

jheinen:
@tbreloff : you put me out of work :-)

tbreloff:
Ha. Don't worry we still need you plenty!

tbreloff:
@jheinen This example works in the gui window:

tmp


but when saving a png, the invisible cells get an background, and the y-axis is different, implying that `drawimage` is different for different outputs:

tmp

I'm on ubuntu, using GKS_WSTYPE "x11" for the first, and "png" for the second.

jheinen:
@tbreloff : X11 does not provide transparency - you have to switch to CairoX11 => "cairox11"

tbreloff:
No that's the opposite of what I see... "x11" **is** masking the cells properly.  It's the "png" that is wrong.

jheinen:
@tbreloff : I can reproduce the y-axis behavior with ``Plots.jl``, but **not** with plain ``GR.jl``. I'll add transparency to ``GR.jl`` (tomorrow) and then compare and check the results. It also turned out, that ``savefig`` can be improved in ``Plots.jl``, as there is no need to close the GUI window.

jheinen:
@tbreloff : The PNG should be transparent with the latest GR master (which uses the Cairographics driver for PNG image output). Unfortunately, currently I could only test this feature with GR as Plots is giving me an error:

```julia
julia> heatmap(z, clim=(30,100), bg=:orange)
ERROR: MethodError: no method matching isdark(::PlotUtils.ColorGradient)
```

Couldn't check the y-axis problem yet - works fine in GR.

jheinen:
... without the background keyword, it works:

screen shot 2016-10-13 at 17 12 26

... but the axes limits still change:

screen shot 2016-10-13 at 17 13 17

tbreloff:
Weird.  Seems like we have different versions of something...

jheinen:
GR is doing what it is told to (I added some ``println`` statements to ``Plots/src/backends/gr.jl``:

```julia
julia> using Plots; gr()
Plots.GRBackend()
julia> z = float((1:10) .* (1:20)');
julia> z[rand(10,20).<.3]=NaN;
julia> heatmap(z, clim=(30,100)) #, bg=:orange)
572: 1.0, 20.0, 1.0, 10.0
610: 1.0, 20.0, 1.0, 10.0
726: 1.0, 20.0, 1.0, 10.0
1012: 1.0, 20.0, 1.0, 10.0
julia> savefig("test.png")
572: 0.5, 20.5, 0.5, 10.5
610: 0.5, 20.5, 0.5, 10.5
726: 0.5, 20.5, 0.5, 10.5
1012: 0.5, 20.5, 0.5, 10.5
```

For any reason, the offset (``0.5``) is applied only in the second pass (``savefig``). I'll check this ...

jheinen:
@tbreloff ... these lines should be removed (in ``gr.jl``)

```julia
            if st == :heatmap
                expand_extrema!(sp[:xaxis], (x[1]-0.5*(x[2]-x[1]), x[end]+0.5*(x[end]-x[end-1])))
                expand_extrema!(sp[:yaxis], (y[1]-0.5*(y[2]-y[1]), y[end]+0.5*(y[end]-y[end-1])))
            end
```

tbreloff:
The "isdark" error is because of PlotThemes (cc: @pkofod).  The `:orange` theme adds an `:orange` color gradient, and so the "orange gradient" is being passed as the background color instead of "orange color".  We'll need to think of a good solution when a gradient name would clash with a color name.

>  these lines should be removed

I expect removing those lines will fix the y-axis issue, so we're down to the transparency in png.

jheinen:
The PNG transparency issue is fixed with the latest GR run-time. Temporary setting ``ENV["GRDIR"] = ""; Pkg.Build("GR")`` and restarting Julia should work ...

tbreloff:
Didn't work:

```
julia> ENV["GRDIR"]=""
""

julia> Pkg.build("GR")
INFO: Building GR
INFO: Downloading pre-compiled GR latest Ubuntu binary
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100 7935k  100 7935k    0     0  1386k      0  0:00:05  0:00:05 --:--:-- 1778k

julia> using Plots; gr()
Unable to load GR framework runtime environment
tom@tom-office-ubuntu:~/.julia/v0.5/Plots$ julia
               _
   _       _ _(_)_     |  A fresh approach to technical computing
  (_)     | (_) (_)    |  Documentation: http://docs.julialang.org
   _ _   _| |_  __ _   |  Type "?help" for help.
  | | | | | | |/ _` |  |
  | | |_| | | | (_| |  |  Version 0.5.1-pre+2 (2016-09-20 03:34 UTC)
 _/ |\__'_|_|_|\__'_|  |  Commit f0d40ec (23 days old release-0.5)
|__/                   |  x86_64-linux-gnu

julia> using Plots; gr()
Plots.GRBackend()

julia> z = float((1:10) .* (1:20)')
10×20 Array{Float64,2}:
  1.0   2.0   3.0   4.0   5.0   6.0   7.0   8.0   9.0   10.0  …   13.0   14.0   15.0   16.0   17.0   18.0   19.0   20.0
  2.0   4.0   6.0   8.0  10.0  12.0  14.0  16.0  18.0   20.0      26.0   28.0   30.0   32.0   34.0   36.0   38.0   40.0
  3.0   6.0   9.0  12.0  15.0  18.0  21.0  24.0  27.0   30.0      39.0   42.0   45.0   48.0   51.0   54.0   57.0   60.0
  4.0   8.0  12.0  16.0  20.0  24.0  28.0  32.0  36.0   40.0      52.0   56.0   60.0   64.0   68.0   72.0   76.0   80.0
  5.0  10.0  15.0  20.0  25.0  30.0  35.0  40.0  45.0   50.0      65.0   70.0   75.0   80.0   85.0   90.0   95.0  100.0
  6.0  12.0  18.0  24.0  30.0  36.0  42.0  48.0  54.0   60.0  …   78.0   84.0   90.0   96.0  102.0  108.0  114.0  120.0
  7.0  14.0  21.0  28.0  35.0  42.0  49.0  56.0  63.0   70.0      91.0   98.0  105.0  112.0  119.0  126.0  133.0  140.0
  8.0  16.0  24.0  32.0  40.0  48.0  56.0  64.0  72.0   80.0     104.0  112.0  120.0  128.0  136.0  144.0  152.0  160.0
  9.0  18.0  27.0  36.0  45.0  54.0  63.0  72.0  81.0   90.0     117.0  126.0  135.0  144.0  153.0  162.0  171.0  180.0
 10.0  20.0  30.0  40.0  50.0  60.0  70.0  80.0  90.0  100.0     130.0  140.0  150.0  160.0  170.0  180.0  190.0  200.0

julia> z[rand(10,20).<.3]=NaN
NaN

julia> heatmap(z, clim=(30,100), bg=:pink)

julia> png("/tmp/tmp")
```

tbreloff:
FYI the gradient name error is fixed on PlotThemes and Plots master

jheinen:
After setting GRDIR (to an empty string), which forces a new download when building the package, you have to quit Julia and start it again (without any GRDIR settings.

tbreloff:
I did quit and reload a second time, still no transparency **with png**.  Note the transparency works in the gui window.  Did you test this exactly (need master of Plots and PlotThemes):

```julia
using Plots; gr()
z = float((1:10) .* (1:20)')
z[rand(10,20).<.3]=NaN
heatmap(z, clim=(30,100), bg=:orange)
```

jheinen:
... yes, that's exactly what I tested. There is now old GR installation on your machine (in $HOME/gr, /opt/gr or /usr/local/gr) ?

tbreloff:
Ok working!  I did:

```
cd
mv gr old_gr
```

and then repeated the original steps, and now it works!  I guess it still references the `~/gr` directory even if you download a new binary.  Sorry for the noise, and thanks!

jheinen:
I think, GR should be more verbose during initialization to avoid such noise ....
A GR.version() function would probably help.

tbreloff:
GR.version() would be a good solution... print out whatever info you have about the currently loaded library.

pkofod:
> The "isdark" error is because of PlotThemes (cc: @pkofod). The :orange theme adds an :orange color gradient, and so the "orange gradient" is being passed as the background color instead of "orange color". We'll need to think of a good solution when a gradient name would clash with a color name.

oh, hadn't thought that could be an issue, for now I will keep this in mind!

PR #556, opened 2016-10-31
Fix markerstrokecolor match

mkborregaard:
Makes `markerstrokecolor` value `:match` default to `seriescolor` as specified in the docs (current setting `foreground_color_subplot`)


mkborregaard:
My preference would have it default to linecolor instead - this would default to seriescolor, except for in bar diagrams (which look strange now):
skaermbillede 2016-10-31 kl 11 57 47

mkborregaard:
I would also suggest only upper intervals shown for barplots, as is the convention.

PR #559, opened 2016-11-01
fix mesh particle size bug and better window creation

SimonDanisch:
This doesn't currently work on atom though :( Not sure whats up with that, but the plot pane won't show and it will change the plot size to 65535x1, which messes things up quite a bit.


tbreloff:
Because master diverged from dev I just force pushed over dev... make sure you re-branch from dev again so that the git history doesn't get messy.

PR #573, opened 2016-11-16
Enable shorthand methods on subplots too

swt30:
Shorthand methods like `title!` and `annotate!` currently only work on fully-fledged plots; this change broadens the method definitions so that they work on subplots too.

```julia
using Plots
funcs = [sin, cos]
p = plot(funcs, -6, 6, layout=2, legend=false)
ylims!(p, -1.5, 1.5) 
annotate!(p[1], -pi/2, -1.2, "Trough")
annotate!(p[2], 0, 1.1, "Peak")
```

figure_1


tbreloff:
Seems like a good change. Thanks

Issue #574, opened 2016-11-19
Legend position with PlotlyJS


tcovert:
I am not able to use the `legend` keyword to move legends using the PlotlyJS backend.  For example, this does not work:
```
julia> using Plots, Blink

julia> plotlyjs()
Plots.PlotlyJSBackend()

julia> p = plot(Plots.fakedata(50,5),w=3)

julia> plot!(p, legend = :bottom)
```
the resulting plot still has 5 legend entries, in the upper left. Is this intentional?

tbreloff:
It's not the behavior I want, but I can't figure out how to have
independent legends in Plotly, or how to reposition it. If anyone knows how
please let me know.

piever:
You can actually specify legend position in x and y coordinates in the layout with PlotlyJS:

```
using PlotlyJS
traces = GenericTrace[]
l = Layout(; legend = attr(x = 0., y= 0.))
for i = 1:5
    trace = scatter(; x = 1:10, y = rand(10), name = "draw $i")
    push!(traces,trace)
end
plot(traces,l)
```

Here for example x = 0, y = 0 corresponds to bottom left. I'm not sure about independent entries.

tbreloff:
Thanks! I'll get to this eventually but a PR from someone else would speed the process. ;)

piever:
I've implemented a simple way to get x,y position of legend from keyword (say [0,0] for bottom left, [0.5, 0] for bottom, [1,0] for bottom right), but it has a couple of issues:

1) The automated positioning of the legend with PlotlyJS is often not very good, because the cartesian axes are inside the figure and the legend will likely end up on top of them. There would need to be some method to programmatically detect a smart position (taking the axis position into account) or the user should be given the possibility of manually inputting coordinates.

<img width="641" alt="screenshot 2016-11-19 21 40 14" src="https://cloud.githubusercontent.com/assets/6333339/20458609/99d8ac76-aea1-11e6-84c3-51e815d19331.png">

2) Given that this happens via the layout, in case of several subplots all the legends are in the same position, which is determined by the position of the legend of the last subplot. This may be an inherent problem of plot.ly

<img width="622" alt="screenshot 2016-11-19 21 50 30" src="https://cloud.githubusercontent.com/assets/6333339/20458631/470c411e-aea2-11e6-9438-f4fc8ccf7cd0.png">

kaslusimoes:
Should this be closed?

piever:
[...]

PR #579, opened 2016-11-20
Implemented legend position in plotly

piever:
I've added a way to change the position of the legend in plotlyjs backend (see #574 ). One can either use the keywords:

screenshot 2016-11-20 15 46 42

Or input coordinates as a tuple:

screenshot 2016-11-20 15 50 12


tbreloff:
Thanks for the PR!

Issue #549, opened 2016-10-23
yerrorbars with categorical x axis in scatter


rand5:
scatter([.5],[.5],yerror=[.001]) produces the expected result but scatter(["A"],[.5],yerror=[.001]) results in an error even though scatter(["A"],[.5]) works.

wkearn:
This has been fixed since #589, so I'll close this issue.

PR #589, opened 2016-11-26
Allow errorbars in plots with categorical x axis (see #549 )

piever:
As mentioned in #549 , it is impossible to add errorbars to a plot if the x axis is categorical. This is simply because Plots.jl uses NaN separated segments to do that and it can't mix NaN with the x-values for a categorical x. This change expands the element-type of the x-axis to allow appending NaNs and now  having categorical variables doesn't create problems anymore:

screenshot 2016-11-26 02 10 18

### DIFF
--- a/src/recipes.jl
+++ b/src/recipes.jl
@@ -526,8 +526,7 @@ end
 
 function error_coords(xorig, yorig, ebar)
     # init empty x/y, and zip errors if passed Tuple{Vector,Vector}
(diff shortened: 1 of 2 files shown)


tbreloff:
Seems like a nice solution. Thanks. It might be a couple days before I have
a chance to merge. @simondanisch feel free to merge if you're around.

PR #591, opened 2016-11-28
Fix embeding of PlotlyJS plots in Jupyter

pabloferz:
This is a solution to the problem in https://github.com/spencerlyon2/PlotlyJS.jl/issues/89, also mentioned in https://github.com/tbreloff/Plots.jl/issues/481#issuecomment-249421012


tbreloff:
Thanks!

pabloferz:
No problem.

pabloferz:
@tbreloff I spoke too soon. This worked a few releases ago. Now, this works just for the first time you draw a plot. I you run the plot command more than one the plots won't be drawn.

Maybe this needs something like the `html_head` and `html_body` you use in the plotly backend instead.

PR #597, opened 2016-12-01
Output/Animation improvements

tbreloff:
- `inline` method and now `show` accepts `:gui`,`:inline`,`:ijulia` values.  see: https://www.youtube.com/watch?v=pLA9H9LzTWs&feature=youtu.be
- animation support for mov/mp4, mp4 will display inline in IJulia


cortner:
thank you - very helpful.

PR #604, opened 2016-12-07
Proper "id" synchronization of plotlyjs plots

pabloferz:
This is an addition that I missed in #591. With this patch things should properly work now.

Cc. @tbreloff

### DIFF
--- a/src/backends/plotlyjs.jl
+++ b/src/backends/plotlyjs.jl
@@ -86,8 +86,8 @@ end
 # ----------------------------------------------------------------
 
 function _show(io::IO, ::MIME"image/svg+xml", plt::Plot{PlotlyJSBackend})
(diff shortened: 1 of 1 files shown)


piever:
It seems that the recently acquired ["save as html"](https://github.com/tbreloff/Plots.jl/pull/590) is no longer working after this and the previous PR on plotlyjs backed (but it works at [this](https://github.com/tbreloff/Plots.jl/commit/90e655d97925b3a91fd73bc9f492371e74a2b171) commit). At least on my machine, now savefig("test.html") outputs an html file that seems incomplete and if I try to open it with a browser I get a blank page. Do you have the same issue?

pabloferz:
@piever Thanks for pointing that out. I updated the PR to fix the issue you're having.

Issue #610, opened 2016-12-11
Plotting a function without specifying range of x-axis fails with confusing error message


ilyagr:
To reproduce:

```julia
julia> using Plots
julia> plot(cos)
ERROR: BoundsError: attempt to access 0-element Array{Plots.Subplot,1} at index [1]
 in getindex(::Plots.Plot{Plots.PyPlotBackend}, ::Int64) at /home/ilyag/.julia/v0.5/Plots/src/types.jl:92
 in macro expansion at /home/ilyag/.julia/v0.5/Plots/src/series.jl:346 [inlined]
 in apply_recipe(::Dict{Symbol,Any}, ::Base.#cos) at /home/ilyag/.julia/v0.5/RecipesBase/src/RecipesBase.jl:238
 in _process_userrecipes(::Plots.Plot{Plots.PyPlotBackend}, ::Dict{Symbol,Any}, ::Tuple{Base.#cos}) at /home/ilyag/.julia/v0.5/Plots/src/pipeline.jl:73
 in _plot!(::Plots.Plot{Plots.PyPlotBackend}, ::Dict{Symbol,Any}, ::Tuple{Base.#cos}) at /home/ilyag/.julia/v0.5/Plots/src/plot.jl:173
 in plot(::Function) at /home/ilyag/.julia/v0.5/Plots/src/plot.jl:46
```

The problem is at `series.jl:346`. It tries to axess the x-range of a non-existing plot object to set the proper range of x-values.  There should either be some default (-10,10 ?), or it should fail with a more helpful error message, e.g. "Cannot determine an approriate range for xaxis. Please add an ... argument".

If there is an existing plot object, there is no problem. E.g., the following succeeds:
```julia
julia> plot([1,-2,3])
julia> plot!(cos)
```

PR #623, opened 2016-12-24
use adapted_grid from PlotUtilities for single functions

KristofferC:
image

### DIFF
--- a/src/series.jl
+++ b/src/series.jl
@@ -343,8 +343,9 @@ end
 
 @recipe function f(f::FuncOrFuncs)
     plt = d[:plot_object]
-    xmin,xmax = axis_limits(plt[1][:xaxis])
-    f, xmin, xmax
+    xmin, xmax = axis_limits(plt[1][:xaxis])
+    xs = adapted_grid(f, (xmin, xmax))
+    xs, f
 end
 
 #


src/series.jl
```diff
@@ -343,8 +343,9 @@ end
 
 @recipe function f(f::FuncOrFuncs)
```
ChrisRackauckas: You have to already have the axis? You can't also pass the axis by a tuple?

ChrisRackauckas: If you do `plot(f,lims=(0.0,1.0))` will that work?

KristofferC: No, see https://github.com/tbreloff/Plots.jl/issues/610, https://github.com/tbreloff/Plots.jl/issues/622

ChrisRackauckas: So would this be usable from a recipe?

KristofferC: I don't really know anything how recipes work.

ChrisRackauckas: I was planning to change `plot(sol)` to return a function, but if `plot(f)` doesn't work, then I am sure `plot(sol)` won't work. If there's an argument to set the axis, then it would still be able to work, but if you have to create a separate plot first, then I'm not sure how to do that in a recipe.

KristofferC: Fixing #610 could be one way forward.

PR #626, opened 2016-12-31
Initial support for InspectDR.

ma-laforge:
Hi Tom,

Not sure if you want this on master, or dev, or something else.

**Comments**
 - Has alot of functionality at this point - but not nearly as complete as Plots.jl/PyPlot, for example.
 - InspectDR is mostly about interactivity.  Though the plots look nice, less emphasis was placed on contolling all aspects of formatting.  Instead, the GUI window is optimized for plot real-estate, readable titles, quick data navigation etc.

PS:
Developed using InspectDR/master (Pkg.checkout("InspectDR").
I can start tagging a new release if you wish... but I am not sure how quickly it will get merged.

### DIFF
--- a/src/backends/inspectdr.jl
+++ b/src/backends/inspectdr.jl
@@ -1,0 +1,411 @@
+
+# https://github.com/ma-laforge/InspectDR.jl
+
+#=TODO:
+Not supported by InspectDR:
+    :foreground_color_title (font), title_location
+    :background_color_legend, :background_color_inside, :background_color_outside,
+    :foreground_color_grid, :foreground_color_legend, :foreground_color_title,
+    :foreground_color_axis, :foreground_color_border, :foreground_color_guide, :foreground_color_text,
+    :polar,
+
+
+Add in functionality to Plots.jl:
+    :annotations, :aspect_ratio,
+=#
+
+# ---------------------------------------------------------------------------
+#TODO: remove features
+const _inspectdr_attr = merge_with_base_supported([
+    :annotations,
+    :background_color_legend, :background_color_inside, :background_color_outside,
+    :foreground_color_grid, :foreground_color_legend, :foreground_color_title,
+    :foreground_color_axis, :foreground_color_border, :foreground_color_guide, :foreground_color_text,
+    :label,
+    :linecolor, :linestyle, :linewidth, :linealpha,
+    :markershape, :markercolor, :markersize, :markeralpha,
+    :markerstrokewidth, :markerstrokecolor, :markerstrokealpha,
+    :markerstrokestyle, #Causes warning not to have it... what is this?
+    :fillcolor, :fillalpha, #:fillrange,
+#    :bins, :bar_width, :bar_edges, :bar_position,
+    :title, :title_location, :titlefont,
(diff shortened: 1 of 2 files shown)


ma-laforge:
Once plots are displayed, feel free to interact with them:
 - Box-zoom: `Right mouse button + drag`
 - Pan: `Shift + left mouse button + drag`
 - Zoom out full: `CTRL + f`
 - (...) more at: https://github.com/ma-laforge/InspectDR.jl#Bindings

tbreloff:
Thanks and sorry for the delay. I've looked at it enough to know that it
isn't merge-able yet but haven't had time to do a proper review.

ma-laforge:
No problem.  I was just giving out a crash course since you probably don't have time for the entire `InspectDR/README.md`.

RE merge-able:
I am a bit surprised that appveyor would be failing on the new code...

**FYI**
I have implemented a bunch more functionality in the "inspectdr" branch of my Plots.jl library, including
 - more color support
 - Arbitrary Shapes/markers.

Support for arbitrary Shapes was surprisingly easier than I thought.

...but
I don't want to mess up what you started, - thus my use of the "inspectdr" branch.

src/backends/inspectdr.jl
```diff
@@ -0,0 +1,411 @@
...
+#Hack: suggested point size does not seem adequate relative to plot size, for some reason.
+_inspectdr_mapptsize(v) = 1.5*v
+
+# ---------------------------------------------------------------------------
+#InspectDR-dependent structures and method signatures.
+#(To be evalutated only once ready to load module)
+const _inspectdr_depcode = quote
+
+import InspectDR
+export InspectDR
```
tbreloff: This is not allowed.  It would require adding InspectDR as a dependency of Plots, which is not an option.  Search for `export` in other backends to see how it's handled.

ma-laforge: Actually, it is equivalent to what is found in GR & PyPlot (It just looks different):

```
function _initialize_backend(::GRBackend; kw...)
    @eval begin
        import GR
        export GR
    end
end
```

I just broke it out to a **quoted expression** (one that does not run), called `_inspectdr_depcode`.  It only runs once `_initialize_backend` gets executed (below):
```
function _initialize_backend(::InspectDRBackend; kw...)
    eval(_inspectdr_depcode)
end
```
I did this because the code was sort of long - and I find indenting to be awkward inside `_initialize_backend` (a personal coding-style thing).

I must admit, I trimed out 80% of the code since then.  I don't mind moving it make things look more like the other backends if you wish.

ma-laforge: BTW, If you do want me to change this before merging:

Should I just make a pull request on my new code that supports more features?

tbreloff:
Ok I see. Yes I'd prefer to keep things consistent.  The indentation
doesn't look as nice but it clearly says "this is not normal code" which I
think is helpful.

tbreloff:
If it's easier to open a fresh PR, that's fine.

ma-laforge:
Superseded by https://github.com/tbreloff/Plots.jl/pull/631

PR #631, opened 2017-01-05
Inspectdr

ma-laforge:
New PR on new branch after initial design review.

Reminder: Once plots are displayed, feel free to interact with them:
 - Box-zoom: Right mouse button + drag
 - Pan: Shift + left mouse button + drag
 - Zoom out full: CTRL + f
 - ... (more at: https://github.com/ma-laforge/InspectDR.jl#Bindings)

### DIFF
--- a/src/backends/inspectdr.jl
+++ b/src/backends/inspectdr.jl
@@ -1,0 +1,411 @@
+
+# https://github.com/ma-laforge/InspectDR.jl
+
+#=TODO:
+Not supported by InspectDR:
+    :foreground_color_title (font), title_location
+    :background_color_legend, :background_color_inside, :background_color_outside,
+    :foreground_color_grid, :foreground_color_legend, :foreground_color_title,
+    :foreground_color_axis, :foreground_color_border, :foreground_color_guide, :foreground_color_text,
+    :polar,
+
+
+Add in functionality to Plots.jl:
+    :annotations, :aspect_ratio,
+=#
+
+# ---------------------------------------------------------------------------
+#TODO: remove features
+const _inspectdr_attr = merge_with_base_supported([
+    :annotations,
+    :background_color_legend, :background_color_inside, :background_color_outside,
+    :foreground_color_grid, :foreground_color_legend, :foreground_color_title,
+    :foreground_color_axis, :foreground_color_border, :foreground_color_guide, :foreground_color_text,
+    :label,
+    :linecolor, :linestyle, :linewidth, :linealpha,
+    :markershape, :markercolor, :markersize, :markeralpha,
+    :markerstrokewidth, :markerstrokecolor, :markerstrokealpha,
+    :markerstrokestyle, #Causes warning not to have it... what is this?
+    :fillcolor, :fillalpha, #:fillrange,
+#    :bins, :bar_width, :bar_edges, :bar_position,
+    :title, :title_location, :titlefont,
+    :window_title,
+    :guide, :lims, :scale, #:ticks, :flip, :rotation,
+    :tickfont, :guidefont, :legendfont,
+    :grid, :legend, #:colorbar,
+#    :marker_z,
+#    :line_z,
+#    :levels,
+ #   :ribbon, :quiver, :arrow,
+#    :orientation,
+    :overwrite_figure,
+#    :polar,
+#    :normalize, :weights,
+#    :contours, :aspect_ratio,
+    :match_dimensions,
+#    :clims,
+#    :inset_subplots,
+    :dpi,
+#    :colorbar_title,
+  ])
+const _inspectdr_style = [:auto, :solid, :dash, :dot, :dashdot]
+const _inspectdr_seriestype = [
+        :path, :scatter #, :steppre, :steppost, :shape,
+    ]
+#see: _allMarkers, _shape_keys
+const _inspectdr_marker = Symbol[
+    :none,
+    :auto,
+    :circle,
+    :rect,
+  :diamond,
+#  :hexagon,
+  :cross,
+  :xcross,
+  :utriangle,
+  :dtriangle,
+  :rtriangle,
+  :ltriangle,
+#  :pentagon,
+#  :heptagon,
+#  :octagon,
+#  :star4,
+#    :star5,
+#  :star6,
+#  :star7,
+  :star8,
+#  :vline,
+#  :hline,
+  :+,
+  :x,
+]
+
+const _inspectdr_scale = [:identity, :ln, :log2, :log10] #Does not really support ln, (plot using log10 instead).
+
+#Do we avoid Map to avoid possible pre-comile issues?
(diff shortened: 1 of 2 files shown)


tbreloff:
I just fetched this PR and tested it, but the most basic command isn't working:

```julia
julia> using Plots; inspectdr(); plot(rand(10))
WARNING: Couldn't initialize inspectdr.  (might need to install it?)
INFO: To do a standard install of inspectdr, copy and run this:

WARNING: No custom install defined for inspectdr
nothing

ERROR: UndefVarError: GlyphPolyline not defined
```

but I was able to get InspectDR working independent of Plots.  I'll try quickly to nail down the problem, but no promises.

src/backends/inspectdr.jl
```diff
@@ -0,0 +1,480 @@
...
+
+# ----------------------------------------------------------------
+
+# Display/show the plot (open a GUI window, or browser page, for example).
+function _display(plt::Plot{InspectDRBackend})
+    mplot = _inspectdr_getmplot(plt.o)
+    if nothing == mplot; return; end
+    gplot = _inspectdr_getgui(plt.o)
+
+    if nothing == gplot && true == plt[:show]
```
tbreloff: Don't check `plt[:show]`... if we're inside the `_display` method then we want to show it, regardless of the attribute.

src/backends/inspectdr.jl
```diff
@@ -0,0 +1,480 @@
...
+        return InspectDR.AxisScale(:log10)
+    elseif :ln == s
+        return InspectDR.AxisScale(:ln)
+    else #identity
+        return InspectDR.AxisScale(:lin)
+    end
+end
+
+# ---------------------------------------------------------------------------
+
```
tbreloff: Since a simple `Pkg.add` doesn't work, it would be helpful to include a method like this:

```julia
function add_backend_string(::PyPlotBackend)
    """
    if !Plots.is_installed("PyPlot")
        Pkg.add("PyPlot")
    end
    withenv("PYTHON" => "") do
        Pkg.build("PyPlot")
    end
    """
end
```

But with whatever install commands are appropriate.

ma-laforge:
> I'm happy to merge whenever you want.

(I'm getting some mixed messages, so I will try to adress what I can)

**ANS:** *I* am ok with you merging in my addition to Plots.jl as is - but in case *you* are not:

>  just fetched this PR and tested it, but the most basic command isn't working:

I am not sure why you had problems.  Is it because I forgot to remind you you need to `Pkg.checkout("InspectDR")` at the moment?

**IMPORTANT:** I just made a PR to tag InspectDR (since you sound more-or-less satisfied): https://github.com/JuliaLang/METADATA.jl/pull/7512.

> Since a simple `Pkg.add` doesn't work, it would be helpful to include a method like this:

I don't think this will still be true once METADATA accepts my new tag of InspectDR.

**Q:** Irrespective: should I still add the following to inspectdr.jl?:

```
function add_backend_string(::InspectDRBackend)
    """
    if !Plots.is_installed("InspectDR")
        Pkg.add("InspectDR")
    end
    """
end
```

ma-laforge:
Also, regarding:
> Don't check plt[:show]... if we're inside the _display method then we want to show it, regardless of the attribute.

That was my original interpretation as well, but when I started using Plots.jl: I got behaviour that I thought to be strange (I might be wrong, though):

First, I run:
```
using Plots
inspectdr()
p1=plot(rand(10), show=true)
```

The plot shows up in a GUI - as expected.

, but I then want to generate a second plot... one that does **not** show.  So, I run the following code:
```
p2=plot(rand(10), show=false)
```

**Problem:** once I have set `show=true`, the second plot seems to ignore my desire to **not** show the plot (ie `show=false`).  Am I doing something wrong/is my expectation incorrect?

I added the check for `true == plt[:show]` as a remedy for what I believed was an undesired behaviour.

**Question:** Was my interpretation wrong?  Do you still want me to make this correction and remove the check for `true == plt[:show]`?

tbreloff:
Sorry about the confusion.  What I mean is that... originally I thought this PR would **break other backends** because of the import/export confusion.  But I just tested and I think any inconsistencies will be InspectDR-backend-only, and so I leave it to you how you want to handle them.  I'm happy to merge now and improve later, or you can keep working on this.

> I am not sure why you had problems. Is it because I forgot to remind you you need to Pkg.checkout("InspectDR") at the moment?

Yes... sorry I thought I posted that but probably forgot.

> Don't check plt[:show]

This should be fixed, as it doesn't display when you call `plot`.

> Q: Irrespective: should I still add the following to inspectdr.jl?:

Up to you... not really necessary if a Pkg.add gets the user running.

tbreloff:
>  but I then want to generate a second plot... one that does not show.

You're thinking about this wrong... the `show` attribute is a helper to automatically call `display` after a plotting command is finished.  This lets you see updates inside a script, etc.

However if the `Plot` object is being returned to the REPL, **it will display** without regard for the `show` attribute.  In fact with `show=true` it will be **displayed twice**.

tl;dr Use a semicolon at the end:

```
p2 = plot();
```

ma-laforge:
Got it.  Will get these things fixed soon.

Question: Do I just re-click on pull request once I am done?  How does this work?

If it is easier: you could merge, and I could create a new pull request.

tbreloff:
> Question: Do I just re-click on pull request once I am done?

Even easier... just push up a new commit which makes your changes and it will automatically be added to the PR. (assuming you're still on the same branch)

ma-laforge:
Perfect, thanks: This should only take a few minutes.

tbreloff:
Thanks!  Whenever you feel ready, you should announce the new backend on Discourse, and maybe consider writing something up here: https://github.com/JuliaPlots/PlotDocs.jl/blob/master/docs/backends.md

ma-laforge:
Great.  I should add something to your backend doc. I suppose I just make a pull request there.

Thanks for the suggestion about "Discourse":
-- I should probably wait until my tag gets merged into METADATA before announcing it.

ma-laforge:
Side question:

Would you consider InspectDR to have a "a small footprint" (like Plotly & UnicodePlots)?

Background: It depends mostly on Cairo & optionally: Gtk.

I personally find those two pacakges to be relatively lightweight (for a GUI toolkit)... and I have not had any installation issues - as of yet.

tbreloff:
Hmm. More "average footprint" if compared witth Plotly and UnicodePlots I
think. I've had plenty of issues withh both dependencies. But I'll be happy
to include any reasonable language you want.

ma-laforge:
> But I'll be happy to include any reasonable language you want.

No need - just curious.  I think InspectDR has enough good points to be preferred by certain users/applications - even without being of "small footprint".

I must admit: It would have been a Herculean effort for a GUI solution to have as small a footprint as a tool like UnicodePlots :).

PR #635, opened 2017-01-06
Use `sans-serif` as default font family (fix #593)

### LINKED ISSUE #93: GLVisualize backend
cc: @dlfivefifty @simondanisch   First PR is merged in dev!  I'll collect a TODO list and any issues/progress here and in the comments.

### LINKED ISSUE #593: `Font family [u'Helvetica'] not found` with PyPlot backend
When using PyPlot, on my system (OS X 10.11, Julia v0.5.0) I always get the following:

```
$ julia -e 'using Plots; pyplot(); plot(rand(10), show=true)'
/Users/nowacki/.julia/v0.5/Conda/deps/usr/lib/python2.7/site-packages/matplotlib/font_manager.py:1288: UserWarning: findfont: Font family [u'Helvetica'] not found. Falling back to Bitstream Vera Sans
  (prop.get_family(), self.defaultFamily[fontext]))
```

This is regardless of whether I use Python and Matplotlib from MacPorts, or after installing it locally with Conda.jl (via `ENV["PYTHON"]=""; Pkg.build("PyPlot")`).

I believe this occurs because the default font set by Plots.jl is Helvetica (in `src/components.jl:242`?), but PyPlot cannot find it.  This is curious, though, because PyPlot on its own has no problem using Helvetica when set in `~/.matplotlib/matplotlibrc` and called separately:

```
$ cat ~/.matplotlib/matplotlibrc 
font.sans-serif : Helvetica
$ julia -e 'using PyPlot; plot(rand(10))'
```

I can confirm that PyPlot is reading this file for defaults:
```
$ julia -e 'using PyPlot; println(PyPlot.matplotlib[:matplotlib_fname]())'
/Users/nowacki/.matplotlib/matplotlibrc
```
and that it's also true when using Plots:
```
$ julia -e 'using Plots; pyplot(); println(PyPlot.matplotlib[:matplotlib_fname]())'
/Users/nowacki/.matplotlib/matplotlibrc
```

Note that deleting `~/.matplotlib` and trying the first command also causes problems, hence my suspicion that it is the way that Plots is setting the default font which causes issues.

Please let me know what else might help fix the issue.  Having to use Bitstream Vera Sans is one of the few impediments to using Plots.jl in publications right now!

anowacki:
Rather than specifying a specific typeface, set the font
*family* in the `family` field of `Font` as `sans-serif`
by default.

Fixes #593.

### DIFF
--- a/src/components.jl
+++ b/src/components.jl
@@ -248,7 +248,7 @@ end
 function font(args...)
 
   # defaults
-  family = "Helvetica"
+  family = "sans-serif"
   pointsize = 14
   halign = :hcenter
   valign = :vcenter


tbreloff:
I just tested and this seems to work for most backends.  Thanks

PR #639, opened 2017-01-07
Fix save as .tex for PGFPlots

crbinz:
From @KristofferC's suggestion [here](https://github.com/tbreloff/Plots.jl/issues/637). I don't actually know if this approach breaks things for other backends or not, so have a look.

### DIFF
--- a/src/output.jl
+++ b/src/output.jl
@@ -52,6 +52,11 @@ function tex(plt::Plot, fn::AbstractString)
 end
 tex(fn::AbstractString) = tex(current(), fn)
 
+function tex(plt::Plot{PGFPlotsBackend}, fn::AbstractString)
+  fn = addExtension(fn, "tex")
+  PGFPlots.save(fn, backend_object(plt), include_preamble=false)
+end
+
 function html(plt::Plot, fn::AbstractString)
     fn = addExtension(fn, "html")
     io = open(fn, "w")


KristofferC:
I think you should just add a method instead of removing the fallback one.

crbinz:
Done. I am not sure, however, what that default method is even for, though. It doesn't work with any of the backends I've tried (understandably).

KristofferC:
Okay, I'm not sure either why it is there but maybe that is for another PR if it should be removed from all other backends? Another thing is if the new method should be in the pgfplot backend file. I guess only Tom knows where he wants it. 

Also,  there is a question of whether the preamble should be included or not. Most of the times I save figures to .tex to include them in a larger document  so not including the preamble is probably correct.

tbreloff:
The correct way to do this is to implement the appropriate show method.
(Which I think gave you a method error originally). The tex method should
be left generic, and pgf-specific code should live in pgfplots.jl. Let me
know if there's anything confusing.

crbinz:
In that case I'll have to leave it up to someone else to fix. I don't even know where to start with implementing the right `show` method.

tbreloff:
No problem... I can add it myself when I have a chance.

crbinz:
Obviously no rush, since the `PGFPlots.save` workaround works fine in the meantime.

PR #640, opened 2017-01-08
WIP: adding surfacecolor attribute in surface

BoundaryValueProblems:
This is my attempt to add ```surfacecolor``` attribute in the ```surface``` function of ```Plots.jl``` using ```PlotlyJS``` as its backend. At least, specifying the ```surfacecolor``` attribute in the ```surface``` function does not throw error or warning message at this point, but the way I modified ```backend/plotly.jl``` is still wrong. I also added ```test/surfacetest.jl```, but it does not plot anything somehow. If I copy the main body of that test to my julia REPL session, it generates a surface plot with still a wrong color on it. I would greatly appreciate your help to complete this task.

Just in case, if one runs the following segment without ```using Plots```, but with ```using PlotlyJS```, it works, and I want to do the same with ```using Plots```.

```julia
using PlotlyJS
m=32; n=5;
r = (0:m)/m
theta = reshape(pi*(-n*m:n*m)/m, 1, 2*n*m+1)
z = r * exp(im*theta)
s = r.^(1/n) * exp(im*theta/n)
x = real(z)
y = imag(z)
u = real(s)
v = imag(s)
surface(x=x,y=y,z=u,surfacecolor=randn(size(v)))
```

### DIFF
--- a/src/backends/plotly.jl
+++ b/src/backends/plotly.jl
@@ -20,7 +20,7 @@ const _plotly_attr = merge_with_base_supported([
     :guide, :lims, :ticks, :scale, :flip, :rotation,
     :tickfont, :guidefont, :legendfont,
     :grid, :legend, :colorbar,
-    :marker_z, :levels,
+    :marker_z, :surfacecolor, :levels,
     :ribbon, :quiver,
     :orientation,
     # :overwrite_figure,
@@ -432,7 +432,7 @@ function plotly_series(plt::Plot, series::Series)
 
     # for surface types, set the data
     if st in (:heatmap, :contour, :surface, :wireframe)
-        for letter in [:x,:y,:z]
+        for letter in [:x,:y,:z,:surfacecolor]
             d_out[letter] = plotly_surface_data(series, series[letter])
         end
     end
--- a/src/arg_desc.jl
+++ b/src/arg_desc.jl
@@ -29,6 +29,7 @@ const _arg_desc = KW(
 :z                 	=> "Various. Input data. Third Dimension. May be wrapped by a `Surface` for surface and heatmap types.",
 :marker_z          	=> "AbstractVector, Function `f(x,y,z) -> z_value`, or nothing. z-values for each series data point, which correspond to the color to be used from a markercolor gradient.",
 :line_z          	=> "AbstractVector, Function `f(x,y,z) -> z_value`, or nothing. z-values for each series line segment, which correspond to the color to be used from a linecolor gradient.  Note that for N points, only the first N-1 values are used (one per line-segment).",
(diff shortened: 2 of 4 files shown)


tbreloff:
Yes this doesn't look right. I'll comment further when I have time, thanks
for the PR though

src/backends/plotly.jl
```diff
@@ -432,7 +432,7 @@ function plotly_series(plt::Plot, series::Series)
 
     # for surface types, set the data
     if st in (:heatmap, :contour, :surface, :wireframe)
-        for letter in [:x,:y,:z]
+        for letter in [:x,:y,:z,:surfacecolor]
```
mkborregaard: This bit of code is for the data series (x, y and z), not for passing keyword arguments.

mkborregaard:
Are you sure `surfacecolor` is implemented in PlotlyJS.jl? (cf @sglyon ) I cannot find the attribute over on the repo.

sglyon:
Yep, `surfacecolor` does work in plotly.js and PlotlyJS.jl.

See here for docs: https://plot.ly/javascript/reference/#surface-surfacecolor

tbreloff:
I should note that it seems a better name for this attribute is `fill_z` for consistency.  For comparison, you set the gradient with `linecolor` but you set numeric values to be mapped using `line_z`.  The same is true of `markercolor`/`marker_z`.  Since the color gradient for the surface would be set by `fillcolor`, I think it makes sense to add the corresponding attribute `fill_z` (which currently uses the actual `z` values)

BoundaryValueProblems:
Tom: I agree with you. ```fill_z``` may be a better name. I want to mention the following fact. It's quite interesting to see that the ```surfacecolor``` attribute in PlotlyJS corresponds to the ```facecolors``` attribute in PyPlot. The difference between them is that in the case of ```surfacecolor``` attribute in PlotlyJS, one can assign any real-valued matrix (of the same size as the z matrix), say, ```data``` to it as ```surfacecolor=data```, and it automatically normalizes the matrix ```data``` to the full dynamic range of whatever the colormap a user specifies for the ```surface``` function. On the other hand, ```facecolors``` attribute in PyPlot needs more work: one has to retrieve a desired colormap (e.g., ```colmap=get_cmap("viridis")```), normalize the data matrix yourself, and then specify the attribute like ```facecolors=colmap(normalized_data)```. Once we succeed in  adding the ```surfacecolor``` or rather ```fill_z``` attribute in ```backends/plotly.jl```, then I can help doing the same for ```backends/pyplot.jl```.

tbreloff:
I spent a bit of time today figuring out the right way to do this.  It was not too obvious and didn't want you to waste your time.  See #641

PR #641, opened 2017-01-09
Add fill_z attribute and plotly impl

tbreloff:
replaces #640


tbreloff:
```julia
using Plots
plotlyjs()
z = rand(20,20)
surface(z)
```

tmp


```julia
surface(z, fz=rand(size(z)))
```

tmp

BoundaryValueProblems:
Thanks, Tom! But something is still wrong. When I supply ```x``` and ```y``` coordinates together with ```z``` coordinates and ```fill_z```, then it displays a wrong color. See the difference between the following two different way of rendering the surface:
```julia
m=32; n=3; r=(0:m)/m; theta = reshape(pi*(-n*m:n*m)/m, 1, 2*n*m+1);
z = r * exp(im*theta); s = r.^(1/n) * exp(im*theta/n); x = real(z); y = imag(z); u = real(s); v = imag(s);
surface(u,fz=randn(size(v))) # Color correct, but the x, y coordinates are wrong.
surface(x,y,u,fz=randn(size(v))) # Color wrong, but the x, y coordinates are correct.
```
Thanks for your further help!

tbreloff:
I see now what you're saying.  I fixed this.

tmp

BoundaryValueProblems:
Thank you so much, Tom!! It's great to see this attribute work now!

PR #642, opened 2017-01-11
[WIP] expand view port of gr backend

### LINKED ISSUE #602: Pie chart legend is cut
When plotting a pie chart using the gr backend, the legend of some pieces of the pie does not appear on screen. The viewport is too small and the legend is off screen. In the following example, "3" is offscreen and also off figure if saved.
 
    julia> using Plots; gr()
     Plots.GRBackend()

    julia> pie(["1", "2", "3"], [.4, .1, .5])

I don't know if the problem also happens with the pyplot backend. I am using Julia Version 0.5.0 (2016-09-19 18:14 UTC) together with Plots 0.10.1 and GR  0.18.0.

lokeshh:
Fixes #602

### DIFF
--- a/src/backends/gr.jl
+++ b/src/backends/gr.jl
@@ -491,7 +491,7 @@ function gr_display(plt::Plot)
 
     # subplots:
     for sp in plt.subplots
-        gr_display(sp, w*px, h*px, viewport_canvas)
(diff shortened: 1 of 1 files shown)


tbreloff:
This is not a good general solution.  Take a look at the `:margin` attribute(s) if you need to fine-tune spacing for a specific plot.

PR #650, opened 2017-01-19
fixed segfault issue when qt>v4.8.5 on Ubuntu using pyplot

huckl3b3rry87:
@tbreloff simply adding:
```
import Conda
Conda.add("qt=4.8.5")
```
to the ``function add_backend_string(::PyPlotBackend)`` did not work. For instance, if you run:
```
import Conda
Conda.add("qt=4.8.6")
```
then restart julia and run:
```
julia> using Plots

julia> plot(rand(30,3))

signal (11): Segmentation fault
while loading no file, in expression starting on line 0
unknown function (ip: 0x32735)
Allocations: 15516666 (Pool: 15515155; Big: 1511); GC: 27
Segmentation fault (core dumped)
febbo@febbo-HP-ZBook-17-G2:~/.julia/v0
```
So, it will never print out your warning message. So, I just automated it and told the user to restart julia.

### DIFF
--- a/src/backends/pyplot.jl
+++ b/src/backends/pyplot.jl
@@ -56,6 +56,34 @@ function add_backend_string(::PyPlotBackend)
         Pkg.build("PyPlot")
     end
     """
+    if !Plots.is_installed("PyPlot")
+        Pkg.add("PyPlot")
+    end
+    withenv("PYTHON" => "") do
+        Pkg.build("PyPlot")
+    end
+    import Conda          # TODO since this only known to occur on Ubuntu we could also check the user's OS here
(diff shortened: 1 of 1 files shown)


tbreloff:
I think I'd prefer if you just printed a warning with  the Conda command
that the user can copy/paste, and an informative message about why. I don't
like the idea of automatically changing Conda versions on backend loading.

huckl3b3rry87:
that is fine. I made another pull request that addresses this

huckl3b3rry87:
For me this is no longer an issue, it just creates an annoying message. Can we delete this warning?

PR #656, opened 2017-01-24
Add support for layout & plot size.

ma-laforge:
Also:
Add basic support for polar plots.
Add support for plot title & background color of entire figure.

**InspectDR tags pending:** https://github.com/JuliaLang/METADATA.jl/pull/7647

Probably best not to merge until tagged?


ma-laforge:
New version InspectDR required for this PR just tagged.

PR #663, opened 2017-02-01
path spaces fix stolen from http://stackoverflow.com/a/13692919

fuzzybear3965:
Spaces in Windows paths resulted in improper DOS command. Fix designed
according to the above-referenced StackOverflow post.


tbreloff:
I trust this works (I don't code in Windows) :smile:

fuzzybear3965:
I didn't like that the build was failing, but then I realized that your builds are failing, too. Now, I don't feel so bad.

Issue #647, opened 2017-01-16
PGFPlots Examples not working.


sebastianpech:
Hi,

I can't get the `bar()` and `histogram()` example from the [pgf examples](https://juliaplots.github.io/examples/pgfplots/) to work. Every call to one of the functions results in:

```
ERROR: The backend must not support the series type Val{:shape}, and there isn't a series recipe defined.
  in apply_recipe(::Dict{Symbol,Any}, ::Type{Val{:shape}}, ::Array{Float64,1}, ::Array{Float64,1}, ::Void) at /Users/***/.julia/v0.5/RecipesBase/src/RecipesBase.jl:238
  in _process_seriesrecipe(::Plots.Plot{Plots.PGFPlotsBackend}, ::Dict{Symbol,Any}) at /Users/***/.julia/v0.5/Plots/src/pipeline.jl:397
  in _process_seriesrecipe(::Plots.Plot{Plots.PGFPlotsBackend}, ::Dict{Symbol,Any}) at /Users/***/.julia/v0.5/Plots/src/pipeline.jl:406 (repeats 2 times)
  in _plot!(::Plots.Plot{Plots.PGFPlotsBackend}, ::Dict{Symbol,Any}, ::Tuple{Array{Float64,1}}) at /Users/***/.julia/v0.5/Plots/src/plot.jl:227
  in (::Plots.#kw##plot)(::Array{Any,1}, ::Plots.#plot, ::Array{Float64,1}) at ./<missing>:0
  in #histogram#360(::Array{Any,1}, ::Function, ::Array{Float64,1}, ::Vararg{Array{Float64,1},N}) at /Users/***/.julia/v0.5/Plots/src/Plots.jl:139
  in (::Plots.#kw##histogram)(::Array{Any,1}, ::Plots.#histogram, ::Array{Float64,1}, ::Vararg{Array{Float64,1},N}) at ./<missing>:0
```

Basic plot, scatter and histogram2d are working fine. I updated every related package to the master branch and I'm working on julia 0.5. The code I'm trying to get up and running is:

```
using Plots
pgfplots()
histogram(randn(100),nbins=10)
```

Can you help me with this issue? It seems that those plots used to work.

sebastianpech:
So I kind of "fixed" it for my case. To get the histogram back to work I had to manipulate the file:`backends/pgfplots.jl`.

[...]

PR #667, opened 2017-02-03
Fix/Extend pgfplots functions

sebastianpech:
As I wrote in #647 I started to fix/extend some features of the pgfplots backend. The changes provide the following:

- Enable histograms, Fix fillcolor bug
- Fix yshift of subplots
- Add annotations and series_annotations
- Add support for polar axis
- Add support for ticks rotation

### DIFF
--- a/src/backends/inspectdr.jl
+++ b/src/backends/inspectdr.jl
@@ -454,9 +454,9 @@ function _update_min_padding!(sp::Subplot{InspectDRBackend})
 
     # add in the user-specified margin to InspectDR padding:
     leftpad   = abs(bb.xmin)*px + sp[:left_margin]
-    toppad    = abs(bb.ymax)*px + sp[:top_margin]
+    toppad    = abs(bb.ymin)*px + sp[:top_margin]
     rightpad  = abs(bb.xmax)*px + sp[:right_margin]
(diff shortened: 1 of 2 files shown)


pkofod:
Cool stuff, good to see some more features added. For reference, could you post some small examples? Code+output

sebastianpech:
Basically I used the examples from the doc and tried getting them to work as expected. I have a [jupyter notebook](https://github.com/sebastianpech/share/blob/master/julia/PGFExamples.ipynb) with my tests (It's large so it takes a while to show up).  Every change in the example code is documented above the cell.

pkofod:
That notebook is exactly what I had in mind, will have a look. Thanks

sebastianpech:
For comparison this is what the current master produces: [notebook](https://github.com/sebastianpech/share/blob/master/julia/PGFExamples-master.ipynb)

pkofod:
Looks great. I was sure I added histogram at some point, but it might have been removed at some point by mistake.

KristofferC:
Very nice

sebastianpech:
I thought so, because to enable the bar and histogram I only needed to add :shape and except for the fillcolor everything was working.

tbreloff:
Thanks :+1:

haampie:
Great work! Really appreciated

PR #669, opened 2017-02-06
enable standalone window for plotlyjs() backend

YongHee-Kim:
enviroment variable `ENV["PLOTS_USE_ATOM_PLOTPANE"]`
is not working for both `plotly()` and `plotlyjs()` backend. 

but since saving to html is possible with #590
plotlyjs could use both `ATOM_PLOTPANE` and standalone window. 

this is simple fix I could've come up with. through It doesn't look so clean...


src/backends/web.jl
```diff
@@ -33,11 +33,15 @@ function open_browser_window(filename::AbstractString)
 end
 
 function write_temp_html(plt::AbstractPlot)
-    html = standalone_html(plt; title = plt.attr[:window_title])
     filename = string(tempname(), ".html")
-    output = open(filename, "w")
-    write(output, html)
-    close(output)
+    if plt.backend == PlotlyJSBackend()
```
tbreloff: Can you instead define `function write_temp_html(plt::Plot{PlotlyJSBackend}) ...` in the backend file?

YongHee-Kim: sure. that was 0880487. 
reverting to 0880487 would be ok?

tbreloff: Sure

YongHee-Kim: It's done 😄

PR #674, opened 2017-02-08
WIP: lookup attributes at the console

mkborregaard:
This PR adds the option to look up plot attributes (including aliases) at the console. It essentially makes this table https://juliaplots.github.io/attributes/ available at the console with the function `plotattr`.
My motivation for this is that when I code I am deeply dependent on looking up functions in the console with `?`. But Plots is essentially just the `plot` function, so this approach does not work, and I find I often go to the webpage looking through attributes. The `plotattr` function just makes it easier. 
Everything here is WIP, from the function name to what is being typed to the code, but it does work now. I shamelessly nicked a lot of the code from PlotDocs. I pass attribute names as strings, overall categories (`attrtype`) as symbols. Some example usage:
```julia
julia>plotattr(:Axis)
Defined Axis attributes are:
discrete_values, flip, foreground_color_axis, foreground_color_border, foreground_color_guide, foreground_color_text, formatter, guide, guidefont, lims, link, mirror, rotation, scale, tickfont, ticks

julia>plotattr(:Series, "ms")
markersize (Series attribute)
Aliases: markersizes, ms, msize 

Default: 4	(Type: Number or AbstractVector)

Description: 
Size (radius pixels) of the markers.

#you don't need the overall category
julia>plotattr("tickfont")
tickfont (Axis attribute)
Aliases:  

Default: Plots.Font("sans-serif",8,:hcenter,:vcenter,0.0,RGB{N0f8}(0.0,0.0,0.0))	(Type: Font)

Description: 
Font of axis tick labels.

### CODE REVIEW

**Comment by mkborregaard** [comment]
Changed the output format:
```julia
julia> plotattr("ms")
markersize {Number or AbstractVector}
markersizes, ms, msize

Size (radius pixels) of the markers.
Series attribute,  default: 4
```

tbreloff:
This is a great idea, and it might make sense to add some bonus info for certain attributes... like a more complete descriptions of the core series types, etc.  Thanks for working on it.

mkborregaard:
Yes that is a nice idea - do you think that should be added to _arg_desc?

PR #679, opened 2017-02-10
gr: added support for hexbin plots

jheinen:


tbreloff:
Thanks @jheinen.  FYI you should feel free to merge any PRs that only touch the GR backend code without my approval.

ohmsweetohm1:
Has this been stripped from `Plots.jl`? I see `GR.jl` would support it.

jheinen:
I think `hexbin` has never found its way into the Plots API ...

PR #683, opened 2017-02-14
Fix halign for annotations in pgfplots backend 

sebastianpech:
PGFPlots uses inverted right and left parameters.

### DIFF
--- a/src/backends/pgfplots.jl
+++ b/src/backends/pgfplots.jl
@@ -98,6 +98,12 @@ const _pgf_series_extrastyle = KW(
     :xsticks => "xcomb",
 )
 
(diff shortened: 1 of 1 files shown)


src/backends/pgfplots.jl
```diff
@@ -98,6 +98,12 @@ const _pgf_series_extrastyle = KW(
     :xsticks => "xcomb",
 )
 
+const _pgf_annotation_halign = KW(
+    :center => "",
+    :left => "right",
```
KristofferC: Doesn't matter much but for something like this I would have put in a short comment.

sebastianpech: You're right I'll add one.

Issue #645, opened 2017-01-14
Clean up temporary files, avoid leaking file handles


tkelman:
Many of the backends create temporary files and/or open file handles with `open` without cleaning up either. While GC will eventually deal with the file handles, some OS's require you to close handles before you're allowed to delete a file. And accumulating an unbounded number of files over time isn't a nice thing to do to users' temporary directories.

tbreloff:
Since you took the time to do a code review, a PR would be really helpful.

tkelman:
I'm not set up to use or test the code here.

YongHee-Kim:
I have few questions to regarding fixing this issue 
1. Searching for every file with prefix `jl_` inside a temporary folder and delete them will do? 
   or since prefix `jl_` can be used in any other Julia packages, should we put `plots_jl` prefix for temporary files?  

2. If above is a right way to do this, Isn't performance cost of searching filename prefix inside the hard drive is expensive? When should I do file searching and deleting?

tkelman:
better to target the specific places where files are created. or run everything in `mktempdir() do` blocks that will clean up automatically

YongHee-Kim:
But sometimes it's not clear when the clean up should happen. 
For example, `ploty()` backend uses OS browser, and I don't know how Julia can know when rendering inside OS browser is completed.
https://github.com/JuliaPlots/Plots.jl/blob/master/src/backends/web.jl#L22-33

Maybe store all temporary files created by plots to pre-defined folder(`"$(tempdir())plots.jl\\"` ), then clean up that folder with `rm` at `__init__` will do the trick?

tkelman:
`__init__` doesn't seem like the right place for this. Are the tests really opening browser windows here?

PR #684, opened 2017-02-15
Set limit for number of temporary files

YongHee-Kim:
This is for #645 
I think problem is that `Plots.jl` can potentially create several hundred temporary files, not that really matters... but just messy.
And since Juno Plot Pane is planning to introduce [plot history feature](https://github.com/JunoLab/atom-julia-client/issues/122) sometime in the future, it makes sense to store 'some' temporary files and make those names predictable for Juno to handle plot history.

### DIFF
--- a/src/output.jl
+++ b/src/output.jl
@@ -117,7 +117,15 @@ function savefig(plt::Plot, fn::AbstractString)
 end
 savefig(fn::AbstractString) = savefig(current(), fn)
 
-
+function Base.tempname()
+    tempfile_limit = UInt32(30)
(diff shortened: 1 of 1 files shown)


tbreloff:
I don't see how I can merge this in any form.  You're overwriting a Base method, and putting an arbitrary limit on temporaries?  What if someone makes an animation out of a bunch of images?

There are certainly places that temporaries can be cleaned up better, but this isn't a good approach IMO.

YongHee-Kim:
I'm sorry it was really bad PR... I wasn't thinking enough

PR #690, opened 2017-02-21
base histograms on StatsBase

mkborregaard:
This is a rewrite of the histogram functionality that passes the calculation of the histograms (the binning, weights etc) to the histogram functionality in StatsBase.

This makes for more intuitive bin edges in histograms, with good alignment to x axis tick marks, and also for better consistency of histogram bins (across Julia and with e.g. R).

All the user-facing interface is unchanged.

### DIFF
--- a/REQUIRE
+++ b/REQUIRE
@@ -7,3 +7,4 @@ Reexport
 FixedSizeArrays
 Measures
 Showoff
+StatsBase
--- a/src/Plots.jl
+++ b/src/Plots.jl
@@ -9,6 +9,7 @@ using Base.Meta
 @reexport using PlotUtils
 @reexport using PlotThemes
 import Showoff
+import StatsBase
 
 export
     grid,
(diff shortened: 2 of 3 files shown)


mkborregaard:
It would be cool to merge this PR https://github.com/JuliaPlots/StatPlots.jl/pull/1/files (currently on StatPlots into this, they are very consistent).

mkborregaard:
@tbreloff would you appreciate having the functionality of supplying a symbol in the bins tuple (such as :exact) and have the bins calculated by the current `calc_edges`? E.g. `histogram(x, bins = (20, :exact))` ?

tbreloff:
Hmm. Feels a little too hacky I think, and might break some logic. Just go
with what you're doing

oschulz:
Close this, in favour of #713 (and add new changes on top of that one, later)?

mkborregaard:
good idea

Issue #675, opened 2017-02-09
grid should be Axis attribute, not Subplot


tbreloff:
... so that you can control horizontal/vertical grid lines separately

PR #695, opened 2017-02-22
Changing grid to Axis attribute from subplot

GuessWhoSamFoo:
#675 Grid is now an Axis attribute so that grid lines can be enabled by x,y axis separately.

`plot(Plots.fakedata(100,10),layout=4,palette=[:grays :blues :heat :lightrainbow],bg_inside=[:orange :pink :yellow :green])`
`plot!(xflip=[true false false false], xgrid=[true false true false], ygrid=[true true false false])`
grid1

`plot(Plots.fakedata(100,10),layout=4,palette=[:grays :blues :heat :lightrainbow],bg_inside=[:orange :pink :yellow :green])`
`plot!(grid=[true true false false])`
grid2

### DIFF
--- a/src/Plots.jl
+++ b/src/Plots.jl
@@ -50,8 +50,6 @@ export
     yflip!,
     xaxis!,
     yaxis!,
-    xgrid!,
-    ygrid!,
 
     xlims,
     ylims,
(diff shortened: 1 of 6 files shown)


tbreloff:
Have you tested this on other backends?  My guess is that this will break every other backend, but I haven't looked closely.

tbreloff:
But thanks for getting this started!

GuessWhoSamFoo:
Thats embarrassing! All the other backends did break. Getting Plotly to work was fairly straightfoward. I'm stuck on GR and InspectDr while trying to get more familiar with the code.

ma-laforge:
For InspectDR:

If I understand the changes correctly, you should be able to get things working by adding the following lines after `xaxis = sp[:xaxis]; yaxis = sp[:yaxis]`:

```
xgrid_show = xaxis[:grid]
ygrid_show = yaxis[:grid]

strip.grid = InspectDR.GridRect(
	vmajor=xgrid_show, vminor=xgrid_show,
	hmajor=ygrid_show, hminor=ygrid_show,
)
```

...But you probably have to use the `master` branches of both Plots.jl & InspectDR.jl to get things working.

PR #697, opened 2017-02-22
Attempt to add support for clims to plotly backend

lwabeke:
Added support for clims, seems to work on heatmap, surface and contour.
On surface zlims can still be set to be different from clims.
For heatmap, only clims work, zlims doesn't have any effect.


tbreloff:
Thanks!

Issue #693, opened 2017-02-22
Scatter plot legend marker duplicate with gr() backend


YongHee-Kim:
```
using Plots; gr()
scatter(eye(2), 1:2, label = ["a" "b"])
```
![image](https://cloud.githubusercontent.com/assets/7087745/23192364/61a0250c-f8e6-11e6-8ecf-d9c6b91a4359.png)

```
using GR
scatter(eye(2), 1:2, labels=["a", "b"])
```
![image](https://cloud.githubusercontent.com/assets/7087745/23192325/29c4700c-f8e6-11e6-853d-3f7410bf4299.png)

`GR` legend works fine so, I figured it's a bug from Plots.jl. And I'm planning to resolve this issue myself(If possible...)
But I found that there is [another gr backend is WIP](https://github.com/JuliaPlots/PlotsGR.jl) 

@tbreloff are you planning to change `GR` backend to `PlotsGR.jl`? and if that's the case, is it not necessary to resolve this?

tbreloff:
I'm confused. What's wrong with the output?

As for the backend questions, please see the new "contributing" page of the
docs

YongHee-Kim:
It's duplicated 

`plotlyjs()` backend 
![image](https://cloud.githubusercontent.com/assets/7087745/23193064/54ab1e20-f8ea-11e6-89c2-df313c96126d.png)

`gr()` backend
![image](https://cloud.githubusercontent.com/assets/7087745/23193071/5d4d9f80-f8ea-11e6-984c-ff773634eb80.png)

tbreloff:
That behavior was certainly intentional, so I don't think there's anything to fix.  Vote: does anyone think this should be changed to a single marker in the center?

mkborregaard:
I do. I have been confused by this as well. What is the purpose of duplicating the marker on gr?

joshday:
My vote is single marker, but my opinion hasn't been strong enough to bring it up before.

tbreloff:
Ok. I vote single marker too actually :)  who wants to change it?

joshday:
Assuming it's in gr.jl, can you give me a pointer/search term to find the relevant code?

mkborregaard:
try https://github.com/JuliaPlots/Plots.jl/blob/master/src/backends/gr.jl#L1020

mkborregaard:
I also vote for reducing the marker size

PR #698, opened 2017-02-22
Change GR label for scatterplot

joshday:
From conversation in #693.  This changes the GR scatterplot labels from two markers to one smaller marker.

screen shot 2017-02-22 at 10 42 26 am


tbreloff:
Thanks!

PR #705, opened 2017-02-25
fix hist for uneven

mkborregaard:


mkborregaard:
Don't merge this yet, as uneven bins are not properly supported by the uneven functions.

PR #713, opened 2017-03-01
Change histogram implementation, use StatsPlots, add new histogram st…

oschulz:
Change histogram implementation, use StatsPlots, add new histogram styles

Supports plotting 1D and 2D StatsBase histograms, seriestype can be set to bar(bins), scatter(bins) or step(bins).

```julia
using Plots, StatsBase

data1d = filter(x -> abs(x) < 2.5, randn(10000))
data2d = (randn(1000000), - 0.1 * log(rand(1000000)))

edge = -4:0.1:4
h1 = fit(Histogram, data1d, edge, closed = :left)
h1_norm = normalize(h1)

edge_func(x) = x + 0.1 * x^3
inhom_edge = map(x -> edge_func(x) / edge_func(4) * 4, -4:0.5:4)
h2 = float(fit(Histogram, data1d, inhom_edge, closed = :left))
h2_err = sqrt.(h2.weights)
h2_norm_density = normalize(h2, mode = :density)
h2_norm_pdf, h2_err_norm = normalize(h2, h2_err, mode = :pdf)

h2d = fit(Histogram, data2d, nbins = (80, 120), closed = :left)


plot(
    plot(
        h2_norm_density,
        fillrange = 0,
        fillalpha = 0.5,
        fillcolor = :red,
        # bar_width = 0.1
    ),
    plot(
        h2_norm_pdf,
        yerror = h2_err_norm,
        seriestype = :scatter,
        # xerror = 0.2
    ),
    plot(
        [h1_norm, h2_norm_pdf],
        xlims = (-3, 3),
        seriestype = :step
    ),
    plot(h2d),
    grid = false,
    size = (1024, 768)
)
```

hists-statsbase-demo

New series recipes for binned data:

* barbins
* scatterbins
* stepbins

New series recipes for histogram:

* barhist (histogram is now an alias for this)
* scatterhist
* stephist

```julia
plot(
    histogram(
        data1d,
        bins = inhom_edge,
        normalize = :density,
        fillrange = 0,
        fillalpha = 0.5,
        fillcolor = :red,
        # bar_width = 0.1
    ),
    scatterhist(
        data1d,
        bins = inhom_edge,
        normalize = :pdf,
    ),
    begin
        stephist(
            data1d,
            bins = edge,
            normalize = true,
            xlims = (-3, 3),
        )
        stephist!(
            data1d,
            bins = inhom_edge,
            normalize = :pdf,
            xlims = (-3, 3),
        )
    end,
    histogram2d(data2d..., bins = (80, 120)),
    grid = false,
    size = (1024, 768)
)
```

hists-modes-demo

Error bars currently don't work correctly for scatterbins and scatterhist, due to problem with manipulating error bars in a series recipe, but do work for `plot(h::StatsBase.Histogram, seriestype = :scatter)` (works around the problem by calling scatter directly, it seems that error bars can be manipulated correctly in a type recipe).

Also adds support for some common auto-binning modes (see http://en.wikipedia.org/wiki/Histogram#Number_of_bins_and_width and http://mathworks.com/help/matlab/ref/histogram.html):

* :sturges, :auto - Sturges' formula
* :sqrt - Square-root choice
* :rice - Rice Rule
* :scott - Scott's normal reference rule
* :fd - Freedman–Diaconis rule

Maybe these could be contributed to StatsBase at some point.

```julia
data_multigauss = vcat(randn(10000) - 2, randn(30000) + 2)

plot(
    histogram(data_multigauss, title = "Default binning"),
    histogram(data_multigauss, bins = :sturges, title = "bins = :sturges"),
    histogram(data_multigauss, bins = :sqrt, title = "bins = :sqrt"),
    histogram(data_multigauss, bins = :rice, title = "bins = :rice"),
    histogram(data_multigauss, bins = :scott, title = "bins = :scott"),
    histogram(data_multigauss, bins = :fd, title = "bins = :fd"),
    grid = false,
    size = (1024, 768)
)
```

autobining-modes-demo

@tbreloff, @mkborregaard, I hope you like it.

### DIFF
--- a/src/recipes.jl
+++ b/src/recipes.jl
@@ -409,7 +409,7 @@ end
 @deps scatterbins scatter
 
 
-_stepbins_path(edge, weights) = begin
+function _stepbins_path(edge, weights)
     nbins = length(linearindices(weights))
     if length(linearindices(edge)) != nbins + 1
         error("Edge vector must be 1 longer than weight vector")
--- a/src/recipes.jl
+++ b/src/recipes.jl
@@ -378,109 +378,256 @@ end
 end
 @deps bar shape
 
+
 # ---------------------------------------------------------------------------
 # Histograms
 
-# edges from number of bins
-function calc_edges(v, bins::Integer)
-    vmin, vmax = extrema(v)
-    linspace(vmin, vmax, bins+1)
+_bin_centers(v::AVec) = (v[1:end-1] + v[2:end]) / 2
+
+
+@recipe function f(::Type{Val{:barbins}}, x, y, z)
+    edge, weights = x, y
+    if (d[:bar_width] == nothing)
+        bar_width := diff(edge)
+    end
+    x := _bin_centers(edge)
+    y := weights
+    seriestype := :bar
+    ()
+end
+@deps barbins bins
+
+
+@recipe function f(::Type{Val{:scatterbins}}, x, y, z)
+    edge, weights = x, y
+    xerror := diff(edge)/2
+    x := _bin_centers(edge)
+    y := weights
+    seriestype := :scatter
+    ()
+end
+@deps scatterbins scatter
+
+
+_stepbins_path(edge, weights) = begin
+    nbins = length(linearindices(weights))
+    if length(linearindices(edge)) != nbins + 1
+        error("Edge vector must be 1 longer than weight vector")
+    end
+
+    it_e, it_w = start(edge), start(weights)
+    px, it_e = next(edge, it_e)
+    py = zero(eltype(weights))
+
+    npathpts = 2 * nbins + 2
+    x = Vector{eltype(px)}(npathpts)
+    y = Vector{eltype(py)}(npathpts)
+
+    x[1], y[1] = px, py
+    i = 2
+    while (i < npathpts - 1)
+        py, it_w = next(weights, it_w)
+        x[i], y[i] = px, py
+        i += 1
+        px, it_e = next(edge, it_e)
+        x[i], y[i] = px, py
+        i += 1
+    end
+    assert(i == npathpts)
+    x[end], y[end] = px, zero(py)
+
+    (x, y)
 end
 
-# just pass through arrays
-calc_edges(v, bins::AVec) = bins
+@recipe function f(::Type{Val{:stepbins}}, x, y, z)
+    edge, weights = x, y
 
-# find the bucket index of this value
-function bucket_index(vi, edges)
-    for (i,e) in enumerate(edges)
-        if vi <= e
-            return max(1,i-1)
+    axis = d[:subplot][Plots.isvertical(d) ? :xaxis : :yaxis]
+
+    xpts, ypts = _stepbins_path(edge, weights)
+    if !Plots.isvertical(d)
+        xpts, ypts = ypts, xpts
+    end
+
+    # create a secondary series for the markers
+    if d[:markershape] != :none
+        @series begin
+            seriestype := :scatter
+            x := Plots._bin_centers(edge)
+            y := weights
+            fillrange := nothing
+            label := ""
+            primary := false
+            ()
         end
+        markershape := :none
+        xerror := :none
+        yerror := :none
     end
-    return length(edges)-1
+
+    x := xpts
+    y := ypts
+    seriestype := :path
+    ylims --> [0, 1.1 * maximum(weights)]
+    ()
+end
+Plots.@deps stepbins path
+
+
+function _auto_binning_nbins{N}(vs::NTuple{N,AbstractVector}, dim::Integer; mode::Symbol = :auto)
+    _cl(x) = max(ceil(Int, x), 1)
+    _iqr(v) = quantile(v, 0.75) - quantile(v, 0.25)
+    _span(v) = maximum(v) - minimum(v)
+
+    n_samples = length(linearindices(first(vs)))
+    # Estimator for number of samples in one row/column of bins along each axis:
+    n = max(1, n_samples^(1/N))
+
+    v = vs[dim]
+
+    if mode == :sqrt  # Square-root choice
+        _cl(sqrt(n))
+    elseif mode == :sturges || mode ==:auto  # Sturges' formula
+        _cl(log2(n)) + 1
+    elseif mode == :rice  # Rice Rule
+        _cl(2 * n^(1/3))
+    elseif mode == :scott  # Scott's normal reference rule
+        _cl(_span(v) / (3.5 * std(v) / n^(1/3)))
+    elseif mode == :fd  # Freedman–Diaconis rule
+        _cl(_span(v) / (2 * _iqr(v) / n^(1/3)))
+    else
+        error("Unknown auto-binning mode $mode")
+    end
 end
 
-function my_hist(v, bins; normed = false, weights = nothing)
-    edges = calc_edges(v, bins)
-    counts = zeros(length(edges)-1)
+_hist_edge{N}(vs::NTuple{N,AbstractVector}, dim::Integer, binning::Integer) = StatsBase.histrange(vs[dim], binning, :left)
+_hist_edge{N}(vs::NTuple{N,AbstractVector}, dim::Integer, binning::Symbol) = _hist_edge(vs, dim, _auto_binning_nbins(vs, dim, mode = binning))
+_hist_edge{N}(vs::NTuple{N,AbstractVector}, dim::Integer, binning::AbstractVector) = binning
 
-    # add a weighted count
-    for (i,vi) in enumerate(v)
-        idx = bucket_index(vi, edges)
-        counts[idx] += (weights == nothing ? 1.0 : weights[i])
-    end
+_hist_edges{N}(vs::NTuple{N,AbstractVector}, binning::NTuple{N}) =
+    map(dim -> _hist_edge(vs, dim, binning[dim]), (1:N...))
 
-    counts = isapprox(extrema(diff(edges))...) ? counts : counts ./ diff(edges) # for uneven bins, normalize to area.
+_hist_edges{N}(vs::NTuple{N,AbstractVector}, binning::Union{Integer, Symbol, AbstractVector}) =
+    map(dim -> _hist_edge(vs, dim, binning), (1:N...))
 
-    # normalize by bar area?
-    norm_denom = normed ? sum(diff(edges) .* counts) : 1.0
-    if norm_denom == 0
-        norm_denom = 1.0
-    end
+_hist_norm_mode(mode::Symbol) = mode
+_hist_norm_mode(mode::Bool) = mode ? :norm : :none
 
-    edges, counts ./ norm_denom
+function _make_hist{N}(vs::NTuple{N,AbstractVector}, binning; normed = false, weights = nothing)
+    edges = _hist_edges(vs, binning)
+    h = float( weights == nothing ?
+        StatsBase.fit(StatsBase.Histogram, vs, edges) :
+        StatsBase.fit(StatsBase.Histogram, vs, weights, edges)
+    )
+    normalize!(h, mode = _hist_norm_mode(normed))
 end
 
 
 @recipe function f(::Type{Val{:histogram}}, x, y, z)
-    edges, counts = my_hist(y, d[:bins],
-                               normed = d[:normalize],
-                               weights = d[:weights])
-    bar_width := diff(edges)
-    x := centers(edges)
-    y := counts
-    seriestype := :bar
+    seriestype := :barhist
     ()
 end
-@deps histogram bar
+@deps histogram barhist
+
+@recipe function f(::Type{Val{:barhist}}, x, y, z)
+    h = _make_hist((y,), d[:bins], normed = d[:normalize], weights = d[:weights])
+    x := h.edges[1]
+    y := h.weights
+    seriestype := :barbins
+    ()
+end
+@deps barhist barbins
+
+@recipe function f(::Type{Val{:stephist}}, x, y, z)
+    h = _make_hist((y,), d[:bins], normed = d[:normalize], weights = d[:weights])
+    x := h.edges[1]
+    y := h.weights
+    seriestype := :stepbins
+    ()
+end
+@deps stephist stepbins
+
+@recipe function f(::Type{Val{:scatterhist}}, x, y, z)
+    h = _make_hist((y,), d[:bins], normed = d[:normalize], weights = d[:weights])
+    x := h.edges[1]
+    y := h.weights
+    seriestype := :scatterbins
+    ()
+end
+@deps scatterhist scatterbins
+
+
+@recipe function f{T, E}(h::StatsBase.Histogram{T, 1, E})
+    seriestype --> :barbins
+
+    st_map = Dict(
+        :bar => :barbins, :scatter => :scatterbins, :step => :stepbins,
+        :steppost => :stepbins # :step can be mapped to :steppost in pre-processing
+    )
+    seriestype := get(st_map, d[:seriestype], d[:seriestype])
+
+    if d[:seriestype] == :scatterbins
+        # Workaround, error bars currently not set correctly by scatterbins
+        xerror --> diff(h.edges[1])/2
+        seriestype := :scatter
+        (Plots._bin_centers(h.edges[1]), h.weights)
+    else
+        (h.edges[1], h.weights)
+    end
+end
+
+
+@recipe function f{H <: StatsBase.Histogram}(hv::AbstractVector{H})
+    for h in hv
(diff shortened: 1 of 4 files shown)


oschulz:
Depends on JuliaStats/StatsBase.jl#243.

See also JuliaPlots/StatPlots.jl#1 - should be merged at the same time.

mkborregaard:
Ha ha will take me a little to review all of this but it looks great. Maybe it could be an idea to split this into smaller logical units (the statsbase support, the normalizations, the new drawing method, the different binning methods) but we will take that when we get to it.

oschulz:
> Maybe it could be an idea to split this into smaller logical units

I'm usually all for modular commits, but this stuff has lot's of inter-dependencies, so I'm not sure it's worth splitting it. Well, have a look and let me know. :-)

mkborregaard:
It would also be nice to have explanations for bins etc in the arg_desc.jl file. And when it gets merged it would be great if you could put a little about this in PlotDocs :-)

src/recipes.jl
```diff
@@ -378,109 +378,256 @@ end
...
+    xerror := diff(edge)/2
+    x := _bin_centers(edge)
+    y := weights
+    seriestype := :scatter
+    ()
+end
+@deps scatterbins scatter
+
+
+_stepbins_path(edge, weights) = begin
```
tbreloff: Any reason you didn't use `function` here?

oschulz:
> It would also be nice to have explanations for bins etc in the arg_desc.jl file. And when it gets merged it would be great if you could put a little about this in PlotDocs

Yes, good point!

src/recipes.jl
oschulz: Sorry, no - force of habit, sometimes I write functions like that. I'll change it to conform with the Plots programming style.

oschulz:
@mkborregaard, a suggestion: If you're happy with this in principle, I suggest we merge it soon. It only goes into the new-histogram branch initially, anyhow. But that would give interested users can the chance to try it out (and maybe give us feedback), while we add things like arg documentation, etc. We have to wait for JuliaStats/StatsBase.jl#243 anyhow before this can go into Plots master.

src/recipes.jl
```diff
@@ -378,109 +378,256 @@ end
...
+
+    n_samples = length(linearindices(first(vs)))
+    # Estimator for number of samples in one row/column of bins along each axis:
+    n = max(1, n_samples^(1/N))
+
+    v = vs[dim]
+
+    if mode == :sqrt  # Square-root choice
+        _cl(sqrt(n))
+    elseif mode == :sturges || mode ==:auto  # Sturges' formula
```
mkborregaard: My preference would be to let the :auto mode relay to StatsBase

mkborregaard:
Good idea. I also think it looks nice from looking over it, and we'll have some time over the next days to look at it more closely before merging the branch. On top of my head it looks like you have my planned additions in there as well, so it may not need a lot more work at this point.
Thanks!

oschulz:
Thanks, Michael!

src/recipes.jl
oschulz: That's a good point. Also, currently we still use the built-in value of 30 bins when nothing is specified for `bins`. Should that default to `:auto` instead?

mkborregaard: yes

oschulz: Btw. the auto-binning in StatsBase uses Sturges' formula, so even though we're not currently using the StatsBase code for `:auto`, the result is the same. I'm kinda planning to contribute the auto-binning modes to StatsBase, but after the normalization stuff is merged and after we've played with them a bit more.

src/recipes.jl
```diff
@@ -378,109 +378,256 @@ end
...
 
+
 # ---------------------------------------------------------------------------
 # Histograms
 
-# edges from number of bins
-function calc_edges(v, bins::Integer)
-    vmin, vmax = extrema(v)
-    linspace(vmin, vmax, bins+1)
+_bin_centers(v::AVec) = (v[1:end-1] + v[2:end]) / 2
```
mkborregaard: This should use the element-wise plus

mkborregaard: Also, is there any reason for removing the `centers` function (below) and defining a new `bin_centers` function? This seems unnecessary.

src/recipes.jl
mkborregaard: Yes, they do belong in StatsBase IMHO.

src/recipes.jl
```diff
@@ -378,109 +378,256 @@ end
...
 
-    edges, counts ./ norm_denom
+_hist_edges{N}(vs::NTuple{N,AbstractVector}, binning::NTuple{N}) =
+    map(dim -> _hist_edge(vs, dim, binning[dim]), (1:N...))
+
+_hist_edges{N}(vs::NTuple{N,AbstractVector}, binning::Union{Integer, Symbol, AbstractVector}) =
+    map(dim -> _hist_edge(vs, dim, binning), (1:N...))
+
+_hist_norm_mode(mode::Symbol) = mode
+_hist_norm_mode(mode::Bool) = mode ? :norm : :none
```
mkborregaard: In StatsBase, I conceded that philosophically you may argue for different normalizations. In Plots, I strongly feel that it should choose a normalization appropriate for naive plotting. That is `:none` (or perhaps better named `:counts`) as the default for even bins, `:density` for uneven bins, and `:pdf` for both if `mode == true`. `:norm` should never be a default but available on request.
You could argue that it can confuse a user that the default depends on input bins. That's an important point, but I think that could be adressed by automatically generating a ylabel (if not supplied to `plot`), saying e.g. "frequency" or "counts" for `:none`, "bin-normalized counts" for `:density` and `"normalized frequency"` for :pdf.

You may

oschulz: I'm more worried now that it'll be obvious that Plots is based on StatsBase, users will get confused if they behave differently. I do see your point about "naive plotting" - on the other hand, will a naive user really make histograms with inhomogenous binning? In general, I'm in favor of consistency, and so Plots should behave like StatsBase. If we really really need it to behave differently, at least we should document the differing default (which would then be `:pdf`, as `:density` is not a strict normalization) very well. @tbreloff , what do you think - `:norm` like in StatsBase as the default, or `:pdf`?

One thing is very important, though - Plots should *never automatically* apply *any normalization* (the user must at least set `normalize = true`), even for inhomogenous binning. Doing otherwise would mean plotting completely different things just because edges may be set one way or the other (all normalizations result in the y-axis having different physical units than counts).

oschulz:
> be very nice get the arg_desc file updated now [...] Really fine work!

Thanks! I'll see that I can get on the arg_desc as soon as possible (busy day, today).

src/recipes.jl
oschulz: > Ok, then I suggest we leave that part as it is, for now, try and tune the auto-bin modes for a while (and maybe even add a few others, there's some recent literature on the subject), and then make a PR for StatsBase when it's mature and has seen some field-testing. I think we'll have a faster turn-around developing the auto-binning modes in Plots than in StatsBase, initially.

src/recipes.jl
oschulz: > Also, is there any reason for removing the centers function (below) and defining a new bin_centers function? This seems unnecessary

`centers` is only used here, and it's not exported - so I wanted to clearly mark it as an internal function and make the name a bit more descriptive. I don't think would be obvious to someone new who just reads `centers(v)` what the function does.

oschulz: > This should use the element-wise plus

Not necessarily, as vector-addition is always element wise - it's different from explicitely defining functions like `sin` for vectors, which is not necessary anymore since Julia has the dot-notation. Of course, in a more complex  expression, `.+` may still have advantages due to fused broadcasting.
In this specific case, though, there was actually a reason why I chose a simple `+`: If you use the current implementation on a `Range` (which is what edges typically are), the result will again be a `Range` - so it's lightweight and fast. If you use `.+`, the result will always be a `Vector`. Not that a bit of memory allocation would be a problem for Plots (performance is obviously limited by other things), but I think it's still elegant.

mkborregaard: OK

src/recipes.jl
mkborregaard: I don't think we should be locked to doing what StatsBase does. I also don't agree with your last point - Plots should apply the :density normalization for inhomogenous bins and mark it clearly in the ylabels. I am quite sure I am not going to change my mind on this, unless @tbreloff rules differently.

tbreloff:
I defer to you guys. You're thinking about it a lot more than me.

src/recipes.jl
oschulz: > I don't think we should be locked to doing what StatsBase does.

I'd very much prefer to conform with StatsBase, but I could live with a different default if we really must do things differently for Plots - as long as it's clearly marked. @tbreloff ?

> I also don't agree with your last point - Plots should apply the :density normalization for inhomogenous bins and mark it clearly in the ylabels.

I'm sorry, but I have to vehemently disagree here. For once the meaning of the plot (and it would be changed) shouldn't depend on axis binning - such auto-magical things tend to be very intransparent to the user, and hence very dangerous. People may spend hours trying to find out why they get counts in some plot and a density on another, just because the edges are different (users tend not to read the full docs first, esp. since Plots is so easy to use). You also can't automatically produce correct y-labels, because Plots can't know what the x-axis units are. And even if you could, automatically setting ylabels and changing y-units depending on how your edges are structured is just, uh, weird.

2d-histgrams are even more problematic - just looking at the colors, the user may never realize that he's not plotting what he thought he was (and where do you put the automatic z-axis label, even if you'd know what to put in it?). And that if just one axis is inhomogenous? In any case, such non-obvious automatic behavior changes could, in the worst case, lead to incorrect scientific results or interpretations.
 
 I stand by my point - inhomogenous axes don't happen by accident, and they're not that common. People who use them will usually know what they want - having plots suddenly change the meaning of the bin values will make their life harder, not easier. Such histograms simply have more complex behaviour than homogenously-binned ones - trying to "hide" that mathematical reality from the users won't do them any favours.

mkborregaard: I know we disagree :-) I am not talking about doing something magical with y-ticks, just let the y-label tell the user what is done.
Our difference is due to different views of what a histogram is. I think of it as primarily a visualization of the underlying statistical distribution. You may think of it as primarily a bin-counting technique.
In the literature, uneven bins are always presented as densities (e.g. http://www.esaim-ps.org/articles/ps/abs/2009/01/ps0721/ps0721.html ) as this is the appropriate visualization. Your best bet at convincing me is to find papers using uneven bins without density normalization. Even statistical packages that don't do that automatically tell you you should. Any advanced user can just supply :counts if he prefers.

mkborregaard: You may also like e-a histograms and dhists (http://www.tandfonline.com/doi/abs/10.1198/jcgs.2009.0002)? But they are probably more relevant for StatPlots, if at all :-)

oschulz: > Our difference is due to different views of what a histogram is. I think of it as primarily a visualization of the underlying statistical distribution. You may think of it as primarily a bin-counting technique

It can be either, of course. However, I think it's pretty undisputed that for "normal" histograms (homogenous edge(s)) the vast majority of users will expect to see counts on the y-axis. Are there any major software packages that automatically normalize homogenously binned histograms by default? Naturally, for inhomogenous bins, plotting counts will often not make sense  (except for cross-checks, etc.). One will usually want to plot a PDF estimate or density - but Plots.jl has no way to guess which of those the user wants. And if the user doesn't know what he wants, he can't interpret the results, anyhow. My argument is not that hists with inhomogenous bins shouldn't be normalized, but that it shouldn't be done automatically.

> Your best bet at convincing me is to find papers using uneven bins without density normalization

In my opinion, this has nothing to do with statistical theory (like I said, of course the user should normalize in some fashion in these cases). It's a question of software design. In general, the behavior of Plots (or any other package) should be predictable for the users - automatically plotting either counts or density goes against that.

There's also another problem: How would we reliably detect inhomogenous edges? The edge definition might very well come from a source with limited numerical precision, e.g. a text file. Now where do we set the numerical threshold for "bins are of equal width"? If a user supplies bin edges directly, instead of using a range, the choice of counts vs. density would really become random, depending on the precision of his values.

tbreloff:
I don't know whose side means what, but... unless one specifies"density"
the heights of a histogram should be the counts within each bin. Anything
else would be confusing, and should be called something else. (IMO)

oschulz:
I agree.

mkborregaard:
Argh.

mkborregaard:
I suppose that means I am suddenly in a minority :-p

mkborregaard:
The last bulwark I will defend with my nails and teeth is that normalized histograms should show densities (method :pdf). No norm-normalization giving inhomogenous bins but summing to 1. I will not be swayed by arguments relating to idempotency.
Also, this behaviour would make the behaviour align with the matlab and python implementations. My preference was the R implementation, reflecting perhaps that R is my native language.

oschulz:
> The last bulwark I will defend with my nails and teeth is that normalized histograms should show densities (method :pdf).

Oh well. :-) Ok, so we make `normalize = true` mean `normalize = :pdf`? So be it.

> I will not be swayed by arguments relating to idempotency.

No, that's no relevant for plots, as users obviously can't run the normalization twice. I was only worried about consistency with StatsBase, but if it's that important to you, I can live with it.

I would *not* put any automatic labels on the y-axis though! Since the user selected `normalize` he already knows it's not counts - but it's the user's job to put proper labels on the axes. After all, we don't do automatic axis label for any other type of plot.

> would make the behaviour align with the matlab and python implementations

That's certainly not a bad choice.

mkborregaard:
Good, let's say that then. I accept your point about the y axis :-)

oschulz:
Ok! :-)

So we just do

    _hist_norm_mode(mode::Bool) = mode ? :pdf : :none

right? Easy enough.

mkborregaard:
Yes, and I'd suggest not including :norm at all for clarity (or at least not documenting it - people who know and love it from StatsBase will still try it probably).

oschulz:
> Yes, and I'd suggest not including :norm at all for clarity

Well, let's not filter it out explicitly (after all, we just forward the  symbol to StatsBase, and some people may want it, like you say), but I'm fine with not documenting it.

src/recipes.jl
mkborregaard: I'll just leave this ref here for future reference. I'd like to add support for especially Wand http://www.esaim-ps.org/articles/ps/abs/2009/01/ps0721/ps0721.html

oschulz:
hey, @mkborregaard - had a very busy week, next one will be too. But it'll try to get on the improvements and doc additions we talked about as soon as I can.

mkborregaard:
That's great, no hurry, we are waiting on the JuliaStats guys, right?

mkborregaard:
@oschulz Congrats on the merge on StatsBase! Could you make a new PR on master by copying new_histogram to your own fork and rebasing on master?

oschulz:
Yes, I'll also have to make a few minor changes - should be quick, though.

oschulz:
About `normalize` and "arg_desc.jl": Should we keep normalize as `Bool` or `Symbol` with default value `false`, or should be deprecate bool values and make `:none` the official default?

mkborregaard:
Let's keep Bool with false as default, and make true equal the full normalisation (pdf)

oschulz:
Adapting "args.jl" to add the new histogram series recipes. Does `:density` still exist? It's still listed in `_allTypes` and also appears in "unicodeplots.jl".

oschulz:
@mkborregaard, @tbreloff: In `_expand_subplot_extrema` (in "pipeline.jl"): Should `:barhist`, `:barbins`; `:stephist`, etc. be added here, next to `:histogram` and `:histogram2d`? I don't quite understand what happens there - as histograms have been recipes for a while now (standard bar-style histograms implemented via `:shape`) the `!(st in (:pie, :histogram, :histogram2d))` seems to be `true`, always, anyhow.

oschulz:
@mkborregaard , @tbreloff : Similar question - in `should_add_to_legend` (in "subplots.jl"), should `:bins2d` be added next to `:histogram2d`? As both `:bins2d` and `:histogram2d` are implemented as recipes, using `:heatmap`, do they need to be listed explicitly here?

mkborregaard:
`density` has been moved to StatPlots but it should be kept in "args.jl".

mkborregaard:
I don't see the issue of adding :bins2d etc in the two places, for consistency. But I haven't looked carefully at whether it is really needed.

oschulz:
Thanks! While I'm at it, I'm also fixing log-scale for the histogram series types. Bit more involved than expected, but I'm nearly there.

oschulz:
Ok, here goes: #787

Issue #714, opened 2017-03-01
Recipe for more "natural" `TimeType` xtick positions


daschw:
Hi @tbreloff and @mkborregaard, I wrote a recipe to choose more "natural" (in my opinion) xtick positions if the first input to plot() is a `Date` or `DateTime` array. More "natural" means that full seconds/minutes/hours/days/months/years are chosen (depending on the time span) - see the figures below for comparison:

standard:
![std_ticks](https://cloud.githubusercontent.com/assets/16589944/23474844/757a17e8-feb5-11e6-8507-fe6c25ccfe2a.png)

with `TimeType` recipe:
![opt_ticks](https://cloud.githubusercontent.com/assets/16589944/23474855/7f731b50-feb5-11e6-879e-ae4ef73b8f34.png)

I'm not sure if this is  of interest/use for anyone. If it is, where would be the best place to put the code?

tbreloff:
This is cool!  If you post the code I might have a better idea of where it would fit.  I get the feeling this can't be added to the existing type recipes.

daschw:
```
using Base.Dates, RecipesBase

# Round DateTime dt to the next "round" DateTime with respect to (Period)Type P
function _round_up(dt::DateTime, P::Type)
    period_hierarchy = [Millisecond, Second, Minute, Hour, Day, Month, Year]
    i = findfirst(period_hierarchy .== P)
    DateTime((PH(dt) for PH in period_hierarchy[end:-1:i])...) + P(1)
end

function _get_ticks(x_min::DateTime, x_max::DateTime, P::Type, steplength::Integer)
    period_hierarchy = [Millisecond, Second, Minute, Hour, Day, Month, Year]
    i = findfirst(period_hierarchy .== P)
    showtype = i >= 5 ? Date : DateTime
    ticks = _round_up(x_min, P):P(steplength):x_max
    return Integer.(ticks), string.(showtype.(ticks))
end

function _get_ticks(x_min::DateTime, x_max::DateTime)
    Δt = x_max - x_min
    # number of ticks is not very flexible and may be set here:
    m = 4 # expected number of ticks
    n = 2 #
    if Δt > Day(365n)
        P = Year
        steplength = Δt / (m * Millisecond(Day(365)))
    elseif Δt > Day(30n)
        P = Month
        steplength = Δt / (m * Millisecond(Day(30)))
    elseif Δt > Day(n)
        P = Day
        steplength = Δt / (m * Millisecond(Day(1)))
[...]

PR #720, opened 2017-03-06
Added logic to choose *round* DateTime tick positions

daschw:
implementation of https://github.com/JuliaPlots/Plots.jl/issues/714
requires https://github.com/JuliaPlots/PlotUtils.jl/pull/8

tested with:

```
using Plots

hourrange = DateTime(2017, 2, 22, 9, 31):Dates.Minute(15):DateTime(2017,2,22,12)
dayrange = DateTime(2017, 1, 30, 9):Dates.Minute(15):DateTime(2017, 2, 2, 14)
weekrange = DateTime(2017, 1, 30, 9):Dates.Minute(15):DateTime(2017, 2, 8, 14)
monthrange = DateTime(2016, 4, 1):Dates.Minute(15):DateTime(2017, 1, 30, 9)
yearrange = DateTime(2017):Dates.Day(1):DateTime(2018)-Dates.Day(1)

hourdata = randn(length(hourrange))
daydata = randn(length(dayrange))
weekdata = randn(length(weekrange))
monthdata = randn(length(monthrange))
yeardata = randn(length(yearrange))

@time plot(hourrange, hourdata)
@time plot(dayrange, daydata)
@time plot(weekrange, weekdata)
@time plot(monthrange, monthdata)
@time plot(yearrange, yeardata)
```
with `gr()` and `pyplot()`

### DIFF
--- a/src/axes.jl
+++ b/src/axes.jl
@@ -156,6 +156,13 @@ function optimal_ticks_and_labels(axis::Axis, ticks = nothing)
     scale = axis[:scale]
     sf = scalefunc(scale)
 
+    # If the axis input was a Date or DateTime use a special logic to find
+    # "round" Date(Time)s as ticks
+    # TODO: maybe: non-trivial scale (:ln, :log2, :log10) for date/datetime
+    if axis[:formatter] in (dateformatter, datetimeformatter) && scale == :identity
+        return optimize_datetime_ticks(amin, amax; k_min = 2, k_max = 4)
+    end
+
     # get a list of well-laid-out ticks
     scaled_ticks = if ticks == nothing
         optimize_ticks(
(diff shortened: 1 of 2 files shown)


mkborregaard:
lgtm

tbreloff:
This doesn't seem right.  You're bypassing most of this function.  What if someone sets `xticks`?  I think it would still do this function (which is wrong).  This should **only** be a replacement for the `optimize_ticks` method.

daschw:
Of course you are right @tbreloff! Sorry!
I did this, because I thought it would make sense if the label format (`DateTime` or `Date`) is chosen based on the tick positions and not based on the input format. (If the input format was `DateTime` but the time span in the plot is over several days I would display `Dates` rather than `DateTimes`). Thus, I had `optimize_datetime_ticks` return ticks and labels.
But of course it's wrong to just ignore and overwrite ticks set by the user.
Would adding `&& ticks == nothing` to the condition be ok?

tbreloff:
> Would adding && ticks == nothing to the condition be ok?

Hmm.  It might be for this specific case.  But I don't love too much special casing.  This method handles hundreds of different attribute combinations generically.  But I guess there are few people that can really understand what the hell I did there. :)

Sure... I'll merge if you add that.

daschw:
>  But I don't love too much special casing.
Sure, I can totally understand that and if you prefer I can also think about how to integrate this better in your existing structure rather than bypassing most of the function

daschw:
ok

Issue #704, opened 2017-02-24
How to change grid appearance in pyplot()?


PedroSapichu:
Hello all!,

I can't figure out why I always get straight lines instead of dashed lines for the grids with pyplot() backend. To the left I show the plot that's used as an example at juliaplots.github, and to the right is the plot I get with the same code (and any other plotting command). With other backends I get the normal dashed line that looks much better.

Thanks!

![grid](https://cloud.githubusercontent.com/assets/25947175/23324142/219ed8e2-faba-11e6-89f2-ba1674fbfb7b.png)

jgbos:
I'll add that I'm trying to figure this out too.  It only happened recently and I'm not sure if it was `conda update` or `Pkg.update` that caused it.  For now I just initialize with `pyplot(grid=false)`.

xiaoweiz:
Use the keyword arg `foreground_color_grid` and assign it to `lightgrey` may alleviate the issue, e.g. 
```julia
plot(x, y, foreground_color_grid=:lightgrey)
```
But I couldn't find a way to change the `linestyle` of the grid to be `dot`.

jgbos:
There doesn't seem to be a keyword to set the grid style.  Here is what you can do.

Set defaults for grid by setting the following values in your `matplotlibrc` file:

```
### GRIDS
grid.color       :   d3d3d3    # grid color
grid.linestyle   :   :         # solid
grid.linewidth   :   0.5       # in points
grid.alpha       :   0.5       # transparency, between 0.0 and 1.0
```

I'm using matplotlib 2.0.  It appears the default `matplotlibrc` file has the grid lines being solid.

xiaoweiz:
@jgbos I'm also using matplotlib 2.0 but its default appearance on my system has no grid.

mkborregaard:
There is a PR to fix here https://github.com/JuliaPlots/Plots.jl/pull/721

mkborregaard:
Closed by #721 but see the discussion there for a recommended enhancement.

PR #721, opened 2017-03-07
set dotted grid lines for pyplot

mkborregaard:
Some people asked for the pyplot grid to be dotted again after the change to matplotlib 2.0. This does that. (It is still black, though, most other backends seem to have it grey).

### DIFF
--- a/src/backends/pyplot.jl
+++ b/src/backends/pyplot.jl
@@ -1059,7 +1059,7 @@ function _before_layout_calcs(plt::Plot{PyPlotBackend})
             end
             if sp[:grid]
                 fgcolor = py_color(sp[:foreground_color_grid])
-                pyaxis[:grid](true, color = fgcolor)
+                pyaxis[:grid](true, color = fgcolor, linestyle = ":")
                 ax[:set_axisbelow](true)
             end
             py_set_axis_colors(ax, axis)


diegozea:
Is It possible to make an optional keyword argument for controlling this (default to dotted line)?

mkborregaard:
You could try to implement this in a PR, if @tbreloff is ready to add another keyword attribute. I never use grid so I'd think of this as non-essential.

mkborregaard:
cf https://github.com/JuliaPlots/Plots.jl/issues/704

mkborregaard:
Making a keyword for the gridlinestyle would have to be done individually on each backend. If you or someone else makes a PR on this, it'd be cool to fix https://github.com/JuliaPlots/Plots.jl/issues/675 at the same time.

tbreloff:
The "correct" approach would be to let grid accept a Plots.Stroke, and to
override components of the default Stroke during preprocessing when the
user passes in keywords like "gridstyle".

mkborregaard:
The correct approach to the entire issue, or just @diegozea 's request?

tbreloff:
Well, both? This should be the solution for all the axis lines. But someone
has to take the time to do it...

mkborregaard:
Feeling inspired @diegozea ? You always make so good PRs :-)

PR #731, opened 2017-03-14
fix plotly error on marker_z

### LINKED ISSUE #730: plotlyjs colormap problem with spy
Hi, somehow, the following very simple command does not work under plotlyjs as a backend: ```spy(eye(10), marker=(:circle,10))```. The pyplot backend displays it correctly and the gr backend also displays it with somewhat larger dots than 10 points. The plotlyjs backend doesn't display it at all and generates the error: ```ERROR: MethodError: no method matching getindex(::ColorTypes.RGBA{Float64}, ::Float64) ...``` If I specify a colormap explicitly, e.g., ```spy(eye(10), c=:viridis, marker=(:circle,10))```, then it doesn't generate error, but the displayed figure is wrong; I cannot see the expected dots (or circles). What's wrong here with plotlyjs?

mkborregaard:
fixes #730
Creates a color scale when marker_z is set in a series recipe

### DIFF
--- a/src/backends/plotly.jl
+++ b/src/backends/plotly.jl
@@ -409,6 +409,13 @@ plotly_surface_data(series::Series, a::AbstractVector) = a
 plotly_surface_data(series::Series, a::AbstractMatrix) = transpose_z(series, a, false)
 plotly_surface_data(series::Series, a::Surface) = plotly_surface_data(series, a.surf)
 
+#ensures that a gradient is called if a single color is supplied where a gradient is needed (e.g. if a series recipe defines marker_z)
+as_gradient(grad::ColorGradient, α) = grad
+as_gradient(grad, α) = cgrad(alpha = α)
+
+# allows passing a ColorGradient where a single color is expected - the other backends allow this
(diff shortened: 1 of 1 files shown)


src/backends/plotly.jl
```diff
@@ -409,6 +409,13 @@ plotly_surface_data(series::Series, a::AbstractVector) = a
 plotly_surface_data(series::Series, a::AbstractMatrix) = transpose_z(series, a, false)
 plotly_surface_data(series::Series, a::Surface) = plotly_surface_data(series, a.surf)
 
+#ensures that a gradient is called if a single color is supplied where a gradient is needed (e.g. if a series recipe defines marker_z)
+as_gradient(grad::ColorGradient, α) = grad
+as_gradient(grad, α) = cgrad(alpha = α)
+
+# allows passing a ColorGradient where a single color is expected - the other backends allow this
+PlotUtils.rgba_string(cg::ColorGradient) = rgba_string(cg[1])
```
mkborregaard: This may fit better in PlotUtils, but it feels a little like a hack that only plotlyjs uses

Issue #739, opened 2017-03-18
default font size too small on pyplot backend?


floswald:
i just updated to matplotlib 2 and found that the default font on plots is very small. how can I change that?

```julia
pyplot()
Plots.PyPlotBackend()

julia> Plots.plot(rand(10),linewidth=2)
```
![figure_1-2](https://cloud.githubusercontent.com/assets/1021558/24072172/095cc520-0be3-11e7-9972-054a9aaeb4ad.png)

for comparison, calling PyPlot directly gives

```julia
julia> using PyPlot
julia> PyPlot.plot(rand(10),linewidth=2)
```
![figure_1-3](https://cloud.githubusercontent.com/assets/1021558/24072190/63642f7c-0be3-11e7-96ac-3cc86a32bdf3.png)

ma-laforge:
Just as a comment:

 - The default font sizes generated by Plots.jl using pyplot() is not meant to match the ones generated directly by using PyPlot.
 - Instead, the font sizes should be calibrated approximately match irrespective of which backend you use (pyplot, gr, plotly, pgfplots, inspectdr, ...)

If you don't like the current size of your plots, you can scale all of the defaults using the solution developed here https://github.com/JuliaPlots/Plots.jl/issues/560:

```
Plots.scalefontsizes(1.5) #Or some other factor
```
That will apply a `*=1.5` to all of the font sizes.

Alternatively, you could use the following solution:

```
fnt = Plots.font("Helvetica", 10.0)
default(titlefont=fnt, guidefont=fnt, tickfont=fnt, legendfont=fnt)
```

floswald:
Ok I see! Thanks!

mkborregaard:
I still think that we should adjust the standard font sizes in the pyplot backend after that updated to matplotlib 2.0. It is an open question how to do it (i.e. should default dpi be changed) - but @ma-laforge is on the case https://github.com/JuliaPlots/Plots.jl/issues/733

mkborregaard:
But pyplot doesn't look quite as small on my computer (as you can see on #733 )

floswald:
That's of course saying much better what I meant.
+1

floswald:
True it does not! Your version is almost ok for me.

ma-laforge:
I thought the same at first... until I looked at your plot a bit closer @floswald.

Both plots are rendered using fonts that are 8 pixels high....

[...]

Issue #555, opened 2016-10-29
Recipes cannot set default fonts


ChrisRackauckas:
In order to set default fonts, one needs to do something like:

``` julia
  xtickfont --> font(11)
  ytickfont --> font(11)
  legendfont --> font(11)
  guidefont  --> font(11)
```

However, `font` is in Plots and not RecipesBase, so changing the default font requires a Plots dependency.

tbreloff:
I've had this thought from other perspectives.  Really the "correct" solution is to have a bunch of attributes: guidefontsize, guidefontfamily, etc... and then guidefont would be a way to set all of those at once, similar to `marker` or `line`.  The downside is that it's a lot of new attributes to maintain/access, and the upside is that this issue wouldn't exist.  :)

ChrisRackauckas:
I think that's the right way to go since `font` doesn't really fit in with the way the other attributes work.

ma-laforge:
A few additional coments:

 1. There should be a way to scale all fonts simultaneously.  Possibly adding the following to the `~/.juliarc.jl` file:
```
PLOTS_DEFAULTS = Dict(:fontadjust => 1.5)
```
 2. It would be nice if this attribute was also available "@ runtime":
```
plt = plot(...)
display(plt) #Want smaller font to maximize real-estate

plot!(size=(1000,800), fontadjust=2.2) #Make more readable on print.
png("morereadbleplot.png")
```

juliohm:
This is a recurring issue and a very nice feature to have. Thank you for reporting it @ChrisRackauckas.

mkborregaard:
All that's needed is for someone to implement this change in practice:
-[ ] unexport the Font type (it can still be used internally)
-[ ] add the new keywords
-[ ] updating the handling of keywords to form the Font object
-[ ] specify a new 'font' magic keyword

ChrisRackauckas:
This is great to see! Thanks!

PR #740, opened 2017-03-19
Fix small issues with response to :dpi value.

ma-laforge:
Now line widths, etc are in pixels (like I assume was the intent for Plots.jl).

:dpi now only affects font sizes (because they are in points - not pixels)...  And grid line widths, I guess... because we don't seem to do the size mapping for those.

...But the main thing that :dpi changes is the size (in inches) of the image metadata...

As with other backends (presumably), the user must use the `size=(w,h)` keyword to control the saved image dimension.

### DIFF
--- a/src/args.jl
+++ b/src/args.jl
@@ -232,7 +232,7 @@ const _plot_defaults = KW(
     :html_output_format          => :auto,
     :inset_subplots              => nothing,   # optionally pass a vector of (parent,bbox) tuples which are
                                                # the parent layout and the relative bounding box of inset subplots
-    :dpi                         => DPI,        # dots per inch for images, etc
+    :dpi                         => DEFAULT_DPI, # dots per inch for images, etc
     :display_type                => :auto,
     :extra_kwargs                => KW(),
 )
--- a/src/backends/pyplot.jl
+++ b/src/backends/pyplot.jl
@@ -327,9 +327,12 @@ py_extents(obj) = obj[:get_window_extent]()[:get_points]()
 
 # compute a bounding box (with origin top-left), however pyplot gives coords with origin bottom-left
 function py_bbox(obj)
-    fl, fr, fb, ft = py_extents(obj[:get_figure]())
+    const fig = obj[:get_figure]()
+    const dpi = fig[:dpi]
+    const LENGTH_PIXEL = LENGTH_POINT*(72/dpi)
+    fl, fr, fb, ft = py_extents(fig) #malaforge: reads GUI size instead of *intended* size?
     l, r, b, t = py_extents(obj)
-    BoundingBox(l*px, (ft-t)*px, (r-l)*px, (t-b)*px)
+    BoundingBox(l*LENGTH_PIXEL, (ft-t)*LENGTH_PIXEL, (r-l)*LENGTH_PIXEL, (t-b)*LENGTH_PIXEL)
 end
 
 # get the bounding box of the union of the objects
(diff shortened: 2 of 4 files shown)


ma-laforge:
Here is a bit of sample code you can use to see how PyPlot now responds:

```
y = rand(10)
x = 1:length(y)
y = ones(x) #Overwrite
y[1] = 0; y[end] *= 2

using Plots
pyplot()
default(overwrite_figure=false)

for dpi in [50, 72, 300]
	@show dpi
	global p = Plots.plot(x, y, linewidth=2, dpi=dpi)
	display(p)
	fname = "pyplot_dpitests_$dpi.png"
	png(fname)
end
```

Notice that the plot line is fixed @ 2 pixels irrespective of the value of dpi....

...But since we don't rescale the grid line widths, those do vary with DPI...

I'm not sure what the point of the DPI setting was in the first place - because most of the input values to Plots.jl are in pixels - so changing the DPI value should simply change the physical image size in the metadata (Ignoring, of course the font sizes - which are typically specified in "points").

mkborregaard:
I think the ability to change the look of the plot with a single call to dpi (as. eg. here https://github.com/JuliaPlots/Plots.jl/issues/739) is a good reason for having it  - would that ability not be lost with this PR?

ma-laforge:
@mkborregaard: This PR will probably do what you want it to:

Fonts will get bigger as you change DPI because I don't actually scale fonts (ie font sizes are in "points" - not pixels).

**However** (and it's a big however :) )

I don't think that we actually want this to be the behaviour.  Instead, we should be implementing what is described in issues:
 - DPI proposal: https://github.com/JuliaPlots/Plots.jl/issues/742
 - Set default fonts/sizes: https://github.com/JuliaPlots/Plots.jl/issues/555

(That said, I have no objection to merging in this solution for now).

ma-laforge:
New implementation now matches much better expected behaviour described in Issue https://github.com/JuliaPlots/Plots.jl/issues/742.

Missing:
 - DPI setting affects both on-screen and saved plots (not certain how to re-structure plots for this).

That said, I don't see this re-structuring to happen any tie soon.

Ready for merge, pending review...

ma-laforge:
@mkborregaard: This new implementation should make it easier to get what you wanted.

For example, you could add the following to your `.juliarc` file:
```
pscale = 2
PLOTS_DEFAULTS = Dict(:size=>(600/pscale,400/pscale), :dpi=>72*pscale)
```

If I understood all the automatic (strange) re-interpretations happening in the background, this should make plots look the same on the retina display as it does with other displays.

To make lines & fonts larger relative to the plot size, you can further increase `pscale`.  Using integer values of `pscale` will result in whole (non-fractional) pixels being used.

ma-laforge:
@mkborregaard: Though I got the plot ouput to look good, the matplotlib GUI seems to ignore the plot size values.

Do you have any input here?
 - Output .png looks good.
 - On-screen GUI is displayed with any size it sees fit (user must manually re-size).

mkborregaard:
I'll be unable to work on julia now and for the next three weeks, but maybe @tbreloff or @pkofod can have a look

ma-laforge:
FYI: These changes altered the definition of "pixel" to something that better matches how images should be defined.

## To all other backend maintainers:

Your layout code might have to be tweaked to adjust for the new size/meaning of "pixel".

mkborregaard:
To all open PRs: Due to the change to 0.6 it appears that all PRs will have to be rebased onto master for Travis tests to run. - also, @ma-laforge sorry for delaying looking at this. What is the current status? Can this be merged regardless of our larger dpi-scaling-discussion?

ma-laforge:
Status:
This PR actually does match the intent on the larger dpi-scaling discussion.

In other words: It is ready to merge.

Just tell me if I need to rebase this first - and if yes: Does it need to be a new PR, or will a `push --force` propagate correctly back to this PR right here?

ma-laforge:
Oh, I get it, I DO have to rebase.  Pending...

ma-laforge:
Ok rebase done.  Everything appears to work as expected.

Travis fails: the :dpi behaviour was changed - meaning the reference images no longer match.

ma-laforge:
*Bump.*  I just want to know if there is something that must be done to maintain working tests.

Technically, I should be updating the images in PlotReferenceImages.jl - but since I cannot do both in a single commit... the checks ***will*** fail.

mkborregaard:
I'm not adressing this because I think we need to have the whole dpi discussion first. This is something where opinion is divided also within Plots, and I don't have the time to take it now, but we'll revisit for sure. 
The approach is to update PlotReferenceImages on your own fork, verify that tests pass, then submit a PR to PlotReferenceImages first remarking that it's linked to this one.

ma-laforge:
> @mkborregaard [...] we need to have the whole dpi discussion first [...]

Fair enough.  I Just did not know what the status was.  I will try to keep an ear out for when the time comes.

&rArr; At least we have a alternative working model for the DPI problem here that people can try out.

NOTE: I also re-defined a few constants - for example:
  - `PX_PER_INCH = DPI = 100` are no longer constants.  They are replaced by `DTPPOINTS_PER_INCH = 72 #Typography (desktop publishing) "points per inch"` - because DPI should be a user-defined quantity - not a constant.

RE: PlotReferenceImages:  Good to know as well.  I will have to do that when it comes time.

asinghvi17:
Bump

daschw:
I think this can be closed. Superceded by #1561

Issue #390, opened 2016-07-16
PlotSpec: Add object to hold backend-independent input specification


ma-laforge:
EDIT: The original purpose of this issue has been achieved in #747 , but it is now an issue for the refactor in the title.

Many plot tools have a way to save/load the plots to/from a file.

...But this solution ties the file type to a plot backend.

Why not have Plots.jl provide a virtual plot backend called :hdf5 (or something similar) that allows you to "render" the plot (and the plot data) to an .hdf5 file?

The plot could then be re-loaded from said .hdf5 file at a later time - then plotted on whichever backend the user desires.

The advantage of the .hdf5 format is that people can still make out the file contents even if you don't have the appropriate reader.

I think users might like this feature.

MA

lobingera:
I fully support. Some time ago (~ 2years) there was a another discussion about plotting tools on julia-users and on my wish list were: 

* externalise internal data -> have the plot as datastructure 
* load/save system independent.

ma-laforge:
Hi @lobingera,

> plot as a datastructure:

Personally, I like this idea.  This works well for my own needs to generate static plot objects (no animations - or anything fancy).  So that's what I did for my own Plots.jl-like module (before I learned of Plots.jl):

	https://github.com/ma-laforge/EasyPlot.jl

To avoid unnecessarily loading the HDF5 module, I moved the .hdf5 writer to a module called `EasyData.jl`

I currently have support for a few "backends" myself under `https://github.com/ma-laforge/`:
 - EasyPlotPlots.jl (Leverages Plots.jl)
 - EasyPlotInspect.jl
 - EasyPlotMPL.jl
 - EasyPlotGrace.jl
 - EasyPlotQwt.jl (From Pierre Raybaut).


**Having said that**
Although I personally prefer representing plots as a datastructure, I am not convinced this will necessarily work well for some of the more "advanced plot features" supported by Plots.jl.

Many of these advanced features seem to work well with the more command-oriented (as opposed to data-oriented) structure used by most plotting packages (including Plots.jl).

Could these features be elegantly tackled using a data-oriented approach?  Would people actually prefer the data-oriented approach once they got used to it?

--> I do not really know - because I don't really deal with those particular problems at this time.  That's why I am not trying to push my own solution onto people.

I think I would rather see Plots.jl support a generic load/save system in an environment that most users are likely to accept.

Maybe someone else out there has more experience on this subject...

tbreloff:
[...]

PR #747, opened 2017-03-27
Initial implementation for HDF5 plots.

ma-laforge:
Works surprisingly well, with a few limitations

- SeriesAnnotations & GridLayout support known to be missing.

### Write to .hdf5 file:
```
p = plot(...)
Plots.hdf5plot_write(p, "plotsave.hdf5")
```

### Read from .hdf5 file:
```
pyplot() #Must first select backend
pread = Plots.hdf5plot_read("plotsave.hdf5")
display(pread)
```

### DIFF
--- a/src/backends.jl
+++ b/src/backends.jl
@@ -277,6 +277,7 @@ end
 @init_backend GLVisualize
 @init_backend PGFPlots
 @init_backend InspectDR
+@init_backend HDF5
 
 # ---------------------------------------------------------
 
--- a/src/backends/hdf5.jl
+++ b/src/backends/hdf5.jl
@@ -1,0 +1,498 @@
+#HDF5 Plots: Save/replay plots to/from HDF5
+#-------------------------------------------------------------------------------
+
+#==Usage
+===============================================================================
+Write to .hdf5 file using:
+    p = plot(...)
+   Plots.hdf5plot_write(p, "plotsave.hdf5")
+
+Read from .hdf5 file using:
+    pyplot() #Must first select backend
+    pread = Plots.hdf5plot_read("plotsave.hdf5")
+    display(pread)
+==#
+
+
+#==TODO
+===============================================================================
+ 1. Support more features
+    - SeriesAnnotations & GridLayout known to be missing.
+ 3. Improve error handling.
+    - Will likely crash if file format is off.
+ 2. Save data in a folder parallel to "plot".
+    - Will make it easier for users to locate data.
+    - Use HDF5 reference to link data?
+ 3. Develop an actual versioned file format.
+    - Should have some form of backward compatibility.
+    - Should be reliable for archival purposes.
+==#
+
+
+import FixedPointNumbers: N0f8 #In core Julia
+immutable HDF5PlotNative; end #Dispatch type
+
+
+#==Useful constants
+===============================================================================#
+const _hdf5_plotroot = "plot"
+const _hdf5_dataroot = "data" #TODO: Eventually move data to different root (easier to locate)?
+const _hdf5plot_datatypeid = "TYPE" #Attribute identifying type
+const _hdf5plot_countid = "COUNT" #Attribute for storing count
+
+#Possible element types of high-level data types:
+const HDF5PLOT_MAP_STR2TELEM = Dict{String, DataType}(
+    "NATIVE" => HDF5PlotNative,
+    "VOID" => Void,
+    "BOOL" => Bool,
+    "SYMBOL" => Symbol,
+    "TUPLE" => Tuple,
+    "RGBA" => ARGB{N0f8},
+    "EXTREMA" => Extrema,
+    "LENGTH" => Length,
+
+    #Sub-structure types:
+    "FONT" => Font,
+    "AXIS" => Axis,
+)
+const HDF5PLOT_MAP_TELEM2STR = Dict{DataType, String}(v=>k for (k,v) in HDF5PLOT_MAP_STR2TELEM)
+
+
+#==
+===============================================================================#
+
+const _hdf5_attr = merge_with_base_supported([
+    :annotations,
+    :background_color_legend, :background_color_inside, :background_color_outside,
+    :foreground_color_grid, :foreground_color_legend, :foreground_color_title,
+    :foreground_color_axis, :foreground_color_border, :foreground_color_guide, :foreground_color_text,
+    :label,
+    :linecolor, :linestyle, :linewidth, :linealpha,
+    :markershape, :markercolor, :markersize, :markeralpha,
+    :markerstrokewidth, :markerstrokecolor, :markerstrokealpha,
+    :fillrange, :fillcolor, :fillalpha,
+    :bins, :bar_width, :bar_edges, :bar_position,
+    :title, :title_location, :titlefont,
+    :window_title,
+    :guide, :lims, :ticks, :scale, :flip, :rotation,
+    :tickfont, :guidefont, :legendfont,
+    :grid, :legend, :colorbar,
+    :marker_z, :line_z, :fill_z,
+    :levels,
+    :ribbon, :quiver, :arrow,
+    :orientation,
+    :overwrite_figure,
+    :polar,
+    :normalize, :weights,
+    :contours, :aspect_ratio,
+    :match_dimensions,
+    :clims,
+    :inset_subplots,
+    :dpi,
+    :colorbar_title,
+  ])
+const _hdf5_seriestype = [
+        :path, :steppre, :steppost, :shape,
+        :scatter, :hexbin, #:histogram2d, :histogram,
+        # :bar,
+        :heatmap, :pie, :image,
+        :contour, :contour3d, :path3d, :scatter3d, :surface, :wireframe
+    ]
+const _hdf5_style = [:auto, :solid, :dash, :dot, :dashdot]
+const _hdf5_marker = vcat(_allMarkers, :pixel)
+const _hdf5_scale = [:identity, :ln, :log2, :log10]
+is_marker_supported(::HDF5Backend, shape::Shape) = true
+
+function add_backend_string(::HDF5Backend)
+    """
+    if !Plots.is_installed("HDF5")
+        Pkg.add("HDF5")
+    end
+    """
+end
+
+
+#==Helper functions
+===============================================================================#
+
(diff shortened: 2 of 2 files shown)


ma-laforge:
## Fun test (Plots.jl samples): Violin plot

```
using Plots
using StatPlots #For 25 & 30

hdf5() #Select HDF5-Plots "backend"
TMPFILE = "myplot.hdf5"
idx = 30 #Which "example" plot to display

ex = Plots._examples[idx]
srand(1234) #Make consistent "random" plots.
plotvector = map(eval, ex.exprs)
curplot = plotvector[1]

#Write to HDF5:
Plots.hdf5plot_write(curplot, TMPFILE)

#Read back from HDF5:
#*****************************************************
inspectdr()
p = Plots.hdf5plot_read(TMPFILE)
display(p)

pyplot()
p = Plots.hdf5plot_read(TMPFILE)
display(p)

gr()
p = Plots.hdf5plot_read(TMPFILE)
display(p)

#...
```

ma-laforge:
### HDF5 for Debugging & More

I look at saving a plot to HDF5 as the "binary equivalent" of saving to ASCII.

The HDF5 plot output is "user-readable" - and could be used to re-create a plot even if your program is not really compatible with this particular HDF5 file format.

**HOW?**

Simply download the free HDF5 viewer available on the HDF Group's website:
https://support.hdfgroup.org/products/java/hdfview/

Said viewer will allow you to peruse the structure of the plot file using a convenient tree view widget.

Presumably, Plots.jl developers could use this viewer to help debug issues with new plot features.

It is much easier to inspect plot structures with HDF5view than current Julia debugging tools I am aware of.

mkborregaard:
This is super cool. It strikes me that it may play really nicely together with the design @tbreloff is trying to push at the `reorg` branch, where the Plots/backend interface is more fully modular - have you had a look at that?

ma-laforge:
I was not aware of the `reorg` branch yet.

I took a peek (comparing with `master`):
It appears that @tbreloff is replacing `FixedSizeArrays` with `StaticArrays` - and is changing how modules are loaded.

@mkborregaard:

I am not certain how HDF5 plots will be affected.  Is there a roadmap for the `reorg` branch?

Are you suggesting using HDF5 plots for debugging of new features, or are you giving me a heads up of upcomming changes to the data structures?

mkborregaard:
The `reorg` branch is intended as the new design of Plots for julia-0.6. @tbreloff started it up but it has been quiet for a little while. It is a reorganisation that makes Plots more fully modular, where each backend lives in its own package. That means you only have to install the backends you want, and also makes it easier for people to take ownership for maintaining a single backend. If I understand the plans right, the idea is then that the main Plots package will always produce the exact same object, regardless of what backend is selected, and then pass that to the downstream package (now the Plots object contains the backend-specific object). I just thought it'd be a cool idea to use HDF5 to save this object.

tbreloff:
To expand, the reorg branch's main goal is to be more compatible with 0.6
precompilation. This means reducing the use of eval by moving backend code
into independent modules that can be separately precompiled.

Another nice benefit (in addition to what Mike said) is that we can add
backend-specific dependecies and custom install steps to each backend
package.

The reorg branch is not, however, meant to do the "PlotSpec" redesign.
(Search the issues for PlotSpec for more info) I expect that to happen
independently, if anyone takes on the challenge.

Anyone can help with the reorg branch. Just look at the changes I made for
GR, along with the new PlotsGR.jl package, and make similar changes for the
other backends.

Please... if you're willing to help, step up to the plate and contribute.

ma-laforge:
I was curious about the purpose of that PlotsGR.jl package...

Ok, I got a good idea what the `reorg` branch does.

I took a similar multi-package approach with my own plotting api (not nearly as extensive as Plots.jl).

FYI: I don't see myself tackling this issue at the moment: I have not yet migrated to 0.6.  I am still waiting on a easy-setup Julia package to be deployed... Then I have a bunch of other migrations to do.

### Important comment
Though I find Julia precompilation does like the multi-package approach better, I find it is quite annoying to keep the core library (Plots.jl), and the backend libraries (ex: PlotsGR.jl) in sync.

Conceptually, there is nothing more difficult with this approach, but syncing up 7+ tightly coupled modules is quite annoying.

I really wish Julia supported multi-module git repositories.

...I wonder what was the reason behind this one-module-per-repository architecture...

### PS

The HDF5 plots backend might\*\* not actually need to be a separate backend...

But this allowed me to add the functionnality without adding HDF5 as a dependency.

\*\*I say "might" because I am not quite certain if the `Plot{}` object hierarchy changes depending on the backend.  The intent of HDF5 plots is to save a backend-agnostic plot information to an `.hdf5` file.

mkborregaard:
I *think* that the only backend-specific element is the .o object. There are plans to change that (as Tom said, look for PlotSpec in the issues). I agree it would be cool to save the backend-agnostic plot information, which is the exact reason I thought this would work best with the PlotSpec refactoring.
BTW, would it be possible save to this (subset of hdf5) format instead: https://github.com/simonster/JLD2.jl ?

ma-laforge:
> @mkborregaard: I think that the only backend-specific element is the .o object.

I think so too.... But that makes me wonder why Plots.jl specializes the plot objects (ex: `Plot{PyPlotBackend}`).

Why not simply have a concrete `Plot` object and dispatch functions on `myfunction(::Type{PyPlotBackend}, plt::Plot)` instead?

The .o is of type `Any` either way.

You could even keep things a bit less verbose by passing an object reference to the backend instead of the type:
```
	#Now you must pass an backend object instead of the type itself:
	function myfunction(::PyPlotBackend, plt::Plot)
		...
	end
```


### All I could find on PlotSpec (https://github.com/JuliaPlots/Plots.jl/issues/390)

> @tbreloff: With that said, I have been wanting to store user inputs separate from a fully processed Plot object, and it's likely that your concept would be
appropriate for that. So you'd serialize a PlotSpec, not a Plot.

Not a trivial thing to implement.  This takes alot of knowledge on the internals of Plots.jl - and also insight on the magic going on inside.  I only use a small subset of plots at the moment.  PlotSpec seems to touch a bit of everything  - and will likely require some significant re-structuring.

...Though I might be able to help with some tasks, I would not be a good lead on this at the moment.

### Not sure on the intent of PlotSpec

Plots.jl already has a seperation of "What to plot" (the Plot object) vs "What the backend needs (contained in Plot.o).

This PlotSpec thing sounds alot like how metafiles are defined:

Instead of describing how the image is composed (has a circle of radius 5, and a square of ...) - one describes ***how to draw*** the image (*draw* a circle of radius 5, and *draw* a square of ...).

So instead of saying I "open" a metafile drawing, you are supposed to say I am "playing" a metafile onto a canvas.

I don't really know what's the advantage of this particular software/mental model.

ma-laforge:
I tried using `JLD` at first (It appears that `JLD2` is simply a pure-Julia implementation).

### Issues with `JLD`:

  1. `JLD` is meant to persist objects to `.hdf5` files.  Though this is an ok stop-gap solution for saving plots, we eventually want to move away from being tightly coupled to a particular object hierarchy.
    - Instead, we should (eventually) be trying to specify a file format that has version info.  That way, `.hdfplots` `v1` files could still be read by the `.hdf5plots` `v2` readers.  A well-defined file format is essential if users are ever going to archive their plots for periods longer than a few months (across release cycles).
  2. Outputs I was able to generate using `JLD` were not what I would describe as "human readable".  There were many temporary "groups" and entries that were hard to decipher by a normal human.
    - The intent of `HDF5 Plots` is for someone with a simple `HDF5` viewer to be able to reverse-engineer a reader - if need be.  In other words, the `.hdf5plots` file format should be the binary equivalent of "***saving to ASCII***".

mkborregaard:
I think for PlotSpec this requires @tbreloff to comment on what his plans and ideas are. @tbreloff , if you would at some point write a specification for what would need to happen that would make it a lot easier for someone to come and take the lead on this refactor.

@ma-laforge About JLD2 I don't have a lot of experience with it, but I understood it to be simply a subset of the hdf5 specification - and furthermore I think it will become the future standard data format for all julia data objects. That's why I thought it could be nice to use that, to align across the ecosystem. Are you sure JLD2 has the same issues as you point out for JLD?

ma-laforge:
> @mkborregaard: About JLD2 [...] I think it will become the future standard data format for all julia data objects. [...] Are you sure JLD2 has the same issues as you point out for JLD?

Not sure.  I was not aware of `JLD2 `- it is not yet an official package.  Also: it appears to be compatible with Juila v0.6 only.  I am not yet building for 0.6.  At least not until I get a easy-to-install distribution.

FYI: The following is a screen shot of the HDF5 files written using JLD vs HDF5-plots output.  I created the JLD output by appending the following commands to the code I posted in the second comment of this very PR:

```
using JLD, HDF5
jldopen("test.jld", "w") do file
    g = g_create(file, "plot") # create a group
    g["attr"] = curplot.attr
end
```

jldvshdf5plots

The output on the left is the `JLD` output - and the one on the right (nested) is the `HDF5-plots` output.

I am all for using the standardized JLD/JLD2 library... but I think you'll agree that the output from JLD is unaccpetable if your goal is for HDF5-plots files to act as the "***binary equivalent of saving to ASCII***".

I will gladly use JLD/JLD2 if the output becomes "human friendly".  I would much rather not have to write specific code just to persist a data structure to file.  It is quite a pain.

mkborregaard:
Good point, I agree the HDF5 version is much nicer there. Not sure how JLD2 would work. 
I think this functionality is nice I'd support merging this PR to Plots, but I lack the skills to give it a good review - this is a job for more able people like @tbreloff or @pkofod :-)

mkborregaard:
@ma-laforge I think this is cool and would be ready to merge, if you think it is ready? But would you be willing to keep an eye out for JLD2 and see if that could provide a way to supercede this functionality?

ma-laforge:
Yes: I think it is ready to merge.  At the very least, it should not break anything else - given that it is implemented as a separate backend.

I am definitely willing to replace my own code with something more standardized - particularly if it makes the output file more human readable.

I will keep an eye out for JLD2.  Hopefully it will satisfy the needs of the "binary, readable, single-plot-file" format in the near future.

mkborregaard:
Nice. I don't think the normal test/example system for Plots backends will apply to this, but can you write some tests that will ensure the integrity of the produced objects? And add a description on PlotDocs under the backends tab?

Issue #750, opened 2017-03-28
make savefig(..., "filename.svg") save self-contained SVG for plotlyjs


tpapp:
MWE:
```julia
using Plots; plotlyjs()
savefig(plot(rand(100)), "/tmp/plot.svg")
```
produces a HTML file which calls Javascript to produce the SVG.

However,
```julia
plt = plot(rand(100))
PlotlyJS.savefig(plt.o, "/tmp/plot.svg")
```
produces an SVG. (workaround suggested by @piever on Gitter). It would be great if `Plots.savefig` and `Plots.svg` would do this instead.

piever:
I thought I'd make a PR: the issue is [here](https://github.com/JuliaPlots/Plots.jl/blob/155587628076c68e1601aea0667354daa50f7f1d/src/backends/plotlyjs.jl#L88) and it'd be fixed if the `show` function was defined for svg as it is defined for the other formats (using the PlotlyJS method). Problem is, displaying plots in Jupyter notebook also calls `_show` with svg format, so if I change it there, the Jupyter plots go from being interactive to just displaying a static svg. I'm not sure what's the cleanest way of solving this and I couldn't figure out exactly what happens when you display in IJulia and why it calls `_show` with svg format whereas plotting in Electron doesn't.

mkborregaard:
Have you tested that interactivity is lost in IJulia if you change the code? I don't think we'd want to lost that.

tpapp:
Would the following upgrade path work:
1. introduce another function or method which Jupyter can use, eg `svg_dynamic`,
2. Make a fix for Jupyter to use that, let it trickle through,
3. in the meantime, define a `show` method for `MIME("image/svg")` that can be used as a workaround,
4. once the changes go through Jupyter, replace `svg` with the static version?

piever:
Found a cleaner solution: added a method to show html for PlotlyJS, so that now it doesn't resort to the svg method for plotting in IJulia. I've changed the svg to use PlotlyJS. I've also slightly changed the display in IJulia and it seems to resolve the flickering I add when using Interact. I'll add a PR right now.

Issue #725, opened 2017-03-10
Plots don't show in IJulia


SimonDanisch:
Anyone has come across the issue, that plots sporadically don't show up in IJulia notebooks?
I have come across this one time myself and it went away after restarting the whole IJulia process.
We currently have students using IJulia and Plots and because of their large number, this sporadic issue pops up relatively often and sometimes doesn't seem to go away as easily.
Any ideas where possible problems could be? We're using plotlyjs as the backend, julia 0.5 and the currently tagged Plots version.

cc: 
@stevengj, you probably know IJulia and it's display stack better than anyone
@alanedelman

alanedelman:
would be good to have this one solved

SimonDanisch:
We just got another data point, that is only fixed by completely restarting the PC!
Might this pin down the problem further? What kind of state persists in IJulia/Plots, that only gets reseted with a restart?

tbreloff:
Do you need it to be PlotlyJS? Try plotly() or gr()?

If it persists you should open an issue with PlotlyJS.  I'm not able to
help.

SimonDanisch:
Are you sure it's a `plotlyjs` issue, though?
Was less reliable in my tests, so we stopped using it, besides the interactivity of `plotlyjs` was more appealing!

mkborregaard:
I have had this recurrently with the plotlyjs() backend. There is also often an issue with `plot!` in atom. That backend would need a little attention to come on a par with the other backends.

ma-laforge:
I don't often use this particular backend, but it might be related to the following issue: https://github.com/JuliaPlots/Plots.jl/issues/699 .

From what I can gather, there appears to be an issue with Julia's `tempname()` function - at least in Windows.

It appears Pedro encounters a similar issue when generating a large number of plots.

I suspect that, if there is a bug with `tempname()`, it is being aggravated by not deleting the temporary files these backends create.

stevengj:
(Could it be a Javascript problem with the browser?  Is it browser-dependent?)

mkborregaard:
I've had it in Firefox only.

ChrisRackauckas:
I believe it's just a PlotlyJS issue. I don't have problems with Plotly, GR, or PyPlot backends for this. But PlotlyJS never seems to show in notebooks for me. I always thought it was a non-issue though since you could use Plotly for it.

mkborregaard:
[...]

PR #751, opened 2017-03-29
wip: svg now saves static svg for plotlyjs + reliable IJulia output

piever:
Solves #750. I've added a specialized method to display html plots for PlotlyJS, so that it doesn't default to svg. Now svg can save the static image. The html can still be saved using for example `savefig("test.html")`.
As an added bonus, this method of showing html on IJulia doesn't have the flickering problem I used to experience when combining it with Interact.

### DIFF
--- a/src/backends/plotlyjs.jl
+++ b/src/backends/plotlyjs.jl
@@ -85,12 +85,9 @@ end
 
 # ----------------------------------------------------------------
 
-function _show(io::IO, ::MIME"image/svg+xml", plt::Plot{PlotlyJSBackend})
-    if isijulia() && !_use_remote[]
-        write(io, PlotlyJS.html_body(PlotlyJS.JupyterPlot(plt.o)))
-    else
-        show(io, MIME("text/html"), plt.o)
-    end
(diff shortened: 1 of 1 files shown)


piever:
EDIT = Actually, it seems to recreate the problem of plots disappearing when refreshing the page. It's probably better to not merge yet and I'll try to see if I can find a solution that works more reliably.

mkborregaard:
This is a recurrent issue with plotlyjs (cf https://github.com/JuliaPlots/Plots.jl/issues/725) - it may not be caused by this PR. But, if you think you can unwind that issue with the changes you're doing here that would be really great!

piever:
At least for me it started not working reliably with #604 (and before that #591 ), which were the PR to fix the issue that plots are gone after refresh. By removing it I think it restarted working better and flickering with interact is gone but I got the refresh issue again. Not really sure what's the way to go.

mkborregaard:
To all open PRs: Due to the change to 0.6 it appears that all PRs will have to be rebased onto master for Travis tests to run.

piever:
I'll close this one and split this PR in two parts (one which takes care of being able to save as svg, the other one will try to remove the existing "hack" to display on Jupyter notebook)

Issue #748, opened 2017-03-27
bar_width broken for Date/DateTime x axis


mkborregaard:
When the x axis is DateTime, it is no longer possible to set `bar_width`. The default value (`nothing`) works, but it is not possible to specify it. The reason seems to be that the x axis becomes reinterpreted to become very large, corresponding to the Int value of the number as DateTime. It would be nice with some more useful defaults.

```julia
bar(Date.(1:5), 1:5) #works
bar(Date.(1:5), 1:5, bar_width = 1) # does not work
bar(Date.(1:5), 1:5, bar_width = 1e10) # sort of works

julia> Int.(Date.(1:5))
5-element Array{Int64,1}:
    1
  366
  731
 1096
 1462
julia> Int.(DateTime.(1:5))
5-element Array{Int64,1}:
     86400000
  31622400000
  63158400000
  94694400000
 126316800000

julia> Int.(DateTime.(1:5)) / 1e10
5-element Array{Float64,1}:
  0.00864
  3.16224
  6.31584
  9.46944
 12.6317 
```
@daschw I think this is caused by https://github.com/JuliaPlots/Plots.jl/pull/720

daschw:
OK, I see. #720 definitely changed the `bar_with` behaviour for `Date` and I can work on reversing this. 

However, I think the issue existed before: I changed the plot recipes for `Date` and `DateTime` on my local Plots install back to
```
@recipe f(::Type{Date}, dt::Date) = (dt -> convert(Int,dt), dt -> string(convert(Date,dt)))
@recipe f(::Type{DateTime}, dt::DateTime) = (dt -> convert(Int,dt), dt -> string(convert(DateTime,dt)))
```
as they were before #720 and
```
bar(Date.(1:5), 1:5, bar_width = 1)
```
did not work as one would expect either. With `bar_width = 1` you would expect the bars to be directlly next to each other without any space between, right?
I could achieve this with
```
bar(Date.(1:5), 1:5, bar_width = 365)
```
so I suppose that a `bar_width` of 1 corresponded to 1 day on the x axis. Now, after #720 it corresponds to 1 millisecond, so you would achieve the same with
```
bar(Date.(1:5), 1:5, bar_width = 365 * 864e5)
```
Anyway, I may be mistaken, so please correct me, if you are sure that this was different before #720!

mkborregaard:
[...]

PR #752, opened 2017-03-29
restore bar_width behavior for date x axis

daschw:
This restores the `bar_width` behavior for `Date` x axis to the way it was before #720 as discussed in #748.
Now
```
bar(Date.(1:5), 1:5)
```
and
```
bar(Date.(1:5), 1:5, bar_width = 365)
```
return the same plot again.


mkborregaard:
To me this looks like the right solution. I will wait till tomorrow with merging to give a chance for others to comment.

PR #758, opened 2017-04-06
Added an animation with subplots example

mcawte:
Here is the added example in the example.jl file.


mkborregaard:
Thanks @mcawte !

PR #769, opened 2017-04-13
Use clims with :surface for PyPlot

blegat:
Looking at [Matplotlib's documentation](https://matplotlib.org/mpl_toolkits/mplot3d/tutorial.html), `vmin` and `vmax` are supported for `plot_surface` but not for `plot_wireframe`.


mkborregaard:
thanks, @blegat!

PR #774, opened 2017-04-15
Recipe for handing arrays of complex values

ChrisRackauckas:
It's very common in many cases to want to visualize results of arrays which are complex over some real axis. Complex-valued variable over time is a clear example. The majority of the time you want to plot the real and imaginary parts. This recipe makes this behavior the default when giving Plots.jl (Array of Real, Array of Complex)

### DIFF
--- a/src/recipes.jl
+++ b/src/recipes.jl
@@ -801,12 +801,10 @@ datetimeformatter(dt) = string(convert(DateTime, dt))
 # -------------------------------------------------
 # Complex Numbers
 
-@userplot ComplexPlot
-@recipe function f(cp::ComplexPlot)
-    xguide --> "Real Part"
-    yguide --> "Imaginary Part"
-    seriestype --> :scatter
-    real(cp.args[1]), imag(cp.args[1])
+@recipe function f{T<:Number}(A::Array{Complex{T}})
+    xguide --> "Re(x)"
+    yguide --> "Im(x)"
+    real.(A), imag.(A)
 end
 
 # Splits a complex matrix to its real and complex parts
--- a/src/recipes.jl
+++ b/src/recipes.jl
@@ -811,8 +811,8 @@ end
 
 # Splits a complex matrix to its real and complex parts
 # Reals defaults solid, imaginary defaults dashed
-# Label defaults are changed to match the real-imaginary reference / indexing 
-@recipe function f{T<:Number,T2<:Number}(x::AbstractArray{T},y::Array{Complex{T2}})
+# Label defaults are changed to match the real-imaginary reference / indexing
+@recipe function f{T<:Real,T2<:Number}(x::AbstractArray{T},y::Array{Complex{T2}})
   A = real.(y)
   B = imag.(y)
   _y = [!iseven(i) ? A[:,i÷2+1] : B[:,i÷2] for i in 1:2size(A,2)]
--- a/src/recipes.jl
+++ b/src/recipes.jl
@@ -810,30 +810,10 @@ end
 # Splits a complex matrix to its real and complex parts
 # Reals defaults solid, imaginary defaults dashed
 # Label defaults are changed to match the real-imaginary reference / indexing
-@recipe function f{T<:Real,T2<:Number}(x::AbstractArray{T},y::Array{Complex{T2}};
-                                       transformation=:split2D)
-  if transformation==:split2D
-    A = real.(y)
-    B = imag.(y)
-    _y = [!iseven(i) ? A[:,i÷2+1] : B[:,i÷2] for i in 1:2size(A,2)]
-    linestyle --> reshape([!iseven(i) ? :solid : :dash for i in 1:2size(A,2)],1,2size(A,2))
-    label --> reshape([!iseven(i) ? "Re(y$(i÷2+1))" : "Im(y$(i÷2))" for i in 1:2size(A,2)],1,2size(A,2))
-    color --> reshape([!iseven(i) ? i÷2+1 : i÷2 for i in 1:2size(A,2)],1,2size(A,2))
-    retval = (x,_y)
-  elseif transformation==:split3D
-    A = real.(y)
-    B = imag.(y)
-    ylabel --> "Re(y)"
-    zlabel --> "Im(y)"
-    retval = (x,A,B)
-  elseif transformation==:modulus
-    _y = norm.(y)
-    label --> reshape(["||y$(i)||" for i in 1:size(_y,2)],1,size(_y,2))
-    retval = (x,_y)
-  else
-    error("Transformation unknown. Please use :split2D, :split3D, or :modulus")
-  end
-  (retval...)
+@recipe function f{T<:Real,T2}(x::AbstractArray{T},y::Array{Complex{T2}})
+  ylabel --> "Re(y)"
+  zlabel --> "Im(y)"
+  x,real.(y),imag.(y)
 end
 
 
--- a/src/recipes.jl
+++ b/src/recipes.jl
@@ -812,14 +812,30 @@ end
 # Splits a complex matrix to its real and complex parts
 # Reals defaults solid, imaginary defaults dashed
 # Label defaults are changed to match the real-imaginary reference / indexing
-@recipe function f{T<:Real,T2<:Number}(x::AbstractArray{T},y::Array{Complex{T2}})
-  A = real.(y)
-  B = imag.(y)
-  _y = [!iseven(i) ? A[:,i÷2+1] : B[:,i÷2] for i in 1:2size(A,2)]
-  linestyle --> reshape([!iseven(i) ? :solid : :dash for i in 1:2size(A,2)],1,2size(A,2))
-  label --> reshape([!iseven(i) ? "Re(y$(i÷2+1))" : "Im(y$(i÷2))" for i in 1:2size(A,2)],1,2size(A,2))
-  color --> reshape([!iseven(i) ? i÷2+1 : i÷2 for i in 1:2size(A,2)],1,2size(A,2))
-  x,_y
+@recipe function f{T<:Real,T2<:Number}(x::AbstractArray{T},y::Array{Complex{T2}};
+                                       transformation=:split2D)
+  if transformation==:split2D
+    A = real.(y)
+    B = imag.(y)
+    _y = [!iseven(i) ? A[:,i÷2+1] : B[:,i÷2] for i in 1:2size(A,2)]
+    linestyle --> reshape([!iseven(i) ? :solid : :dash for i in 1:2size(A,2)],1,2size(A,2))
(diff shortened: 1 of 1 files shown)


ChrisRackauckas:
To make it clear that this is happening, the imaginary values are plotted as dashed lines and the default labels are adjusted to match the new indexing / meaning.

newplot 23

These are set as default that the user can override.

KristofferC:
Wouldn't it be better for a pair to have the same color and only differ on the line style?

ChrisRackauckas:
Yeah, that's a good idea. How do you iterate through the default colors to do that?

daschw:
@ChrisRackauckas I think you can just use integers for the default colors as in `plot(rand(10), color = 4)` or `plot(rand(10, 6), color = [1 1 2 2 3 3])`

ChrisRackauckas:
Thanks @daschw 

newplot 24

mkborregaard:
Nice. A few questions:
1. Is this the behaviour everyone would intuitively expect for a complex number? In that case IMO it belongs in Plots - if it is a clever but non-standard visualization it may be better in PlotRecipes.
2. Would it make sense to also have a user recipe for when x is not supplied? What about a version with three args - should that give two planes?
3. What about the seriestype - does this only make sense for `:path`? (In which case that should be fixed).
4. What is the behaviour for the user passing in values for color, linestyle and label?

ChrisRackauckas:
>Is this the behaviour everyone would intuitively expect for a complex number?

I would believe so. Generally you want to visualize complex numbers in a 2D plane, split the real-imaginary parts, or plot the modulus. 2D plane is only when all you're plotting is the complex numbers. When there's a real axis involved, this approach makes sense unless you want to default to a 3D plot, which would be a cool other option.

So if we wanted to expand this a bit, a toggle for 3D and a toggle for modulus would be good options to have.

>Would it make sense to also have a user recipe for when x is not supplied? 

There already is one.

 >What about a version with three args - should that give two planes?

I'm not sure what that would do. Or what the common use of that would be.

>What about the seriestype - does this only make sense for :path?

No, this makes sense for other series types. Scatter plots of complex values over time for complex regressions, etc.

> What is the behaviour for the user passing in values for color, linestyle and label, other than the defaults?

They need to specify a 2N vector for the real then the imaginary part.

ChrisRackauckas:
If I put toggles on the recipe, where would they be documented?

ChrisRackauckas:
Added the other options:

```julia
A = rand(5,2) + im*rand(5,2)
using Plots
t = linspace(0,1,5)
plot(t,A)
plot(t,A,transformation=:split3D)
plot(t,A,transformation=:modulus)
```

newplot 25

newplot 26

newplot 27

I think this covers the vast majority of what people would want to do if they wanted to plot (Real,Complex)

ChrisRackauckas:
For reference, single arrays of complex are handled as

```julia
A = rand(5,2) + im*rand(5,2)
using Plots
complexplot(A)
```

newplot 28


This is a user recipe already. I think it should be a type recipe as well, since I am not sure what else you would plot with an array of complex numbers. Just each column is an (x,y) series.

ChrisRackauckas:
I made that complex plot a type recipe, and took away the default as a scatter:

```julia
A = rand(5,2) + im*rand(5,2)
using Plots
plot(A)
```

newplot 29

This makes it work just like an (x,y) input, which is what I would assume would be the most common interpretation.

ChrisRackauckas:
And (Complex,Complex) is inherently 4-dimensional, so we can leave that for later. I think a good idea would be to have transformation=(:Option1,Option2). 4-d output could occur, but then with dimensional reduction recipes that would be fine. But that's a much more rare case, since most people realize that a 4D plot isn't the easiest to understand.

mkborregaard:
If this should go into Plots, there shouldn't be an additional `transformation` keyword. I see the usefulness of that functionality, though - maybe an argument for putting this in PlotRecipes instead?

mkborregaard:
> They need to specify a 2N vector for the real then the imaginary part.

Also, for this to be in Plots, it should take all the standard values for the attributes - such as `color = :blue`

ChrisRackauckas:
>PlotRecipes instead?

This has a dependency on Plots itself, which will kill the utility here (most packages will have that precompilation issue if they depend on it, and this is definitely a recipe which other recipes are built off of and not as much of an end-user thing). These are moreso transformations of complex to real vectors that I think most people would expect as automatic in any plot.

>there shouldn't be an additional transformation keyword

Any idea on how else to handle it?

>Also, for this to be in Plots, it should take all the standard values for the attributes - such as color = :blue

Will those not just pass right on through and work? This just transform the array of complex to arrays of real values, so if that works on arrays of real values then it should work here.

mkborregaard:
I haven't tested the colors, if it does work that's great.
I do see the issue that you'd like to be able to include this in a recipe, but why is it a problem that PlotRecipes would need to be loaded to use it?
No good alternative ideas on how to handle the transformations keyword, which is why i suggested PlotRecipes - it is difficult to insert new complex functionality into Plots directly.

mkborregaard:
Mind you, I am also seeing the utility of this in Plots (as I started saying) - but if I understand @tbreloff 's philosophy about Plots correctly the idea is that the interaction should be as general as possible and a new keyword should apply across seriestypes and input types.

mkborregaard:
An alternative suggestion is to keep the complexplot recipe, and for you to extend that with the array/complex case. Then you can add the transformation keyword, and keep it all in Plots.
(BTW I preferred the scatter default).

mkborregaard:
BTW they are both user recipes, just two different kinds (a userplot is also a kind of user recipe, and yes, I also find the terminology confusing). A type recipe has the signature
`@recipe function f{T<:MyType}(::Type{T}, val::T)` and would simply transform a complex vector to a non-complex one. But using a type recipe would not allow you to have different behavior for the one-argument and two-argument cases.

ChrisRackauckas:
>The alternative suggestion is to keep the complexplot recipe, and for you to extend that with the array/complex case. Then you can add the transformation keyword, and keep it all in Plots.

But then if a recipe returns complex vectors, it wouldn't be applied automatically?

>(BTW I preferred the scatter default).

Plotting (Real,Real) doesn't default to a scatter plot. I am not sure why (Complex) should be treated any differently. It's the same thing with a multiplication.

Honestly, the main point of this recipe would be to stop people from having to do `plot(real.(A),imag.(A))`. That must be a very common repeated code. This just makes that automatic and adds some sensible labels, but keeps it almost the exactly the same as though you did the splitting yourself. I find it odd that the current default plot for complex numbers is to instead error.

>I do see the issue that you'd like to be able to include this in a recipe, but why is it a problem that PlotRecipes would need to be loaded to use it? It wouldn't incur a dependency on PlotRecipes, right?

PlotRecipes has a dependency on Plots itself, which causes the precompilation problem. So packages which precompile (DifferentialEquations, ApproxFun, etc.) can't depend on it (see the previous issues about this stuff). So it's a great end-user library, but not something that can really be built off of. That's the main issue.  And plotting complex numbers is so common that I think it's important for Plots.jl to not just throw an error. If you really don't want a recipe here, it should at least throw an informative error that complex numbers are not supported.

>but if I understand @tbreloff 's philosophy about Plots correctly the idea is that the interaction should be as general as possible and a new keyword should apply across seriestypes and input types.

General as possible doesn't mean every argument has to be general. `match_dimensions` only works on Heatmaps, `normalize` and `bins` only works on histograms, etc. Those are much less general than an attribute that applies to every series any time complex numbers are involved. And it has ready generalizations to other number systems like quaternions and dual numbers (dual numbers could use the same exact recipe)

mkborregaard:
> But then if a recipe returns complex vectors, it wouldn't be applied automatically?

That's right, yes.

> Plotting (Real,Real) doesn't default to a scatter plot.

It does on my system :-) But as I said I'm not that experienced with complex numbers, so I cannot say if the case is the same or different. Your argument is sound.

> Honestly, the main point of this recipe would be to stop people from having to do plot(real.(A),imag.(A))

But you can do that and put it in Plots. But that doesn't include the option of an extra `transformation` keyword.

> So packages which precompile (DifferentialEquations, ApproxFun, etc.) can't depend on it 

You don't need to depend on PlotRecipes to define a recipe that uses a recipe there. Just tell the user to load PlotRecipes rather than Plots to do the plot.

>  If you really don't want a recipe here,

What gives you this impression? I think it is a great and very useful recipe. I am just saying that if you put it in PlotRecipes you can add options like an extra keyword attribute and document that. Whereas in Plots I think there should be a simple translation of the complex array to a non-complex array.

> Those are much less general than an attribute

Those are different seriestypes. That is different from a differen input type though.

mkborregaard:
If it's very important to have the transformationans in there, one possibility is to define the various transformations as different seriestypes: `complex`, `complexplane`, `complexmodulo`, then handling and dispatching on those in your recipe. Adding a new seriestypes that requires the input values to be of a certain type is in line with Plots logic.

ChrisRackauckas:
>You don't need to depend on PlotRecipes to define a recipe that uses a recipe there. Just tell the user to load PlotRecipes rather than Plots to do the plot.

That would be nearly impossible for me to document since any routine can output complex number. If anything, it should go into the Plots.jl docs, kind of like how DataFrame usage is mentioned a few places.

>It does on my system :-)

Really? I would've thought the default was the same everywhere...

```julia
using Plots
plot(rand(5),rand(5))
```

newplot 30

>If it's very important to have the transformationans in there, one possibility is to define the various transformations as different seriestypes

But then how would the series argument be handled? Would it have to implement an option for every single individually? The current way "just works" with all of the series.


What about reusing the `normalize` keyword? `norm` can be applied to any number. Is there any other multi-dimensional example directly in Plots.jl?

ChrisRackauckas:
I made the recipes a simple split to real and complex parts, and put the more advanced support in a separate library: https://github.com/JuliaDiffEq/DimensionalPlotRecipes.jl . That will allow me to keep going ham, but Plots.jl will have very basic support.

mkborregaard:
👍 Will review and probably merge tonight.

mkborregaard:
It'd also be fine (great even) to put the more advanced functionality in the `complexplot` recipe in Plots. In my understanding, a costum-named plot function is a signal that you have more complex functionality and extra non-standard keywords. It is just recipes meant to be called directly with `plot` that needs to conform closely to the standard interface.

mkborregaard:
> I would've thought the default was the same everywhere...

It is, I was unclear. I have made it the default in my juliarc.jl file. I wasn't claiming this as the standard, rather I was admitting to being partial - as I said, your argument was sound.

mkborregaard:
Nice one, I think this works great for very basic functionality as you say. I think it is a nice idea to develop your ideas in an independent package - which could possibly later be added to Plots as  a `complexplot` userplot recipe, if you'd like.

mkborregaard:
Thanks @ChrisRackauckas!

Issue #462, opened 2016-08-28
Set number of ticks without specifying their location


ahwillia:
Matplotlib has a nice feature ([`locator_params`](http://matplotlib.org/api/pyplot_api.html#matplotlib.pyplot.locator_params)) that lets you set the number of ticks on an axis without specifying their actual location.

It would be nice to have something like `plot(randn(100,2); xaxis=(3))` produce something like this by default:

<img width="598" alt="screen shot 2016-08-28 at 11 51 56 am" src="https://cloud.githubusercontent.com/assets/636625/18035980/d9780ab6-6d15-11e6-89c9-65feeb74cc87.png">

mkborregaard:
Closed in #780

PR #780, opened 2017-04-23
WIP: Allow specifying the number of ticks by passing an Integer to the ticks attribute

daschw:
The discussion in https://github.com/JuliaPlots/PlotUtils.jl/pull/11 reminded me of this request https://github.com/JuliaPlots/Plots.jl/issues/462.
I think this would also provide an easy workaround for users in such degenerate cases, where only one tick is shown in a plot.

This is the first attempt to implement the option to set the desired number of ticks for an axis by passing an Integer to the `ticks`/`xticks`/`yticks`/`zticks` attributes. However, I did not yet manage to find the right configuration when calling `optimize_ticks` in order to always return correct number of ticks.

Here is the output of
```julia
plot([begin
        scatter(s[:,1], s[:,2])
        plot!(e1[:,1], e1[:,2], ticks = i, title = "ticks = $i")
    end for i in [0:5; 10; 15; 20]]..., label = "")
```
with the data of the example discussed in https://github.com/JuliaPlots/PlotUtils.jl/pull/11:

nticks_strictspan_false_filter

There are too few ticks chosen in most cases. The issue is, that the algorithm in `optimize_ticks` returns ticks that can in general be outside the axis limits. I doubt that we want to extend the axis limits in those cases, thus the outlying ticks are dropped.

cc: @crbinz Maybe you have some useful input here, because you did such a great job identifying the problem in your degenerate example and because you are already familiar with the ticks scoring algorithm.

### DIFF
--- a/src/axes.jl
+++ b/src/axes.jl
@@ -188,6 +188,23 @@ function optimal_ticks_and_labels(axis::Axis, ticks = nothing)
             k_min = 5, # minimum number of ticks
             k_max = 8, # maximum number of ticks
         )[1]
+    elseif typeof(ticks) <: Int
+        # only return ticks within the axis limits
+        filter(
+            ti -> sf(amin) <= ti <= sf(amax),
+            optimize_ticks(
+                sf(amin),
+                sf(amax);
+                # TODO: find a better configuration to return the chosen number
+                # of ticks
+                k_min = ticks + 1, # minimum number of ticks
+                k_max = ticks + 2, # maximum number of ticks
+                k_ideal = ticks + 2,
+                # `strict_span = false` rewards cases where the span of the
+                # chosen  ticks is not too much bigger than amin - amax:
+                strict_span = false,
+            )[1]
+        )
     else
         map(sf, filter(t -> amin <= t <= amax, ticks))
     end
@@ -226,7 +243,7 @@ function get_ticks(axis::Axis)
     elseif ticks == :auto
         # compute optimal ticks and labels
         optimal_ticks_and_labels(axis)
-    elseif typeof(ticks) <: AVec
+    elseif typeof(ticks) <: Union{AVec, Int}
         # override ticks, but get the labels
         optimal_ticks_and_labels(axis, ticks)
     elseif typeof(ticks) <: NTuple{2}
--- a/src/axes.jl
+++ b/src/axes.jl
@@ -241,7 +241,7 @@ function get_ticks(axis::Axis)
     elseif typeof(ticks) <: Union{AVec, Int}
         # override ticks, but get the labels
         optimal_ticks_and_labels(axis, ticks)
-    elseif typeof(ticks) <: NTuple{2}
+    elseif typeof(ticks) <: NTuple{2, Any}
         # assuming we're passed (ticks, labels)
         ticks
     else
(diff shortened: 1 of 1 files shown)


diegozea:
Axis should be draw when `ticks = 0` IMHO

daschw:
Thanks @diegozea, I agree. I will try to modify it accordingly.

mkborregaard:
Why not let `optimize_ticks` expand the axis limits, when these are not specified explicitly?

daschw:
You are right, that would be an option ... I suppose only in the cases when the number of ticks is specified, though? And I guess axis limits specifications would "beat" tick number specifications, when both are set?

daschw:
Letting `optimize_ticks` expand the limits would result in the following plots:
nticks_expand_lims
I'm not really sure if this is better ... especially `ticks = 1` is surprising. 3 to 5 seem ok and the correct number of ticks is chosen in each case.

src/axes.jl
```diff
@@ -181,15 +181,27 @@ function optimal_ticks_and_labels(axis::Axis, ticks = nothing)
...
+            sf(amin),
+            sf(amax);
+            k_min = ticks, # minimum number of ticks
+            k_max = ticks, # maximum number of ticks
+            k_ideal = ticks,
+            # `strict_span = false` rewards cases where the span of the
+            # chosen  ticks is not too much bigger than amin - amax:
+            strict_span = false,
+        )
+        axis[:lims] = map(invscalefunc(scale), (viewmin, viewmax))
```
daschw: Is there another way to do this? Setting an axis attribute here kind of does not seem right.

mkborregaard:
Ugh, no I think your original implementation was better.

mkborregaard:
@daschw I think you should just run with this and merge when you feel it's ready.

src/axes.jl
```diff
@@ -241,7 +241,7 @@ function get_ticks(axis::Axis)
     elseif typeof(ticks) <: Union{AVec, Int}
         # override ticks, but get the labels
         optimal_ticks_and_labels(axis, ticks)
-    elseif typeof(ticks) <: NTuple{2}
+    elseif typeof(ticks) <: NTuple{2, Any}
```
mkborregaard: It's probably better to not do this, and the rebase this PR onto master before merging (or now, if you want the functionality at this point). I'm fairly sure this will cause git conflicts.

daschw: I accidently commited the "passing tuple to ticks on 0.6" fix on this wip branch instead of a new one yesterday. Then I undid it again. Will this cause conflicts?

Also, I think we have more important stuff to deal with right now than this new feature and I think I will not work on this in the near future. Is there anything speaking against me closing this wip PR and opening a new one when I find time to continue working on it, in order to avoid these kind of "accidents"?

mkborregaard: No that shouldn't cause conflicts.
No need to close the PR, but it shouldn't cause problems either.

mkborregaard:
It appears that all PRs will have to be rebased onto master for Travis tests to run.

mkborregaard:
Should we merge this?

daschw:
Yes, I think it can be merged. It could be improved if anyone wants to do this, but the basic functionality is there, and it should not break any existing plots or functionality.

PR #786, opened 2017-04-29
Move the @userplot recipes

ChrisRackauckas:
https://github.com/JuliaPlots/RecipesBase.jl/pull/16


src/backends/glvisualize.jl
```diff
@@ -826,7 +826,7 @@ function gl_bar(d, kw_args)
     hw = if bw == nothing
         mean(diff(x))
     else
-        Float64[cycle(bw,i)*0.5 for i=1:length(x)]
+        Float64[_cycle(bw,i)*0.5 for i=1:length(x)]
```
mkborregaard: WRT to the history it would look a lot clearer if the _cycle change was separate from the movement of the functions.

ChrisRackauckas: Alright sounds good

mkborregaard: 👍 I guess the PR also needs a rebase before eventually merging to sort out the conflicts

ChrisRackauckas: I'll cherry pick for the cycle change and the `@userplot` change, and close this as a messy pre-PR.

mkborregaard: Perfect

PR #787, opened 2017-04-29
Rewrite of histogram recipes, using StatsBase.histogram

oschulz:
Continuation of #713.

Also fixes histograms with log-axes (currently badly broken, esp. for the y-axis):

```
using Plots, StatsBase
pyplot();


data1d = filter(x -> abs(x) < 2.5, randn(10000))
data2d = (randn(1000000) - 1, - 0.1 * log(rand(1000000)))

edge = -4:0.1:4
h1 = fit(Histogram, data1d, edge, closed = :left)
h1_norm = normalize(h1)

edge_func(x) = x + 0.1 * x^3
inhom_edge = map(x -> edge_func(x) / edge_func(4) * 4, -4:0.5:4)
h2 = float(fit(Histogram, data1d, inhom_edge, closed = :left))
h2_err = sqrt.(h2.weights)
h2_norm_density = normalize(h2, mode = :density)
h2_norm_pdf, h2_err_norm = normalize(h2, h2_err, mode = :pdf)

h2d = fit(Histogram, data2d, nbins = (80, 120), closed = :left)



plot(
    plot(
        h2_norm_density,
        fillrange = 0,
        fillalpha = 0.5,
        fillcolor = :red,
        # bar_width = 0.1,
        yscale = :log10,
    ),
    plot(
        h2_norm_pdf,
        yerror = h2_err_norm,
        seriestype = :scatter,
        # xerror = 0.2,
        yscale = :log10,
    ),
    plot(
        [h1_norm, h2_norm_pdf],
        xlims = (-3, 3),
        seriestype = :step,
        yscale = :log10,
    ),
    plot(h2d),
    grid = false,
    size = (1024, 768)
)
```

hists-statsbase-logy-demo

```
plot(
    histogram(
        data1d,
        bins = inhom_edge,
        normalize = :density,
        fillrange = 0,
        fillalpha = 0.5,
        fillcolor = :red,
        # bar_width = 0.1,
        yscale = :log10,
    ),
    scatterhist(
        data1d,
        bins = inhom_edge,
        normalize = :pdf,
        yscale = :log10,
    ),
    begin
        stephist(
            data1d,
            bins = edge,
            normalize = :pdf, # true
            xlims = (-3, 3),
            markershape = :circle,
            yscale = :log10,
        )
        stephist!(
            data1d,
            bins = inhom_edge,
            normalize = :pdf,
            xlims = (-3, 3),
            markershape = :circle,
            yscale = :log10,
        )
    end,
    histogram2d(data2d..., bins = (80, 120)),
    grid = false,
    size = (1024, 768)
)
```

hists-modes-logy-demo


mkborregaard:
Nice - I'll review with the aim of merging this week. 
Just a question - what about log-scale on the x axis? That is also broken, IMHO, as I think the automatic binning should be applied after the transformation (rather than before, as it is now). You may find that is outside the scope, I just thought, while you're at it :-)

oschulz:
> Just a question - what about log-scale on the x axis? That is also broken

It is, and this PR also fixes that, to a certain extent: As long as there are no negative bins, all should be well, now (There's specific handling if the first bin starts at zero, I basically copied what ROOT does in such cases). I'm going to improve this further - but like you say, it's probably best if we concentrate on getting the  the major changes merged, first.

mkborregaard:
Great. Yeah, setting the x axis log should just fail on negative bins.

oschulz:
In general, we may want to improve support for log-plots of non-negative data, though. It's just quite common to have a few zeros in the data, shouldn't prevent log-plots. I'm handling this by treating zeros as NaN, for histograms: I think we should extend that to other plot-types (like `:path` in general). For example, a user may want to plot absolute or squared values of data on log-y - perfectly valid concept, so we should handle zeros (and they should be visible as a gap in the plot, as currently implemented for `:stephist`).

> Yeah, setting the x axis log should just fail on negative bins.

Not necessarity: ROOT does allow this, for example - of course, only the positive part is drawn. But I think this is valid, it's just an equivalent of setting xlims. Something to thing about ... but not an important issue, I guess. Normally, users shouldn't log-plot negative data (on any axis), of course.

mkborregaard:
Worth discussing definitely. But I agree, not important here.

mkborregaard:
Converting zeros to NaNs should probably elicit a warning, though.

oschulz:
> Converting zeros to NaNs should probably elicit a warning, though.
 I don't think so - the warning would be annoying (for histograms, for example, they would be very frequent, as zero-filled bins aren't exactly a rarity). I think as long as it's obvious from the plot that there's a value that couldn't be displayed, the internal way it's handled doesn't warrant printing out warnings. I'd rather suggest documenting the behavior, instead (if it's expected behavior, no warning is necessary, after all).

mkborregaard:
Let's just go ahead with this PR - we can always look at additional functionality later

REQUIRE
```diff
@@ -7,3 +7,4 @@ Reexport
 FixedSizeArrays
 Measures
 Showoff
+StatsBase
```
Evizero: Is there a known lower version requirement for this?

mkborregaard: Good point - I have been meaning to ask the StatsBase people when they'll make the release, so we can time the Plots release. I'll do that straigth away.

mkborregaard:
@oschulz I hope you'll be ready to take the final important step - and document this? I think it would be great to have the different seriestypes documented under their own heading in the Plots docs. It would be great, for a start, to have a shortn'sweet entry about 'histogram-like' series, detailing the input data, relevant attributes, possibilities etc.

REQUIRE
oschulz: If we put a future StatPlots version in - could that result in Pkg conflicts when running both on master and installing/updating other packages? I'm not sure if Pkg checks deps for packages on master.

oschulz:
> I hope you'll be ready to take the final important step - and document this

Sure, this will definitely need some documentation. I'll start working on it while you review.

Concerning this PR - some people I know would like to start using the new histograms, but running on a PR is obviously a bit more tricky than running on master. It would be great if we could merge this before next Monday, for certain reason (related to internal projects).

REQUIRE
mkborregaard: 0.14.0 - it is already in METADATA. I'll try to review quickly and push a new minor version release.

oschulz: > 0.14.0 - it is already in METADATA

Nice, hadn't seen it yet. I've changed our requirement accordingly.

src/recipes.jl
```diff
@@ -378,109 +384,323 @@ end
...
+
+    n_samples = length(linearindices(first(vs)))
+    # Estimator for number of samples in one row/column of bins along each axis:
+    n = max(1, n_samples^(1/N))
+
+    v = vs[dim]
+
+    if mode == :sqrt  # Square-root choice
+        _cl(sqrt(n))
+    elseif mode == :sturges || mode ==:auto  # Sturges' formula
```
mkborregaard: I know everybody (including StatsBase) uses sturges by default, but why? I have the impression that there are many better algorithms today. Now is certainly the time to decide on a default for the future.

oschulz: > I know everybody (including StatsBase) uses sturges by default, but why?

Good question - inertia? I'd really like to contribute this to StatsBase in the future, though, and I don't know if they would agree to change their default. So it might be easier to stick with sturges for compatibility. I don't have a very strong objection to use different defaults in Plots and StatsBase, but it might confuse users.

mkborregaard: Yes, I remember we had that conversation. It is a good idea to contribute this to StatsBase instead. OK let's stick with this now, but you'd be very welcome to add additional (good) algorithms to test this out with in practice before transferring to StatsBase.

oschulz:
Rebased on current master.

mkborregaard:
Thanks a lot for this - great work! I am really happy with this improved histogram capability.

oschulz:
Thanks, @mkborregaard! I'll start documenting this - and I guess we should make a release, soon, to get the new histograms out there? Afterwards, I have some ideas for further improvements / fine tunning (maybe also for other plot types).

mkborregaard:
I get 
```julia
histogram(abs(randn(10000)), xscale = :log10)
ERROR: DomainError:
 in nan_dom_err at ./math.jl:196 [inlined]
 in log10(::Float64) at ./math.jl:202
 in (::Plots.##114#115{Symbol})(::Float64) at /Users/michael/.julia/v0.5/Plots/src/axes.jl:148
 in optimal_ticks_and_labels(::Plots.Axis, ::Void) at /Users/michael/.julia/v0.5/Plots/src/axes.jl:185
 in get_ticks(::Plots.Axis) at /Users/michael/.julia/v0.5/Plots/src/axes.jl:228
 in tick_padding(::Plots.Axis) at /Users/michael/.julia/v0.5/Plots/src/backends.jl:71
 in _update_min_padding!(::Plots.Subplot{Plots.GRBackend}) at /Users/michael/.julia/v0.5/Plots/src/backends.jl:104
 in _collect(::Array{Plots.AbstractLayout,2}, ::Base.Generator{Array{Plots.AbstractLayout,2},Plots.#_update_min_padding!}, ::Base.EltypeUnknown, ::Base.HasShape) at ./array.jl:320
 in _update_min_padding!(::Plots.GridLayout) at /Users/michael/.julia/v0.5/Plots/src/layouts.jl:302
 in prepare_output(::Plots.Plot{Plots.GRBackend}) at /Users/michael/.julia/v0.5/Plots/src/plot.jl:254
 in display at /Users/michael/.julia/v0.5/Plots/src/output.jl:133 [inlined]
 in gui at /Users/michael/.julia/v0.5/Plots/src/output.jl:123 [inlined]
 in display(::Base.REPL.REPLDisplay{Base.REPL.LineEditREPL}, ::MIME{Symbol("text/plain")}, ::Plots.Plot{Plots.GRBackend}) at /Users/michael/.julia/v0.5/Plots/src/output.jl:138
 in display(::Base.REPL.REPLDisplay{Base.REPL.LineEditREPL}, ::Plots.Plot{Plots.GRBackend}) at ./REPL.jl:135
 in display(::Plots.Plot{Plots.GRBackend}) at ./multimedia.jl:143
 in print_response(::Base.Terminals.TTYTerminal, ::Any, ::Void, ::Bool, ::Bool, ::Void) at ./REPL.jl:154
 in print_response(::Base.REPL.LineEditREPL, ::Any, ::Void, ::Bool, ::Bool) at ./REPL.jl:139
 in (::Base.REPL.##22#23{Bool,Base.REPL.##33#42{Base.REPL.LineEditREPL,Base.REPL.REPLHistoryProvider},Base.REPL.LineEditREPL,Base.LineEdit.Prompt})(::Base.LineEdit.MIState, ::Base.AbstractIOBuffer{Array{UInt8,1}}, ::Bool) at ./REPL.jl:652
 in run_interface(::Base.Terminals.TTYTerminal, ::Base.LineEdit.ModalInterface) at ./LineEdit.jl:1579
 in run_interface(::Base.Terminals.TTYTerminal, ::Base.LineEdit.ModalInterface) at /Applications/Julia-0.5.app/Contents/Resources/julia/lib/julia/sys.dylib:?
 in run_frontend(::Base.REPL.LineEditREPL, ::Base.REPL.REPLBackendRef) at ./REPL.jl:903
 in run_repl(::Base.REPL.LineEditREPL, ::Base.##950#951) at ./REPL.jl:188
 in _start() at ./client.jl:363
 in _start() at /Applications/Julia-0.5.app/Contents/Resources/julia/lib/julia/sys.dylib:?
```

mkborregaard:
Also, `bins` is not defaulting to `:auto`.

mkborregaard:
and #799

oschulz:
> Also, bins is not defaulting to :auto.

What is it defaulting to, at the moment?

oschulz:
> Also, bins is not defaulting to :auto 

Currently, it defaults to `30`, which was the previous behaviour. I think we decided to keep the default, back when we started this - but I would be fine with changing it. `30` is not such a horrible default, though.

mkborregaard:
I think if we do have an `:auto` category, that should be the default. We can discuss what `:auto` should be then. IMHO it should be one of the auto schemes, given that we have them. Personally I don't like Sturges' very much, though it is compelling to have the same behaviour as StatsBase and most other plotting packages.

oschulz:
>  think if we do have an :auto category, that should be the default.

I agree.

> Personally I don't like Sturges' very much

Me neither. Here's a demo with different number of data points (low vs. high statistics) - let's see how they perform. The input data is

```julia
    data_multigauss = vcat(randn(div(1*n, 4)) - 2, randn(div(3*n, 4)) + 2)
```

### n = 10^2

autobining-demo-100-datapts-histogram

### n = 10^3

autobining-demo-1000-datapts-histogram

### n = 10^4

autobining-demo-10000-datapts-histogram

### n = 10^5

autobining-demo-100000-datapts-histogram

### n = 10^6

autobining-demo-1000000-datapts-histogram

### n = 10^6, `seriestype = :stephist`

autobining-demo-1000000-datapts-stephist

Here's my conclusion:

* Neither `:sturges` nor `:sqrt` are great choices (no surprise).

* `:rice`, `:scott` and `:fd` do well both for low and high statistics, except for `:scott` and `:fd` at very low statistics (n = 10^2). If a lot of data is available, they will set a high (but not too high) number of bins, and produce a histogram that is both detailed and "smooth" at the same time. Unfortunately, our default histogram plotting style `:barhist` can't really handle a high number of bins.

* The current default, a fixed number of 30 bins, actually produces a decent histogram in all cases, but doesn't take advantage of high-statistics cases, of course.

So I see two possibilities (for the new default `bins = :auto`):

* Make `bins = :auto` equivalent to the current default (30 bins)

* Make `bins = :auto` equivalent to  `:rice`, `:scott` or `:fd`, and also change the default histogram drawing style from `:barhist` to `:stephist`, as the latter can handle both coarse and fine binning.

pkofod:
in the presented cases, I think `:rice` proves itself as a prette decent choice, right?

mkborregaard:
I'd agree, though the general scientific consensus appears to be that it yields too many bins. 
As I've expressed before, I don't think we should change the default seriestype :-)

oschulz:
> @pkofod: in the presented cases, I think :rice proves itself as a prette decent choice, right?

For this data  - a simple non-gaussian case where methods like `:sturges` that assume a normal distribution will fail - yes. But for even for complex data, `:fd` may be better than `:rice. Apparently, Freedman-Diaconis was developed to be less sensitive to outliers in data (according to https://en.wikipedia.org/wiki/Histogram#Freedman.E2.80.93Diaconis.27_choice). There's even more modern methods, I found quite a few publications, but they can get very complex.

But yes, `:rice` would be a good default - sadly, it's not compatible with bar histograms.

oschulz:
@mkborregaard so we go with `:auto` -> `30`?

mkborregaard:
The Sturges method is from 1926 and has only the claim to fame of being the first. One of the most influential more recent methods is Wand's http://www.jstor.org/stable/2684697 , which was selected as generally optimal in this recent comparison https://www.esaim-ps.org/articles/ps/abs/2009/01/ps0721/ps0721.html (along with a few others). I'd personally prefer to go with that eventually, but it is not yet implemented here. 
Let's stick with 30 for the moment, for backwards consistency.

oschulz:
> Let's stick with 30 for the moment, for backwards consistency.

PR is ready: #802

mkborregaard:
Am I right in my impression that all issues we have identified with the new histograms have now been adressed?

oschulz:
I think so - except for one thing: log-scale for heatmaps and 2D-histograms is still broken - but that isn't due to the new histograms, really. I want to look into it - if it's an easy fix, maybe it could also go into the next release.

oschulz:
> log-scale for heatmaps and 2D-histograms is still broken

I mean for `zscale`.

mkborregaard:
That'd be cool, thanks!

PR #785, opened 2017-04-28
updated warnings for a) ImageMagick and b) Conda

huckl3b3rry87:

PR #790, opened 2017-05-01
changed warning in imagemagick

huckl3b3rry87:
split https://github.com/JuliaPlots/Plots.jl/pull/785 and updated the warning message


mkborregaard:
Happy to merge, but a quick naive question - is it OK to just install ImageMagick or is it necessary to install the command line tools explicitly?

huckl3b3rry87:
@mkborregaard I think that it would be OK, but I will leave this up to you guys.

Issue #670, opened 2017-02-06
Make tests pass


tbreloff:
The VRT tests fail because of the PyPlot.jl upgrade to Matplotlib 2.0.

daschw:
Ok, so I've run the VRT tests locally today and I have found basically two issues for pyplot/matplotlib2:

1. The colorbar labels are cropped, as can be seen in the following pictures:
![4_colors](https://cloud.githubusercontent.com/assets/16589944/24019779/c39d266e-0a99-11e7-9671-bc92a600b1b1.png)
![10_histogram2d](https://cloud.githubusercontent.com/assets/16589944/24019786/ca6eeb44-0a99-11e7-8ad3-abe169b4545e.png)
![22_contours](https://cloud.githubusercontent.com/assets/16589944/24019789/cf52121c-0a99-11e7-9d12-61769a34e70a.png)
![24_3d](https://cloud.githubusercontent.com/assets/16589944/24019797/d46cf80c-0a99-11e7-9fad-af497fe24704.png)
![28_heatmap](https://cloud.githubusercontent.com/assets/16589944/24019807/dcd35644-0a99-11e7-93df-1eda1bcfcd17.png)
Should I create a seperate issue for this?

2. (probably minor) The display of line styles (dot, dashdot, ...) has changed:
![12_line_styles](https://cloud.githubusercontent.com/assets/16589944/24020041/dfc4c7b0-0a9a-11e7-8bb7-b078cda356b4.png)
I don't know, if this is a big issue, but in the legend the dashed and dashdotted lines cannot really be identified as such.

cc: @mkborregaard

daschw:
I've also run into some problems with GR:

1. Images are displayed transposed. This is probably already known via https://github.com/JuliaPlots/Plots.jl/issues/432
![6_images](https://cloud.githubusercontent.com/assets/16589944/24020427/68354150-0a9c-11e7-9f27-57f69bc584e0.png)

2. The custom marker shapes are weirdly stretched:
![21_custom_markers](https://cloud.githubusercontent.com/assets/16589944/24020824/259b7510-0a9e-11e7-80e7-1e6a4d803372.png)

3. Colorbars are not shown:
![24_3d](https://cloud.githubusercontent.com/assets/16589944/24020916/96447dd4-0a9e-11e7-873b-3c8084d755bd.png)
[...]

Issue #603, opened 2016-12-07
Margins after colorbar


tcovert:
Any idea how I might make this figure wide enough so that the text on the colorbar is not cutoff for the negative values?

![sih](https://cloud.githubusercontent.com/assets/4276172/20951004/ddefacc6-bbe8-11e6-889d-0484921f348c.png)

When I try to adjust margins using, `right_margin`, all that does is move the colorbar further to the right of the contour, without adding extra padding after the colorbar itself.

mkborregaard:
Should be fixed

PR #794, opened 2017-05-02
Avoid PyPlot colorbar label cropping

daschw:
This should solve the issue of colorbar label cropping in pyplot sins matplotlib2.0 discussed in https://github.com/JuliaPlots/Plots.jl/issues/670. (I'm not sure if it also solves https://github.com/JuliaPlots/Plots.jl/issues/603)

The issue was that the colorbar ticks were previously inside the colorbar, and now they are outside. Thus the width added to rightpad had to be increased. I found 2.3mm to be the minimal value without label cropping in all VRT images with colorbars:

pyplot_4

pyplot_10

pyplot_22

pyplot_24

pyplot_28

The old (cropping) behavior can be viewed in https://github.com/JuliaPlots/Plots.jl/issues/670.


mkborregaard:
awesome

mkborregaard:
I think we should tag a new release today or tomorrow, are you ready to merge this and your other PRs, @daschw ?

PR #795, opened 2017-05-03
Add pgfplots support for zcolor and colorbar

sebastianpech:
Extension of the pgfplots backend to support zcolor for 2D plots.
[Examples here](https://gist.github.com/sebastianpech/6bfb181ab8265b9b109810c6155e6440)

As seen in the examples there is a problem when calculating the bounding box as the colorbar is not taken into account. I'm not sure yet how this is going to work. Any hints besides reducing the width by a fixed length? 

I tried the 3D example from the doc as well, but the pgfplots julia wrapper apparently does not support colormaps with 3D scatter or lines plots.
I managed to enable the colormap by using the mesh option but this only works for paths and not for markers

### DIFF
--- a/src/backends/pgfplots.jl
+++ b/src/backends/pgfplots.jl
@@ -22,8 +22,8 @@ const _pgfplots_attr = merge_with_base_supported([
     :guide, :lims, :ticks, :scale, :flip, :rotation,
     :tickfont, :guidefont, :legendfont,
     :grid, :legend,
-    # :colorbar,
-    # :marker_z, :levels,
+    :colorbar,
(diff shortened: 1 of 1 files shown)


mkborregaard:
Is this WIP or ready to merge?

sebastianpech:
In my opinion it's ready to merge. Of course some of the extensions I made are not applicable to every possible use case (i.e. 3D scatter, ... ) but so is the whole pgfplots backend and the pgfplots julia wrapper.
So it's also WIP.

mkborregaard:
Thanks @sebastianpech !

PR #804, opened 2017-05-08
Change cycle => _cycle

ChrisRackauckas:
There is a Base function named `cycle` which this is not. This makes the code rather confusing, especially since `Base.cycle` is used in other places. Additionally, this is then imported in other libraries, meaning it's an important function. This changes the name of this function to `_cycle` instead of shadowing the Base function to improve code clarity.

### DIFF
--- a/src/backends/glvisualize.jl
+++ b/src/backends/glvisualize.jl
@@ -783,7 +783,7 @@ function gl_bar(d, kw_args)
     hw = if bw == nothing
         ignorenan_mean(diff(x))
     else
-        Float64[cycle(bw,i)*0.5 for i=1:length(x)]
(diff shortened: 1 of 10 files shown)


ChrisRackauckas:
Yes, though it's breaking to StatPlots which explicitly imports it, so that PR would need to follow.

mkborregaard:
good point - let's make sure the merges are simultaneous and followed by a concerted release.

mkborregaard:
To all open PRs: Due to the change to 0.6 it appears that all PRs will have to be rebased onto master (and conflicts resolved) for Travis tests to run.

mkborregaard:
and this?

ChrisRackauckas:
Yes.

Issue #734, opened 2017-03-15
0.6 issues with PyPlot and IJulia


cstjean:
At the REPL, `PyPlot`'s master works for me on 0.6, but `Plots`'s master fails:

```
julia> using Plots

julia> plot([1,2])
WARNING: Base.cycle is deprecated.
  likely near no file:0
WARNING: Base.cycle is deprecated.
  likely near no file:0
WARNING: Base.cycle is deprecated.
  likely near no file:0
WARNING: Base.cycle is deprecated.
  likely near no file:0
WARNING: Base.cycle is deprecated.
  likely near no file:0
WARNING: Couldn't initialize pyplot.  (might need to install it?)
INFO: To do a standard install of pyplot, copy and run this:

if !Plots.is_installed("PyPlot")
    Pkg.add("PyPlot")
end
withenv("PYTHON" => "") do
    Pkg.build("PyPlot")
end
import Conda
Conda.add("qt=4.8.5")

# now restart julia!


ERROR: MethodError: no method matching ioff()
The applicable method may be too new: running in world age 21556, while current world is 21576.
Closest candidates are:
  ioff(::Any...; kws...) at /Users/cedric/.julia/v0.6/PyPlot/src/PyPlot.jl:169 (method too new to be called from this world context.)
```

In IJulia:

```julia
> using Plots

InitError: MethodError: no method matching set_ijulia_output(::String)
The applicable method may be too new: running in world age 21560, while current world is 21563.
Closest candidates are:
  set_ijulia_output(::AbstractString) at /Users/cedric/.julia/v0.6/Plots/src/output.jl:261 (method too new to be called from this world context.)
during initialization of module Plots

Stacktrace:
 [1] setup_ijulia() at /Users/cedric/.julia/v0.6/Plots/src/output.jl:274
 [2] __init__() at /Users/cedric/.julia/v0.6/Plots/src/Plots.jl:229
 [3] _include_from_serialized(::String) at ./loading.jl:157
 [4] _require_from_serialized(::Int64, ::Symbol, ::String, ::Bool) at ./loading.jl:194
 [5] _require_search_from_serialized(::Int64, ::Symbol, ::String, ::Bool) at ./loading.jl:224
 [6] require(::Symbol) at ./loading.jl:409
```

Possibly similar to [this issue](https://github.com/JuliaLang/IJulia.jl/pull/513)

mkborregaard:
cf #745

darwindarak:
[...]

PR #805, opened 2017-05-08
RFC: Update Travis configuration

darwindarak:
`Plots.jl` tests are actually running now, instead of failing during installation.  The tests seem to be failing because the images produced are actually slightly different from the reference ones.  For example:

The reference image
ref1

compared to the generated image
ref1

Maybe due to the changes in Matplotlib 2.0?

### DIFF
--- a/test/install_wkhtmltoimage.sh
+++ b/test/install_wkhtmltoimage.sh
@@ -7,7 +7,7 @@ sudo apt-get -qq update
 
 sudo apt-get install -y xfonts-75dpi
 wget http://download.gna.org/wkhtmltopdf/0.12/0.12.2/wkhtmltox-0.12.2_linux-trusty-amd64.deb
-sudo dpkg -i wkhtmltox-0.12.2_linux-trusty-amd64.deb || true
+sudo dpkg -i --depends wkhtmltox-0.12.2_linux-trusty-amd64.deb
 sudo apt-get -f install
 wkhtmltoimage http://www.google.com test.png
 ls
--- a/test/install_wkhtmltoimage.sh
+++ b/test/install_wkhtmltoimage.sh
@@ -7,7 +7,7 @@ sudo apt-get -qq update
 
 sudo apt-get install -y xfonts-75dpi
 wget http://download.gna.org/wkhtmltopdf/0.12/0.12.2/wkhtmltox-0.12.2_linux-trusty-amd64.deb
-sudo dpkg -i --depends wkhtmltox-0.12.2_linux-trusty-amd64.deb
+sudo dpkg -i --force-depends wkhtmltox-0.12.2_linux-trusty-amd64.deb
 sudo apt-get -f install
 wkhtmltoimage http://www.google.com test.png
 ls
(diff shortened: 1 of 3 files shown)


mkborregaard:
Yes, see https://github.com/JuliaPlots/Plots.jl/issues/670 . That issue is not related to the PR, though, right?

pkofod:
This includes v0.6 which is fine, but what is actually solved here? Sorry for being dense.

darwindarak:
@mkborregaard You're right, that is a separate issue.  This PR makes it so that the tests at least get to running some Julia code.  It was stuck installing `wkhtmltox`.

@pkofod Maybe nothing, since even tests on 0.5 have been ignored for the last 4 months.  But 0.6 is around the corner and there seems to be a couple 0.6 specific bugs that make Plots unusable on almost any backend.  Whether these tests pass or fail, they provide useful information for debugging.  For example, the test [results](https://travis-ci.org/JuliaPlots/Plots.jl/jobs/230019825) show the PyPlot backend working just fine.  It turns out that setting the Matplotlib backend to `Agg` bypasses the [world age problem](https://github.com/JuliaPlots/Plots.jl/issues/734).

test/install_wkhtmltoimage.sh
```diff
@@ -7,7 +7,8 @@ sudo apt-get -qq update
 
 sudo apt-get install -y xfonts-75dpi
 wget http://download.gna.org/wkhtmltopdf/0.12/0.12.2/wkhtmltox-0.12.2_linux-trusty-amd64.deb
-sudo dpkg -i wkhtmltox-0.12.2_linux-trusty-amd64.deb
+sudo dpkg -i wkhtmltox-0.12.2_linux-trusty-amd64.deb || true
```
tkelman: why was this failing, wouldn't it be better to address the issue instead of letting it fail and ignoring it?

darwindarak: It was missing a dependency, which the next line resolves automatically.  I figured that if that doesn't work, then the tests will still fail when it tries to run `wkhtmltoimage`.

tkelman: ah. debs are screwy. is there a specific flag to dpkg that could tell it to only ignore dependency issues?

darwindarak: You're right, there's a `--depends` flag to turn all dependency issues into warnings.

mkborregaard:
@daschw What do you say to this PR? Is the `force-depends` an alternative way of solving the issues we've had with wkhtmltoimage? I'm for adding a 0.6 test now. I don't think we should remove the checkout of PlotUtils, but we can install StatPlots normally.

daschw:
>  Is the force-depends an alternative way of solving the issues we've had with wkhtmltoimage?

We had 2 issues with wkhtmltoimage. 1. The server http://download.gna.org/... was/is not available. So I changed the link. 2. There was a missing dependency for wkhtmltox (xfonts-base). I added this dependency [here](https://github.com/JuliaPlots/Plots.jl/blob/master/test/install_wkhtmltoimage.sh#L8), and @darwindarak turned the dependency error to a warning with `--force-depends` and then installed all dependencies with `sudo apt-get -f install`.

>  I don't think we should remove the checkout of PlotUtils, but we can install StatPlots normally.

I agree. I don't know what the reason for removing `Pkg.checkout("PlotUtils")` was, though.

mkborregaard:
All in all I think the conclusion is that this PR can be closed. Thanks for the PR, though, @darwindarak nothing wrong with it we just chose a different way around the issues.

PR #806, opened 2017-05-08
0.6 deprecations

dpsanders:


dpsanders:
To fix deprecations with `abstract` type declarations, I will need to use `Compat.jl`. Is that acceptable?

mkborregaard:
We'll push a new release in a few days, and then master and future releases will be 0.6 only. So there is no reason for Compat. We'll just merge this after the release. 
And thanks a lot for this!

daschw:
It is really only line 9 in series.jl. Changing that back to `typealias ...` allows precompilation again.

daschw:
cf. #823 and #824

PR #807, opened 2017-05-08
gr: use correct colorbar scaling

jheinen:


mkborregaard:
@jheinen This worked for heatmap and contourf, but the other seriestypes using colors via marker_z (and line_z?) are still bounded between 0 and 1. Not sure where to insert the `GR.setspace(zmin, zmax, 0, 90)` command to fix this :-)

PR #815, opened 2017-05-11
Updates for release 0.11

mkborregaard:


REQUIRE
```diff
@@ -1,8 +1,8 @@
 julia 0.5
 
 RecipesBase
-PlotUtils
-PlotThemes
+PlotUtils 0.4
```
darwindarak: Should we remove `Pkg.checkout(PlotUtils)` from `travis_commands.jl` then?

mkborregaard: @pkofod is the guy who knows about Travis :-)

darwindarak: I've added the change to #805.

Issue #826, opened 2017-05-15
yaxis limits cut in muliple series bar plots - unable to set ylims


daschw:
```julia
using Plots
bar([2 1; 2 1])
```
produces this:
![bar_cropped](https://cloud.githubusercontent.com/assets/16589944/26058641/042c874c-397f-11e7-8d35-66deed76336f.png)
The yaxis limits are set based on the last series and not based on the one with the highest input.
```julia
bar([2 1; 2 1], ylims = (0, 2))
```
does not change anything. The only workaround is:
```julia
bar([2 1; 2 1])
ylims!((0, 2))
```
cc: @oschulz

oschulz:
I'll look into it - it should, of course, be possible to override the ylims.

oschulz:
See #827.

oschulz:
Now looks like this:

```julia
bar([2 1; -2 -1])
```

![bars-two-plots](https://cloud.githubusercontent.com/assets/546147/26080397/b078b876-39c6-11e7-8a98-e902cbfde0c9.png)

oschulz:
Thanks for reporting, @daschw!

daschw:
Awesome

Issue #511, opened 2016-09-30
Add `widen` keyword argument


mkborregaard:
Bar plots in the pyplot and gr backends fit the bars exactly to the sides of the subplot, which is not very aesthetically nice. I would suggest adding an option to
- [ ] widen the plot along the x axis to give a little space on each side of the bars
- [ ] widen the plot along the top, but not the bottom of the y axis for bar plots and histograms, to give a little space there.

On gitter I mentioned the specification for this general type of behaviour in R:

```
‘xaxs’ The style of axis interval calculation to be used for the
x-axis. Possible values are ‘"r"’, ‘"i"’, ‘"e"’, ‘"s"’,
‘"d"’. The styles are generally controlled by the range of
data or ‘xlim’, if given.
Style ‘"r"’ (regular) first extends the data range by 4
percent at each end and then finds an axis with pretty labels
that fits within the extended range.
Style ‘"i"’ (internal) just finds an axis with pretty labels
that fits within the original data range.
Style ‘"s"’ (standard) finds an axis with pretty labels
within which the original data range fits.
Style ‘"e"’ (extended) is like style ‘"s"’, except that it is
also ensures that there is room for plotting symbols within
the bounding box.
Style ‘"d"’ (direct) specifies that the current axis should
be used on subsequent plots.
(Only ‘"r"’ and ‘"i"’ styles have been implemented in R.)
```

One issue is that the seriestype is not always known at the time of calling the backend(?), so it is difficult to give barplots a special behaviour (in this case, not giving space on the bottom of the y axis) compared to other plot types.

An example here of a really simple barplot as they often look (from today's Nature)
![simple barplot example](https://cloud.githubusercontent.com/assets/8429802/18989981/5ad9cd2a-870f-11e6-8073-d0cc7645ca00.png)

mkborregaard:
gr() leaves no space between bars either, which is misleading - that should be standard behaviour for histograms, but never bar plots, as bars represent categorical variables.

tbreloff:
This was my earlier thinking:

> maybe it can be an axis attribute widen, defaulting to :auto which keeps current behavior, true would always widen, false would never, and a number would widen by that percentage?

mkborregaard:
Oh, yes, that bit from our gitter chat was important to include here. That is a great interface I think. `:top_and_sides` might be the interface for what I suggest for bar plots?

pfitzseb:
What would you think about doing this the like css: `widen` is a tuple with either 
[...]

PR #827, opened 2017-05-15
Remove ylims default setting for bar and histogram plots

oschulz:
Setting `ylims -> ...` doesn't play well with multiple series and cannot be properly overridden by user (see #826).

I don't like the fact that by default the plots don't leave a little bit of space above the highest value, but looks like `ylims` isn't the right way to address it.

### DIFF
--- a/src/recipes.jl
+++ b/src/recipes.jl
@@ -323,7 +323,7 @@ end
 
 # create a bar plot as a filled step function
 @recipe function f(::Type{Val{:bar}}, x, y, z)
-    procx, procy, xscale, yscale, baseline, wautolims = _preprocess_barlike(d, x, y)
+    procx, procy, xscale, yscale, baseline = _preprocess_barlike(d, x, y)
     nx, ny = length(procx), length(procy)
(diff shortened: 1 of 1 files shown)


daschw:
Great, thanks!

mkborregaard:
@oschulz I agree, there should be more space (see this issue from way back when, which is still open https://github.com/JuliaPlots/Plots.jl/issues/511). 
The right way to do it is not in the recipe, but in https://github.com/JuliaPlots/Plots.jl/blob/master/src/axes.jl#L368-L416 .
The issue is here that 1) behaviour should be different for x and y axes, and be different for seriestypes - e.g. `bar` should only expand ylims above and not below. So the logic behind this code slightly collides with our intuitive aesthetic preference. Perhaps, suggested in the issue, an extra `widen` keyword that could be set in the recipe and be used by the code in axes.jl could be a practical solution?

Oh, and a hint - if you write ": Fixes #826 " in the title of the PR, it becomes automatically linked to the issue and the issue will close when the PR gets merged.

PR #829, opened 2017-05-16
merge master into reorg

mkborregaard:


mkborregaard:
The reorg branch is an attempt to resolve the issues with precompilation on 0.6. I suggest merging master rather than rebasing, as this is a public branch.

PR #834, opened 2017-05-17
Move new PlotExample to the end for VRT testing

daschw:
I moved the new PlotExample from #758 to the end.
Otherwise in the VRT you have to compare apples to oranges (starting with example 19).
Is there something speaking against this "change", like some intended order of the PlotExamples?

Also I skipped this new example in the VRT, because being an animation it takes a little longer than the other examples. However, it's not that long and it's probably a good thing to have more ReferenceImages for testing, so I could also include it. What do you think @mkborregaard ?

### DIFF
--- a/src/examples.jl
+++ b/src/examples.jl
@@ -172,24 +172,6 @@ PlotExample("",
     end)]
 ),
 
-PlotExample("Animation with subplots",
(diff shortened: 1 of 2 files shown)


mkborregaard:
I think it's great to do like you've done here

PR #835, opened 2017-05-17
WIP Allow widen on log axes

mkborregaard:


src/axes.jl
```diff
@@ -366,18 +366,20 @@ end
...
-    lmin-eps, lmax+eps
+    isf(lmin-eps), isf(lmax+eps)
 end
 
 # figure out if widening is a good idea.  if there's a scale set it's too tricky,
 # so lazy out and don't widen
 function default_should_widen(axis::Axis)
     should_widen = false
-    if axis[:scale] == :identity && !is_2tuple(axis[:lims])
+    if (!(axis[:scale] in _logScales && axis[:extrema].emin <= 0)) && !is_2tuple(axis[:lims])
```
oschulz: Does that mean that widening doesn't happen for `axis[:scale] == :identity` any longer (I don't quite know what widening does, I have to admit).

mkborregaard: No it doesn't, it means it would allow it for log scales as well, but with a provision for not doing it if min is 0 (which it never should be in the first place if the axis is log).
`widen` is the function that creates a little space around the points, so that the limit of the plotted objects don't intersect the axis limits.

oschulz:
Currently, this gives me trouble with log plots (maybe to be expected, as this is still WIP?).

Current master:

hists-modes-logy-demo-master

master merged with #835:

hists-modes-logy-demo-835

mkborregaard:
Thanks for reporting, I haven't tested this carefully. It is definitely still WIP.

mkborregaard:
I think this has been deprecated by #1572 and can be closed?

daschw:
Yes.

PR #837, opened 2017-05-17
resetfontsizes: fixes #836

### LINKED ISSUE #836: Feature Request: `Plots.resetfontsizes()`
I have this setup:

```julia
module test
    using Plots
    # want: Plots.resetfontsizes()
    Plots.scalefontsizes(2)
    pyplot()
    f() = plot(rand(10))
end
```

the problem is that each time I reload the module, the fontsize increases by a factor of 2. I am looking for a function `Plots.resetfontsizes()` that resets all fonts to default. This is related to issue #560

floswald:
added const copies of _all_defaults and _axis_defaults to be able to reset fonts to initial values with new  `resetfontsizes` method. 

## Example

```julia
Plots.plot(rand(10))
Plots.scalefontsizes(2)
Plots.plot(rand(10))
Plots.resetfontsizes()
Plots.plot(rand(10))
```

### DIFF
--- a/src/args.jl
+++ b/src/args.jl
@@ -336,10 +336,10 @@ const _all_defaults = KW[
 ]
 
 # to be able to reset font sizes to initial values
-const _initial_fontsizes = Dict(:titlefont  => deepcopy(_subplot_defaults[:titlefont]),
-                                :legendfont => deepcopy(_subplot_defaults[:legendfont]),
-                                :tickfont   => deepcopy(_axis_defaults[:tickfont]),
-                                :guidefont  => deepcopy(_axis_defaults[:guidefont]))
+const _initial_fontsizes = Dict(:titlefont  => _subplot_defaults[:titlefont].pointsize,
+                                :legendfont => _subplot_defaults[:legendfont].pointsize,
+                                :tickfont   => _axis_defaults[:tickfont].pointsize,
+                                :guidefont  => _axis_defaults[:guidefont].pointsize)
 
 const _all_args = sort(collect(union(map(keys, _all_defaults)...)))
 
@@ -523,25 +523,6 @@ end
 
 function default(d::KW, k::Symbol)
     get(d, k, default(k))
-end
-
-# reset the defaults globally to values at startup
-
-"""
-`initial(key)` returns the intial value for that key
-"""
-
-function initial(k::Symbol)
-    k = get(_keyAliases, k, k)
-    for defaults in _all_initial_defaults
-        if haskey(defaults, k)
-            return defaults[k]
-        end
-    end
-    if haskey(_axis_initial_defaults, k)
-        return _axis_initial_defaults[k]
-    end
-    k in _suppress_warnings || error("Unknown key: ", k)
 end
 
 
--- a/src/args.jl
+++ b/src/args.jl
@@ -335,9 +335,11 @@ const _all_defaults = KW[
     _axis_defaults_byletter
 ]
 
-# to be able to reset things to initial values
-const _all_initial_defaults = deepcopy(_all_defaults)
-const _axis_initial_defaults = deepcopy(_axis_defaults)
+# to be able to reset font sizes to initial values
+const _initial_fontsizes = Dict(:titlefont  => deepcopy(_subplot_defaults[:titlefont]),
+                                :legendfont => deepcopy(_subplot_defaults[:legendfont]),
+                                :tickfont   => deepcopy(_axis_defaults[:tickfont]),
+                                :guidefont  => deepcopy(_axis_defaults[:guidefont]))
 
 const _all_args = sort(collect(union(map(keys, _all_defaults)...)))
 
--- a/src/args.jl
+++ b/src/args.jl
@@ -335,6 +335,10 @@ const _all_defaults = KW[
     _axis_defaults_byletter
 ]
 
+# to be able to reset things to initial values
+const _all_initial_defaults = deepcopy(_all_defaults)
+const _axis_initial_defaults = deepcopy(_axis_defaults)
+
 const _all_args = sort(collect(union(map(keys, _all_defaults)...)))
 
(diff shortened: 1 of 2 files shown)


mkborregaard:
You instead of adding a new function, define a new method of scalefontsizes that takes a keyword, e.g. the symbol `:default`. Maybe write a small docstring for the overall function while you're at it?

mkborregaard:
I've added some practical text to the title - wip prevents it from getting merged prematurely, "fixes" links it to the issue so they are updated together and the issue closed when this is merged.

src/args.jl
```diff
@@ -335,6 +335,12 @@ const _all_defaults = KW[
     _axis_defaults_byletter
 ]
 
+# to be able to reset font sizes to initial values
+const _initial_fontsizes = Dict(:titlefont  => deepcopy(_subplot_defaults[:titlefont]),
```
mkborregaard: You don't need deepcopy

src/args.jl
```diff
@@ -519,6 +525,24 @@ function default(d::KW, k::Symbol)
     get(d, k, default(k))
 end
 
+# reset the defaults globally to values at startup
+
+"""
+`initial(key)` returns the intial value for that key
+"""
```
mkborregaard: No space after docstrings

src/components.jl
```diff
@@ -303,6 +303,17 @@ function scalefontsizes(factor::Number)
     end
 end
 
+"Resets font sizes to initial default values"
+function scalefontsizes()
+  for k in (:titlefont, :guidefont, :tickfont, :legendfont)
+      f = default(k)
+      if haskey(_initial_fontsizes,k)
+        factor = f.pointsize / _initial_fontsizes[k].pointsize
```
mkborregaard: use `k in keys(_initial_fontsizes)` instead - `haskey` is deprecated on 0.6

src/args.jl
```diff
@@ -519,6 +525,24 @@ function default(d::KW, k::Symbol)
...
+
+function initial(k::Symbol)
```
mkborregaard: this function looks like it's not needed at all?

src/args.jl
floswald: are you sure about that? just doing `Dict(:titlefont  =>_subplot_defaults[:titlefont])` refers to the original type, hence gets changed along with `default(k)`. I tried `copy` but there is no method for the `fonts` type. maybe i dont understand.

floswald: yes, I can confirm this. you need a deepcopy (if you copy the entire type like I do). i'll chnage that.

src/components.jl
```diff
@@ -307,8 +307,8 @@ end
 function scalefontsizes()
   for k in (:titlefont, :guidefont, :tickfont, :legendfont)
       f = default(k)
-      if haskey(_initial_fontsizes,k)
-        factor = f.pointsize / _initial_fontsizes[k].pointsize
+      for k in keys(_initial_fontsizes)
```
mkborregaard: Did I write `for`? I meant `if k in keys`

mkborregaard: althout here you actually know that k is in keys, so the check can be deleted.

floswald: oh sorry, i didn't get that. will change.

src/components.jl
```diff
@@ -303,6 +303,17 @@ function scalefontsizes(factor::Number)
     end
 end
 
+"Resets font sizes to initial default values"
```
mkborregaard: This doctext will not display properly - when you `?scalefontsizes` at the REPL it will print out all the docstrings, but not which methods they're associated with. So you need to do something like
"""
    scalefontsizes([x])
Increase the default font size by a factor `x`. The non-argument form restores the defaults
"""

mkborregaard:
It's starting to look great, but let me just throw out a thought here - is the behaviour where font sizes keep scaling every time the function is called really the ideal? The alternative is to always set font sizes to the initial defaults * x, supplying 1 would recreate the defaults. I haven't thought this through, just thought I'd throw this out while we're at it.

floswald:
yeah i agree. it's a bit confusing to to `scalefontsizes(2)` and then `scalefontsizes(0.9)` if you realized it was too big. i think that your suggestion would be the right way to do it, yes. so i change `scalefontsizes` to that effect.

mkborregaard:
Let us just think it through. Here is the issue with doing that - imagine that you set the default legendfont to be smaller, in comparison to the rest. Calling scalefontsizes in the implementation I suggested above would then override that default and give all the values the hardcoded defaults set in args.jl.

floswald:
yes that's right. you would always return to the initial setup of font sizes. hmm, maybe a second method `resetfontsizes!` would be much clearer in the end, instead of the keyword? we would keep the current scalefontsizes, no keyword, same functionality. if you call `resetfontsizes!` you go back to the start, if you want to keep your personal settings, you'll have to fiddle around with the right factor for scalefontsizes.

mkborregaard:
For now, I suggest you implement the PR as it is now (with the docstring). That's the least intrusive operation, it's documented and it creates no nasty surprises. You could open the different functionality as an issue - how does that sound?

mkborregaard:
Did you loose steam a little here @floswald ? I am happy to be persuaded if you think my suggestion wasn't good :-)

floswald:
I'm working on a big job at the moment with a deadline so had to
prioritize. Happy to finish this PR but not before end of next week.

I like your suggestion! :-)

mkborregaard:
perfect!

mkborregaard:
To all open PRs: Due to the change to 0.6 it appears that all PRs will have to be rebased onto master for Travis tests to run.

mkborregaard:
Nice. Just let me know when this is ready for final review and merge.

floswald:
That's it, good to go!

mkborregaard:
Thanks @floswald ! welcome to the club of Plots contributors :-)

PR #843, opened 2017-05-19
Change default backends

mkborregaard:
The backend order is out of date. PyPlot and PlotlyJS will error on 0.6

### DIFF
--- a/src/backends.jl
+++ b/src/backends.jl
@@ -148,7 +148,7 @@ function pickDefaultBackend()
     # the ordering/inclusion of this package list is my semi-arbitrary guess at
     # which one someone will want to use if they have the package installed...accounting for
     # features, speed, and robustness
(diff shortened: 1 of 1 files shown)


ChrisRackauckas:
And the GR GUI will stall if you open it, close it, and then try to open it again. If you're looking for a stable default to put on top, that isn't it.

ChrisRackauckas:
Also, its default install doesn't present itself with a good image to Windows users:

https://github.com/jheinen/GR.jl/issues/60

mkborregaard:
I'll merge this as a temporary fix until things work better on 0.6 with PyPlot. I am happy to take a larger discussion of default backends later. My candidate would be GR in the future too, if JSTerm lives up to its promises but not with the current GKS term. If someone would take responsibility of adressing the open Plotly issues I'd be happy with that instead.

Issue #844, opened 2017-05-19
`Pkg.build("Plots")` fails on v0.6 and v0.7 (Windows?)


ChrisRackauckas:
```julia
julia> plot(rand(10,10))
WARNING: Base.cycle is deprecated, use Base.Iterators.cycle instead.
  likely near no file:0
WARNING: Base.cycle is deprecated, use Base.Iterators.cycle instead.
  likely near no file:0
WARNING: Base.cycle is deprecated, use Base.Iterators.cycle instead.
  likely near no file:0
WARNING: Base.cycle is deprecated, use Base.Iterators.cycle instead.
  likely near no file:0
WARNING: Couldn't initialize plotly.  (might need to install it?)
INFO: To do a standard install of plotly, copy and run this:

Pkg.build("Plots")


ERROR: ArgumentError: Module JSON not found in current path.
Run `Pkg.add("JSON")` to install the JSON package.
Stacktrace:
 [1] _require(::Symbol) at .\loading.jl:428
 [2] require(::Symbol) at .\loading.jl:398
 [3] _initialize_backend(::Plots.PlotlyBackend) at C:\Users\Chris\.julia\v0.7\Plots\src\backends\plotly.jl:64
 [4] backend() at C:\Users\Chris\.julia\v0.7\Plots\src\backends.jl:182
 [5] backend(::Symbol) at C:\Users\Chris\.julia\v0.7\Plots\src\backends.jl:209
 [6] pickDefaultBackend() at C:\Users\Chris\.julia\v0.7\Plots\src\backends.jl:158
 [7] backend() at C:\Users\Chris\.julia\v0.7\Plots\src\backends.jl:171
 [8] Plots.Plot() at C:\Users\Chris\.julia\v0.7\Plots\src\types.jl:85
 [9] #plot#273 at C:\Users\Chris\.julia\v0.7\Plots\src\plot.jl:50 [inlined]
 [10] plot(::Array{Float64,2}) at C:\Users\Chris\.julia\v0.7\Plots\src\plot.jl:46
```

I can't get the automatic build to work on v0.6 or v0.7 on a Windows 10 machine (didn't try others yet). Then trying `Pkg.build("Plots")` doesn't work either. A manual `Pkg.add("Plotly")` does work, but that should be unnecessary given the Plots build script.

ChrisRackauckas:
I checked my `deps` folder and `plotly-latest.min` successfully downloaded.

[...]

PR #845, opened 2017-05-19
Add JSON to REQUIRE

ChrisRackauckas:
Fixes https://github.com/JuliaPlots/Plots.jl/issues/844 . JSON should really be required if this backend is installed by default and requires it to work.

### DIFF
--- a/REQUIRE
+++ b/REQUIRE
@@ -8,3 +8,4 @@ FixedSizeArrays
 Measures
 Showoff
 StatsBase 0.14.0
+JSON


ChrisRackauckas:
I think something before might have actually been doing this anyways, which is why it "accidentally" worked.

mkborregaard:
This should be merged at the 0.6 transition IMO.

ChrisRackauckas:
Why is that so? I think this should even be backported: the older Plotly stuff only worked because a dependency seemingly installed JSON for Plots. That kind of dependency handling is supposed to not happen. Plots gets a little weird because of conditional dependencies, but Plotly installs with Plots and so isn't really a conditional dependency, and this only worked through a fluke of the package system which no longer works. This also means that if whatever package was giving it JSON tags v0.5-compatible release that removes the requirement, the Plotly default will retroactively break on v0.5.

mkborregaard:
If that's the case, I agree it should be merged immediately. Thanks @ChrisRackauckas !

ChrisRackauckas:
I found out where the mysterious JSON.jl was coming from before: Juno and other editor tools like VSCode. All of those use Electron tools, so they have to have JSON.jl. So if someone was using an IDE, JSON was already installed,  and `Pkg.build("Plots")` would work. That covered most cases, so I don't think this got noticed. But I had the failures when I had a REPL only installation, without any other backends. While rare, this is the case where the backup Plotly itself would've failed, because in this case the user would not have accidentally gotten JSON.jl from an IDE. Just thought it should be documented.

PR #808, opened 2017-05-09
Hack around world age issue PyPlot (fix #734)

darwindarak:
This currently uses `Base.invokelatest` to get around the world age problem.  It decreases the performance by quite a bit though.  Before the PR (but after adding an `eval` before [`PyPlot.ioff()](https://github.com/JuliaPlots/Plots.jl/blob/master/src/backends/pyplot.jl#L90)), we have

```julia
julia> backend()
Plots.PyPlotBackend()

julia> @benchmark plot($rand(10))
BenchmarkTools.Trial:
  memory estimate:  117.05 KiB
  allocs estimate:  2008
  --------------
  minimum time:     426.443 μs (0.00% GC)
  median time:      432.476 μs (0.00% GC)
  mean time:        446.854 μs (2.76% GC)
  maximum time:     2.963 ms (81.89% GC)
  --------------
  samples:          10000
  evals/sample:     1
```
and after the PR, we have
```julia
julia> @benchmark plot($rand(10))
BenchmarkTools.Trial:
  memory estimate:  120.58 KiB
  allocs estimate:  2063
  --------------
  minimum time:     718.450 μs (0.00% GC)
  median time:      727.692 μs (0.00% GC)
  mean time:        751.499 μs (2.02% GC)
  maximum time:     4.293 ms (73.99% GC)
  --------------
  samples:          6627
  evals/sample:     1
```

Hopefully this isn't the permanent solution.  This PR is mainly to get a discussion going around the issue.

PR #848, opened 2017-05-20
RFC: fix world-age problem in PyPlot initialization

darwindarak:
First step in #808.  After this PR, PyPlot still needs to be explicitly initialized.


mkborregaard:
Thanks, @darwindarak !

mkborregaard:
With this, I get the "_show is not defined for this backend" error from pyplot on 0.6, requiring a call to `gui()`

PR #849, opened 2017-05-21
Fix plot(::Matrix{RGB{Float64}}) in plotly backend

dlfivefifty:


mkborregaard:
Thanks, @dlfivefifty !

PR #853, opened 2017-05-21
fix passing ticks tuple on 0.6

daschw:
@kescobo reported in the gitter channel that passing a tuple to the ticks attribute like `scatter(1:3, xticks = (1:3, ["a","b","c"]))` is not working. This happens only on 0.6 and the reason for this is that
```julia
(1:3, ["a","b","c"]))::NTuple{2}
```
gives an error on 0.6, while it does not on 0.5.
This PR should resolve this issue.

### DIFF
--- a/src/axes.jl
+++ b/src/axes.jl
@@ -229,7 +229,7 @@ function get_ticks(axis::Axis)
     elseif typeof(ticks) <: AVec
         # override ticks, but get the labels
         optimal_ticks_and_labels(axis, ticks)
-    elseif typeof(ticks) <: NTuple{2}
+    elseif typeof(ticks) <: NTuple{2, Any}
         # assuming we're passed (ticks, labels)
         ticks
     else


kescobo:
@daschw wow, nice! Just tweaked that file on my local system and can confirm it works now

👍

mkborregaard:
great :-)

mkborregaard:
Should we also fix the case where just a Vector{String} is passed?

daschw:
Yes, I suppose. Do you know how to resolve this exactly? Otherwise I can look into this later today

mkborregaard:
No I'm not sure was just thinking about it. It's got to do with the handling here https://github.com/daschw/Plots.jl/blob/ce169edd08f273ba62eaa197aea3f38797b33ed0/src/axes.jl#L229 , I suppose.

daschw:
Yep, we probably need to split this into two different cases for strings and numbers.

daschw:
I have been trying to implement this, but then it occured to me that I was not even sure what we would expect from this case.

I think it's quite clear what we expect from something like
```julia
using Plots
xstring = ["x$i" for i in 1:10]
lab = ["lab&i" for i in 1:10]
x = 10:10:100
y = rand(10)

plot(y, xticks = lab)
plot(xstring, y, xticks = lab)
plot(x, y, xticks = lab)
```
and we could implement the functionality for those cases. However, all three commands result in the same plot which can already be easily created with `plot(lab, y)`.

If our input, on the other hand, is an arbitrary numeric vector, let's say `x = rand(10)`. What would we expect from 
```julia
plot(x, y, xticks = lab)
```
?
Should the xticks be at the data points positions? What if `lab` and `x` had different lengths? Should the xticks be evenly distrubuted over the plot area? Is there a reasonable default for such cases or shouldn't we just let the user handle such cases by additionally specifying the tick positions?

mkborregaard:
So, yes. good point. To specify labels you also need to specify positions.

PR #857, opened 2017-05-23
Replace `text` with `glvisualize_text`

mkborregaard:
glvisualize defined a `text` function that shadowed a method of Plots exported `text` function

### DIFF
--- a/src/backends/glvisualize.jl
+++ b/src/backends/glvisualize.jl
@@ -692,7 +692,7 @@ function draw_ticks(
     text, positions, offsets
 end
 
-function text(position, text, kw_args)
+function glvisualize_textposition, text, kw_args)
     text_align = alignment2num(text.font)
     startpos = Vec2f0(position)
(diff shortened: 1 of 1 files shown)


src/backends/glvisualize.jl
```diff
@@ -692,7 +692,7 @@ function draw_ticks(
     text, positions, offsets
 end
 
-function text(position, text, kw_args)
+function glvisualize_textposition, text, kw_args)
```
daschw: Isn't there a bracket missing?

mkborregaard: OMG!

mkborregaard: Sorry everyone. I should have tested first - trying to cram too much into too little time. (Also I am really looking forward to those tests working again :-) )

mkborregaard: Fixed in #858

PR #862, opened 2017-05-24
fix glvisualize backend 

SimonDanisch:
There is a bug in reactive, not allowing me to dynamically adapt the speed of rendering depending on user interaction. So I removed that and just use a fixed renderloop.
I also removed some shady signal clean up code which was also not working anymore. But that means, signals will get cleaned up less well, possibly slowing down event processing!

### DIFF
--- a/src/backends/glvisualize.jl
+++ b/src/backends/glvisualize.jl
@@ -98,41 +98,6 @@ end
 # end
 const _glplot_deletes = []
 
-function close_child_signals!(screen)
-    for child in screen.children
-        for (k, s) in child.inputs
-            empty!(s.actions)
-        end
-        for (k, cam) in child.cameras
-            for f in fieldnames(cam)
-                s = getfield(cam, f)
-                if isa(s, Signal)
-                    close(s, false)
-                end
-            end
-        end
-        empty!(child.cameras)
-        close_child_signals!(child)
-    end
-    return
-end
-function empty_screen!(screen)
-    if isempty(_glplot_deletes)
-        close_child_signals!(screen)
-        empty!(screen)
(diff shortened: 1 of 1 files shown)


mkborregaard:
So you think of this as a temporary patch? Still great to have it working!

SimonDanisch:
Should be!

mkborregaard:
Let's merge! 😺

mkborregaard:
I just tried out this branch (working from the REPL), and the glvisualize window behaves extremely strangely when resizing - are you aware of this @SimonDanisch ?

mkborregaard:
Some screenshots:
Default popup-window:
skaermbillede 2017-05-25 kl 09 43 55
After resizing the window by dragging with the mouse:
skaermbillede 2017-05-25 kl 09 44 10
Plotting again to the resized window:
skaermbillede 2017-05-25 kl 09 44 24

I wonder if part of the strange behaviour could be due to the 'pixel doubling' on my retina screen (Apple :heart_eyes:) ?

mkborregaard:
So sorry to keep pestering you, but... now that you're here: I guess we can now replace this warning with the standard installation instructions (`Pkg.add("GLVisualize")`)?
```
WARNING: Couldn't initialize glvisualize.  (might need to install it?)
INFO: To do a standard install of glvisualize, copy and run this:

For those incredibly brave souls who assume full responsibility for what happens next...
There's an easy way to get what you need for the GLVisualize backend to work (until Pkg3 is usable):

Pkg.clone("https://github.com/tbreloff/MetaPkg.jl")
using MetaPkg
meta_checkout("MetaGL")

See the MetaPkg readme for details...
```

PR #870, opened 2017-05-27
add ability to specify legend placement to gr backend

wfrgra:


mkborregaard:
@jheinen did you mention that you planned to base this functionality on GR's `location` keyword instead? How does this relate to that idea?

jheinen:
Yes. we are implementing a "complete" convenience layer. But, this will still take some time ... In a first step (and as a quick solution for `Plots.jl`), you should merge your commit ...

mkborregaard:
Great, will you merge this when it's ready? :-)

jheinen:
Yes. This will allow us to use the same code base for different wrappers (Julia, Python) and environments (console, browser). But it'll take some further development ...

jheinen:
... and that's why the JS wrapper takes some more time than expected :-(

mkborregaard:
thanks, @wfrgra !

PR #876, opened 2017-05-30
RFC: Ignore NaNs on 0.6 (like on 0.5): fixes #796

### LINKED ISSUE #796: xlim and ylim defaults are incorrect on 0.6
On 0.5, the defaults for `xlim` and `ylim` take into account all elements of the plot and try to accommodate them all; on 0.6, they seem to default to (0, 1).

mkborregaard:
Bases summary functions on NaNMath:
min, max, extrema, minimum, maximum

### DIFF
--- a/src/Plots.jl
+++ b/src/Plots.jl
@@ -10,6 +10,7 @@ using Base.Meta
 @reexport using PlotThemes
 import Showoff
 import StatsBase
+import NaNMath # define functions that ignores NaNs. To overcome the destructive effects of https://github.com/JuliaLang/julia/pull/12563
 
 export
     grid,
@@ -103,15 +104,6 @@ export
     BezierCurve,
 
     plotattr
-
-# ---------------------------------------------------------
-
-import NaNMath
-# define functions (e.g. `_extrema`, that ignores NaNs, when the type is applicable. To overcome the destructive effects of https://github.com/JuliaLang/julia/pull/12563
-for fun in (:extrema, :minimum, :maximum)
-       @eval $(Symbol(string("_",fun))){F<:AbstractFloat}(x::AbstractArray{F}) = NaNMath.$(fun)(x)
-       @eval $(Symbol(string("_",fun)))(x) = Base.$(fun)(x)
-end
 
 # ---------------------------------------------------------
 
--- a/src/Plots.jl
+++ b/src/Plots.jl
@@ -106,9 +106,10 @@ export
 
 # ---------------------------------------------------------
 
+import NaNMath
 # define functions (e.g. `_extrema`, that ignores NaNs, when the type is applicable. To overcome the destructive effects of https://github.com/JuliaLang/julia/pull/12563
 for fun in (:extrema, :minimum, :maximum)
-       @eval $(Symbol(string("_",fun))){F<:AbstractFloat}(x::AbstractArray{F}) = Base.$(fun)(filter(!isnan,x))
+       @eval $(Symbol(string("_",fun))){F<:AbstractFloat}(x::AbstractArray{F}) = NaNMath.$(fun)(x)
        @eval $(Symbol(string("_",fun)))(x) = Base.$(fun)(x)
 end
 
--- a/src/Plots.jl
+++ b/src/Plots.jl
@@ -106,8 +106,8 @@ export
 
 # ---------------------------------------------------------
 
-# define functions (e.g. `_extrema`, that uses the NaNMath version (which ignores NaNs)) when the type is applicable
-for fun in (:extrema, :minimum, :maximum, :mean)
+# define functions (e.g. `_extrema`, that ignores NaNs, when the type is applicable. To overcome the destructive effects of https://github.com/JuliaLang/julia/pull/12563
+for fun in (:extrema, :minimum, :maximum)
        @eval $(Symbol(string("_",fun))){F<:AbstractFloat}(x::AbstractArray{F}) = Base.$(fun)(filter(!isnan,x))
        @eval $(Symbol(string("_",fun)))(x) = Base.$(fun)(x)
 end
--- a/src/Plots.jl
+++ b/src/Plots.jl
@@ -10,7 +10,7 @@ using Base.Meta
 @reexport using PlotThemes
 import Showoff
 import StatsBase
-import NaNMath: extrema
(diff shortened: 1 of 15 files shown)


src/Plots.jl
```diff
@@ -106,6 +106,15 @@ export
 
 # ---------------------------------------------------------
 
+import NaNMath
+# define functions (e.g. `_extrema`, that ignores NaNs, when the type is applicable. To overcome the destructive effects of https://github.com/JuliaLang/julia/pull/12563
+for fun in (:extrema, :minimum, :maximum)
```
mkborregaard: An alternative to the eval magic would be to just write it out - less concise, but probably easier for people wondering where `_minimum` is defined and what it does. Thoughts?

mkborregaard:
Marked WIP because this awaits a release of NaNMath

src/Plots.jl
```diff
@@ -106,6 +106,15 @@ export
...
+       @eval $(Symbol(string("_",fun))){F<:AbstractFloat}(x::AbstractArray{F}) = NaNMath.$(fun)(x)
```
dpsanders: You don't need `string` here:
```jl
julia> f = :hello
julia> Symbol("_", f)
:_hello
```

src/Plots.jl
dpsanders: Probably preferable to write it out explicitly.

mkborregaard: I know, right? It just feels like you should use the classic metaprogramming example when there is a chance :-) Oh well.

daschw: Just out of curiosity: Why didn't you define a `_min` and `_max` here?

mkborregaard:
Great! I didn't define those because the min and max functions in NaNMath are compliant will all of the types of the Base functions, so no reason to dispatch via a local function in Plots. But I am actually right now looking into removing it for the other functions too, so it calls the NaNMath versions explicitly where applicable. It'll require a bit more review, but be more transparent and easier to maintain, I think.

src/backends/glvisualize.jl
```diff
@@ -1382,7 +1382,7 @@ end
 
 
 function gl_heatmap(x,y,z, kw_args)
-    get!(kw_args, :color_norm, Vec2f0(extrema(z)))
+    get!(kw_args, :color_norm, Vec2f0(NaNMath.extrema(z)))
```
mkborregaard: @SimonDanisch I have replaced `minimum`, `mean`, `maximum`, `max` and `min` with versions from the NaNMath package that ignore `NaN`, to maintain the julia-0.5 behaviour of these functions. I have only done the changes where I believe the argument is likely to be a subtype of `AbstractFloat` or `AbstractArray{<:AbstractFloat}` as I believe this is the most likely place to find `NaN`s. Can you check that 1. all of the changes appear valid, and 2. that we don't need to identify NaN-compliant versions for other types? E.g. in line 1428 you take the `minimum` of an `AABB` object, but I'm guessing that will never be NaN?

src/backends/gr.jl
```diff
@@ -824,7 +824,7 @@ function gr_display(sp::Subplot{GRBackend}, w, h, viewport_canvas)
             # create the colorbar of contour levels
             if sp[:colorbar] != :none
                 gr_set_viewport_cmap(sp)
-                l = round(Int32, 1000 + (h - minimum(h)) / (maximum(h) - minimum(h)) * 255)
+                l = round(Int32, 1000 + (h - NaNMath.minimum(h)) / (NaNMath.maximum(h) - NaNMath.minimum(h)) * 255)
```
mkborregaard: @SimonDanisch I have replaced minimum, mean, maximum, max and min with versions from the NaNMath package that ignore NaN, to maintain the julia-0.5 behaviour of these functions. I have only done the changes where I believe the argument is likely to be a subtype of AbstractFloat or AbstractArray{<:AbstractFloat} as I believe this is the most likely place to find NaNs. Can you check that 1. all of the changes appear valid, and 2. that we don't need to identify NaN-compliant versions for other types? I am particular uncertain about this line.

src/components.jl
```diff
@@ -645,8 +645,8 @@ function (bc::BezierCurve)(t::Real)
     p
 end
 
-Base.mean(x::Real, y::Real) = 0.5*(x+y)
-Base.mean{N,T<:Real}(ps::FixedSizeArrays.Vec{N,T}...) = sum(ps) / length(ps)
+# mean(x::Real, y::Real) = 0.5*(x+y) #commented out as I cannot see this used anywhere and it overwrites a Base method with different functionality
```
mkborregaard: @tbreloff I am removing two functions here that 1. overwrite Base types, 2. are a little hassle to define NaN-compliant functions for, and 3. don't appear to be used anywhere in the code.

src/recipes.jl
```diff
@@ -550,7 +550,7 @@ Plots.@deps stepbins path
 function _auto_binning_nbins{N}(vs::NTuple{N,AbstractVector}, dim::Integer; mode::Symbol = :auto)
     _cl(x) = max(ceil(Int, x), 1)
```
mkborregaard: @oschulz the `ceil(Int, x)` but will fail if x is ever NaN. I assume this is not a problem? (just checking)

oschulz: Hm, maybe we better do

    _cl(x) = ceil(Int, max(x, one(x)))

instead, just in case. Well spotted!

mkborregaard: So, on 0.6 we'll have to do `_cl(x) = ceil(Int, NaNMath.max(x, one(x)))` - that is the whole point of this PR :-D
I'll put in the change.

oschulz: Sounds good!

src/backends/gr.jl
jheinen: I checked the GR backend code and I don't see any reason why not use the NaN-compliant functions. You can safely replace them. There is another call in line 642:

```julia
rmin, rmax = GR.adjustrange(minimum(r), maximum(r))
```

mkborregaard: Thank you!

mkborregaard:
I am going to go ahead and merge this for some forward progress towards the 0.6 version.

dpsanders:
Thanks for your great work on this!

mkborregaard:
Thanks a lot for the help!

PR #877, opened 2017-05-31
Make tests pass for GR and PyPlot

daschw:
I also edited the travis commands and added the packages, which I think are required. I hope that this will work.

This also requires the new PlotReferenceImages https://github.com/JuliaPlots/PlotReferenceImages.jl/pull/1.

### DIFF
--- a/test/runtests.jl
+++ b/test/runtests.jl
@@ -114,12 +114,12 @@ facts("Axes") do
     @fact typeof(axis) --> Plots.Axis
     @fact Plots.discrete_value!(axis, "HI") --> (0.5, 1)
     @fact Plots.discrete_value!(axis, :yo) --> (1.5, 2)
-    @fact extrema(axis) --> (0.5,1.5)
+    @fact Plots.NaNMath.extrema(axis) --> (0.5,1.5)
     @fact axis[:discrete_map] --> Dict{Any,Any}(:yo  => 2, "HI" => 1)
 
     Plots.discrete_value!(axis, ["x$i" for i=1:5])
     Plots.discrete_value!(axis, ["x$i" for i=0:2])
-    @fact extrema(axis) --> (0.5, 7.5)
+    @fact Plots.NaNMath.extrema(axis) --> (0.5, 7.5)
 end
 
 
--- a/test/runtests.jl
+++ b/test/runtests.jl
@@ -121,12 +121,12 @@ facts("Axes") do
     @fact typeof(axis) --> Plots.Axis
     @fact Plots.discrete_value!(axis, "HI") --> (0.5, 1)
     @fact Plots.discrete_value!(axis, :yo) --> (1.5, 2)
-    @fact Plots.NaNMath.extrema(axis) --> (0.5,1.5)
+    @fact Plots.ignoreNaN_extrema(axis) --> (0.5,1.5)
     @fact axis[:discrete_map] --> Dict{Any,Any}(:yo  => 2, "HI" => 1)
 
     Plots.discrete_value!(axis, ["x$i" for i=1:5])
     Plots.discrete_value!(axis, ["x$i" for i=0:2])
-    @fact Plots.NaNMath.extrema(axis) --> (0.5, 7.5)
+    @fact Plots.ignoreNaN_extrema(axis) --> (0.5, 7.5)
 end
 
 
--- a/test/runtests.jl
+++ b/test/runtests.jl
@@ -19,30 +19,70 @@ img_eps = isinteractive() ? 1e-2 : 10e-2
 #     image_comparison_facts(:gadfly, skip=[4,6,23,24,27], eps=img_eps)
 # end
 
-facts("PyPlot") do
-    @fact pyplot() --> Plots.PyPlotBackend()
-    @fact backend() --> Plots.PyPlotBackend()
-
-    image_comparison_facts(:pyplot, skip=[6,25,30,31], eps=img_eps)
-end
+# facts("GR") do
+#     @fact gr() --> Plots.GRBackend()
+#     @fact backend() --> Plots.GRBackend()
+#
+#     if is_linux() && isinteractive()
+#         image_comparison_facts(:gr, eps=img_eps)
+#     end
+# end
 
 facts("GR") do
     @fact gr() --> Plots.GRBackend()
     @fact backend() --> Plots.GRBackend()
 
-    if is_linux() && isinteractive()
-        image_comparison_facts(:gr, skip=[2,25,30,31], eps=img_eps)
-    end
+    image_comparison_facts(:gr,
+        skip=[
+            13, # :rtriangle not found
+        ],
+        eps=img_eps)
 end
 
-facts("Plotly") do
-    @fact plotly() --> Plots.PlotlyBackend()
-    @fact backend() --> Plots.PlotlyBackend()
+facts("PyPlot") do
+    @fact pyplot() --> Plots.PyPlotBackend()
+    @fact backend() --> Plots.PyPlotBackend()
 
-    # # until png generation is reliable on OSX, just test on linux
-    # @static is_linux() && image_comparison_facts(:plotly, only=[1,3,4,7,8,9,10,11,12,14,15,20,22,23,27], eps=img_eps)
+    image_comparison_facts(:pyplot, skip=[2,31], eps=img_eps)
 end
 
+facts("PlotlyJS") do
+    @fact plotlyjs() --> Plots.PlotlyJSBackend()
+    @fact backend() --> Plots.PlotlyJSBackend()
+
+    image_comparison_facts(:plotlyjs,
+        skip=[
+            2,  # animation
+            10, # match_dimensions not defined
+            27, # (polar plots) takes very long / not working
+            31, # animation
+        ],
+        eps=img_eps)
+end
+
+facts("InspectDR") do
+    @fact inspectdr() --> Plots.InspectDRBackend()
+    @fact backend() --> Plots.InspectDRBackend()
+
+    image_comparison_facts(:inspectdr,
+        skip=[
+            6,  # heatmap not defined
+            10, # heatmap not defined
+            22, # contour not defined
+            23, # pie not defined
+            28, # heatmap not defined
+        ],
+        eps=img_eps)
+end
+
+# facts("Plotly") do
+#     @fact plotly() --> Plots.PlotlyBackend()
+#     @fact backend() --> Plots.PlotlyBackend()
+#
+#     # # until png generation is reliable on OSX, just test on linux
+#     # @static is_linux() && image_comparison_facts(:plotly, only=[1,3,4,7,8,9,10,11,12,14,15,20,22,23,27], eps=img_eps)
+#     image_comparison_facts(:plotly, eps=img_eps)
+# end
 
 # facts("Immerse") do
 #     @fact immerse() --> Plots.ImmerseBackend()
--- a/test/runtests.jl
+++ b/test/runtests.jl
@@ -30,7 +30,12 @@ facts("PyPlot") do
     @fact pyplot() --> Plots.PyPlotBackend()
     @fact backend() --> Plots.PyPlotBackend()
 
-    image_comparison_facts(:pyplot, skip=[2,31], eps=img_eps)
+    image_comparison_facts(:pyplot,
+        skip=[
+            2,  # animation (skipped for speed)
+            31, # animation (skipped for speed)
+        ],
+        eps=img_eps)
 end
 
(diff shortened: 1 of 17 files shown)


daschw:
The travis check fails before the tests even start, because travis cannot download http://download.gna.org/wkhtmltopdf/0.12/0.12.2/wkhtmltox-0.12.2_linux-trusty-amd64.deb in test/install_wkhtmltoimage. How about replacing the link with https://downloads.wkhtmltopdf.org/0.12/0.12.2/wkhtmltox-0.12.2_linux-precise-amd64.deb?

mkborregaard:
I am clueless about these linux things, but if it works I am all for it. Isn't `precise` a fairly old dist-release though?

test/runtests.jl
```diff
@@ -19,30 +19,59 @@ img_eps = isinteractive() ? 1e-2 : 10e-2
...
+
+    image_comparison_facts(:gr, eps=img_eps)
+end
+
 facts("PyPlot") do
     @fact pyplot() --> Plots.PyPlotBackend()
     @fact backend() --> Plots.PyPlotBackend()
 
-    image_comparison_facts(:pyplot, skip=[6,25,30,31], eps=img_eps)
+    image_comparison_facts(:pyplot, skip=[2,31], eps=img_eps)
```
mkborregaard: I think 25 and 30, at least, depend on StatPlots (25 uses a DataFrame and 30 a violin plot), so they should be skipped, right? 

Also, It'd be nice to keep a record of which are skipped because they take time, and which aren't compatible with a given backend (so we can include those when processing examples for the documentation site).

test/runtests.jl
```diff
@@ -19,30 +19,59 @@ img_eps = isinteractive() ? 1e-2 : 10e-2
 #     image_comparison_facts(:gadfly, skip=[4,6,23,24,27], eps=img_eps)
 # end
 
+facts("GR") do
+    @fact gr() --> Plots.GRBackend()
+    @fact backend() --> Plots.GRBackend()
+
+    image_comparison_facts(:gr, eps=img_eps)
```
mkborregaard: no skipping?

test/runtests.jl
```diff
@@ -19,30 +19,59 @@ img_eps = isinteractive() ? 1e-2 : 10e-2
...
-    @fact backend() --> Plots.PlotlyBackend()
-
-    # # until png generation is reliable on OSX, just test on linux
-    # @static is_linux() && image_comparison_facts(:plotly, only=[1,3,4,7,8,9,10,11,12,14,15,20,22,23,27], eps=img_eps)
+facts("InspectDR") do
+    @fact inspectdr() --> Plots.InspectDRBackend()
+    @fact backend() --> Plots.InspectDRBackend()
+
+    image_comparison_facts(:inspectdr,
+        skip=[
```
mkborregaard: There are a lot of plots that look strange with inspectdr - when running the documentation I ended up skipping [4,6,8,10,22,23,24,25,27,28,29,30,31] . I guess skipping as few as possible, like you do here, makes sense. Still, I'd think at least 2, 25, 27, 30, 31 (like the others) should be skipped?

test/runtests.jl
```diff
@@ -19,30 +19,59 @@ img_eps = isinteractive() ? 1e-2 : 10e-2
...
-        image_comparison_facts(:gr, skip=[2,25,30,31], eps=img_eps)
+facts("PlotlyJS") do
+    @fact plotlyjs() --> Plots.PlotlyJSBackend()
+    @fact backend() --> Plots.PlotlyJSBackend()
+
+    if is_linux()
+        image_comparison_facts(:plotlyjs,
+            skip=[
+                2,  # animation
+                10, # match_dimensions not defined
```
mkborregaard: match_dimensions is just a warning, right? It still makes sense to skip it (we don't want warnings in the tests when they run right) but I am thinking whether we should adress this in the code - there is nothing wrong with just adding it to the supported keywords for the non-pyplot backends, I think (at least those, like GR and Pyplot, that call `transpose_z` in the backend code).

test/runtests.jl
```diff
@@ -19,30 +19,59 @@ img_eps = isinteractive() ? 1e-2 : 10e-2
...
+            28, # heatmap not defined
+        ],
+        eps=img_eps)
 end
 
+# facts("Plotly") do
+#     @fact plotly() --> Plots.PlotlyBackend()
+#     @fact backend() --> Plots.PlotlyBackend()
+#
+#     # # until png generation is reliable on OSX, just test on linux
```
mkborregaard: I think png generation on OSX has improved

test/runtests.jl
daschw: Oh, I did not consider the StatPlots dependency because everything just worked out-of-the-box locally. Thanks for the hint! I will skip 25 and 30 for all backends then.

test/runtests.jl
daschw: Everything worked for GR. But I will skip 25 and 30.

test/runtests.jl
daschw: Agreed!

test/runtests.jl
daschw: I think that this testimage really errored. I have look at the exact error message when I'm at home. Otherwise, I'm not sure whether we don't want warnings in the tests. They still pass.

test/runtests.jl
daschw: The tests for plotly did not work on my machine and I thought it would be OK to skip plotly if plotlyjs is working.

mkborregaard: agreed

daschw:
Yes, precise is a LTS version from 2012. If you look in system info of the travis log, you can see that travis uses precise.
I cannot assure you that it will work. I just looked for another available package file. There could still be dependency issues like in [this](https://travis-ci.org/JuliaPlots/Plots.jl/builds/234506742?utm_source=github_status&utm_medium=notification) travis log from 15 days ago. I think then we would also have to add `sudo apt-get install xfonts-base`.

mkborregaard:
Sounds like you're completely on top of that :-)

test/runtests.jl
daschw: StatPlots is used in [imgcomp.jl](https://github.com/JuliaPlots/Plots.jl/blob/master/test/imgcomp.jl#L17) which is included by runtests and installed in [travis_commands.jl](https://github.com/JuliaPlots/Plots.jl/blob/master/test/travis_commands.jl#L11) anyway, though.

mkborregaard: That's... slightly confusing :-) But thanks for noticing, you really have a sharp eye here.
Maybe it's good to do that, though we're only testing a bit of the functionality in StatPlots, we're at least testing the functionality that's in the juliaplots documentation.

What do you say, include this?
If we do, we need to remove the 'using Plots' statements though. StatPlots reexports Plots so this will/may cause issues with function versions.

daschw: I say we can try it and if it does not work we can still skip those testimages.

test/runtests.jl
daschw: No, you are right, it's just a warning. I will include it then.

mkborregaard: Not sure why `match_dimensions` isn't a keyword though. I think it just an omission from the argument lists for gr and plotlyjs. Here, it tests for it in this line https://github.com/JuliaPlots/Plots.jl/blob/master/src/backends/plotly.jl#L409 (inside the `transpose_z` function). And it obeys the keyword (it transposes a matrix in heatmap, because PyPlot (without Plots) does it differently from GR and PlotlyJS, but all Plots backends do it the same way)

daschw:
OK, the tests are now running but they fail:
PlotlyJS gives a connection error on travis (`connect: connection refused (ECONNREFUSED)`)
InspectDR returns that error:
```
ERROR: LoadError: InitError: Cannot open display: 
 in Gtk.GLib.GError(::Gtk.##229#230) at /home/travis/.julia/v0.5/Gtk/src/GLib/gerror.jl:17
```
PyPlot and GR would run, but for some images they return errors like this
```
INFO: Testing plot: pyplot:19:Open/High/Low/Close
  Error :: (line:272)
    Expression: image_comparison_tests(pkg,i,debug=debug,sigma=sigma,eps=eps) |> success --> true
    MethodError: no method matching mean(::Array{Int64,1})
    you may have intended to import Base.mean
    Closest candidates are:
      mean{T<:AbstractFloat}(!Matched::AbstractArray{T<:AbstractFloat,N}) at /home/travis/.julia/v0.5/NaNMath/src/NaNMath.jl:168
```
for `mean`, `minimum` and `extrema`

Also
```julia
using Plots
test_examples(:gr, 8)
```
fails now on master for example. This is probably related to https://github.com/JuliaPlots/Plots.jl/pull/876, right @mkborregaard ?

daschw:
NaNMath.jl functions should probably be defined for
```julia
myfunc{T<:Real}(x::AbstractArray{T})
```
for our purposes?

mkborregaard:
I hope not - they are defined as `AbstractFloat` in that package. Widening that distinction may open a can of worms, because NaN is Float64.

daschw:
ah, ok, I see the problem ...

mkborregaard:
That means I have to go with my first solution - to define a local Plots function _minimum, pass to NaNMath for AbstractArray{<:AbstractFloat} and otherwise dispatch to the Base type :-(

mkborregaard:
Let's just say this shows the usefulness of tests...

daschw:
or we could define
```julia
NaNMath.f{T<:Integer}(x::AbstractArray{T}) = Base.f(x)
```
in Plots for mean, minimum, maximum, extrema, etc.
... but I don't know if that's better

daschw:
Yep, the first useful application of "working" tests ;-)

mkborregaard:
https://github.com/JuliaPlots/Plots.jl/pull/881

mkborregaard:
No I think in this case extending the NaNMath methods will be too confusing.

daschw:
All checks have passed!

mkborregaard:
Holy moly!! Fantastic, @daschw , great work.

mkborregaard:
By johve, I hope the 0.6 tests pass too - I'd say they might.

daschw:
Thanks!
Should I wait some time for reviews before merging?

mkborregaard:
Maybe wait a day, give @tbreloff and @pkofod a chance to give their input?

pkofod:
Great to have tests running again. I would probably have used lowercases `nan` in the function name (as in the `isnan` function from Base), but... Thanks for doing this!

mkborregaard:
That's a fair point, although it is strictly for internal use. We can change that once this is merged.

mkborregaard:
I think it will cause conflicts with master if changed here - @daschw you copied the commit from master rather than rebasing, right?

daschw:
No, I rebased

mkborregaard:
Great, then it shouldn't be a problem. I'm not really sure what the different operations look like on the outside :-) On my GitKraken it looks as if you added the commit at the end, and there appear to be commits from me, simondanisch and oshulz on your branch:
skaermbillede 2017-06-02 kl 09 13 48

mkborregaard:
On my own branch, if I do a rebase it has all the commits coming out on top of master.

daschw:
I did
```
git fetch origin
git checkout ds-testing
git rebase origin/master
git pull --rebase
git push -u forked ds-testing
```
Maybe that's not the right procedure (?)

But we can still add the ignorenan function changes in a separate PR.

mkborregaard:
No, you need to force push after doing a rebase, so you should skip `git pull --rebase` and push with "--force". I'm sorry for appearing pedantic about this, but I think it is really good for a project as involved as Plots to be careful with the git. In this case, of course we can change ignorenan later, but the bigger problem is that it will look from the git history as if you've added all commits and PRs on master after you forked your branch. 
The idea of git is to keep changes together as logical coherent units, and know who to blame for different code lines, but that will not work if the commits aren't assigned correctly.

Fortunately there's an extremely simple fix for this: copy all non-hidden files from the current state (with ds-testing checked out) Plots folder somewhere on your desktop (but not the whole folder, as you don't want to copy hidden git files). Create a new branch off current master and check it out. Put back all the files, and commit all the changes again as a single commit, or several commits, whatever you like. Make a new PR from that and close the present one. 

There are also more complicated ways that involve a soft reset then recommiting then rebasing or creating a new branch and cherry-picking commits but they're not better :-)

One thing you might consider is to try out gitKraken instead of using command-line git - I find it makes these operations so much more intuitive (the power of visualizations, right?).

Oh and 
git

daschw:
:-) Thanks a lot for this guide! I will make a new clean PR and close this one.

daschw:
Closing this in favour of #884

PR #878, opened 2017-05-31
Remove remnant debug output about binning

oschulz:
Ooops, that debug info was never meant to stay in ...


mkborregaard:
Ah, nice :-) I noticed and thought we'd keep it in for a few weeks to make people notice that the binning mode had changed.

PR #885, opened 2017-06-02
Changed GR legend markers for plots with fillrange

daschw:
Before this PR
```julia
using Plots
gr()

y = [0 0 0;
     2 4 6]
f = [0 0 0;
     1 3 5]

plot(
    plot(y),
    plot(y, fillrange = f, linewidth = 0),
    plot(y, fillrange = f, fillcolor = [4 5 6], linewidth = 3),
    layout = (1, 3)
    )
```
used to produce this:
gr_legend_old
Now the legend markers look like this:
gr_legend_new


mkborregaard:
Nice :+1:

mkborregaard:
A question is - when adding new functionality like this, should we try to make a habit of including an example of it in the examples/tests? It would make good sense, I think, as it both adds a test and documents the new functionality.

daschw:
In general I agree. Could there also be an issue with the PlotReferenceImages repo size? In that case an option would be to bundle multiple new functionalities in one new example from time to time.

PR #886, opened 2017-06-02
Changed spy recipe and added spy example to tests (fix #883)

### LINKED ISSUE #883: feature request: spy function for matrices
There was a feature request for a ``spy`` function to graphically represent matrices (especially useful for sparse matrices): [issue 40](https://github.com/JuliaPlots/Plots.jl/issues/40). It was closed a while back. Is that spy function available? Is there a preferred alternative? Thanks.

daschw:
as discussed in #883

### DIFF
--- a/src/examples.jl
+++ b/src/examples.jl
@@ -318,6 +318,16 @@ PlotExample("Animation with subplots",
     end)]
 ),
 
+PlotExample("Spy",
+    "For a matrix `mat` with unique nonzeros `spy(mat)` returns a colorless plot. If `mat` has various different nonzero values, a colorbar is added. The colorbar can be disabled with `legend = nothing`. As always, the marker shape and size can be changed with `spy(mat, markersize = 5, markershape = :diamond)`",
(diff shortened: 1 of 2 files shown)


mkborregaard:
Hm, I disagree. I think the reason I didn't notice the markersize before is that the grid was covering them. I think we should just set grid --> false in the recipe, but not be too noisy in the terminal.

mkborregaard:
One thing the documentation really needs is a folder with a page for each of the (groups of) seriestypes. That would be a nice place to have this info. And perhaps we should add a `spy` example to the examples/tests?

daschw:
Yes, that's a better solution.

daschw:
I removed the command line info, turned the colorbar off for matrices with unique nonzeros and added an example. This now requires https://github.com/JuliaPlots/PlotReferenceImages.jl/pull/2 to pass.

mkborregaard:
Nice touch with the superquick response here and with legendtitle :-)

Issue #882, opened 2017-06-02
Add label/legend title 


arturgower:
Would be great to specify a label/legend title. For publications this can be a necessity. 

[Here is an example](https://matplotlib.org/examples/pylab_examples/legend_demo3.html) from matplotlib

![legend_demo3](https://cloud.githubusercontent.com/assets/24454863/26721852/6ed38852-4785-11e7-9c45-41137582e344.png)

daschw:
`plot(rand(10, 3), legendtitle = "My Awesome Title")` works now on master for GR and PyPlot

PR #887, opened 2017-06-02
Added legend_title attribute for Pylot backend

daschw:
This implements the feature request of #882 for the PyPlot backend.
```julia
using Plots; pyplot()
plot(rand(10, 3), legend_title = "My Awesome Title")
```
now produces:
pyplot_legend_title

@jheinen Would this also be possible for GR?


daschw:
... or should the attribute rather be named `legendtitle`, `labeltitle` or `label_title`

mkborregaard:
Great! 👍 for `legendtitle`, though the others could/should be added as aliases.

daschw:
Now the default attribute is `legendtitle` and the aliases are:
`legend_title`, `labeltitle`, `label_title`, `leg_title`, `key_title`

lazarusA:
Is there a way to change the size of the legend title?

daschw:
> Is there a way to change the size of the legend title?

No, that's not (yet) implemented in Plots.

PR #900, opened 2017-06-07
amend bins description, fixes #898

### LINKED ISSUE #898: Histograms have the wrong number of bins
The documentation says of the `bins` option, "For histogram-types, defines the number of bins, or the edges, of the histogram."  But `histogram` usually doesn't use the correct number of bins specified with this option.  I tried `histogram(rand(1000), bins = n)` for n = 1 through 50, and found that it only output the correct number of bins for n = 1, 2, 5, 10, 20, or 50.  For any other value of n, it displays a histogram whose number of bins was in that list, somewhere near the number set using `bins`.

mkborregaard:


ttparker:
Looks good!  Much clearer.  You may want want to add the example "e.g. `linspace(extrema(x)..., 25)`" at the end, or maybe that's too much detail.  Up to you.

mkborregaard:
Right, I've added the example.

PR #916, opened 2017-06-08
WIP: Better Juno integration

pfitzseb:
by using Requires.jl. Also cleaned up the Juno related code a bit.

TODO:
- [x] use Requires.jl for IJulia too
- [x] test if this works properly
- [x] check if this works around the Media issue

### DIFF
--- a/src/backends.jl
+++ b/src/backends.jl
@@ -135,7 +135,7 @@ function pickDefaultBackend()
             Pkg.installed(env_default)  # this will error if not installed
             sym = Symbol(lowercase(env_default))
             if haskey(_backendType, sym)
-                return backend(sym)
+                return @eval backend($sym)
             else
                 warn("You have set PLOTS_DEFAULT_BACKEND=$env_default but it is not a valid backend package.  Choose from:\n\t",
                      join(sort(_backends), "\n\t"))
@@ -150,12 +150,12 @@ function pickDefaultBackend()
     # features, speed, and robustness
     for pkgstr in ("GR", "PyPlot", "PlotlyJS", "PGFPlots", "UnicodePlots", "InspectDR", "GLVisualize")
         if Pkg.installed(pkgstr) != nothing
-            return backend(Symbol(lowercase(pkgstr)))
+            return @eval backend(Symbol(lowercase($pkgstr)))
         end
     end
 
     # the default if nothing else is installed
-    backend(:plotly)
+    @eval backend(:plotly)
 end
 
 
--- a/src/backends.jl
+++ b/src/backends.jl
@@ -135,7 +135,7 @@ function pickDefaultBackend()
             Pkg.installed(env_default)  # this will error if not installed
             sym = Symbol(lowercase(env_default))
             if haskey(_backendType, sym)
-                return @eval backend($sym)
+                return backend(sym)
             else
                 warn("You have set PLOTS_DEFAULT_BACKEND=$env_default but it is not a valid backend package.  Choose from:\n\t",
                      join(sort(_backends), "\n\t"))
@@ -150,12 +150,12 @@ function pickDefaultBackend()
     # features, speed, and robustness
     for pkgstr in ("GR", "PyPlot", "PlotlyJS", "PGFPlots", "UnicodePlots", "InspectDR", "GLVisualize")
         if Pkg.installed(pkgstr) != nothing
-            return @eval backend(Symbol(lowercase($pkgstr)))
+            return backend(Symbol(lowercase(pkgstr)))
         end
     end
 
     # the default if nothing else is installed
-    @eval backend(:plotly)
+    backend(:plotly)
 end
 
 
--- a/REQUIRE
+++ b/REQUIRE
@@ -11,3 +11,4 @@ Showoff
 StatsBase 0.14.0
 JSON
 NaNMath
+Requires
--- a/src/Plots.jl
+++ b/src/Plots.jl
@@ -245,8 +245,6 @@ function __init__()
     # for compatibility with Requires.jl:
     @init
 
-    setup_ijulia()
-
     if isdefined(Main, :PLOTS_DEFAULTS)
         for (k,v) in Main.PLOTS_DEFAULTS
             default(k, v)
--- a/src/Plots.jl
+++ b/src/Plots.jl
@@ -241,10 +241,8 @@ end
 
 const CURRENT_BACKEND = CurrentBackend(:none)
 
-function __init__()
-    # for compatibility with Requires.jl:
-    @init
-
+# for compatibility with Requires.jl:
+@init begin
     if isdefined(Main, :PLOTS_DEFAULTS)
         for (k,v) in Main.PLOTS_DEFAULTS
             default(k, v)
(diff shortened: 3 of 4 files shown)


mkborregaard:
Thanks a lot for this @pfitzseb ! It would be so cool to have this working. If it does, it would be worth looking into using Requires to handle the conditional loading of backends too.

mkborregaard:
I just tried out this with StatPlots, that often gives me the Media error, and I didn't get it. Retried with master and got the Media error.
The bad news - I got the GR worldage error (I don't in Juno on master); as usual, fixable by calling `gr()` explicitly first.

daschw:
This also supports your suggestion here: https://github.com/JuliaPlots/Plots.jl/pull/808#issuecomment-307152746

pfitzseb:
I'll look into using Requires for the backends tomorrow or over the weekend. Might take a while because I'm not familiar with the backend code at all.
> I just tried out this with StatPlots, that often gives me the Media error, and I didn't get it. 

Nice!

mkborregaard:
Awesome! Yes, the world age issue should be solvable elsewhere, so it looks really promising.

pfitzseb:
Okay, this approach has one problem I'm not sure how to solve: If you have a script like
```julia
using Plots
plot([1,2,3,4])
```
and evaluate that at once, you won't see any plot because the `@require` block will only be evaluated after that. Not sure if (or how) we can work around that, or even if it's that important to do so.

mkborregaard:
Great! ready to merge when you are.

mkborregaard:
Hmm some people do seem to use that functionality

pfitzseb:
Yeah, I guess we could check whether `Juno in names(Main)` in `__init__` and call the initialization then or something?

mkborregaard:
Yes, that's a nice solution.

pfitzseb:
The thing is that the actual problem is the # 265 fix. What I've come up with so far is littering `invokelatest`s around the eval code in Atom.jl. That seems to mostly work, but I'll do some more testing before saying anything definite ;)

pfitzseb:
Alright, so as far as I can tell this works fine in most cases. Exception: first plot might not display in the plot pane if there's no idle period between `using Plots` and the first actual `plot` command (as described above). Still not entirely sure on the world age fix, but that one seems to be working fine too.

Let me know what you guys think! :)

daschw:
Thanks @pfitzseb, this looks great!

>Yeah, I guess we could check whether Juno in names(Main) in __init__ and call the initialization then or something?

Do you think this can be used to avoid not displaying the the first plot, or did I misunderstand this?

Are you still planning on implementing the usage of Requires.jl for IJulia or do you think this is ready to merge?

(Oh, and sorry that you had to wait so long for an answer!)

pfitzseb:
>>Yeah, I guess we could check whether Juno in names(Main) in init and call the initialization then or something?
>
>Do you think this can be used to avoid not displaying the the first plot, or did I misunderstand this?

I thought it could be, but that was because I didn't understand Requires' `@init` properly. We can work around it on the Juno side, but not here. Also note that this is only relevant if the `using Plots` and a subsequent call to  `plot` are *immediately* executed after each other (by selecting both lines, or putting both statements on one line, or by evaluating the whole file). This *shouldn't* matter for the first plot in general.

Will see what I can do regarding IJulia -- haven't used that in ages, but I guess it's setup is way easier now anyways ;)

pfitzseb:
Alright, plotting in IJulia is now handled via Requires as well, but I have only superficially tested it with GR:
screenshot_20170712_103013

Unfortunately I spoke too soon regarding my "fix" to the `GR.emergencyclosegks` world age error though -- in Juno I'm getting the same error message as before, just displayed in the plotpane. Maybe I'll find some solution to that though...

pfitzseb:
Alright, with that Atom.jl PR linked above something like `using Plots; plot([1,2,3])` just works. 🎉

pfitzseb:
Now with 100% less warnings shown during precompilation. 
Someone who actually understands how Plots works should probably check that the `_show` definition doesn't break anything, but I'm fairly confident all works as it should now.

src/backends.jl
```diff
@@ -135,7 +135,7 @@ function pickDefaultBackend()
             Pkg.installed(env_default)  # this will error if not installed
             sym = Symbol(lowercase(env_default))
             if haskey(_backendType, sym)
-                return backend(sym)
+                return @eval backend($sym)
```
daschw: This causes the warning `You have set PLOTS_DEFAULT_BACKEND=GR but it is not installed.` for me, because `$sym` evals to `Plots.gr` instead of `:gr` here and `backend(Plots.gr)` throws a `MethodError`.

Besides, these `@eval`s in `pickDefaultBackend()` don't seem to solve the world age issue for GR. I don't know about the other backends.

Solutions for GR (and probably analogously for other backends) I found to work would be either
* to replace each function `GR.f(args...)` in `src/backends/gr.jl` with `Base.invokelatest(GR.f, args...)` (no idea if this has an effect on performance - it definitely worsens readability),
* to add `backend()` in `@init` in `src/Plots.jl` (This however throws the warning `WARNING: using Plots.GR in module Main conflicts with an existing identifier.`) or
* to expect from the users to call `backend()` or specify their backend before the first plot.

pfitzseb: Hm, I do not see that `You have set PLOTS_DEFAULT_BACKEND=GR but it is not installed.` warning anywhere -- neither in Juno, IJulia nor the REPL.
And I'm also not seeing any world age issues anymore, but that might just be because of the Atom.jl PR that makes use of `invokelatest`. Or did that appear for anyone not using Juno?

daschw: > Hm, I do not see that You have set PLOTS_DEFAULT_BACKEND=GR but it is not installed. warning anywhere

Yes, I have specified GR as my local default backend in my `.juliarc` with `ENV["PLOTS_DEFAULT_BACKEND"] = "GR"`. If you have not explicitly defined your default backend, you don't end up here.
You can reproduce the issue I described with:
```julia
using Plots
sym = :gr
@eval backend($sym)
```
`@eval backend($sym())` would work though.

> And I'm also not seeing any world age issues anymore, but that might just be because of the Atom.jl PR that makes use of invokelatest. Or did that error appear for anyone not using Juno?

Oh, you are right! I was not aware that this was only in Juno, because I was only using Juno when testing. Are the world age issues now fixed in Atom.jl even without the `@eval`s here?

pfitzseb: Right, actually have read the warning :)
I've reverted the `@eval` stuff, since using `invokelatest` in Atom does indeed fix the worldage bug.

That also should make this PR fairly uncontroversial all in all now.

daschw:
Great! I think this is ready to merge. What do you say @pfitzseb, should we wait for others' comments?

pfitzseb:
I think this should be good to go. :shipit: 
It also doesn't seem super likely anyone else will chime in, since the PR has been open for a month now.

daschw:
One more thing: this does not require Atom or Juno master, does it?

pfitzseb:
No, this works fine on the latest releases (except for the world age error of course).

daschw:
Thanks a lot @pfitzseb!

mkborregaard:
Thanks @pfitzseb, amazing to come back from holiday to see this merged :-) I've been totally AFK so sorry I didn't chime in.

PR #926, opened 2017-06-12
Reduced colorbar size on PyPlot for big z values (fix #925)

### LINKED ISSUE #925: pyplot: Heatmap legend cut off on top
The legend of the resulting heatmap is cut off on top for large numbers.

Example:
```julia
using Plots; pyplot()
x = y = 1:10
z = 600000 * x * y'
heatmap(x, y, z, aspectratio = true)
```
image
You can see a part of `1e7` above the legend.

Exists in both Julia 0.5.2 & Plots 0.11 and Julia 0.6, Plots 0.12.

In an older version, there was a similar bug where the right side was cut off.

(cc @daschw)

daschw:
```julia
using Plots; pyplot()
x = y = 1:10
z = 1e5 * x * y'
heatmap(x, y, z, aspectratio = true)
```
used to produce this:
pyplot_colorbar_big_old

Now it looks like this:
pyplot_colorbar_big

The colorbar axes limits remain the same for smaller `z` values like, e.g. `z = 1e4 * x * y'`:
pyplot_colorbar_normal

### DIFF
--- a/src/backends/pyplot.jl
+++ b/src/backends/pyplot.jl
@@ -1221,7 +1221,7 @@ function _update_plot_object(plt::Plot{PyPlotBackend})
             cbw = sp.attr[:cbar_width]
             # this is the bounding box of just the colors of the colorbar (not labels)
             has_toplabel = sp[:zaxis][:extrema].emax >= 1e7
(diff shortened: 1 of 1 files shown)


daschw:
I omitted changing the colorbar axes bottom corners. The first plot now looks like this:
pyplot_colorbar_big_new

PR #929, opened 2017-06-13
replace d by plotattributes inside recipes

jw3126:
This PR is part of the plan to replace d inside `@recipes`. See [here](https://github.com/JuliaPlots/RecipesBase.jl/pull/20).


mkborregaard:
I'm in favour of this, but it's a massive breaking change across many packages - so I suggest that if we merge it we do it across the board when we bump the julia version requirement to 0.7 (since we just missed the 0.6 window).

PR #936, opened 2017-06-16
Conform to changes in InspectDR.PlotLayout.

ma-laforge:
InspectDR needed structural changes to improve how layouts are controlled/allow for user-defined stylesheets/presets.

Waiting for METADATA tag so users don't get error messages:
https://github.com/JuliaLang/METADATA.jl/pull/9843

### DIFF
--- a/src/backends/inspectdr.jl
+++ b/src/backends/inspectdr.jl
@@ -362,30 +362,29 @@ function _inspectdr_setupsubplot(sp::Subplot{InspectDRBackend})
         a.xlabel = xaxis[:guide]; a.ylabels = [yaxis[:guide]]
 
     l = plot.layout
-        l.frame.fillcolor = _inspectdr_mapcolor(sp[:background_color_subplot])
-        l.framedata.fillcolor = _inspectdr_mapcolor(sp[:background_color_inside])
-        l.framedata.line.color = _inspectdr_mapcolor(xaxis[:foreground_color_axis])
-        l.fnttitle = InspectDR.Font(sp[:titlefont].family,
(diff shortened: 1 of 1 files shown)


mkborregaard:
Hey, do you want me to add you as a collaborator so you can commit freely to backends/inspectdr.jl?

ma-laforge:
Sure, but is it be possible to limit my privileges to merging in pull requests, or something?

I would rather not have the ability to accidentally break more things than I need to.

mkborregaard:
No, the system does not allow such fine-grained control. If you're not comfortable with we'll just continue as now.

ma-laforge:
Well, if you want me to take on a bit of responsibility because you have too much to do, than I guess I'm ok becoming a collaborator.

Otherwise: I find the team is already pretty responsive at the moment - so I don't *personally* really see the need to become a collaborator at this point in time.

ma-laforge:
PS: New version of InspectDR is tagged in METADATA.  Ready to merge whenever the team wishes.

mkborregaard:
There's something wrong with Travis at the moment. Does `Pkg.test()` succeed locally for you on this?

ma-laforge:
I see...  Regarding `Pkg.test()`:

It seems successful for me, but only if I manually:

```
Pkg.add("VisualRegressionTests")
#and/or
Pkg.clone("https://github.com/JuliaPlots/PlotReferenceImages.jl.git")
```
Before running the tests.

...But apparently, InspectDR is not tested against in these tests anyhow (so it is probably not the cause of the test failures).

Issue #920, opened 2017-06-09
y-axis log scale bug


jebej:
The y-axis logscale option does not work in PyPlot and GR for log values smaller than or equal to 1:

```julia
heatmap([1,2], [10, 32], rand(2,2), yscale = :log10)
ERROR: DomainError:
 in nan_dom_err at .\math.jl:196 [inlined]
 in log10(::Float64) at .\math.jl:202
 in (::Plots.##114#115{Symbol})(::Float64) at C:\Users\Jeremy\.julia\v0.5\Plots\src\axes.jl:148
...
```

Not sure why this is getting a domain error. Plots must be transforming the y-axis values.

jebej:
This has to do with how Plots calculate the axes limits, which for heatmaps in GR and PyPlot are not the same as the limits of the given y vector.

daschw:
Thanks @jebej for reporting this, but I think the issue a little different. It has nothing to do with the axis (`xaxis` or `yaxis`) or on the values of `z`, but with the vallues of `x`, or `y` respectively, and the fact that you are plotting a heatmap.
If e.g. `x[2] >= 2 * x[1]` on a `log10` `xaxis` the first edge of the heatmap is not positive and `log10` throws a `DomainError`.
You can try this by comparing
```julia
heatmap([1, 3], [1, 2], rand(2,2), xscale = :log10)
```
to
```julia
heatmap([1, 2.99], [1, 2], rand(2,2), xscale = :log10)
```

jebej:
You're right that there are no differences between x and y. However, I do think the issue is in how the limits are calculated. In fact they are calculated differently for plotly (or actually, they are not calculated), which I assume is because heatmaps are supported natively.

daschw:
Yes they are calculated [here](https://github.com/JuliaPlots/Plots.jl/blob/master/src/utils.jl#L336). What would the expected behaviour for a plot on a log axis reaching into its non-positve area be? Should it be cropped at 0 or should there be an error (maybe more verbose about the reason)?

jebej:
I'm not sure what you mean, there are no negative numbers involved here... If the x values passed are valid, ie there are no zeros or negative values, it should work, just like plotly does right now:

![image](https://user-images.githubusercontent.com/11141558/27014811-27ac70ba-4ece-11e7-9e96-aec6c22ed0be.png)

jebej:
[...]

PR #940, opened 2017-06-18
Avoid non-positive heatmap edges for log axis on GR and PyPlot

daschw:
```julia
using Plots
x = y = 1:3:30
heatmap(x, y, x * y', xscale = :log10, yscale = :log10)
```
used to return a `log10` `DomainError`, although all input data is positve. That is because the first calculated heatmap edge is `-0.5` (cf. #920).

PlotlyJS solves this by removing the first values of `x` and `y` as well as the first row and column of the data matrix:
heatmap_log10_plotlyjs

PyPlot and GR now increase the first heatmap edge values to a positive value to avoid `DomainError`s and keep all matrix data.
PyPlot:
heatmap_log10_pyplot_new
GR still has some issues:
heatmap_log10_gr_new
On GR heatmaps are drawn as images with the uniformly sized rectangles, so the log scale does not affect the heatmap edge positions. Instead the colorbar ticks are affected. (cc: @jheinen )

### DIFF
--- a/src/backends/gr.jl
+++ b/src/backends/gr.jl
@@ -585,7 +585,7 @@ function gr_display(sp::Subplot{GRBackend}, w, h, viewport_canvas)
         end
         if st == :heatmap
             outside_ticks = true
-            x, y = heatmap_edges(series[:x]), heatmap_edges(series[:y])
+            x, y = heatmap_edges(series[:x], sp[:xaxis][:scale]), heatmap_edges(series[:y], sp[:yaxis][:scale])
             expand_extrema!(sp[:xaxis], x)
             expand_extrema!(sp[:yaxis], y)
             data_lims = gr_xy_axislims(sp)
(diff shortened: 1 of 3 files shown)


jebej:
Thanks for this. I'll check it out soon (on vacation), and like I  mentionned on #920 there might be two different ways of doing this, so it might be worth it to discuss it briefly.

mkborregaard:
cf https://github.com/JuliaPlots/Plots.jl/issues/942 and this PR https://github.com/JuliaPlots/Plots.jl/pull/835 which adresses a different issue - that axis widening on log axes can also cause the value to go sub-zero.

jebej:
The change looks good and the PyPlot examples looks fine! Is it possible to keep the full matrix with Plotly as well, or is it something done by the js library itself?

oschulz:
A question related to this: Which backends support z-log-scale for heatmap now?

daschw:
@oschulz I don't think any backend does.

Issue #942, opened 2017-06-19
heatmaps don't respect lims


mkborregaard:
```julia
x, y = 1:20, 1:10
z = y*x'
using Plots; gr()
heatmap(x, y, z, xlim = (5,15))
```
![skaermbillede 2017-06-19 kl 10 56 42](https://user-images.githubusercontent.com/8429802/27277360-46ec4c96-54de-11e7-91d0-21a1f8c95957.png)

The lims only applies to the axis ticks, but isn't applied to the heatmap.
```julia
pyplot()
heatmap(x, y, z, xlim = (5,15))
```
![skaermbillede 2017-06-19 kl 10 56 34](https://user-images.githubusercontent.com/8429802/27277366-4a6eec52-54de-11e7-8e25-3efa7e821378.png)

Works on pyplot.

mkborregaard:
cc @jheinen

mkborregaard:
Wow, thanks for the quick fix!

PR #944, opened 2017-06-20
gr: heatmaps now respect xlim/ylim specifications

jheinen:
Fixes https://github.com/JuliaPlots/Plots.jl/issues/942


jheinen:
I'll check this ...

PR #965, opened 2017-07-05
Switch from FixedSizeArrays to StaticArrays.FixedSizeArrays

rdeits:
FixedSizeArrays.jl [has been deprecated](https://github.com/SimonDanisch/FixedSizeArrays.jl#this-package-doesnt-support-03-and-its-not-planned-to-update-it-to-06-use-staticarrays-instead) in favor of `StaticArrays.FixedSizeArrays` which has better v0.6 support and is actively maintained.

I've successfully run the unit tests using the scripts referenced in `.travis.yml`, and I've sanity-checked that basic plotting from Jupyter works, but I'm not sure of what other tests I need to run to be sure that this is OK. Any advice?

### DIFF
--- a/REQUIRE
+++ b/REQUIRE
@@ -4,7 +4,7 @@ RecipesBase 0.2.0
 PlotUtils 0.4.1
 PlotThemes 0.1.3
 Reexport
-FixedSizeArrays
+StaticArrays 0.5
 FixedPointNumbers 0.3
 Measures
 Showoff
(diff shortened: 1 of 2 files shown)


daschw:
Looks good to me, thanks!
Maybe @SimonDanisch can provide some insights, whether this or Tom's approach [here](https://github.com/JuliaPlots/Plots.jl/commit/9d038ea05e96ef093de0ea995157af5aae630411) is the way to go.

SimonDanisch:
Both seem valid, but this change is definitely shorter ;)

daschw:
Ok, great :-)

rdeits:
Thanks!

PR #970, opened 2017-07-13
fix glvisualize backend for 0.6

SimonDanisch:
We should wait for this until GLVisualize is tagged, but can probably already merge to have things tagged faster!

### DIFF
--- a/REQUIRE
+++ b/REQUIRE
@@ -12,3 +12,4 @@ StatsBase 0.14.0
 JSON
 NaNMath
 Requires
+Contour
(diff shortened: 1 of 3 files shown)


SimonDanisch:
and it's tagged :)
https://github.com/JuliaLang/METADATA.jl/pull/10247

SimonDanisch:
Great! I tested it on a fresh install on my pc and seems to be okay!
I added the Contour package as a dependency. Should be lightweight and might make sense anyways to have it here. (ref https://github.com/JuliaPlots/Plots.jl/issues/974)

mkborregaard:
Thanks, awesome work @SimonDanisch !

PR #975, opened 2017-07-20
drop FactCheck in tests (fix #915)

### LINKED ISSUE #915: Drop FactCheck in tests
It's deprecated. We should just rely on Base.Test

daschw:


test/imgcomp.jl
```diff
@@ -43,11 +42,7 @@ function image_comparison_tests(pkg::Symbol, idx::Int; debug = false, popup = is
     fn = "ref$idx.png"
 
     # firgure out version info
-    G = glob(joinpath(relpath(refdir), "*"))
-    # @show refdir fn G
-    slash = (@static is_windows() ? "\\" : "/")
-    versions = map(fn -> VersionNumber(split(fn, slash)[end]), G)
-    versions = reverse(sort(versions))
+    versions = sort(VersionNumber.(readdir(refdir)), rev = true)
```
mkborregaard: This change has broken local testing on Mac, because Macs have a hidden file in all folders called ".DS_Store", which comes out first in the sort.

PR #976, opened 2017-07-20
Line legend marker for ribbons in GR (fix #896)

### LINKED ISSUE #896: GR Legend Marker for Ribbons
I believe https://github.com/JuliaPlots/Plots.jl/pull/885 changed the markers for ribbons (it used to just show a line).  Should a line also be part of the marker here?

screen shot 2017-06-06 at 10 51 14 am

daschw:
Resets the GR legend marker for plots with ribbons to the behaviour before #885.
```julia
using Plots; gr()
rib1, rib2, rib3, rib4 = 0.1, (0.1, 0.2), 0.1 * rand(10), (0.1 * rand(10), 0.2 * rand(10))
plot([plot(rand(10), ribbon = rib, fillalpha = 0.5) for rib in (rib1, rib2, rib3, rib4)]...)
```
produced these legend markers:
ribbon_old

Now only lines are drawn again:
ribbon_new

### DIFF
--- a/src/backends/gr.jl
+++ b/src/backends/gr.jl
@@ -1049,7 +1049,7 @@ function gr_display(sp::Subplot{GRBackend}, w, h, viewport_canvas)
                 should_add_to_legend(series) || continue
                 st = series[:seriestype]
                 gr_set_line(series[:linewidth], series[:linestyle], series[:linecolor]) #, series[:linealpha])
-                if st == :path && series[:fillrange] == nothing
+                if st == :path && (series[:fillrange] == nothing || has_ribbon(series))
                     GR.polyline([xpos - 0.07, xpos - 0.01], [ypos, ypos])
                 elseif st == :shape || series[:fillrange] != nothing
                     gr_set_fill(series[:fillcolor]) #, series[:fillalpha])
--- a/src/backends/gr.jl
+++ b/src/backends/gr.jl
@@ -1049,7 +1049,7 @@ function gr_display(sp::Subplot{GRBackend}, w, h, viewport_canvas)
                 should_add_to_legend(series) || continue
                 st = series[:seriestype]
                 gr_set_line(series[:linewidth], series[:linestyle], series[:linecolor]) #, series[:linealpha])
-                if st == :path && (series[:fillrange] == nothing || has_ribbon(series))
+                if st == :path && (series[:fillrange] == nothing || series[:ribbon] != nothing)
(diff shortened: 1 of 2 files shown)


joshday:
Thanks for working on this!  Is there a way to get the color in the legend to match the line color?  It looks like it's using the alpha color.

daschw:
I don't think it is, but I agree it looks like so.
How about this?
```julia
using Plots; gr()
y, rib = rand(10), rand(10)
plot(
    plot(y, ribbon = rib),
    plot(y, ribbon = rib, fillalpha = 0.5),
    plot(y, ribbon = rib, fillcolor = 2),
    plot(y, ribbon = rib, fillalpha = 0.5, fillcolor = 2),
    )
```

gr_ribbon_new

daschw:
No, you were right. It's really using the `fillalpha` somehow. I don't know how to fix this, though. It seems like `gr_set_line(...)` used to accept an alpha argument, but does not any more. @jheinen probably knows more about this.

daschw:
I also changed the legend marker for path with fillrange. Here are the legend markers for different series types after this PR:

```julia
using Plots; gr()
y, rib = rand(10), rand(10)
plot(
    bar(y, fillalpha = 0.5, title = "Bar / Shape"),
    plot(y, title = "Path"),
    plot(y, ribbon = rib, fillcolor = 2, title = "Path with ribbon"),
    plot(y, fillrange = -rand(10), fillalpha = 0.5, fillcolor = 2, title = "Path with fillrange"),
    )
```

gr_markers

jheinen:
The transparency should be set to 1 (in the legend part, around line 1042) - at least for the text output, e.g.
```julia
            if sp[:legendtitle] != nothing
                GR.settextalign(GR.TEXT_HALIGN_CENTER, GR.TEXT_VALIGN_HALF)
                gr_set_textcolor(sp[:foreground_color_legend])
                GR.settransparency(1)
                gr_text(xpos - 0.03 + 0.5*w, ypos, string(sp[:legendtitle]))
                ypos -= dy
            end
```

jheinen:
... probably also for the lines, markers and fill areas - but that's a matter of taste

jheinen:
@daschw : I think, only the surrounding rectangle and the text should be drawn without any transparency (=> 1). Lines, markers etc. should be drawn with the same attributes as the series. What do you think?

daschw:
I agree and I'm on it. Thanks for the hint with `GR.settransparency()`!

daschw:
Now we have
gr_markers_new

Issue #983, opened 2017-07-24
don't use pywrap


stevengj:
Currently, you use `pywrap` to [import a whole bunch of modules](https://github.com/JuliaPlots/Plots.jl/blob/dbc7bc4253d9aeb376eacfc1a4bc3d53e57e4c45/src/backends/pyplot.jl#L63-L89).  

I would urge that you *not* do this.  Just use `pyimport`, and then use `module[:symbol]` (or better yet `module["symbol"]` for things like functions where you don't need auto-conversion to Julia types, to avoid the type-introspection overhead).  The reason is that `pywrap` imposes a big startup cost, since it loads every symbol in the module and converts it to a Julia object.

PR #988, opened 2017-07-26
don't use pywrap

stevengj:
As explained in #983, this eliminates the `pywrap` calls from the PyPlot backend initialization.  On my machine, this cuts the time for `Plots.pyplot()` down from 12 seconds to 6.5 seconds.

### DIFF
--- a/src/backends/pyplot.jl
+++ b/src/backends/pyplot.jl
@@ -70,18 +70,18 @@ function _initialize_backend(::PyPlotBackend)
         append!(Base.Multimedia.displays, otherdisplays)
 
         export PyPlot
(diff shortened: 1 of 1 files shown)


daschw:
I don't have time today to take a closer look. but some of the testimages fail, e.g.
```julia
using Plots; pyplot()
test_examples(:pyplot, 9) # Error: type PyObject has no field pymember
test_examples(:pyplot, 10) # Error: type PyObject has no field ma
```
Any idea, what's causing this and how to fix it?

stevengj:
Whoops, I forgot to commit all of the changes.  Hopefully fixed now.

PR #989, opened 2017-07-27
RFC: fix markershape in recipes

mkborregaard:
This fixes a bug in recipes, where if a series recipe defines `seriestype := scatter`, the markers will not be shown. This is because the default value of `markershape` is `:none`, which is changed to `:circle` in `add_defaults` if the seriestype is a `:scatter` type, but `add_defaults` is called before series recipes are processed. This behaviour requires the recipe to define markershape manually, as e.g. here: [https://github.com/JuliaPlots/StatPlots.jl/blob/master/src/boxplot.jl#L94-L96](https://github.com/JuliaPlots/StatPlots.jl/blob/master/src/boxplot.jl#L94-L96) , which is really confusing and shouldn't be necessary.
A problem with the suggested solution (which also exists now, but at least isn't fixed by this solution) is that you can't define `scatter` then turn off markers (which could perhaps sometimes be smart to do, e.g. for fixing the coordinates of `series_annotations`.

### DIFF
--- a/src/args.jl
+++ b/src/args.jl
@@ -706,6 +706,9 @@ function preprocessArgs!(d::KW)
     delete!(d, :marker)
     if haskey(d, :markershape)
         d[:markershape] = _replace_markershape(d[:markershape])
+        if d[:markershape] == :none && d[:seriestype] in (:scatter, :scatterbins, :scatterhist, :scatter3d) #the default should be :auto, not :none, so that :none can be set explicitly and would be respected
+            d[:markershape] = :circle
+        end
     elseif anymarker
         d[:markershape_to_add] = :circle  # add it after _apply_recipe
     end


daschw:
> A problem with the suggested solution (which also exists now, but at least isn't fixed by this solution) is that you can't define `scatter` then turn off markers

Would this be possible if `markershape` defaulted to `:auto` rather than `:none`?

mkborregaard:
I think so - if the default would be `:auto`, and then Plots would now never to plot markers when `markershape` is either `:auto` or `:none`, but the automatic conversion of `markershape` to `:circle` would only be implemented for `:auto`? In that way I think `:none` would be a flag that the user had specified this explictly. I guess.

mkborregaard:
But the application of things like markershape is so complex in Plots because of recipes and the like that I feel a little uncertain how best to mess with it.

daschw:
I think so too. Do you see any issues with this?

daschw:
Ah, I just saw your second comment now ...
Have you already tried defaulting to `:auto` with a few examples?
I can totally understand your uncertainty ... The only one completely comprehending all the interconnections and consequences is probably Tom.

daschw:
... I don't see an issue at first glance, though

mkborregaard:
Right, I'll go ahead and merge this and leave the suggested change for another time.

PR #999, opened 2017-08-07
WIP Generalized array indexing: fixes #954

### LINKED ISSUE #954: Plots should implement the generalized Arrays interface, to be compatible with e.g. OffsetArrays
Plots doesn't work with OffsetArrays: 
```Julia
julia> plot(OffsetArray(rand(100), -50))
ERROR: size not supported for arrays with indices (-49:50,); see http://docs.julialang.org/en/latest/devdocs/offset-arrays/
```

mkborregaard:
Plots should use the generalized arrays interface for Array indices, thus automatically supporting anything that implements the AbstractArray interface. This involves e.g.:
 -  replace many uses of `size` with `indices`
 -  replace `1:length(A)` with `linearindices(A)`, and `length(A)` with `length(linearindices(A))`
 -  replace explicit allocations like `Array{Int}(size(B))` with `similar(Array{Int}, indices(B))`
See https://docs.julialang.org/en/latest/devdocs/offset-arrays/#Arrays-with-custom-indices-1


asinghvi17:
Are the only changes to be made those in the original post?

diegozea:
`linearindices` doesn't exist in Julia 1.1.1. Should it be [`LinearIndices`](https://pkg.julialang.org/docs/julia/THl1k/1.1.1/devdocs/offset-arrays.html#Linear-indexing-(LinearIndices)-1) ?

daschw:
Superceded by #2304

PR #1003, opened 2017-08-09
gr: no longer handle '_' and '^' in text

jheinen:


daschw:
@jheinen why did you leave the `'\\'` case? What is a use case for this?

```julia
using Plots; gr()
plot(rand(10), title = "my_title")
```
now works as expected.
However 
```julia
plot(rand(10), title = "my_title with \\")
```
shows no title at all and with
```julia
plot(rand(10), title = "my_title with \\ this")
```
`"title with this"` is subscripted.

I'm not saying this is wrong, I just find it confusing / don't understand it. :smile:

jheinen:
It allows you to insert greek characters (without using LaTeX), e.g.
```julia
plot(rand(10), title = "my title with an \\gamma character")
```
There's still a lot of things to do - GR currently has **no** full utf-8 support. Text strings are limited to the basic Latin (ASCII) and the Latin-1 supplement unicode subset, depending on the underlying GKS driver. There's still a lot of work :-(

daschw:
OK, I see. Thanks for the clarification!
This means  that we have to accept for now that the usage of `_` in a title string yields different behaviour for
```julia
plot(rand(10), title = "my_title with \\gamma") # "title with γ" as subscript of "my"
```
and
```julia
plot(rand(10), title = "my_title with y") # "my_title with y"
```
right?

jheinen:
We could probably use another notation, ignore `\\` or start a discussion on [Gitter](https://gitter.im/tbreloff/Plots.jl).

mkborregaard:
@jheinen after this PR the scientific formatter is broken (https://github.com/JuliaPlots/Plots.jl/issues/1203) - do you have a suggestion for the best way to reinstate that?

Issue #296, opened 2016-06-03
Detailed control of looks


pkofod:
To make themes look exactly like their targets (ggplot2 for example) more detailed control of the looks of plots is needed. For example, I need to change the line width and style of the grid lines. What is the plans regarding such detailed control of the appearance of plots in Plots?

tbreloff:
Similar to how foreground colors have cascading defaults set to 'match', I
think I'll probably change border, grid, tick lines, etc to be a
Plots.Stroke, which defaults to match. So 'grid_stroke_vertical' would
match 'grid_stroke' by default, which would match 'foreground_stroke'. The
associated color args would become aliases to the appropriate stroke arg.
You would be able to pass any valid inputs to the 'stroke' method, and the
remainder would get defaults.

Then we just need to support everything in the backends...

pkofod:
Originally I was weary of even asking because of the whole 5 backend do it in 7 different ways ;)

daschw:
closing via #1006

PR #1006, opened 2017-08-10
grid should be axis attribute (fix #675)

### LINKED ISSUE #675: grid should be Axis attribute, not Subplot

daschw:
This makes `grid` and `foreground_color_grid` Axis attributes instead of Subplot attributes (cf #675).
A lot of this work has already been done by @GuessWhoSamFoo in #695, so credits to him!!!
I just tried to make it work on all different backends and also made `foreground_color_grid` an Axis attribute.

I tested the changes with the following code:
```julia
using Plots

x = rand(10)
plot(
    plot(x, xgrid = false, title = "xgrid = false"),
    plot(x, ygrid = nothing, title = "ygrid = nothing"),
    plot(x, grid = false, title = "grid = false"),
    plot(x, xforeground_color_grid = :red, title = "xfcg = :red"),
    plot(x, foreground_color_grid = :red, title = "fcg = :red"),
    plot(x, xforeground_color_grid = :red, yforeground_color_grid = :green, title = "xfcg = :red, yfcg = :green"),
    label = "",
    )
```

Here are the resulting plots for the different backends:
* **GR**
grid_axis_gr

* **PyPlot**
grid_axis_pyplot

* **Plotly(JS)**
grid_axis_plotlyjs

* **GLVisualize**
grid_axis_glvisualize

The following backends do not support `foreground_color_grid`:

* **InspectDR**
grid_axis_inspectdr

* **PGFPlots**
grid_axis_pgfplots

@jheinen, @SimonDanisch and @ma-laforge could you please check the changes to your respective backends?

### DIFF
--- a/src/args.jl
+++ b/src/args.jl
@@ -163,6 +163,24 @@ const _scaleAliases = Dict{Symbol,Symbol}(
     :log  => :log10,
 )
 
+const _allGridSyms = [:x, :y, :z,
+                    :xy, :xz, :yx, :yz, :zx, :zy,
+                    :xyz, :xzy, :yxz, :yzx, :zxy, :zyx,
+                    :all, :both, :on,
+                    :none, :off,]
+const _allGridArgs = [_allGridSyms; string.(_allGridSyms); nothing; false; true]
+hasgrid(arg::Void, letter) = false
+hasgrid(arg::Bool, letter) = arg
+function hasgrid(arg::Symbol, letter)
+    if arg in _allGridSyms
+        arg in (:all, :both, :on) || contains(string(arg), string(letter))
+    else
+        warn("Unknown grid argument $arg; $letter\grid was set to `true` instead")
+        true
+    end
+end
+hasgrid(arg::AbstractString, letter) = hasgrid(Symbol(arg), letter)
+
 # -----------------------------------------------------------------------------
 
 const _series_defaults = KW(
@@ -285,6 +303,9 @@ const _axis_defaults = KW(
     :mirror => false,
     :grid                     => true,
     :foreground_color_grid    => :match,            # grid color
+    :gridalpha                => 0.3,
+    :gridstyle                => :dot,
+    :gridlinewidth                => 1,
 )
 
 const _suppress_warnings = Set{Symbol}([
@@ -414,6 +435,7 @@ add_aliases(:linealpha, :la, :lalpha, :lα, :lineopacity, :lopacity)
 add_aliases(:markeralpha, :ma, :malpha, :mα, :markeropacity, :mopacity)
 add_aliases(:markerstrokealpha, :msa, :msalpha, :msα, :markerstrokeopacity, :msopacity)
 add_aliases(:fillalpha, :fa, :falpha, :fα, :fillopacity, :fopacity)
+add_aliases(:gridalpha, :ga, :galpha, :gα, :gridopacity, :gopacity)
 
 # series attributes
 add_aliases(:seriestype, :st, :t, :typ, :linetype, :lt)
@@ -469,6 +491,8 @@ add_aliases(:series_annotations, :series_ann, :seriesann, :series_anns, :seriesa
 add_aliases(:html_output_format, :format, :fmt, :html_format)
 add_aliases(:orientation, :direction, :dir)
 add_aliases(:inset_subplots, :inset, :floating)
+add_aliases(:gridlinewidth, :gridwidth, :grid_linewidth, :grid_width, :gridlw, :grid_lw)
+add_aliases(:gridstyle, :grid_style, :gridlinestyle, :grid_linestyle, :grid_ls, :gridls)
 
 
 # add all pluralized forms to the _keyAliases dict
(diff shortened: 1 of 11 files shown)


mkborregaard:
Nice! great effort! But what's with the PlotlyJS grid, it looks awful?

daschw:
Yes, now that you say it. It was not like this before, right? I'll have a look what I messed up here.

mkborregaard:
I'm not sure that's this PR, it looks like that on master as well (on my system) but it didn't always. Compare this from our docs
skaermbillede 2017-08-11 kl 10 30 15
with this from today's master
skaermbillede 2017-08-11 kl 10 30 25
I wonder if something went amiss in Plots or a setting was changed in PlotlyJS?

daschw:
Weird, this does not happen on my install with plots and plotlyjs master ...

mkborregaard:
But it does happen with this (ds-grid-axes) branch?

daschw:
Yes, but that's expected (as I just found out).
Here's the issue with PlotlyJS: Before this PR `foreground_color_grid` was not a supported attribute for plotly(js). Thus the plotly default (this very transparent gray) was chosen for the grid lines. The user, however, had no control of the grid colors. With the support of `foreground_color_grid` the grid color now defaults to match `foreground_color_subplot` which is black I think. I see three options here:
* Keep the old behaviour (no `foreground_color_grid` support for plotly(js))
* The current implementation of this branch: Most adjustment options by the user but maybe not the nicest defaults.
* Force an alpha on the grid color like in gr or glvisualize, so the user can choose the color, but there will always be the same alpha value defined by us.

I would go for the 3rd option as a compromise between the first two and try to find out the right alpha value so the plots without grid color set look the same as they do now. But maybe you prefer another option or see another alternative?

mkborregaard:
I'd tend to prefer 1 - just keep that as it is for plotlyjs

mkborregaard:
BTW, I find I get the ugly grid lines in Juno, but not in the REPL.

daschw:
Hm, I don't get it in either.

mkborregaard:
weird.

daschw:
Ok, Plotly(JS) now looks like this (first option from above - no grid color support):
grid_axis_plotlyjs_repl

BTW, now I am also experiencing your issue. The above image is from the REPL when running the same code in Juno an image with all black lines is displayed, and saving it as PNG from Juno results in something inbetween:
grid_axis_plotlyjs_juno

I don't think however, that this is related to this PR. Still, very strange 😕 !

mkborregaard:
It's probably unrelated, yeah.

daschw:
One more thing before merging: Currently the (x,y)grid is turned off for the attribute values `false` and `nothing`. Any opinions on whether the `grid` attribute should only accept `Bool` or maybe also act on symbols (and possibly strings) like `:none`,  `:off`, ... ?
A further possibility could be that `grid = :x` or `grid = :y` only shows the grid on the respective axis.

mkborregaard:
Yes, I like both of these options. Very much in Plots' spirit.

mkborregaard:
In principle, the best way might be to do something different - to make `grid` a magic argument, that can take the `x`, `foreground_color_grid` etc, allowing you to specify `grid = (:x, :red, :dot)`. @tbreloff suggested something similar here https://github.com/JuliaPlots/Plots.jl/pull/721#issuecomment-284759627 (though not exactly the same - I just think a magic argument might be better than asking for `Stroke`). That would be more versatile. 

But that may be overkill?

daschw:
I like your suggestions with the magic arguments and being able to set the stroke, but it might take me a few more days. :smile:

mkborregaard:
It doesn't have to be in this PR - it could be a separate project after this is merged.

daschw:
OK, I implemented the magic argument `grid` now - I hope mostly in Plots style ;-)
I introduced new axis attributes: `gridalpha`, `gridlinewidth` and `gridstyle`, which can be set either directly or by passing a tuple to the grid attribute. E.g: 
```julia
plot(..., grid = (:x, # xgrid = true, ygrid = false
    :solid, # gridstyle
    :blue, # foreground_colour_grid
    0.4, # gridalpha
    2 # gridlinewidth
    ))
```
All grid arguments are supported for GR, PyPlot and GLVisualize.
Plotly(JS) supports all but `gridstyle`. (Note: This leads to a new default look for plotly(js) grid lines unless we do not support foreground_color_grid, see also the discussion above)
I have not implemented the new attributes for InspectDR and PGFPlots.
I added a test example for the magic grid arguments:
```julia
x = rand(10)
p1 = plot(x, title = "Default looks")
p2 = plot(x, grid = (:y, :olivedrab, :solid, 0.5), title = "Modified y grid")
p3 = plot(deepcopy(p2), title = "Add x grid")
xgrid!(:on, :cadetblue, 2, :dashdot)
plot(p1, p2, p3, layout = (1, 3), label = "", fillrange = 0, fillalpha = 0.3)
```
If we want to merge this I will add the respective images to PlotReferenceImages soon.
Here is the testexample for
* **GR**
magicgrid_gr

* **PyPlot**
magicgrid_pyplot

* **Plotly(JS)**
magicgrid_plotlyjs

I guess we can still discuss the defaults and the PlotlyJS behaviour.

mkborregaard:
Awesome. This is great!
Only two things - WRT plotlyjs I guess there is no way of getting most of this behaviour but keeping the default lines? They are a lot prettier iMHO.
Also just one thing to consider - `grid` has a second meaning in Plots, used to create a regular subplot layout. I don't know if this may cause some confusion?

src/args.jl
```diff
@@ -163,6 +163,24 @@ const _scaleAliases = Dict{Symbol,Symbol}(
     :log  => :log10,
 )
 
+const _allGridSyms = [:x, :y, :z,
```
mkborregaard: nice

mkborregaard:
Nice code. This turned out to be a pretty big effort! Apart from my small considerations above, this is ready to go in IMHO.

daschw:
Thanks a lot for your feedback Michael!
I have one more suggestion regarding the plotlyjs issue. We could choose different defaults on all backends, so that the plotlyjs default grid looks nicer and the grid attributes can be supported by plotly.
I'd propose something like `gridstyle = :solid, gridalpha = 0.2, gridlinewidth = 0.5, foreground_color_grid = :match`:
* **GR**
grid_defaults_gr
* **PyPlot**
grid_defaults_pyplot
* **PlotlyJS**
grid_defaults_plotly

If you think this is too big a change to the default looks of GR and PyPlot, I am fine with not supporting the grid attributes in plotlyjs as well.

mkborregaard:
In that case I'd prefer the stippled-style grid lines. I am a bit on the fence here, but leaning towards putting full control in Plots, overriding the defaults for all the backends, and using something very much like you last example but stippled lines. But I am not really sure...

daschw:
Hm `gridstyle = ?`
* **GR**
grid_defaults_gr_style

* **PyPlot**
grid_defaults_pyplot_style

* **PlotlyJS** does not support gridstyle so it's the same as above.

We can also ask others for their opinion. Either the Plots members, or in the Gitter channel.

mkborregaard:
gridstyle = dot 👍 
Def nice with some more views.

daschw:
Does this also close #296?

mkborregaard:
Oh, nice old issue :-) @pkofod ?

pkofod:
I guess it pretty much does, and if it doesn't completely, then new, specific issues can be opened !

PR #1009, opened 2017-08-15
Add stride to wireframe/surface

JackDevine:
I only have this for PyPlot at the moment, I would be more than happy
to add other backends if people find it useful. Also, if people have
any ideas for aliases, then I would be happy to do that too.


JackDevine:
I find this really useful for making 3d plots that are not too cluttered. Also, using this, one can create the closest thing that there is to a waterfall plot in Plots.jl
```julia
using Plots; pyplot()

xx = linspace(-1,2,100)
tt = linspace(0,1,100)
arr = [sin(2*pi*t)*exp(-(x - 2t)^2/0.05) for t in tt, x in xx]

surface(xx, tt, arr, rstride=10, cstride=100, w=1, linecolor=:black, color=cgrad([:white,:white]))
```
figure_1

mkborregaard:
Looks good. It would be really useful if you take a look at whether this would be easy enough to implement for at least gr and plotlyjs, so we don't end up adding two attributes that only work with pyplot. 
BTW, it might be more Plots syntax to call it `stride` and take a tuple (where 0 could be the default value). wirefame_stride, surface_stride, surf_str, str might be good aliases?

JackDevine:
Thanks for those suggestions! I made the easy changes, so now the interface is:
```julia
julia> xx = yy = linspace(-2, 2, 100)
-2.0:0.04040404040404041:2.0

julia> A = [exp(-(x^2 + y^2)) for x in xx, y in yy];

julia> surface(xx, yy, A, w=0.5, color=:black, stride=(5,5))

```
Which is much nicer. But now that I look into it I realize that I don't actually know how to do gr and plotly. Sorry about that...

mkborregaard:
Hi @JackDevine I don't think we should add officially documented keywords for something that is only supported on a single backend - but we can merge this now so the functionality will be available, and then add the documentation on it when there are implementations for at least GR and Plotly.
Could you rebase on master and resolve the conflicts?

JackDevine:
@mkborregaard Good point about the documentation. It turned ou that this was harder to achieve in other backends than I had thought.

mkborregaard:
Thanks!

Issue #25, opened 2015-09-19
xlim and ylim


joshday:
I hope this isn't something I missed.  Feature request for something similar to R's `xlim` and `ylim` plot arguments for specifying the endpoints for the axes.

Ticks:
- [x] Gadfly/Immerse
- [x] PyPlot
- [x] Qwt
- [ ] UnicodePlots (not possible?)
- [ ] Winston (not possible?)

Limits:
- [x] Gadfly/Immerse
- [x] PyPlot
- [x] Qwt
- [x] UnicodePlots
- [x] Winston

Tick labels:
- [x] Gadfly/Immerse
- [x] PyPlot
- [ ] Qwt (not possible?)
- [ ] UnicodePlots (not possible?)
- [ ] Winston (not possible?)

tbreloff:
You didn't miss it. I also need to add the ability to set tick labels and grid lines. 

In general, you can always access the underlying plotting package directly. I export Gadfly for example, and you can get the underlying plot object from the "o" field. So if there's something missing you should still be able to access the underlying functionality.

tbreloff:
Mostly done with limits/ticks for Gadfly/Immerse... I'll add some checkboxes to your comment to keep track.  For now, there are lots of ways to update.  I'll push to master soon:

```
using Plots
plot(rand(100), title="Changing ticks and axis limits")
plot!(xlims=(25,75),xticks=25:5:75)
ylims!((0.6,0.8))
yticks!([0.6,0.63,0.7,0.8])
```
![tmp](https://cloud.githubusercontent.com/assets/933338/10051432/5f007d4a-61ef-11e5-926b-47d0640747d4.png)

joshday:
:+1:

tbreloff:
This is as complete as I'm going to make it on first pass.  I'll close the issue, and if anyone wants more extensive functionality (and it's possible for that backend) I'll open separate issues.

PR #1011, opened 2017-08-15
Enforce plot order in user recipes

martinbiel:
Process user recipes depth-first rather than breadth-first to enforce the plot order given in the recipe. Fixes the problems addressed in [#25](https://github.com/JuliaPlots/RecipesBase.jl/issues/25)

Passes all tests on GR and PyPlot backends.

### DIFF
--- a/src/pipeline.jl
+++ b/src/pipeline.jl
@@ -60,29 +60,26 @@ function _process_userrecipes(plt::Plot, d::KW, args)
     args = _preprocess_args(d, args, still_to_process)
 
     # for plotting recipes, swap out the args and update the parameter dictionary
-    # we are keeping a queue of series that still need to be processed.
+    # we are keeping a stack of series that still need to be processed.
     # each pass through the loop, we pop one off and apply the recipe.
     # the recipe will return a list a Series objects... the ones that are
-    # finished (no more args) get added to the kw_list, and the rest go into the queue
-    # for processing.
+    # finished (no more args) get added to the kw_list, the ones that are not
+    # are placed on top of the stack and are then processed further.
     kw_list = KW[]
     while !isempty(still_to_process)
-        # grab the first in line to be processed and pass it through apply_recipe
-        # to generate a list of RecipeData objects (data + attributes)
+        # grab the first in line to be processed and either add it to the kw_list or
+        # pass it through apply_recipe to generate a list of RecipeData objects (data + attributes)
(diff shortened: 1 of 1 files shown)


mkborregaard:
Just to give some context from https://github.com/JuliaPlots/RecipesBase.jl/issues/25 , the issue is that series order is changed when the series dispatches to a new user recipe. Here is an MWE:

If you have a recipe that does not dispatch to another recipe the series plot in order
```
struct Solu
       v
       t
 end

# this recipe does not call any other type recipe
@recipe function f(sol::Solu, v::AbstractVector)
       lw := 5
       @series begin
            label := "first series"
            sol.v, sol.t  # sol is resolved to its components here
       end
       @series begin
            label := "second series"
            1:10,v
       end
 end
```
skaermbillede 2017-08-15 kl 21 41 38
If you dispatch to a different recipe:
```julia
@recipe function f(sol::Solu)
       sol.v, sol.t
  end

@recipe function f(sol::Solu, v::AbstractVector)
       lw := 5
       @series begin
            label := "first series"
            sol # here
       end
       @series begin
            label := "second series"
            1:10,v
       end
 end
```
the order of series is reversed:
skaermbillede 2017-08-15 kl 21 42 29

mkborregaard:
I am in favour of the change, and I think the fix looks solid. Also thanks @martinbiel for running the tests yourself.
@daschw ? @tbreloff ?

src/pipeline.jl
```diff
@@ -60,29 +60,26 @@ function _process_userrecipes(plt::Plot, d::KW, args)
...
-
-            if isempty(recipedata.args)
-                _process_userrecipe(plt, kw_list, recipedata)
-            else
-                # args are non-empty, so there's still processing to do... add it back to the queue
-                push!(still_to_process, recipedata)
-            end
+        # recipedata should be of type RecipeData.  if it's not then the inputs must not have been fully processed by recipes
+        if !(typeof(next_series) <: RecipeData)
+            error("Inputs couldn't be processed... expected RecipeData but got: $recipedata")
```
daschw: I think `$recipedata` should be replaced with `$next_series` in the error here.

mkborregaard: Nice catch!

mkborregaard:
Thanks @martinbiel , great contribution!

daschw:
Agreed, thanks a lot!

PR #1014, opened 2017-08-16
import plot! placeholder from RecipesBase

daschw:
Requires https://github.com/JuliaPlots/RecipesBase.jl/pull/26

### DIFF
--- a/src/Plots.jl
+++ b/src/Plots.jl
@@ -8,7 +8,7 @@ import StaticArrays
 using StaticArrays.FixedSizeArrays
 
 @reexport using RecipesBase
-import RecipesBase: plot, animate
+import RecipesBase: plot, plot!, animate
 using Base.Meta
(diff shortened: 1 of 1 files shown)


mkborregaard:
so this should be merged now?

daschw:
I think as soon as a new RecipesBase version is tagged, right?

daschw:
I guess it can also be merged now. Then you'll get a warning in the tests that `RecipesBase.plot!` could not be imported.

daschw:
So yes, I think this should be merged before we tag a new RecipesBase release and we also should also tag a new Plots release before that. Right?

mkborregaard:
I'll roll the releases tomorrow if you don't get to it tonight

daschw:
I do Plots tonight and we can do RecipesBase after Plots is tagged.

PR #1015, opened 2017-08-16
only flip colorbar for marker_z on gr

daschw:
With #1005 not only the colorbar is flipped for marker_z on GR but also the yaxis. See, e.g `test_examples(:gr, 32)`:

gr32_old

With this PR only the colorbar is flipped:

gr32_new


mkborregaard:
You're a machine!

PR #969, opened 2017-07-12
Add fillrange for Plotly backend

pearlzli:
These two commits resolve plotly/Plotly.jl#34, allowing the user to fill between two lines using `fillrange`. Previously, using `fillrange` gave the following warning:

```julia
julia> using Plots

julia> Plots.plotly();

julia> x = linspace(-6, 6, 100);

julia> y = sin.(x);

julia> p = plot(x, y .+ 1, fillrange = y .- 1, label = "Hello")
WARNING: fillrange ignored... plotly only supports filling to zero. fillrange: [-0.720585,-0.606535,-0.49826,-0.397347,-0.305278,-0.223403,-0.152925,-0.0948765,-0.0501105,-0.0192835,-0.002848,-0.00104508,-0.0139012,-0.0412278,-0.0826237,-0.137482,-0.204996,-0.284178,-0.373863,-0.472736,-0.579347,-0.692131,-0.809432,-0.92953,-1.05066,-1.17105,-1.28893,-1.40257,-1.5103,-1.61054,-1.70183,-1.78281,-1.85231,-1.9093,-1.95294,-1.98261,-1.99785,-1.99845,-1.9844,-1.95591,-1.91338,-1.85745,-1.78895,-1.70886,-1.61837,-1.51881,-1.41163,-1.29841,-1.18082,-1.06057,-0.939431,-0.819182,-0.701586,-0.58837,-0.481193,-0.38163,-0.291141,-0.211055,-0.142545,-0.0866182,-0.0440947,-0.0155985,-0.00154777,-0.00214877,-0.0173926,-0.0470557,-0.0907026,-0.147693,-0.21719,-0.298175,-0.389458,-0.489701,-0.597433,-0.711071,-0.82895,-0.949338,-1.07047,-1.19057,-1.30787,-1.42065,-1.52726,-1.62614,-1.71582,-1.795,-1.86252,-1.91738,-1.95877,-1.9861,-1.99895,-1.99715,-1.980
```

and produced this plot:

capture2

With these commits, the warning is no longer thrown and the result is:

capture

PR #1023, opened 2017-08-19
RFC: added ribbon to plotly

piever:
Added a slightly modified version of #969 to allow ribbon plots in plotly(js). Now `plot(sin, ribbon = 0.1)` gives:

newplot-5

### DIFF
--- a/src/utils.jl
+++ b/src/utils.jl
@@ -496,6 +496,7 @@ function make_fillrange_from_ribbon(kw::KW)
     rib1, rib2 = -first(rib), last(rib)
     # kw[:ribbon] = nothing
     kw[:fillrange] = make_fillrange_side(y, rib1), make_fillrange_side(y, rib2)
+    (get(kw, :fillalpha, nothing) == nothing) && (kw[:fillalpha] = 0.5)
 end
 
 #turn tuple of fillranges to one path
--- a/src/utils.jl
+++ b/src/utils.jl
@@ -498,6 +498,14 @@ function make_fillrange_from_ribbon(kw::KW)
     kw[:fillrange] = make_fillrange_side(y, rib1), make_fillrange_side(y, rib2)
 end
 
+#turn tuple of fillranges to one path
+function concatenate_fillrange(x,y::Tuple)
+    rib1, rib2 = first(y), last(y)
+    yline = vcat(rib1,(rib2)[end:-1:1])
+    xline = vcat(x,x[end:-1:1])
+    return xline, yline
+end
+
 function get_sp_lims(sp::Subplot, letter::Symbol)
     axis_limits(sp[Symbol(letter, :axis)])
 end
--- a/src/backends/gr.jl
+++ b/src/backends/gr.jl
@@ -1096,7 +1096,7 @@ function gr_display(sp::Subplot{GRBackend}, w, h, viewport_canvas)
                 st = series[:seriestype]
                 gr_set_line(series[:linewidth], series[:linestyle], series[:linecolor]) #, series[:linealpha])
 
-                if st == :shape || series[:fillrange] != nothing
+                if (st == :shape || series[:fillrange] != nothing) && series[:ribbon] == nothing
                     gr_set_fill(series[:fillcolor]) #, series[:fillalpha])
                     l, r = xpos-0.07, xpos-0.01
                     b, t = ypos-0.4dy, ypos+0.4dy
--- a/src/backends/plotly.jl
+++ b/src/backends/plotly.jl
@@ -589,9 +589,9 @@ function plotly_series(plt::Plot, series::Series)
         # if hasfillrange is true, return two dictionaries (one for original
         # series, one for series being filled to) instead of one
         d_out_fillrange = deepcopy(d_out)
+        d_out_fillrange[:showlegend] = false
         if isa(series[:fillrange], AbstractVector)
             d_out_fillrange[:y] = series[:fillrange]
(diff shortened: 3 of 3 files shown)


piever:
There is only one thing I'm not sure about. Now the plot has a line in the middle of the ribbon, so that if I click on the legend symbol (using plotly interactivity), the ribbon goes away but the central line remains (you can also see it playing with fillalpha). Similar behaviour happens with the fillrange plots. Should I leave as is or remove the line?

mkborregaard:
Yeah, :ribbon is not a seriestype, it is something in addition to a line. So the line should be shown, with a semi-transparent ribbon around, just like in GR. If you can't get the ribbon to display in the legend that's no problem. I'd venture that it's preferable to not show the ribbon in the legend in fact.

mkborregaard:
But the ribbon doesn't look transparent in the example you posted - can you post an example of the current behaviour?

src/utils.jl
```diff
@@ -498,6 +498,14 @@ function make_fillrange_from_ribbon(kw::KW)
     kw[:fillrange] = make_fillrange_side(y, rib1), make_fillrange_side(y, rib2)
 end
 
+#turn tuple of fillranges to one path
```
mkborregaard: If this is only meant to be used by the plotlyjs backend I'd think it makes sense to keep it there.

mkborregaard: Or how was your thinking here?

piever: You're right, I thought maybe it would be useful to also implement ribbon for other backends, but I guess it can stay in the plotly file as long as it's only used by plotly.

piever:
To get the transparency you need to give an explicit `fillalpha` (same as in gr()):
`plot(sin, ribbon = 0.1, fillalpha = 0.5)`
newplot-6

In the legend I can either show only the ribbon or only the line. You are right that the latter makes a lot more sense (otherwise the ribbon in the legend also gets the transparency, as you see in this png).
Only thing is: GR shows both ribbon and line in the legend, maybe it should be corrected to only show the line?

mkborregaard:
No let's keep the GR behaviour as it is. I'm surprised the fillalpha isn't default - IMHO it should be.

piever:
I've changed the legend to be a line in case of ribbon plot for plotly. Now on plotly():
`plot(sin, ribbon = 0.1, fillalpha = 0.5)`
sinjs

whereas on gr:
`plot(sin, ribbon = 0.1, fillalpha = 0.5)`
singr

Regarding where to keep the extra utility function needed, I've made up my mind that I would prefer it to be in the general plots utils.jl, because it basically is the way that I used to get shaded error in StatPlots and last time I checked PGFPlots supported `shadederror` but not ribbon, so it could be used to get ribbon in PGFPlots as well.

Now that also plotly supports `ribbon` I think `shadederror` in StatPlots doesn't have much reason to exist. I would just wait a bit that this `ribbon` implementation is part of a Plots release and then also groupapply can switch to just using ribbon plots to display error.

The only slightly annoyance that I would have to completely remove `shadederror` is that GR displays the legend with also the shaded area. My understanding is that for publication quality figures, this is a bit unorthodox and it'd be nice to at least have an option to only display the line, but I'm really not sure how to implement that.

Whether `fillalpha` should default to 0.5 I think is a separate question and no matter what we decide, it should be uniform across backends.

mkborregaard:
1. Your arguments for keeping the function in utils is good.
2. With regards to what the legend shows, this should be consistent among backends. gr's current behaviour was added in https://github.com/JuliaPlots/Plots.jl/pull/976 . I'm not sure there's a generally accepted  option for publication quality figures. I tried to look around (e.g. google image search "figure confidence site:sciencemag.org") and found very few graphs that had an in-plot legend. Those that did only had the lines in the legend, not the shaded error.
3. In that PR there's also a discussion of the fillalpha. I think that's also where the current fillalpha default of 1 was added? I definitely think having a ribbon with fillalpha default of 1 is not the right decision.

After making up our mind on the fillalpha, this can be merged, IMHO.

@daschw , @jheinen any comments?

daschw:
My objective in #976 was to make the legend marker look as similar as possible to the respective series, but I see now that this is probably not the preferred option for a ribbon plot. It could be changed to only show the line in the legend by changing https://github.com/JuliaPlots/Plots.jl/blob/master/src/backends/gr.jl#L1099 to `(st == :shape || series[:fillrange] != nothing) && series[:ribbon] == nothing`.
The fillalpha of the ribbon itself was not changed by #976. I agree that having a default alpha for ribbons would be nicer. This can be changed for all backends by adding something like `kw[:fillalpha] == nothing && kw[:fillalpha] = 0.5` somewhere around https://github.com/JuliaPlots/Plots.jl/blob/master/src/utils.jl#L498.
@piever Could you include these changes in this PR or should we open seperate PRs?

piever:
Thanks for the instructions, I hadn't thought one could change the fillalpha when translating from ribbon to fillrange, that's the best way. I'll try to include everything in this PR

piever:
PlotlyJS:
`plot(sin, ribbon = 0.1)`
sinjs

GR:
`plot(sin, ribbon = 0.1)`

singr

I agree that these are much better default: as soon as this is in a released version, I can take the now unnecessary shadederror from StatPlots and use this one for groupapply

mkborregaard:
Perfect, thanks a lot @piever and thanks @daschw for jumping in so quickly with high-quality comments. We pushed a new release a few days ago, but I'll attempt not to be too lazy about pushing a new one :-)

Issue #302, opened 2016-06-06
Guides/Titles should be PlotText


tbreloff:
Remove titlefont, title_location, etc, and just change title/guide and similar to be PlotText objects (which contain the text string, font, positioning, etc)

daschw:
Thanks for the hint @mkborregaard! I think we should also consider https://github.com/JuliaPlots/Plots.jl/issues/555 here. Although these are seperate issues, the fix could be implemented at once.

PR #1029, opened 2017-08-20
Allow turning on/off the axes border (fix #1025)

### LINKED ISSUE #1025: gr(): add right and top axis
I wonder if there is any way to add the two missing right and top axis when using the gr backend, to have a plot like in pyplot for instance with the boxed area.

daschw:
This implements the possibility of turning on/off the axes border in GR, PyPlot and GLVisualize by adding the new subplot arrdibute `draw_axes_border` with its aliases `show_axes_border` and `axes_border`. Currently the default is set to `draw_axes_border = false`. Note that this changes the default looks for PyPlot.

- **GR**
axes_border_gr

- **PyPlot**
axes_border_pyplot

### DIFF
--- a/src/arg_desc.jl
+++ b/src/arg_desc.jl
@@ -93,7 +93,7 @@ const _arg_desc = KW(
 :bottom_margin            => "Measure (multiply by `mm`, `px`, etc) or `:match` (matches `:margin`).  Specifies the extra padding on the bottom of the subplot.",
 :subplot_index            => "Integer.  Internal (not set by user).  Specifies the index of this subplot in the Plot's `plt.subplot` list.",
 :colorbar_title           => "String.  Title of colorbar.",
-:draw_axes_border         => "Bool.  Draw a border around the axes.",
+:framestyle               => "Symbol.  Style of the axes frame. Choose from $(_allFramestyles)",
 
 # axis args
 :guide     				  => "String. Axis guide (label).",
--- a/src/arg_desc.jl
+++ b/src/arg_desc.jl
@@ -93,6 +93,7 @@ const _arg_desc = KW(
 :bottom_margin            => "Measure (multiply by `mm`, `px`, etc) or `:match` (matches `:margin`).  Specifies the extra padding on the bottom of the subplot.",
 :subplot_index            => "Integer.  Internal (not set by user).  Specifies the index of this subplot in the Plot's `plt.subplot` list.",
 :colorbar_title           => "String.  Title of colorbar.",
+:draw_axes_border         => "Bool.  Draw a border around the axes.",
 
 # axis args
 :guide     				  => "String. Axis guide (label).",
--- a/src/args.jl
+++ b/src/args.jl
@@ -432,7 +432,7 @@ add_aliases(:foreground_color_title, :fg_title, :fgtitle, :fgcolor_title, :fg_co
 add_aliases(:foreground_color_axis, :fg_axis, :fgaxis, :fgcolor_axis, :fg_color_axis, :foreground_axis,
                             :foreground_colour_axis, :fgcolour_axis, :fg_colour_axis, :axiscolor)
 add_aliases(:foreground_color_border, :fg_border, :fgborder, :fgcolor_border, :fg_color_border, :foreground_border,
-                            :foreground_colour_border, :fgcolour_border, :fg_colour_border, :bordercolor, :border)
+                            :foreground_colour_border, :fgcolour_border, :fg_colour_border, :bordercolor)
 add_aliases(:foreground_color_text, :fg_text, :fgtext, :fgcolor_text, :fg_color_text, :foreground_text,
                             :foreground_colour_text, :fgcolour_text, :fg_colour_text, :textcolor)
 add_aliases(:foreground_color_guide, :fg_guide, :fgguide, :fgcolor_guide, :fg_color_guide, :foreground_guide,
--- a/src/args.jl
+++ b/src/args.jl
@@ -282,6 +282,7 @@ const _subplot_defaults = KW(
     :bottom_margin            => :match,
     :subplot_index            => -1,
     :colorbar_title           => "",
+    :draw_axes_border         => false,
 )
 
 const _axis_defaults = KW(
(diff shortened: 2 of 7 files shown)


mkborregaard:
I like it. Maybe add keywords :box and :frame?

daschw:
Good idea!

mkborregaard:
But maybe we should take the opportunity to think more long-term? The default gr lower-left and pyplot full box are not the only axes possibilities. There is also 
1) a box, but with thicker lower left
gglowleft

2) no box, only axes, but the axes don't meet
gggap

3) no axes, only gridlines (Gadfly style)
ggaxes

4) no axes or box at all (e.g. for maps)
ggvoid

All of these would be nice to set in connection with PlotThemes.jl etc. So maybe we shouldn't add a keyword with a bunch of aliases, but instead an axes_style / framestyle keyword that took a symbol? I'm just talking interface, we could always implement the gory details as we go along.

daschw:
Actually, I really like @tbreloff 's proposal in https://github.com/JuliaPlots/Plots.jl/issues/296#issuecomment-223563502 and https://github.com/JuliaPlots/Plots.jl/pull/721#issuecomment-284759627 to have Strokes as arguments for the border/axes/grid/tick lines as an underlying structure. Something like `foreground_color_axis`, e.g. could then be an alias of `axis_stroke` and when passed to `Plots.stroke()` it is identified as color.
I think this would provide the most user control on individual axis/border/grid lines in a clean way without introducing too many new arguments like `border_color`, `border_linestyle`, `border_alpha`, `border_width`. They would just be aliases for border_stroke.
Then it would be easy to introduce a new "meta" keyword like `framestyle` in your proposal above which acts on the border (and maybe axes) strokes.
The only issue I see right now with this approach is that currently you cannot set the linewidth of a Stroke to something between 0 and 1 with `Plots.stroke()` because it would identify it as an alpha. If `axis_stroke_alpha` and `axis_stroke_width` would be aliases for `axis_stroke`, you would no longer be able to set the stroke width to e.g. `0.5`.
A possible solution for this could be to restrict alphas to `Float`s in `Plots.allAlphas()`, which would allow to set the alpha with 0.5 and the linewidth with 1//2.

mkborregaard:
I like the proposal even though I am not sure I follow in detail what you mean. Except for the last bit - that's too magical for people to remember. You actually already have the problem now for magic argument `line` - in that case `alpha` wins https://github.com/JuliaPlots/Plots.jl/blob/master/src/args.jl#L596-L602 . Not completely sure how to deal with this.

daschw:
Yes, I know, we have this for all magical arguments. The difference, however, to my proposal is that this problem can be "easily" resolved by the user by specifying e.g. `grid_linewidth = 0.5`.
In my proposal above `grid_linewidth` would be an alias for, let's say, `grid_stroke`, as would be `grid_alpha`. Thus both, `grid_linewidth = 0.5` and `grid_alpha = 0.5` would result in `stroke(0.5)` in https://github.com/JuliaPlots/Plots.jl/blob/master/src/components.jl#L374 and alpha wins. Thus `grid_linewidth = 0.5` would actually change the alpha value.
I also see now that my possible solution above is not a solution at all, because `grid_linewidth = 0.5` would still change the alpha and `grid_alpha = 1//2` would change the linewidth ...
So, I don't really have an idea right now, how this could be done.
I just liked the idea of only having the attribute (border/axis/grid)_stroke instead of (border/axis/grid)_width, _color, _style, _alpha.

daschw:
Still your proposal with `framestyle` is a good idea and it can also be implemented without this Stroke structure change.
I'm just not sure whether to introduce new Plots attributes `_width/color/alpha/style` for `axis/border/tick` with cascading defaults and have `framestyle` act on them or to implement `framestyle` individually for different backends.
I think the first option plays better with PlotThemes and gives more control to the users. The downside is a lot of new attributes (border could even be left/right/bottom/top_).

daschw:
OK, I implemented your proposal from above now (except the 2nd plot): The attribute is now called `framestyle` instead of `draw_axes_border` and it can take the values `:box`, `:semi`, `:axes`, `:grid` and `:none`, defaulting to `axes`.
```julia
histogram(fill(randn(1000), 5), framestyle = [:box :semi :axes :grid :none],
    title = [":box" ":semi" ":axes" ":grid" ":none"], layout = 5, label = "")
```
now produces:
* **GR**:
framestyle_gr

* **PyPlot:**
framestyle_pyplot

I could not yet test it on GLVisualize.
I am open to any suggestion regarding different names and also maybe more aliases.

mkborregaard:
I love it.
Maybe aliases "frame" and also realias "border" (which currently is an alias for  "foreground_color_border")?
I just realised that currently `ticks = nothing` sets the same behaviour as `framestyle = :none`, but IMHO it actually shouldn't - it should remove the ticks and keep the frame IMHO. So no real conflict there.

daschw:
Yes, I agree, this PR changes the behaviour of `ticks = nothing` to
gr_noticks
which I think makes more sense than the current behaviour (hiding the axes). This was also mentioned here https://github.com/JuliaPlots/Plots.jl/pull/780#issuecomment-296523864.
Can you also think of aliases (or better defaults) for the framestyle attribute values (`:box`, `:semi`, `:axes`, `:grid`, `:none`?

mkborregaard:
I think they are great and don't need aliases.

daschw:
OK, great, just in case I initialized an alias `Dict` here https://github.com/JuliaPlots/Plots.jl/pull/1029/files#diff-e70c578e40cc32f982144c9a373ba5eaR185. So something like `box = :on` should also work (equivalent to `framestyle = :box`. If we get some ideas in time, we can extend it.

daschw:
I looked around a little, but I did not find the functionality for for plotly so far. I don't really like their docs...
GLVisualize also works:
framestyle_glvisualize
I don't know much about InspectDR. Maybe I can look into PGFPlots this weekend.

daschw:
I think the tests fail because of #1038. I'm gonna merge this and then update the reference images, if that's ok for you?

daschw:
Now framestyle = :none also works on GLVisualize:
framestyle_glvisualize

mkborregaard:
good idea

mkborregaard:
:tada:

mkborregaard:
> Actually, I really like @tbreloff 's proposal in #296 (comment) and #721 (comment) to have Strokes as arguments for the border/axes/grid/tick lines as an underlying structure

See also https://github.com/JuliaPlots/Plots.jl/issues/302 . It'd be good to continue thinking about this.

mkborregaard:
There's actually a common axis type we're missing - the one where the axes cross at 0,0 rather than at xmin, ymin

daschw:
Very good point! I will look into this.

mgiugliano:
May I ask why there is no trace of "framestyle" in the documentation?

mkborregaard:
Huh? The example with `framestyle` (no 34) indeed seems to have disappeared - and it also does not appear in the attribute table. I couldn't quite work out what's going on, but this seems weird.

pkofod:
Wow, didn't realize this was in, very nice!

Issue #962, opened 2017-07-03
`group` attribute spoils `markersize`


heliosdrm:
If data points of a scatter plot are grouped by the `group` attribute, mapping the marker size to a variable does not work properly -- except in the first group. 

See small example:

```julia
v = 1:10
g = repeat(["A","B"], inner=5)
scatter(v,v, markersize=2v, color=:lightgrey)
scatter!(v,v, markersize=2v, group=g)
```

Output:
![newplot](https://user-images.githubusercontent.com/10651028/27806911-b8b961fa-603d-11e7-959f-6b11b388c9ef.png)

Grey circles have the correct size; the reddish points ("A") just overlap the grey ones, but the green ones ("B") have a different scale.

heliosdrm:
However, when the data series are manually split, instead of using the `group` attribute, the behaviour is the one I expected (at least in this case). Namely

```julia
scatter(v[g.=="A"],v[g.=="A"], markersize=2v[g.=="A"])
scatter!(v[g.=="B"],v[g.=="B"], markersize=2v[g.=="B"])
```

![newplot](https://user-images.githubusercontent.com/10651028/28693868-29874112-7327-11e7-9438-9cff7470a34b.png)

My intuition is that both ways of grouping the data should provide the same result. In this particular case I'm grateful that it doesn't, so that I can choose how to define the size of the markers. But this looks like hacking an ambiguous behaviour, rather than the proper way of doing it.

I suspect that the size of the markers in the legend has something to do with this issue, but I don't understand how this works.

heliosdrm:
Ok, forget the previous comment. I've just understood what you meant.

Anyway, for what is worth, I would expect that `group` splits all vector arguments that are as long as the data series `x`, `y`, etc.

mkborregaard:
I see what you mean, that could work, I think (but wouldn't be trivial to implement).

piever:
Closed by #1049

Issue #56, opened 2015-10-29
TimeSeries support


milktrader:
Not sure how this is implemented, but following https://github.com/tbreloff/Plots.jl/issues/16 the interface would be to load TimeSeries with `timeseries()` and then use multiple dispatch for the `plot` method.

Ideally, one should be able to do this:
````julia
julia> using MarketData

julia> plot(AAPL, style="ohlc")
````
And it would just work. The style options could be the standard
- [ ] line
- [ ] ohlc
- [ ] candlestick

**Comment by tbreloff** [comment]
There are 2 ways you could consider implementing this:

```
function Plots.plot(arr::TimeArray; linetype=:path, kw...)
  # extract x/y arrays from arr
  # if linetype is :ohlc or :candlestick do the necessary processing
  plot(x, y; kw..., otherkwargs...)
end
```

or you could make a PlotRecipe like in: https://github.com/tbreloff/Plots.jl/blob/master/src/recipes.jl#L8

I'm thinking either way, the ohlc/candlestick plots might make sense as PlotRecipes.

**Comment by milktrader** [comment]
So is `:path` a simple line?

Would the idea be to write a method like:
````function Plots.plot(arr::TimeArray; linetype=:path, kw...)
  x = arr.timestamp
  y = arr.values
  # if linetype is :ohlc or :candlestick do the necessary processing
  plot(x, y; kw..., otherkwargs...)
end
````

**Comment by tbreloff** [comment]
`:path` and `:line` are similar... the difference is that `:line` implies you want to sort the data by x-value, whereas `:path` will allow "backtracking".  Example:

```
julia> using Plots

julia> x, y = rand(20),rand(20);

julia> plot(x,y, lab="path", w=3)

julia> plot!(x,y, l=:line, lab="line", w=3)
```

![tmp](https://cloud.githubusercontent.com/assets/933338/10828058/b54fbc72-7e48-11e5-8c3b-0b30297ef3c3.png)

tbreloff:
I'm going to close this, as it likely belongs as an implementation in another package.  Feel free to ask questions if you need help implementing it.

Issue #44, opened 2015-10-13
Animations


tbreloff:
In Qwt, I have an interface that looks like:

```
a = animation(plt, "/tmp/png")
for x in 0:0.1:5
    push!(plt, 1, x, sin(x))
    refresh(plt)
    saveframe(a)
end
makegif(a)
```

It creates a directory to store pngs of the snapshots, then at the end calls out to the system to run ffmpeg to create an animated gif.

This can't be the best way to do it.  Does anyone know of a better solution??  cc: @timholy

lobingera:
Actually it's not a Bad interface to provide animations like this. Maybe some timing interface could be provided so you can control the default playback speed in writing already.

tbreloff:
The part I'm really not happy with is the hacky dependence on the external program ffmpeg.  This is what I do now in Qwt (which is not a good solution for the general user):

```
makegif(animation::PlotAnimation) = run(`ffmpeg -framerate 20 -i $(animation.dir)/qtanim%04d.png -y $(animation.dir)/out.gif`)
```

lobingera:
@tbreloff, would you like to have a native ffmpeg implemented in julia? https://github.com/kmsquire/VideoIO.jl exists.

tbreloff:
Close, but it seems like only video playback is supported.

Evizero:
:100: for creating GIFs out of plots

tbreloff:
This was fairly easy (just need to document the ffmpeg dependency):

```
julia> using Plots
INFO: Recompiling stale cache file /home/tom/.julia/lib/v0.4/Plots.ji for module Plots.

julia> xy() = randn(10), randn(10)
xy (generic function with 1 method)

julia> p = scatter(xy()..., size=(500,300))
[Plots.jl] Initializing backend: immerse
WARNING: Method definition draw(Compose.Backend, Compose.Context) in module Compose at /home/tom/.julia/v0.4/Compose/src/container.jl:404 overwritten in module Immerse at /home/tom/.julia/v0.4/Immerse/src/compose.jl:33.
WARNING: Method definition display(Base.REPL.REPLDisplay{R<:Base.REPL.AbstractREPL}, Base.Multimedia.MIME{:text/html}, Gadfly.Plot) in module Gadfly at /home/tom/.julia/v0.4/Gadfly/src/Gadfly.jl:969 overwritten in module Immerse at /home/tom/.julia/v0.4/Immerse/src/display_gadfly.jl:144.
[...]

PR #1033, opened 2017-08-21
Group now groups kw arguments as well

### LINKED ISSUE #1031: grouping seem to spoil zcolor
Potentially related with #962. 
Example:

```
using DataFrames
import RDatasets
using StatPlots
gr()
iris = RDatasets.dataset("datasets","iris");
scatter(iris[:PetalLength], iris[:PetalWidth], zcolor = iris[:SepalLength], group = iris[:Species], layout = 3)
```
gives:

group

whereas

```
plt = plot(layout = 3)
counter = [0]
by(iris, :Species) do dd
    counter[1] += 1
    scatter!(plt, dd[:PetalLength], dd[:PetalWidth], zcolor = dd[:SepalLength], subplot = counter[1])
end
```

gives 
groupby

I'd be potentially in favor of the proposed improvement in #962: all args of the same length as the series also get split by `group`. If there's consensus that it's the way to go, I can try and implement that.

piever:
With this PR, when grouping, also keyword arguments get grouped, provided they are long enough in the first dimension. Fixes #1031 and #962 
New behavior:

```
using Plots
using DataFrames
import RDatasets
gr()
iris = RDatasets.dataset("datasets","iris");
scatter(iris[:PetalLength], iris[:PetalWidth], group = iris[:Species],
layout = 3, marker_z = iris[:SepalLength], markersize = iris[:SepalWidth])
```

group

Also, there was a forgotten method to group, which gives a Dict of indices instead of a column: it is fixed with this PR. Example syntax:

`scatter(iris[:PetalLength], iris[:PetalWidth], group = Dict(:a => collect(1:50), :b => collect(51:150)))`

Issue with this PR: in case of overlaid plots with colorbar, we get superimposed colorbars. We'd need a working keyword to make sure the user can set them to be the same:

`scatter(iris[:PetalLength], iris[:PetalWidth], group = iris[:Species], marker_z = iris[:SepalLength])
`
groupbad

### DIFF
--- a/src/args.jl
+++ b/src/args.jl
@@ -854,7 +854,7 @@ end
 # expecting a mapping of "group label" to "group indices"
 function extractGroupArgs{T, V<:AVec{Int}}(idxmap::Dict{T,V}, args...)
     groupLabels = sortedkeys(idxmap)
-    groupIds = VecI[collect(idxmap[k]) for k in groupLabels]
+    groupIds = Vector{Int}[collect(idxmap[k]) for k in groupLabels]
     GroupBy(groupLabels, groupIds)
 end
 
--- a/src/args.jl
+++ b/src/args.jl
@@ -840,16 +840,24 @@ end
 
 
 # this is when given a vector-type of values to group by
-function extractGroupArgs(v::AVec, args...)
+function extractGroupArgs(v::AVec, args...; legendEntry = string)
     groupLabels = sort(collect(unique(v)))
     n = length(groupLabels)
     if n > 100
         warn("You created n=$n groups... Is that intended?")
     end
     groupIds = Vector{Int}[filter(i -> v[i] == glab, 1:length(v)) for glab in groupLabels]
-    GroupBy(map(string, groupLabels), groupIds)
+    GroupBy(map(legendEntry, groupLabels), groupIds)
 end
 
+legendEntryFromTuple(ns::Tuple) = string(("$n " for n in ns)...)
+
+# this is when given a tuple of vectors of values to group by
+function extractGroupArgs(vs::Tuple, args...)
+    (vs == ()) && return GroupBy([""], [1:size(args[1],1)])
+    v = collect(zip(vs...))
+    extractGroupArgs(v, args...; legendEntry = legendEntryFromTuple)
+end
 
 # expecting a mapping of "group label" to "group indices"
 function extractGroupArgs{T, V<:AVec{Int}}(idxmap::Dict{T,V}, args...)
--- a/src/series.jl
+++ b/src/series.jl
@@ -515,6 +515,13 @@ end
         @series begin
             label     --> string(glab)
             idxfilter --> groupby.groupIds[i]
+            for (key,val) in d
+                length(args) == 0 && break
+                if key != :group && isa(val, AbstractArray) && size(val,1) == size(args[1],1)
+                    n = ndims(val)
+                    :($key) := val[groupby.groupIds[i], fill(Colon(), n-1)...]
+                end
+            end
             args
         end
     end
--- a/src/series.jl
+++ b/src/series.jl
@@ -511,13 +511,13 @@ end
 
 # split the group into 1 series per group, and set the label and idxfilter for each
 @recipe function f(groupby::GroupBy, args...)
+    lengthGroup = maximum(union(groupby.groupIds...))
     for (i,glab) in enumerate(groupby.groupLabels)
         @series begin
             label     --> string(glab)
             idxfilter --> groupby.groupIds[i]
             for (key,val) in d
-                length(args) == 0 && break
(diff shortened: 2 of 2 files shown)


mkborregaard:
Very nice.

mkborregaard:
Does it work with StatPlots, and with passing markershapes as a 1-row Matrix?

src/args.jl
```diff
@@ -854,7 +854,7 @@ end
 # expecting a mapping of "group label" to "group indices"
 function extractGroupArgs{T, V<:AVec{Int}}(idxmap::Dict{T,V}, args...)
     groupLabels = sortedkeys(idxmap)
-    groupIds = VecI[collect(idxmap[k]) for k in groupLabels]
+    groupIds = Vector{Int}[collect(idxmap[k]) for k in groupLabels]
```
mkborregaard: Must be a leftover from some old code

src/series.jl
```diff
@@ -511,10 +511,17 @@ end
 
 # split the group into 1 series per group, and set the label and idxfilter for each
 @recipe function f(groupby::GroupBy, args...)
+    lengthGroup = maximum(union(groupby.groupIds...))
     for (i,glab) in enumerate(groupby.groupLabels)
         @series begin
             label     --> string(glab)
             idxfilter --> groupby.groupIds[i]
+            for (key,val) in d
+                if key != :group && isa(val, AbstractArray) && size(val,1) >= lengthGroup
```
mkborregaard: why `>=`?

piever:
`markershape` works:
```
scatter(iris[:PetalLength], iris[:PetalWidth], group = iris[:Species],
markersize = iris[:SepalWidth], markershape = [:circle :star :diamond])
```
group

It also works with StatPlots but only with the DataFramesMeta syntax:

`@with iris scatter(:PetalLength, :PetalWidth, group = :Species, layout = 3, marker_z = :SepalLength, markersize = :SepalWidth)`

This one doesn't work:

`scatter(iris, :PetalLength, :PetalWidth, group = :Species, layout = 3, marker_z = :SepalLength, markersize = :SepalWidth)`

But occasionally the StatPlots syntax with the dots behaves a bit strangely, I'm not sure what it is.

src/series.jl
piever: It is possible to give a group dictionary with less elements than the actual data:
`plot(rand(10), group = Dict(:a => [1,3], :b = [2,4]), ms = rand(10))`
In that case I believe we still want to group the markersize.

piever:
Maybe there actually is one way to get it to work with StatPlots: I could try and do it [here](https://github.com/JuliaPlots/Plots.jl/blob/b585a0ef61b1af1308cf7ebf0f746c42b39d3869/src/args.jl#L864) instead and then it would also work with the StatPlots syntax. I'll give it a shot and make a second attempt at a PR.

piever:
And actually there's one extra corner case I hadn't taken into account: also tuples of vectors of the correct length should be grouped, for example, when passing asymmetric ribbons:
`plot(rand(10), ribbon = (ones(10), 0.5*ones(10)))`

Given that we're fixing `group`, I think we take the opportunity to also fix `:groupedbar` (see [#44](https://github.com/JuliaPlots/StatPlots.jl/issues/44) and [#56](https://github.com/JuliaPlots/StatPlots.jl/issues/56) ). The change would have to happen in the same section I'm modifying now.

src/series.jl
mkborregaard: I think I disagree here, it seems like it could cause more trouble than benefit.

piever: You're probably right, I'm changing it to `==`

piever:
Added `tuple` support (for asymmetric ribbons) and addressed your concern. I think all is well now considering the new syntax in DataFrames.
One last thing: I also wanted to add a new method to group, for grouping on more than one column, in the same way as I do with `groupapply`. For example:

```
using DataFrames
import RDatasets
using StatPlots
gr()
school = RDatasets.dataset("mlmRev","Hsb82")
@given school density(:MAch, group = (:Sx, :Sector))
```

would split in 4 groups depending on all the combinations of `Male` `Female` and `Public` `Private` and the legends would be

```
Male Public
Male Private
Female Public
Female Private
```

is something that I end up doing pretty often with my data (for example when having 3 separate experimental groups and splitting by treatment vs no treatment). It's pretty easy to implement, if you're also happy to have this feature I can add it here.

mkborregaard:
yes, sounds great :-)

piever:
Feature added:

`@given school density(:MAch, group = (:Sx, :Sector))
`

density

piever:
I'll merge later today if everybody is happy with the changes.

mkborregaard:
:tada:

mkborregaard:
I feel like the example https://github.com/JuliaPlots/Plots.jl/pull/1033#issuecomment-325214217 should be added to the StatPlots readme. It's really cool.

Issue #1032, opened 2017-08-21
Error bars interact weirdly with zcolor in GR


piever:
GR tries to match the color of the error bars with the color from `zcolor`, but it doesn't quite work (plus, I can't really find the correct attribute to change them to black).
MWE:

`scatter(1:2, [1,1], zcolor = [0,1], xerr = 0.05, markersize = 8, markerstrokewidth= 3)`

![error](https://user-images.githubusercontent.com/6333339/29516808-d9a655d8-8669-11e7-9006-897d066f550a.png)

Maybe a simple solution would be to just plot the error bars of the same color as `markerstroke` ?

mkborregaard:
That's just a big. Error bars get their color from `markerstrokecolor`

daschw:
Closed by #1035

PR #1035, opened 2017-08-23
don't draw fill for 1D markers in gr

piever:
Solves #1032 : GR was trying to draw the "fill" part of markers even for 1D markers (i.e. `:hline, :vline, :+, :x, :cross, :xcross`)

### DIFF
--- a/src/backends/gr.jl
+++ b/src/backends/gr.jl
@@ -326,7 +326,10 @@ function gr_draw_markers(series::Series, x, y, msize, mz)
                 cfuncind(ci)
                 GR.settransparency(_gr_gradient_alpha[ci-999])
             end
-            gr_draw_marker(x[i], y[i], msi, shape)
+            # don't draw filled area if marker shape is 1D  
+            if !(shape in (:hline, :vline, :+, :x, :cross, :xcross))
+                gr_draw_marker(x[i], y[i], msi, shape)
+            end
         end
     end
 end


daschw:
Great, thanks!

mkborregaard:
Nice!

daschw:
@piever If you added fix #1032 in the PR title, merging it would automatically close the issue.

piever:
> If you added "fix #1032" in the PR title, merging it would automatically close the issue.

Thanks for the pro tip, had no idea!

daschw:
Maybe `:+, :x, :cross, :xcross` should be excluded here?
See https://github.com/JuliaPlots/PlotReferenceImages.jl/blob/master/Plots/gr/0.12.3/ref25.png and compare it to https://github.com/JuliaPlots/PlotReferenceImages.jl/blob/master/Plots/gr/0.12.1/ref25.png

piever:
It's true that now to change the color of the cross you need to change `markercolor` rather than `markerfillcolor` as there is no fill. On the other hand, a 1D `:+` sign with fill and boundary looks a bit strange to me, but if it's common to put it in plots together with other markers that have fill and stroke, then I should probably put it back.

mkborregaard:
So, the issue is that `seriescolor`, which is controlled by the series, only controls  the fill of markers, but not the markerstroke. A better solution would be to ensure that for non-fill markers, the markerstroke defaults to the seriescolor.

daschw:
see for instance the pyplot behaviour: https://github.com/JuliaPlots/PlotReferenceImages.jl/blob/master/Plots/pyplot/0.12.3/ref25.png
I think `:cross` was designed to also have a fill.

mkborregaard:
Oh thanks :flushed: I did go to check pyplot but only on the website, not to PlotReferenceImages.

piever:
Is it possible to get `:cross` to mean the same fat cross marker in GR as it does in pyplot?

daschw:
Looking at src/components.jl I really think `:cross` and `:xcross` have a fill being `Plots.Shape`s.  So I propose no fill drawing on gr for `:+`, `:x`, `:hline`, `:vline`, but keeping it for `:cross` and `:xcross`

piever:
Ok, if it's a shape, then in the future one can also make it look like the pyplot shape right? I'll make a PR with the two GR fixes that are still required (correct tick vertical alignment and putting back fill in `:cross` and `:xcross`

PR #1037, opened 2017-08-23
changes default histogram bins

mkborregaard:
It used to be 30, but that works very poorly when you have very few data points (which we often have). Long-time I'd like to change this to Ward's criterion, but scott could be fine until then. It gives more bins than the usual :sturges criterion, which is the default in StatsBase and many other packages (but nobody likes). cc @oschulz

Also updates the description with the new normalization mode.


src/args.jl
```diff
@@ -203,7 +203,7 @@ const _series_defaults = KW(
     :markerstrokewidth => 1,
     :markerstrokecolor => :match,
     :markerstrokealpha => nothing,
-    :bins              => :auto,        # number of bins for hists
+    :bins              => :scott,        # number of bins for hists
```
oschulz: Shouldn't `:auto` default to `:scott`, then?

oschulz:
I'm fine with changing it - it was you who wanted to

@mkborregaard > stick with 30 for the moment, for backwards consistency.

back then (#787). ;-)

There's a problem, though: Looking back at our old discussion, one issue was that while `:scott` and friends do nicely, they will choose a fine binning if there's enough data - too fine for the default histogram drawing style (`:barhist`). So if  `:scott` becomes the new default, then we either have to change the default drawing style to `:stephist`, or we have to set an upper limit on the number of bins for `:auto`-binning.

mkborregaard:
Ha ha.
Wondering if it would make sense to let the default style depend on the (inferred) bin number. Might be a little too automagical.

src/args.jl
mkborregaard: Good question - that'd make it a little less transparent what's going on to the user, especially if defaults change, though.

mkborregaard:
I've now 1) changed the auto-binning to Freedman-Diaconis (after a bit of reading), and 2) made the default for :histogram change from :barhist to :stephist when there are more than a million data points. I thought a lot about how to do that best - different n will lead to different numbers of bins under certain criteria, and the effect of bins also depend on plot size and line width etc. But we're just looking for a default here, the user can always choose what (s)he wants explicitly.
Thoughts?

oschulz:
I like it!

mkborregaard:
Thanks!

mkborregaard:
I've got a question for you @oschulz - the current binning for multidimensional histograms appears to just apply a binning rule on the first dimension as if it were a 1-d histogram. But is that consistent with the theoretical best way of defining 2d-histograms? It seems to me as if it risks having too few points in each bin. Maybe it's not necessary to have an n-dimensional code (though I see the prettiness of it) when we only offer 1d and 2d histograms.

oschulz:
> the current binning for multidimensional histograms appears to just apply a binning rule on the first dimension as if it were a 1-d histogram

Ooops - it should actually do it for all dimensions, at least that was my intent ...

mkborregaard:
But even if it does that, is that valid?

oschulz:
Auto-binning separately for every axis? Yes, I think it's valid.

mkborregaard:
I wrote too fast. I've reread your code and realize you dealt with my concern in `n = max(1, n_samples^(1/N))`. Sorry for the noise!

oschulz:
:-)

PR #1038, opened 2017-08-23
Adjust margin (ticks + margin given by user) with GR ( fix #1028 )

### LINKED ISSUE #1028: xticks being cut with GR even with extra margin
With GR, if I do: `plot(1:10,rand(10), right_margin = 10mm)` the `10.0` tick on the x axis get cut, even if I give extra margin

test

Actually, it seems like, at least on my machine, specifying a margin has no effect wih GR

piever:
Now GR takes in consideration the margin input by the user when making the plot + the ticks.

`plot(["String $i" for i in 1:10], rand(10), xrotation = 90)`

longname

`plot(["String $i" for i in 1:10], rand(10), xrotation = 90, left_margin = 10mm)`

longnameleft

I recommend playing with it to make sure that the default margins look good / it works robustly. I only tested on Juno as my GKS installations is a bit disfunctional.

### DIFF
--- a/src/backends/gr.jl
+++ b/src/backends/gr.jl
@@ -326,7 +326,7 @@ function gr_draw_markers(series::Series, x, y, msize, mz)
                 cfuncind(ci)
                 GR.settransparency(_gr_gradient_alpha[ci-999])
             end
-            # don't draw filled area if marker shape is 1D  
+            # don't draw filled area if marker shape is 1D
             if !(shape in (:hline, :vline, :+, :x, :cross, :xcross))
                 gr_draw_marker(x[i], y[i], msi, shape)
             end
@@ -543,10 +543,10 @@ end
 
 
 function _update_min_padding!(sp::Subplot{GRBackend})
-    leftpad   = 10mm
-    toppad    = 2mm
-    rightpad  = 2mm
-    bottompad = 6mm
+    leftpad   = 10mm + sp[:left_margin]
+    toppad    = 2mm  + sp[:top_margin]
+    rightpad  = 2mm  + sp[:right_margin]
+    bottompad = 6mm  + sp[:bottom_margin]
     if sp[:title] != ""
         toppad += 5mm
     end
@@ -558,10 +558,11 @@ function _update_min_padding!(sp::Subplot{GRBackend})
                         valign = :top,
                         color = sp[:xaxis][:foreground_color_axis],
                         rotation = sp[:xaxis][:rotation])
-            h = 0
+            h = 0.0
             for (cv, dv) in zip(xticks...)
                 tbx, tby = gr_inqtext(0, 0, string(dv))
-                h = max(h, tby[2] - tby[1])
+                tby_min, tby_max = extrema(tby)
+                h = max(h, tby_max - tby_min)
             end
             bottompad += 1mm + gr_plot_size[2] * h * px
         else
--- a/src/backends/gr.jl
+++ b/src/backends/gr.jl
@@ -542,6 +542,30 @@ function gr_display(plt::Plot)
 end
 
 
+function gr_set_xticks_font(sp)
+    flip = sp[:yaxis][:flip]
+    mirror = sp[:xaxis][:mirror]
+    gr_set_font(sp[:xaxis][:tickfont],
+                halign = (:left, :hcenter, :right)[sign(sp[:xaxis][:rotation]) + 2],
+                valign = (mirror ? :bottom : :top),
+                color = sp[:xaxis][:foreground_color_axis],
+                rotation = sp[:xaxis][:rotation])
+    return flip, mirror
+end
+
+
+function gr_set_yticks_font(sp)
+    flip = sp[:xaxis][:flip]
+    mirror = sp[:yaxis][:mirror]
+    gr_set_font(sp[:yaxis][:tickfont],
+                halign = (mirror ? :left : :right),
+                valign = (:top, :vcenter, :bottom)[sign(sp[:yaxis][:rotation]) + 2],
+                color = sp[:yaxis][:foreground_color_axis],
+                rotation = sp[:yaxis][:rotation])
+    return flip, mirror
+end
+
+
 function _update_min_padding!(sp::Subplot{GRBackend})
     # Add margin given by the user
     leftpad   = 10mm + sp[:left_margin]
@@ -747,13 +771,7 @@ function gr_display(sp::Subplot{GRBackend}, w, h, viewport_canvas)
 
         if !(xticks in (nothing, false))
             # x labels
-            flip = sp[:yaxis][:flip]
-            mirror = sp[:xaxis][:mirror]
-            gr_set_font(sp[:xaxis][:tickfont],
-                        halign = (:left, :hcenter, :right)[sign(sp[:xaxis][:rotation]) + 2],
-                        valign = (mirror ? :bottom : :top),
-                        color = sp[:xaxis][:foreground_color_axis],
-                        rotation = sp[:xaxis][:rotation])
+            flip, mirror = gr_set_xticks_font(sp)
             for (cv, dv) in zip(xticks...)
                 # use xor ($) to get the right y coords
                 xi, yi = GR.wctondc(cv, xor(flip, mirror) ? ymax : ymin)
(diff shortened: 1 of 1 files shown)


src/backends/gr.jl
```diff
@@ -542,32 +542,74 @@ function gr_display(plt::Plot)
...
 
+function gr_set_xticks_font(sp)
+    flip = sp[:yaxis][:flip]
+    mirror = sp[:xaxis][:mirror]
+    gr_set_font(sp[:xaxis][:tickfont],
+                halign = (:left, :hcenter, :right)[sign(sp[:xaxis][:rotation]) + 2],
+                valign = (mirror ? :bottom : :top),
+                color = sp[:xaxis][:foreground_color_axis],
+                rotation = sp[:xaxis][:rotation])
+    return flip, mirror
```
mkborregaard: It seems unintuitive that `gr_set_xticks_font` should return `flip, mirror` - what is the reasoning behind this design?

src/backends/gr.jl
```diff
@@ -542,32 +542,74 @@ function gr_display(plt::Plot)
...
+end
+
+
+function gr_set_yticks_font(sp)
```
mkborregaard: This function is so similar to gr_set_xticks_font that it seems more reasonable to have a `gr_set_ticks_font(letter)` function, a pattern used many places in axes.jl.

src/backends/gr.jl
```diff
@@ -542,32 +542,74 @@ function gr_display(plt::Plot)
...
+    # Add margin given by the user
+    leftpad   = 2mm  + sp[:left_margin]
+    toppad    = 2mm  + sp[:top_margin]
+    rightpad  = 4mm  + sp[:right_margin]
+    bottompad = 2mm  + sp[:bottom_margin]
+    # Add margin for title
     if sp[:title] != ""
         toppad += 5mm
     end
-    if sp[:xaxis][:guide] != ""
```
mkborregaard: Why'd you take this out?

mkborregaard:
You're getting a Travis failure on this one. The reason is that this will change the look of the reference images. To address this, the approach is 1) install PlotReferenceImages.jl and VisualRegressionTests, 2) go to Plots/tests/runtests.jl and run the tests in a local session. This will open a Gtk window for every image that does not match the reference image, and gives you the opportunity to overwrite. 3) after overwriting images, ensure that Pkg.test("Plots") succeeds on your machine, 4) open a PR on PlotReferenceImages with your new images, and link to this one.

It's a bit of a hassle the first time, but very useful :-)

mkborregaard:
And oh, thanks for this!

src/backends/gr.jl
piever: As it was, GR only added space for the ticks if there was also an axis guide, which would be a problem if there is no axis guid but ticks are big.

jheinen:
Thanks for your help. Looks (nearly) perfect. Shouldn't the rotated texts be vertically aligned?

mkborregaard:
thanks @piever !

piever:
Thanks @jheinen for merging! I've kept the previous code for displaying the ticks, and the rule seems that if it's not rotated it stays centered, otherwise it aligns to the side on which it is rotated. However, when rotating of 90 degrees, this rule looks a bit silly. I can play with it a bit tomorrow and, if I come up with a better rule, I'll make a new PR.

daschw:
FYI this made the subplots with xrotation really small:
https://github.com/JuliaPlots/PlotReferenceImages.jl/blob/master/Plots/gr/0.12.3/ref29.png
compared to
https://github.com/JuliaPlots/PlotReferenceImages.jl/blob/master/Plots/gr/0.12.1/ref29.png
Although it looks similar in pyplot (https://github.com/JuliaPlots/PlotReferenceImages.jl/blob/master/Plots/pyplot/0.12.3/ref29.png), I think the old behaviour was nicer.
@piever Do you think it is possible to avoid this but still having the fixes from this PR?

jheinen:
Unfortunately, there is now another problem. Repeating the same plot leads to strange results:
```julia
using Plots
plot(randn(5))
```
The plot origin changes on each `plot()` call.

Didn't find a reason yet ...

jheinen:
Problem with plot origins fixed in https://github.com/JuliaPlots/Plots.jl/pull/1039

mkborregaard:
> FYI this made the subplots with xrotation really small

This is the reason we have the test suite based on PlotReferenceImages. Given the number of users who work on Plots master, it is a lot better to catch plotting regressions before merging rather than after. I think we should try to ensure that if Travis fails on a PR that Pgk.test does succeed on a local version of PlotReferenceImages, giving us the chance to discuss any changes to the plots befor merging.

 In this case it'd be nice to find a fix for this. Perhaps adjusting the spacing between subplots is the way to go, though?

piever:
@daschw that's a bit unexpected, what is the exact code you used for that plot?
One of the advantages now is that the user can reduce the margin by hand (e.g. with `bottom_margin = -5mm`), but it'd be nice to have a better default. On the other hand, the previous code worked by chance, as it was putting a constant margin no matter what...

daschw:
The code for all examples that are used for the Plots tests is in /src/examples.jl. This special example can be found here: https://github.com/JuliaPlots/Plots.jl/blob/master/src/examples.jl#L288-L295

jheinen:
It doesn't look SO bad (with the latest patches):
screen shot 2017-08-25 at 04 04 32

piever:
Well, the plot command is giving extra extra margin by hand:

```julia
plot(rand(100,6),layout=@layout([a b; c]),title=["A" "B" "C"],
                               title_location=:left, left_margin=[20mm 0mm],
                               bottom_margin=50px, xrotation=60)
```

If you simply give:

```julia
plot(rand(100,6),layout=@layout([a b; c]),title=["A" "B" "C"],
                               title_location=:left, xrotation=60)
```

You get the correct margin:

gr

And thanks a lot @jheinen for the repeated plots bug: for some reason it doesn't happen in Juno and I completely missed it!

daschw:
OK, thanks @piever and @jheinen for clarifying this. I did not look into the code at all (manual margins), just compared the images and liked the old one better. But I guess this is expected ... my fault.

mkborregaard:
Yes, I am also fine with this now. Great, thanks!

mkborregaard:
Perhaps it'd be a nice idea to change the example?

daschw:
Yep, I think we should reduce the margin.

piever:
Agreed: it's weird to test regressions on an example that, if the plot command works correctly, would actually look bad...

piever:
@jheinen There actually is still a weird issue on which I'm going to need your help: the conversion to screen coordinates happening here:

https://github.com/JuliaPlots/Plots.jl/blob/master/src/backends/gr.jl#L598

Is not always correct, for two possible reasons:

1) The `gr.jl` file uses this `gr_plot_size` variable to take into account the plot size which gets modified, but I'm not sure it always gets updated at the right time

2) Some times, the window gets some extra resizing (for example, due to the size of the plot pane in Juno) which should be taken into account

Something like this seems to work better, but I really don't know enough to understand what's going on:

`bottompad += 1mm + l/(GR.wctondc(1,1)[2]- GR.wctondc(0,0)[2]) * sp.plt[:size][2] * px`

piever:
Or to put the question more succinctly, what it the most reliable way to convert the output of `gr_inqtext` to pixel based Plots.jl coordinate, given that [this](https://github.com/JuliaPlots/Plots.jl/blob/master/src/backends/gr.jl#L598) simplistic approach fails? It's being hard for me to find out with trial and error...

jheinen:
GR uses a metric specification (`setwsviewport`) to map normalized device coordinates (NDC => [0,1]x[0,1]) to device coordinates/pixels (DC). Using `inqdspsize()` you can obtain the display size. For vector out put formats, this size is fixed.

```julia
using GR
inqdspsize()
```

On a MacBook, the result is `(0.286449334905011, 0.17903083431563188, 1280, 800)`, for SVG, you get `(0.254, 0.1905, 1024, 768)`. The first two numbers describe the metric size, the second two numbers the size in pixels.

So, if you want 800 x 600 pixels, you would set the viewport using:
`setwsviewport(0, 800 * 0.254 / 1024, 0, 600 * 0.1905 / 768)`
... and the workstation window with the same aspect ratio:
 `setwswindow(0, 1, 0, 600/800`

If the `plt.size()` is updated in Plots when using Juno, the reverse transformation should work, too.

The problem is, if we start thinking in pixels (e.g. for margins, font / marker sizes, etc.), the resulting graphics is always device dependent, unless we also take care of a DPI setting, which is currently not supported in GR :-(.

piever:
Still there is something mysterious going on about the `gr_inqtext` calculation after #1039 and I can't figure out what it is.

If I run the command:

`plot(["String $i" for i in 1:10], rand(10), xrotation = 90, size = (600, 200))`

the first time I get as `gr_get_ticks_size` for the `xticks` the value 0.079 (which is too little, so actually now #1028 is still an issue). From the second time on, when I run the plot command, I get 0.059, which is way too small.

If instead I select `GR.selntran(1)` when calling `gr_inqtext`, the first time I get 0.11, which gives an appropriate margin, whereas from the second time on, it goes down to 0.084 which is too little. Also, this causes the bug of the shifting origin in GSK but not Juno.

I'm completely lost, any idea what's going on?

jheinen:
`gr_plot_size` is not set correctly on the first call.

jheinen:
@piever : problem should be fixed in https://github.com/JuliaPlots/Plots.jl/pull/1044 - I did not merge the PR because most tests failed, which is probably not related to GR!?

mkborregaard:
The GR tests fail too. It may not be a problem, though we just need to check the PlotReferenceImages. If noone else goes through them in the meantime I'll check them tomorrow morning :-)

mkborregaard:
This PR appears to have caused a World Age issue: https://github.com/JuliaPlots/Plots.jl/issues/1047

PR #1042, opened 2017-08-25
make linecolor default to :auto

daschw:
this allows the users to easily ~~get rid of the borders~~ match the linecolor to the seriescolor in shape/histogram/bar-like plots with `linecolor = :match` while still defaulting to the old behaviour:
```julia
x = randn(1000) .+ (0:2:6)'
plot(histogram(x, title = "default"), histogram(x, linecolor = :match, title = "linecolor = :match"))
```
blackborder

### DIFF
--- a/src/args.jl
+++ b/src/args.jl
@@ -198,7 +198,7 @@ const _series_defaults = KW(
     :seriestype        => :path,
     :linestyle         => :solid,
     :linewidth         => :auto,
-    :linecolor         => :match,
+    :linecolor         => :auto,
     :linealpha         => nothing,
     :fillrange         => nothing,   # ribbons, areas, etc
     :fillcolor         => :match,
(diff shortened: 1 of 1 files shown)


mkborregaard:
Hmm I'm not sure I like this. Having a border of the same color of the fill area can make the color harder to see, but it doesn't remove it, and it can create some fuzzy shapes because the line has a width. That's an issue for instance with shapes where you sometimes need the edges to be exact (e.g. for shapefiles). In these cases I prefer `linewidth = 0` which removes the line.

daschw:
The way I see it, this is just an additional option. You can still always set `linewidth = 0`. I was just always irritated, since I started using Plots, that `linecolor = :match` does not "match" for shapes, bars and histograms. But maybe that's just me ...

mkborregaard:
Maybe - or maybe others :-) I guess I can't see a reason for linecolor to match the fill - why would you then want the line in the first place?

daschw:
For instance for something like `histogram(randn(1000), linecolor = :match, fillalpha = 0.5)`. I probably should have used this as example 😆

mkborregaard:
> I probably should have used this as example

Yes :-) 
Good point, I see that use.

PR #1044, opened 2017-08-25
gr: defer the calculation of padding constraints

jheinen:


daschw:
I tried to run the tests locally with this and this undoes the fix of @piever in #1038 regarding the cut xlabels:
```julia
using Plots
test_examples(:gr, 1)
```
gr_xticklabel_cut1
(look at the xticklabel 50)

PR #1046, opened 2017-08-27
Remove hidden files when checking version numbers

mkborregaard:
Addresses this problem: https://github.com/JuliaPlots/Plots.jl/pull/975/files#r135406391


daschw:
Thanks for fixing this!

PR #1048, opened 2017-08-28
Reduce margin to 10 px

mkborregaard:
Based on the discussion here https://github.com/JuliaPlots/Plots.jl/pull/1038


mkborregaard:
Depends on https://github.com/JuliaPlots/PlotReferenceImages.jl/pull/13

PR #1053, opened 2017-08-29
Make wand edges possible in statplots

mkborregaard:
Required for https://github.com/JuliaPlots/StatPlots.jl/pull/77

### DIFF
--- a/src/recipes.jl
+++ b/src/recipes.jl
@@ -532,8 +532,8 @@ function _auto_binning_nbins{N}(vs::NTuple{N,AbstractVector}, dim::Integer; mode
 end
 
 _hist_edge{N}(vs::NTuple{N,AbstractVector}, dim::Integer, binning::Integer) = StatsBase.histrange(vs[dim], binning, :left)
-_hist_edge{N}(vs::NTuple{N,AbstractVector}, dim::Integer, binning::AbstractVector) = binning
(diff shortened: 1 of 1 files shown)


src/recipes.jl
```diff
@@ -527,8 +528,11 @@ function _auto_binning_nbins{N}(vs::NTuple{N,AbstractVector}, dim::Integer; mode
 end
 
 _hist_edge{N}(vs::NTuple{N,AbstractVector}, dim::Integer, binning::Integer) = StatsBase.histrange(vs[dim], binning, :left)
-_hist_edge{N}(vs::NTuple{N,AbstractVector}, dim::Integer, binning::Symbol) = _hist_edge(vs, dim, _auto_binning_nbins(vs, dim, mode = binning))
 _hist_edge{N}(vs::NTuple{N,AbstractVector}, dim::Integer, binning::AbstractVector) = binning
+function _hist_edge{N}(vs::NTuple{N,AbstractVector}, dim::Integer, binning::Symbol)
```
mkborregaard: Because the logic is slightly different from the existing binning methods (it returns an edge vector) it has a slightly unclean interface with the existing code. @oschulz would probably say it looks like a hack :-)

mkborregaard:
I've simplified this massively, with the issue that `wand_edges` and `_auto_binning_nbins` have now become type-unstable. I don't think it's an issue as there is a function barrier, and the assignment location (`edges`)  was never more type certain that AbstractVector{<:Number} anyway. Benchmarks confirm there's no performance loss. And the code is a lot simpler.

mkborregaard:
I'll also merge this today if I get no comments.

PR #1056, opened 2017-08-29
Fix incorrect escape sequence error on Julia nightly

jgoldfar:


src/args.jl
```diff
@@ -175,7 +175,7 @@ function hasgrid(arg::Symbol, letter)
     if arg in _allGridSyms
         arg in (:all, :both, :on) || contains(string(arg), string(letter))
     else
-        warn("Unknown grid argument $arg; $letter\grid was set to `true` instead.")
+        warn("Unknown grid argument $arg; $letter\\grid was set to `true` instead.")
```
daschw: Thanks! I was not aware that this errors on julia nightly.
How about `$(Symbol(letter, :grid))`?
Your change changes the warning shown in julia 0.6.

jgoldfar: That's what I get for making a change without checking how the message looked... Your suggested change is now implemented. Thanks!

daschw:
Great, thanks!

PR #1057, opened 2017-08-29
RFC: Add :origin framestyle

daschw:
This adds the `:origin` option and its alias `:zeroline` to the newly introduced `framestyle` attribute (cf #1029).
```julia
x, y = [randn(10) rand(10) + 1], [randn(10) rand(10) + 1]
scatter(x, y, framestyle = :origin,
    title = ["Origin inside data" "Origin outside data"], layout = 2, label = "")
```
**GR**
origin_gr

**PyPlot**
origin_pyplot

**GLVisualize**
origin_glvisualize

cc: @mkborregaard

### DIFF
--- a/src/args.jl
+++ b/src/args.jl
@@ -181,13 +181,14 @@ function hasgrid(arg::Symbol, letter)
 end
 hasgrid(arg::AbstractString, letter) = hasgrid(Symbol(arg), letter)
 
-const _allFramestyles = [:box, :semi, :axes, :grid, :none]
+const _allFramestyles = [:box, :semi, :axes, :origin, :grid, :none]
 const _framestyleAliases = Dict{Symbol, Symbol}(
     :frame              => :box,
     :border             => :box,
     :on                 => :box,
     :transparent        => :semi,
     :semitransparent    => :semi,
+    :zeroline           => :origin,
 )
 # -----------------------------------------------------------------------------
 
--- a/src/args.jl
+++ b/src/args.jl
@@ -189,6 +189,7 @@ const _framestyleAliases = Dict{Symbol, Symbol}(
     :transparent        => :semi,
     :semitransparent    => :semi,
     :zeroline           => :origin,
+    :zero               => :origin,
 )
 # -----------------------------------------------------------------------------
 
--- a/src/args.jl
+++ b/src/args.jl
@@ -181,15 +181,13 @@ function hasgrid(arg::Symbol, letter)
 end
 hasgrid(arg::AbstractString, letter) = hasgrid(Symbol(arg), letter)
 
-const _allFramestyles = [:box, :semi, :axes, :origin, :grid, :none]
+const _allFramestyles = [:box, :semi, :axes, :origin, :zerolines, :grid, :none]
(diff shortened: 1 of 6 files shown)


daschw:
I guess it can be discussed whether the ticks and tick labels should be directly at the axes instead ...

mkborregaard:
I'm in favour of that

daschw:
Of the ticks and tick labels directly at the axes?

mkborregaard:
Yes, well in favour of all of it, it looks very nice :-) I do think numbers and labels are best on the axes, I was trying to find some examples of this, but it doesn't seem very common in the published literature. Here's one, though
bar

daschw:
Hm, an example that definitely supports your preference is `scatter(rand(10) - 2, rand(10) -2, framestyle = :origin)`
origin_bad
That really does not look nice (maybe it could be checked, if all data is negative)
An example that would favour the current implementation is maybe `histogram(randn(1000), framestyle = :origin)` (or other plots that are very "crowded" around zero).
Maybe we should have two different framestyles for both use cases?

mkborregaard:
I think in the second case you wouldn't choose the origin framestyle. IMHO ticks should always go on the axes.

daschw:
* **GLVisualize**

origin_glvisualize

* **GR**

origin_gr

* **PyPlot**

origin_pyplot

mkborregaard:
Very very nice! I guess this is not easy to implement in plotly? Do you think we should suppress the 0.0 ticks?

daschw:
I was thinking about suppressing the 0.0 ticks, too.
* **GR**
origin_gr

* **PyPlot**
origin_pyplot

(GLVisualize is the same logic as GR) I think that's better. What's your opinion?

Regarding plotly I will think about this when I find some time to dig into the docs and then I'll try to implement all framestyles at once (if possible). Except someone else beats me to it :-)

mkborregaard:
Yes, much better. I've been thinking about learning more plotly myself so perhaps I'll have a chance but no promises at the moment :-)

mkborregaard:
Do you know what, it turns out that the framestyle where the axes meet at the origin but the ticks are outside the plots (yes, the one where I said "you'd never do that") is actually a standard setting in plotly:
https://plot.ly/python/scatterplot-matrix/
newplot 4
And appears to be very common http://help.qlik.com/en-US/sense/June2017/Subsystems/Hub/Content/Visualizations/ScatterPlot/scatter-plot.htm

In fact the version with the tick marks on the axis appears much rarer. Sorry. I still like the other one better but I was clearly wrong to dismiss this type 😳

daschw:
It looks nicer without the black ticks though (compared to my initial implementation).
I'm not against two separate framestyles, but I can't think of two different descriptive names right now.

daschw:
Now there are two new framestyles `:origin` and `:zerolines`:

* **GLVisualize**

zerolines_glvisualize

* **GR**

zerolines_gr

* **PyPlot**

zerolines_pyplot

I'm not sure about the names though. Any ideas/opinions?

mkborregaard:
:1 from me. I think the names are great, though I suspect this might be an area where many will have to look in the docs.

PR #1058, opened 2017-08-29
Make it possible for downstream users (e.g. StatPlots) to define that a seriestype is "histogram-like", "line-like" or "surface-like"

mkborregaard:


mkborregaard:
Thanks!

PR #1060, opened 2017-08-30
gr: suppress default output in CI or Jupyter mode

jheinen:
_update_min_padding() is called before the backend is "started".
In GR, this triggers an automatic initialization and starts the
default output driver (X11, Quartz or GDI). This doesn't make
sense in headless testing environments (CI) or when using Plots
in Jupyter notebooks.

These patches correct this behavour - but it's unclear (to me),
why the call of _update_min_padding() can't be postponed.

### DIFF
--- a/src/backends/gr.jl
+++ b/src/backends/gr.jl
@@ -580,6 +580,9 @@ function gr_get_ticks_size(ticks, i)
 end
 
 function _update_min_padding!(sp::Subplot{GRBackend})
+    if !haskey(ENV, "GKSwstype") && isijulia()
+        ENV["GKSwstype"] = "svg"
+    end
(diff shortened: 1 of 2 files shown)


mkborregaard:
There's the same issue in Juno, right?

jheinen:
Unfortunately - yes. The `isjulia()` check doesn't help here.

jheinen:
Would work, but for the first plot, we again have the world age problem. As mentioned by others, the only workaround is to explicitly set the backend using `gr()` - strange!

mkborregaard:
I wonder if there's an interaction with using Requires in Juno? Could something be done in the init function?

PR #1061, opened 2017-08-30
WIP - fix 32-bit errors

mkborregaard:
This branch is an attempt to address 32 bit errors in Plots. So far just based on appveyor. It is possible to build a 32 bit julia for testing on a 64-bit linux but I don't have that at the moment.

### DIFF
--- a/src/components.jl
+++ b/src/components.jl
@@ -89,7 +89,7 @@ end
 "create a shape by picking points around the unit circle.  `n` is the number of point/sides, `offset` is the starting angle"
 function makeshape(n; offset = -0.5, radius = 1.0)
     z = offset * π
-    Shape(partialcircle(z, z + 2π, n+1, radius))
+    Shape(partialcircle(z, z + 2π, n+1., radius))
 end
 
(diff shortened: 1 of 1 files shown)


pkofod:
What is the problem here? That `Int32` overflows?

mkborregaard:
No idea :joy: I just opened the PR because the 32-bit version fails and I thought I could fix it easily.

pkofod:
I was just wondering if `Int64(n)` would fix it as well...

mkborregaard:
Good idea. It would be nice to have Plots working on 32-bit, especially if it's just a few adjustments. Though the fact that this only shows up  in Appveyor and has exactly 0 issues tells me not that many are currently running the package on a 32-bit system.

pkofod:
Is there a link to the failing 32bit output?

mkborregaard:
Just follow the appveyor badge - we've allowed 32-bit failures.

mkborregaard:
This should not be necessary anymore

Issue #83, opened 2015-11-22
Add tests/examples/docs


tbreloff:
- [ ] 3D line/scatter
- [ ] Contours (and Surface type)
- [ ] copy
- [ ] stroke/brush, etc
- [ ] spy
- [ ] PlotRecipe
- [ ] reference MLPlots/ExamplePlots
- [ ] colorschemes
- [ ] dataframes
- [ ] density
- [ ] linked subplots

pkofod:
I figure this is the correct place to ask. Do you have any plans for a "readthedocs"-type documentation, or do you plan to provide jupyter notebooks as there are now? I think everyone appreciates all the functionality effort that you throw into this project, but sometimes I find it a bit difficult to find example of the listed keywords being used. I'd be happy to help, but I just need to know which direction you're planning to go in.

tbreloff:
The short answer is yes... I'd like to have a full website that is a rewrite of the readme, the examples, and some short tutorials which cleanly show some basic usage and some "tricky hacks" for power users. The current notebooks were not written as documentation or tutorials... They are mostly just tests and experiments I created while developing. The better place to look is in the examples linked from the readme. 

I'm not a web developer, and it seemed easier to just have markdown files when I first started. Any ideas and collaboration for new docs will be very appreciated!  My one hope is that all docs can be easily/automatically regenerated without heavyweight dependencies. Maybe I should just use Jekyll and continue to use markdown?  Please post any ideas you have... Thanks.

mmagnuski:
readthedocs supports Mkdocs which is very simple - you just create a `doc` or `docs` directory where you keep all the `.md` files and readthedocs does the rest. This is copied from readthedocs documentation:

> When you choose *Mkdocs as your Documentation Type*, we will first look for a `mkdocs.yml` file in your repository. If we don’t find one, we will generate one for you. We will look inside a `doc` or `docs` directory first, and then default to the top-level of your documentation.
> Then Mkdocs will build any files with an `.md` extension. If you have a `README.md`, it will be transformed into an `index.md` automatically.

pkofod:
I only mentioned readthedocs because a lot of packages seem to use it, I have no experience with it myself. It does seem to be quite easy to do, though.

tbreloff:
> you just create a doc or docs directory where you keep all the .md files and readthedocs does the rest

This sounds pretty reasonable, but it would take me a little effort to familiarize myself with readthedocs and get it set up for the first time.

[...]

PR #1062, opened 2017-08-31
WIP: allow to group user recipes

piever:
This PR allows to choose custom grouping behavior for user recipes (a notable example being `groupedbar`, see [#83](https://github.com/JuliaPlots/StatPlots.jl/pull/83) ). Now if the user recipe is set to `group_as_matrix`, it will be grouped as a matrix so `x,y` will go from being two vector to being a 1D array and a matrix, as required by some user recipes.

With the PRs:

```julia
using StatPlots
groupedbar([1 ,2 ,1 ,2 ,1 ,2], rand(6), group = [1,1,2,2, 3, 3])
```

bar

### DIFF
--- a/src/series.jl
+++ b/src/series.jl
@@ -516,8 +516,9 @@ splittable_kw(key, val::Tuple, lengthGroup) = all(splittable_kw.(key, val, lengt
 split_kw(key, val::AbstractArray, indices) = val[indices, fill(Colon(), ndims(val)-1)...]
 split_kw(key, val::Tuple, indices) = Tuple(split_kw(key, v, indices) for v in val)
 
-function build_arg_mat(x_ind, x, y, groupby)
-    y_mat = fill(NaN, length(keys(x_ind)), length(groupby.groupLabels))
+function groupedvec2mat(x_ind, x, y::AbstractArray, groupby, def_val = y[1])
+    y_mat = Array{promote_type(eltype(y), typeof(def_val))}(length(keys(x_ind)), length(groupby.groupLabels))
+    fill!(y_mat, def_val)
     for i in 1:length(groupby.groupLabels)
         xi = x[groupby.groupIds[i]]
         yi = y[groupby.groupIds[i]]
@@ -525,6 +526,8 @@ function build_arg_mat(x_ind, x, y, groupby)
     end
     return y_mat
 end
+
+groupedvec2mat(x_ind, x, y::Tuple, groupby) = Tuple(groupedvec2mat(x_ind, x, v, groupby) for v in y)
 
 group_as_matrix(t) = false
 
@@ -558,7 +561,12 @@ group_as_matrix(t) = false
         end
         x_u = unique(x)
         x_ind = Dict(zip(x_u, 1:length(x_u)))
+        for (key,val) in d
+            if splittable_kw(key, val, lengthGroup)
+                :($key) := groupedvec2mat(x_ind, x, val, groupby)
+            end
+        end
         label --> reshape(groupby.groupLabels, 1, :)
-        typeof(g)((x_u, (build_arg_mat(x_ind, x, arg, groupby) for arg in last_args)...))
+        typeof(g)((x_u, (groupedvec2mat(x_ind, x, arg, groupby, NaN) for arg in last_args)...))
     end
 end
--- a/src/series.jl
+++ b/src/series.jl
@@ -516,6 +516,16 @@ splittable_kw(key, val::Tuple, lengthGroup) = all(splittable_kw.(key, val, lengt
 split_kw(key, val::AbstractArray, indices) = val[indices, fill(Colon(), ndims(val)-1)...]
 split_kw(key, val::Tuple, indices) = Tuple(split_kw(key, v, indices) for v in val)
 
+function build_arg_mat(x_ind, x, y, groupby)
+    y_mat = fill(NaN, length(keys(x_ind)), length(groupby.groupLabels))
+    for i in 1:length(groupby.groupLabels)
+        xi = x[groupby.groupIds[i]]
+        yi = y[groupby.groupIds[i]]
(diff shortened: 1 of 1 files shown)


src/series.jl
```diff
@@ -519,16 +519,31 @@ split_kw(key, val::Tuple, indices) = Tuple(split_kw(key, v, indices) for v in va
...
+                        :($key) := split_kw(key, val, groupby.groupIds[i])
+                    end
                 end
+                args
             end
-            args
         end
+    else
+        g = args[1]
+        x, y = g.args
```
mkborregaard: What about z? What if there is only y?

mkborregaard:
Could you please say a little more about why this is necessary and why you've chosen this approach to deal with it?

mkborregaard:
The reason I'm asking is that Plots already has some mechanisms for communicating how to deal with input arguments (wrapping as Surface, SliceIt) and I want to make sure we don't use very different designs to do similar things - but I can't say for sure just by looking at the code.

src/series.jl
piever: The "WIP" label is for real: I just got this to work and wanted to see if we're happy with the approach, I still need to:
- Figure it out in case of only `y` or `x`, `y`, `z`
- Add the same fix for "column like" keyword arguments as in here: #1033

piever:
Yes, the design is not at all obvious but it seems to me that it's the only solution. The issue is that, when a type recipe (like `groupedbar` for example) arrives at the [point in the pipeline where the grouping happens](https://github.com/JuliaPlots/Plots.jl/pull/1062/files#diff-db35f720028daec1866d4fdb8a1627a6R520), the only available information to Plots.jl, unless the user gives extra keywords, is the type of the recipe, in this case, `GroupedBar`. In order for the function to do something depending on the type of the input, the cleanest is multiple dispatch, so there is an auxiliary function that dictates how the grouping should proceed. I think two types of groupings makes sense: the normal one, into separate series, and the `group_as_matrix` style, as in `groupedbar`. The standard should of course be the first option. So the auxiliary function defaults to the first style of grouping, unless it is overloaded for a specific type (for example, I believe both `groupedbar` and `groupedboxplot` should overload it). This auxiliary function has to be available to the whole Plots.jl ecosystem, and the common dependency of everything is RecipesBase, so it should live there. With these PRs, if a random package wants to add a type recipe with `group_as_matrix` they can do so without depending on Plots or StatPlots.

Sorry for the wall of text, hope it helps clarify!

mkborregaard:
I've given this some thought - an obvious alternative would be to use new keywords, e.g. a new keyword to Plots that behaved differently, or just bespoke keywords in the user recipe. But I think there's something to be said for having this behaviour consistent. From the user's perspective, `group` behaves intuitively in your call above (so does `x`, though using x like this in bar and boxplots is something that some new users find difficult to intuit), even if the internal implementation is different.
So :+1: from me. 
With regards the `group_as_matrix` function, you could put it temporarily in Plots so it's available to StatPlots, and then we can move to RecipesBase with the 0.7 shift.
(As an aside, WRT the groupedbar recipe, we need to make the tick marks work too - e.g. if `x` is a vector of Symbols, that Symbol needs to be under the middle bar.)

piever:
100% agreed on the first part: users shouldn't have to memorize different keywords because of implementation details.

As to where to put `group_as_matrix`, I'm not sure I understand the reasoning. An important thing to point out is that this change to RecipesBase is non breaking. The `group_as_matrix` function is already programmed to default to `false`, which means that for all existing recipes, the change in RecipesBase doesn't do anything. For this change to do something, the `group_as_matrix` function needs to actively be overloaded. Couldn't we simply make a point release in RecipesBase and then add the correct lower bound for RecipesBase in Plots and StatPlots? The rest of the echosystem could keep the older RecipesBase 0.2.0 and everything would keep working as usual. Then, as soon as the user updates to a newer Plots.jl, a newer RecipesBase.jl would be required, ensuring that things still work.

mkborregaard:
Of course the package manager should be able to handle this.

It's just an update philosophy we've decided to follow on RecipesBase, to ensure that all versions of RecipesBase work with all versions of Plots on a given julia release. Part of our selling argument is that RecipesBase is a dep where extra care is being taken to avoid conflicts.

But I don't see the problem either by just putting it into Plots now and transferring to RecipesBase later. It will be a few months where downstream users can't use this functionality to define recipes in packages that only depend on RecipesBase (StatPlots and PlotRecipes and any non-package usages would have it available though).

piever:
I understand your point more clearly now: I'll just move the function here to Plots.jl. There may still be some uncertainty as to what exactly the `group_as_matrix` function should do with 1 or 3 arguments, I'll think more carefully about it and will ask for feedback here in case of doubt.

mkborregaard:
Awesome, keep up the great work!

piever:
```
julia> using StatPlots, DataFrames

julia> df = DataFrame(a=[1,2,3,4], b=["foo","bar","foo","bar"], c = rand(4))
4×3 DataFrames.DataFrame
│ Row │ a │ b     │ c          │
├─────┼───┼───────┼────────────┤
│ 1   │ 1 │ "foo" │ 0.00892433 │
│ 2   │ 2 │ "bar" │ 0.914354   │
│ 3   │ 3 │ "foo" │ 0.033187   │
│ 4   │ 4 │ "bar" │ 0.357899   │

julia> @df df groupedbar(:a, group = :b, err = :c)
```
groupedbar

Now it can also take only the `y` argument (or `x,y,z`, thought there really isn't  an application for that yet) as well as kw arguments. The change is non-breaking as all existing recipes except `groupedbar` in StatPlots default to the previous grouping mechanism. I'll merge tomorrow if there are no more concerns.

mkborregaard:
Cool. I still think it should have sensible xticks, but that's outside the scope of this PR :-)

piever:
That's for sure, we should probably add the rule that if the `x` variable is integer valued, so should be the ticks.

PR #1064, opened 2017-08-31
Change default bar width to 0.8

mkborregaard:
The current default bar width in Plots is that adjacent bars touch, like in a histogram. This should not be the default for bars.


mkborregaard:
I'll merge if I don't get any other comments. @daschw ?

PR #1065, opened 2017-09-01
gr: suppress default output in Juno

jheinen:


mkborregaard:
This works on my system too. It doesn't address the world age issue, so I guess that's unrelated.

PR #1067, opened 2017-09-02
A couple more generalized array fixes

timholy:
Not quite sure how you want to add tests, but these get
```julia
using OffsetArrays
y = OffsetArray(rand(21), -10:10)
plot(y)
z = OffsetArray(rand(6,21), 0:5, -10:10)
heatmap(z)
```
working on PlotlyJS (along with https://github.com/JuliaIO/JSON.jl/pull/212). Tested also on PyPlot, and it throws an error, but in the 10 minutes I allowed myself to devote to this I couldn't figure out how to fix it (seems to be an error thrown inside a `try/catch` or something, there isn't a good backtrace).

Note this is against @mkborregaard's `generalized_arrays` branch (#999).

### DIFF
--- a/src/series.jl
+++ b/src/series.jl
@@ -69,13 +69,13 @@ end
 # TODO: can we avoid the copy here?  one error that crops up is that mapping functions over the same array
 #       result in that array being shared.  push!, etc will add too many items to that array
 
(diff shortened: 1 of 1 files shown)


mkborregaard:
Thanks a lot for this @timholy ! I did some changes this morning too, and with this it should be getting close.

PR #1073, opened 2017-09-05
fix 2dhistogram bins

mkborregaard:
The current estimator selected too few bins for 2-dimensional histograms - edited using a principle derived from the multivariate Scott's rule (Scott 1992 - Multivariate Density Estimation: Theory, Practice, and Visualization).

### DIFF
--- a/src/recipes.jl
+++ b/src/recipes.jl
@@ -506,7 +506,8 @@ function _auto_binning_nbins{N}(vs::NTuple{N,AbstractVector}, dim::Integer; mode
 
     n_samples = length(linearindices(first(vs)))
     # Estimator for number of samples in one row/column of bins along each axis:
-    n = max(1, n_samples^(1/N))
+    nd = n_samples^(1/(2+N))
+
 
     v = vs[dim]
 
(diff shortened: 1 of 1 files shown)


src/recipes.jl
```diff
@@ -515,15 +517,15 @@ function _auto_binning_nbins{N}(vs::NTuple{N,AbstractVector}, dim::Integer; mode
     end
 
     if mode == :sqrt  # Square-root choice
-        _cl(sqrt(n))
+        _cl(sqrt(n_samples))
     elseif mode == :sturges  # Sturges' formula
-        _cl(log2(n)) + 1
+        _cl(log2(n_samples)) + 1
```
oschulz: Let's use

    _cl(log2(n_samples) + 1)

instead.

src/recipes.jl
```diff
@@ -515,15 +517,15 @@ function _auto_binning_nbins{N}(vs::NTuple{N,AbstractVector}, dim::Integer; mode
...
-        _cl(2 * n^(1/3))
+        _cl(2 * nd)
     elseif mode == :scott  # Scott's normal reference rule
-        _cl(_span(v) / (3.5 * std(v) / n^(1/3)))
+        _cl(_span(v) / (3.5 * std(v) / nd))
     elseif mode == :fd  # Freedman–Diaconis rule
-        _cl(_span(v) / (2 * _iqr(v) / n^(1/3)))
+        _cl(_span(v) / (2 * _iqr(v) / nd))
     elseif mode == :wand
         wand_edges(v)  # this makes this function not type stable, but the type instability does not propagate
```
oschulz: I know it's not in this PR, but if we'd define

```
wand_edges(x...) = error("Load the StatPlots package in order to use :wand bins. Defaulting to :auto", once = true)
```
It would keep things type stable, and users would get an error instead of a binning they don't expect (If this is used by some automatic plotting routine, the warning may not be seen). I think an exception would be better than a warning, here.

src/recipes.jl
mkborregaard: Sure! why?

src/recipes.jl
mkborregaard: The type stability would stop the moment the user loaded StatPlots, as wand_edges returns a Vector of edges, not a number. When you wrote the code, you intended the final binning to always be done by StatPlots.histrange, so didn't allow for that possibility.

The other reason I don't do this is I think statistically minded users would have `:wand` set as the default binning in their juliarc file (I have that).

src/recipes.jl
mkborregaard: Done.

src/recipes.jl
oschulz: > as StatPlots.wand_edges returns a Vector of edges

Ah, right. Oh well, the type instability here isn't that much of a problem. Long term, we should definitely thing about something of a plug-in concept for things like binning methods, but there's no rush.

> The other reason I don't do this is I think statistically minded users would have :wand set as the default

Personally, I'm not fond of automatically overriding a user's specified intent because some package isn't loaded, as it can (in the worst case) lead to wrong results. But I guess in this case, the risk isn't great.

mkborregaard: I agree. That's also why I asked you to review this - I'm not completely certain it's sound. Will wait a little with merging.

mkborregaard: Oh, that bit's already merged, so I'll merge this one but keep it in mind.

PR #1084, opened 2017-09-09
Only one colorbar per subplot on pyplot

daschw:
This changes
```julia
using Plots; pyplot()
y1 = rand(10)
y2 = rand(10) + 2
scatter([y1 y2], marker_z = [y1 y2], markershape = [:star5 :circle], clims = (0, 5), markersize = 10)
```
from
marker_z_pyplot_old
to
marker_z_pyplot_new

### DIFF
--- a/src/backends/gr.jl
+++ b/src/backends/gr.jl
@@ -341,7 +341,7 @@ function gr_draw_markers(series::Series, x, y, clims)
     mz = normalize_zvals(series[:marker_z], clims)
     GR.setfillintstyle(GR.INTSTYLE_SOLID)
     gr_draw_markers(series, x, y, series[:markersize], mz)
-    if mz != nothing
+    if hascolorbar(series[:subplot])
         GR.setscale(0)
         gr_colorbar(series[:subplot], clims)
     end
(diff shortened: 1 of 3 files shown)


KenziTrader:
This patch doesn't only remove the double colorbar but also the second contour.

```julia
using Plots
pyplot()

x = 1:0.5:20
y = 1:0.5:10
X = repmat(x', length(y), 1)
Y = repmat(y, 1, length(x))
Z = (3X + Y .^ 2) .* abs.(sin.(X) + cos.(Y))
p1 = contour(x, y, Z, fill=true)
contour!(x, y, Z, fill=false, levels=[0., 150.])
```
download 1
download

daschw:
No, it does not. You just cannot see it because you are plotting the same series (same colors).
Try e.g.
```julia
p1 = contour(x, y, Z)
contour!(x, y, Z, levels=[0., 150.])
```
to see that another contour level is added.

daschw:
or better:
```julia
contour!(x, y, Z, levels=[0., 150.], color = :grays)
```

KenziTrader:
Yes, this works! Thanks!

PR #1086, opened 2017-09-12
Use show instead of writemime for eps

anriseth:


mkborregaard:
Hi, what's the rationale behind the change?

anriseth:
writemime was deprecated in Julia 0.4(or 0.5?), and was been replaced by
show. I guess  nobody saves anything as EPS, seeing that there have been no
complaints since 0.6 came out.

anriseth:
To add to that: writemime does not exist on 0.6.

mkborregaard:
ha ha, thanks @anriseth !

PR #1089, opened 2017-09-15
activate clims for inactive colorbar in pyplot

daschw:
fixes this issue brought up in the gitter channel:

Andrew Palugniok
@apalugniok
11:39

Hey all, I've recently did Pkg.update() and since that a call to colorbar = false has been causing clims to be ignored. For example

data = rand(10,2)
heatmap(data, clims = (0,2), colorbar = true) 
heatmap(data, clims = (0,2), colorbar = false)

And these two commands produce differently coloured plots. Haven't checked out master or anything yet, but I couldn't find this in the recent issues. On a bright note I've loved using Plots so far, hopefully I'll be able to contribute one day :smiley:

### DIFF
--- a/src/backends/pyplot.jl
+++ b/src/backends/pyplot.jl
@@ -530,7 +530,6 @@ function py_add_series(plt::Plot{PyPlotBackend}, series::Series)
                     lc
                 end
                 push!(handles, handle)
-                needs_colorbar = true
             end
 
             a = series[:arrow]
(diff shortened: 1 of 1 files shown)


daschw:
Thanks a lot @apalugniok for the reporting this!

mkborregaard:
Nice and succinct fix!

src/backends/pyplot.jl
```diff
@@ -445,7 +445,7 @@ function py_add_series(plt::Plot{PyPlotBackend}, series::Series)
 
     # handle zcolor and get c/cmap
     needs_colorbar = hascolorbar(sp)
-    extrakw = if needs_colorbar
+    extrakw = if needs_colorbar || is_2tuple(sp[:clims])
```
mkborregaard: A user might also pass in e.g. `clims = 0:1` so I'd prefer `|| get(sp, :clims, nothing) != nothing`

daschw: I agree, this would allow more input. However, `clims` defaults to `:auto` and this would have to be changed for all other backends too.

mkborregaard: good point

Issue #1095, opened 2017-09-19
Problem with :log10 scale for small numbers


fedxa:
If you plot in :log10 scale, and output falls below 1e-16, some craziness happen:
```
x=logspace(-9,-7,20);plot(x,x.^2,xscale=:log10,yscale=:log10)
```
Correct plot is just a line, as you can see:
```
x=logspace(-9,-7,20);plot(log10.(x),log10.(x.^2))
```

The bug is very manifest in PyPlot backend. The GR backends just fails to produce the plot at all.  Plotly backend works correctly.

apalugniok:
For pyplot
https://github.com/JuliaPlots/Plots.jl/blob/master/src/backends/pyplot.jl#L906
```
kw[Symbol(:linthresh,letter)] = NaNMath.max(1e-16, py_compute_axis_minval(axis))
```
seems fishy.

matplotlib documentation for linthresh:
inthreshx/linthreshy:
The range (-x, x) within which the plot is linear (to avoid having the plot go to infinity around zero).

fedxa:
Oh, cool! "max"->"min" solves it. Will make a pull request.  Let us see what is the problem with GR.

Anyway, a strange option for matplotlib...

apalugniok:
I wonder if the whole line is even necessary at this point. Since if we take the min every time the whole scale will be a logscale which should be the case if the `linthresh` keyword isn't provided, although I'm no matplotlib expert so I don't know if that's how it'd behave.

But I think `min()` is a good solution until we get to the bottom of this :D

fedxa:
That was my first idea -- just remove the line.  Somehow (no idea why!) in this case the default for linthresh turns to be something like 0.1, which is completely crazy.

apalugniok:
Oh wow matplotlib really throwing a curveball there...

fedxa:
It seems more complicated :) In plain PyPlot.jl I don't see any problems.  I am lazy to check with python :) So, there is something in the way matplotlib is called in the backend.

apalugniok:
So I tried a few things with GR. I get a load of error `GKS: Rectangle definition is invalid in routine SET_WINDOW`.
It seems like it comes back to the `GR.setwindow(xmin,xmax,ymin,ymax)` function.
I tried running this function on it's own
```
using GR
GR.setwindow(1e-9, 1e-7, 1e-18, 1e-14) # This gives the error
[...]

PR #1096, opened 2017-09-19
Fix for wrong setting of logscale for values below 1e-16

fedxa:
This fixes the issue https://github.com/JuliaPlots/Plots.jl/issues/1095 for PyPlot backend


mkborregaard:
Sweet, thanks @fedxa !

PR #1098, opened 2017-09-20
Edited Colorbar Font

AnkurD:
Pulls font family and size for Colorbar label from guidefont and
colorbar ticks from ytickfont

### DIFF
--- a/src/backends/pyplot.jl
+++ b/src/backends/pyplot.jl
@@ -999,9 +999,14 @@ function _before_layout_calcs(plt::Plot{PyPlotBackend})
             # create and store the colorbar object (handle) and the axis that it is drawn on.
             # note: the colorbar axis is positioned independently from the subplot axis
             fig = plt.o
-            cbax = fig[:add_axes]([0.8,0.1,0.03,0.8], label = string(gensym()))
-            cb = fig[:colorbar](handle; cax = cbax, kw...)
-            cb[:set_label](sp[:colorbar_title])
+			axis = sp[Symbol(:y,:axis)]
+			cbax = fig[:add_axes]([0.8,0.1,0.03,0.8], label = string(gensym()))
+			cb = fig[:colorbar](handle; cax = cbax, kw...)
+			cb[:set_label](sp[:colorbar_title],size=py_dpi_scale(plt, axis[:guidefont].pointsize),family=axis[:guidefont].family)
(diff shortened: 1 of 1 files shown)


mkborregaard:
Thanks for making this PR! I'm not certain, though, that we want the colorbar labels to have the same fonts as guidefont and ytickfont. Could you post some illustrative examples?

AnkurD:
Sure, here are a couple heatmaps I've made with this:
magpotmaxphase
magpotminphase

I understand the font objects should be different, I was just unsure how to do that so I referenced existing objects.

mkborregaard:
Hm, it does look nice. Why did you post two examples with the same font, though?

src/backends/pyplot.jl
```diff
@@ -999,9 +999,14 @@ function _before_layout_calcs(plt::Plot{PyPlotBackend})
             # create and store the colorbar object (handle) and the axis that it is drawn on.
             # note: the colorbar axis is positioned independently from the subplot axis
             fig = plt.o
-            cbax = fig[:add_axes]([0.8,0.1,0.03,0.8], label = string(gensym()))
-            cb = fig[:colorbar](handle; cax = cbax, kw...)
-            cb[:set_label](sp[:colorbar_title])
+			axis = sp[Symbol(:y,:axis)]
```
mkborregaard: I don't think `axis` is the best variable name here.

daschw:
Hi @AnkurD, Thanks for this PR! I agree that the matching colorbar font looks really nice.
If you're not feeling comfortable with PlotReferenceImages.jl I can run the tests and make a PR there for you.
However, I agree with @mkborregaard that the variable name `axis` is not the best choice in https://github.com/JuliaPlots/Plots.jl/pull/1098/files#diff-66b5ae06f1355e5f28a17f044f2941e7R1002.
How about using `yaxis` or skipping the line and using `sp[:yaxis]` below?
Also it would be great if you could align the identations of your changed lines accordingly.
After that I'd see no reason not to merge this PR 😄

AnkurD:
Hi there,
Thank you both for the feedback. I'll work on testing this locally with PlotsReferenceImages.jl before I submit the new pull request. I apologize for the delay.

AnkurD:
I have made a new pull request with the edited code and updated plots

daschw:
I'm closing this. The new implementation is in #1108

PR #1108, opened 2017-09-26
Colorbar takes font style from Y Axis

AnkurD:
Added some examples for different font types
Source Code Pro:
source_code_pro
Noto Sans:
noto_sans
DejaVu Serif
dejavu_serif

### DIFF
--- a/src/backends/pyplot.jl
+++ b/src/backends/pyplot.jl
@@ -999,13 +999,13 @@ function _before_layout_calcs(plt::Plot{PyPlotBackend})
             # create and store the colorbar object (handle) and the axis that it is drawn on.
             # note: the colorbar axis is positioned independently from the subplot axis
             fig = plt.o
-		cbax = fig[:add_axes]([0.8,0.1,0.03,0.8], label = string(gensym()))
-		cb = fig[:colorbar](handle; cax = cbax, kw...)
-		cb[:set_label](sp[:colorbar_title],size=py_dpi_scale(plt, sp[:yaxis][:guidefont].pointsize),family=sp[:yaxis][:guidefont].family)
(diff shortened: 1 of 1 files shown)


daschw:
Looks good!
Just FYI: You could have also pushed these changes to the master branch you were using in https://github.com/JuliaPlots/Plots.jl/pull/1098

Since this PR changes the appearance of some test plots, the tests will probably not pass. The next steps would be to install VisualRegressionTests.jl and clone PlotReferenceImages.jl and run the tests locally (https://juliaplots.github.io/contributing/#testing). For some images (in this case the ones with colorbars) you will be prompted a comparison of the old and the new image, and a dialog asking whether to replace the old image with the new one. If you don't see any regression in the new image you can replace the old one. Once you have replaced the images and the tests pass locally you can add the new images (git add), commit and submit a PR to PlotReferenceImages.jl. In that PR please reference this PR. This will ensure that the Plots tests will pass in future.

This probably sounds a little complicated, so if you need any support, just ask here!

AnkurD:
Ah so I did actually make the PR to PlotReferenceImages before I submitted this one, see here: https://github.com/JuliaPlots/PlotReferenceImages.jl/pull/15

daschw:
Ok, great, I did not see this.

mkborregaard:
@AnkurD would you be willing to have a look at implementing this for GR and PlotlyJS if possible? (I couldn't tell if it would be).

AnkurD:
I will do my best to implement this for GR and PlotlyJS as well. It might take a couple weeks as I am about to travel, but I'm glad to still help.

mkborregaard:
awesome!

PR #1110, opened 2017-09-26
RFC: Implement line_z and fill_z for GR

daschw:
This allows to have lines with colors corresponding to a value, multi-colored lines, "heatmaps" with custom shapes, ... e.g.
```julia
using Plots; gr()

σs = 0.5:0.1:2
normal_x = linspace(-5, 5, 100)
normal_y = [exp.(-normal_x.^2 / (2σ^2)) / (2π * σ^2) for σ in σs]

xs, ys = Plots.unzip(Plots.partialcircle(0, 2π, 50, 0.5))
shapes = [Shape(xs + i, ys + j) for i in 1:10, j in 1:10]
zvals = [i * j for i in 1:10, j in 1:10]

y = cumsum(randn(1000))

plot(
    plot(normal_x, normal_y, line_z = σs'),
    plot(shapes, fill_z = zvals),
    plot(y, line_z = y),
    plot(y, fillrange = 0, fill_z = y, fillcolor = :grays),
    label = ""
    )
```
used to look like this:
gr_line_z_old

Now it produces:
gr_line_z
If this behaviour is desirable implementations for PyPlot and Plotly will follow.

### DIFF
--- a/src/args.jl
+++ b/src/args.jl
@@ -1371,12 +1371,15 @@ function _add_defaults!(d::KW, plt::Plot, sp::Subplot, commandIndex::Int)
         getSeriesRGBColor(d[:markerstrokecolor], d[:markerstrokealpha], sp, plotIndex)
     end
 
-    # if marker_z or line_z are set, ensure we have a gradient
+    # if marker_z, fill_z or line_z are set, ensure we have a gradient
     if d[:marker_z] != nothing
         ensure_gradient!(d, :markercolor, :markeralpha)
     end
     if d[:line_z] != nothing
         ensure_gradient!(d, :linecolor, :linealpha)
+    end
+    if d[:fill_z] != nothing
+        ensure_gradient!(d, :fillcolor, :fillalpha)
     end
 
     # scatter plots don't have a line, but must have a shape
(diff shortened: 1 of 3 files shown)


mkborregaard:
Sweet! 💯

mkborregaard:
I think `fill_z` is implemented for PyPlot, though, but to do something different - to custom color individual facets of surface plots (https://github.com/JuliaPlots/Plots.jl/pull/641)

daschw:
> I think fill_z is implemented for PyPlot and Plotly, though, but to do something different - to custom color individual facets of surface plots (#641)

OK, thanks for the info! Then I should probably implement this for GR as well. I guess the behaviour of the last subplot is probably not required that often, but I can still implement the fill_z behaviour of the top right subplot.

mkborregaard:
Cool. Might it create some confusion that fill_z means two different things then? (not sure)

daschw:
Does it? I always understood the marker_z, line_z and fill_z attributes as possibilities to provide an additional value which is translated into colors (of a gradient) and can be interpreted via a colorbar. For markers it's marker_z, for lines line_z and for filled areas fill_z. Filled areas can appear in surfaces, shapes and fillranges. <- That's how I would read it ... Not sure if it's too confusing or not consistent enough (shapes also have lines/strokes, markers have lines/strokes and filled areas ...)

mkborregaard:
No, I agree. But @tbreloff 's use of the term when used to color the facets in a surfaceplot is not consistent with that logic...

mkborregaard:
But maybe it's consistent from the user's perspective, as you point out 👍

mkborregaard:
I really love this enhancement.

daschw:
Cool, I don't know how to implement Tom's facet color fill_z from PyPlot and Plotly for GR, though. I could not figure out how to color the facets of a surface in GR yet ...

mkborregaard:
I guess that's why that is not implemented yet...

PR #1120, opened 2017-10-01
showaxis attribute

daschw:
* This implements the `showaxis` attribute to control whether an individual axis is shown.
* Furthermore I added some minor axis argument preprocessing fixes (one can now e.g. do `plot(rand(10), axis = ("my label", [], :red, font(20)))` with the attributes applying to all axes) and added showaxis attribute to the magic axis arg.
* Possible values are :x, :y, :z, :xy, :xz, :yx, :yz, :zx, :zy, :xyz, :xzy, :yxz, :yzx, :zxy, :zyx, :all, :both, :on, :yes, :show, :off, :no, :hide as Symbol or String and `true` or `false`.

This allows e.g. `xshowaxis = false`, `xaxis = :hide`, `axis = :y`, ... to hide the x axis in a plot:
This has been implemented for GR, PyPlot, PlotlyJS, GLVisualize and PGFPlots and should make it easier to reproduce something like facets:
```julia
plot([plot(rand(10), axis = sa) for sa in (:y, :hide, :both, :x)]..., layout = (2, 2), label = "", link = :all)
```
* **GR**
hide-axis-gr

* **PyPlot**
hide-axis-pyplot

* **Plotly(JS)**
hide-axis-plotlyjs
It seems like `link` is not working for plotly(js). Is this a known issue?

* **GLVisualize**
hide-axis-glvisualize

* **PGFPlots**
hide-axis-pgfplots

Furthermore I implemented four basic framestyles for PGFPlots here:
framestyle_pgfplots

### DIFF
--- a/src/arg_desc.jl
+++ b/src/arg_desc.jl
@@ -110,11 +110,11 @@ const _arg_desc = KW(
 :foreground_color_text    => "Color Type or `:match` (matches `:foreground_color_subplot`). Color of tick labels.",
 :foreground_color_guide   => "Color Type or `:match` (matches `:foreground_color_subplot`). Color of axis guides (axis labels).",
 :mirror                   => "Bool.  Switch the side of the tick labels (right or top).",
-:grid                     => "Bool, Symbol, String or `nothing`. Show the grid lines? `:x`, `:y`, `:z`, `:xy`, ..., `:all`, `:none`, `:off`",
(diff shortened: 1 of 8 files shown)


mkborregaard:
More generally, `lims` argument family don't work for plotlyjs - I think this is related.

daschw:
That makes sense, thanks for the info!

daschw:
Now the axis arrows are removed in pgfplots and the `:zerolines` and `:origin` framestyles are implemented:

pgfplots_showaxis

pgfplots_framestyles

PR #1122, opened 2017-10-02
Fix barwidth to minimum(diff) not mean(diff)

mkborregaard:


mkborregaard:
This adjusts automatic bar width as 80% of the smallest, rather than the mean, distance between bars. 
```
bar([1,2,4,5], 4 .+ rand(4))
```
Before:
skaermbillede 2017-10-02 kl 08 51 37
After:
skaermbillede 2017-10-02 kl 08 52 37

Evizero:
nice

daschw:
Great!

PR #1115, opened 2017-09-29
Function default x limits: fix #1109

mkborregaard:
Tries to find defined x axis limits where the function is defined. Tries (-5,-1,0,0.01) for the minimum, and (5,1,0.99) for the maximum value. Thus, it will not work for functions that only are defined for negative numbers.
Should we add that? It would require a bit fiddling to ensure that the max value was still always larger than the min, and I don't know how common such functions are.

PR #1129, opened 2017-10-04
Apply automatic limits for vectors of functions

mkborregaard:


mkborregaard:
#1115 did not work when passing a vector of functions. This fixes that.
```julia
plot([sin, sqrt])
```
skaermbillede 2017-10-04 kl 11 38 06

mkborregaard:
There seems to be a slight issue (unrelated) with `adapted_grid`, as sqrt doesn't really asymptote towards 0.

PR #1130, opened 2017-10-04
RFC: Use existing colorbar axis if it already exists (for PyPlot)

### LINKED ISSUE #412: Colorbars inside plot areas
When I run the following code...

``` julia
using Plots
y = cumsum(randn(100,3))
scatter(y, zcolor=y)
```

...I see a colorbar for the first series outside the plot area, and two colorbars for the other two series superimposed inside the plot area:
colorbar1
I was expecting the three series to share a single colorbar placed outside the plot area. Thinking I could keep the first series's colorbar on and disable the others, I added `colorbar=[true false false]`, but the plot output did not change.

When I add `layout=2`, the first two colorbars are placed outside the two plot areas, and the third is superimposed on the second plot area:

``` julia
scatter(y, zcolor=y, layout=2)
```

colorbar2

What is the syntax for having multiple series share one colorbar?

EDIT: I'm using the PyPlot backend with Plots 0.8.1. When I switch to GR, there is one colorbar located outside the plot area.

darwindarak:
This should fix #412

### DIFF
--- a/src/backends/pyplot.jl
+++ b/src/backends/pyplot.jl
@@ -1000,7 +1000,14 @@ function _before_layout_calcs(plt::Plot{PyPlotBackend})
             # create and store the colorbar object (handle) and the axis that it is drawn on.
             # note: the colorbar axis is positioned independently from the subplot axis
             fig = plt.o
-            cbax = fig[:add_axes]([0.8,0.1,0.03,0.8], label = string(gensym()))
+
+            if haskey(fig, :cbar_ax)
+                cbax = fig[:cbar_ax]
+                cbar[:clear]()
+            else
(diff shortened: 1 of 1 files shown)


daschw:
I think #412 is already fixed on current master. At least I cannot reproduce it.

daschw:
It has been fixed in https://github.com/JuliaPlots/Plots.jl/pull/1084.
Thanks for the effort, though!
I think we can close #412, or can anyone still reproduce the issue reported there?

darwindarak:
You’re right, it works just fine on master

mkborregaard:
Thanks otherwise! Should work on the release version too - sorry for not closing the issue in time!

PR #1137, opened 2017-10-05
WIP: errorbars behind markers (fix  #1106)

### LINKED ISSUE #1106: Error bars are on top of points

daschw:
cf #1106 
errorbar

### DIFF
--- a/src/pipeline.jl
+++ b/src/pipeline.jl
@@ -100,14 +100,14 @@ function _process_userrecipe(plt::Plot, kw_list::Vector{KW}, recipedata::RecipeD
     plt.n += 1
     kw[:series_plotindex] = plt.n
 
-    if get(kw, :seriestype, :path) == :bar
-        # errorbar in front of bars
-        push!(kw_list, kw)
-        _add_errorbar_kw(kw_list, kw)
-    else
+    if get(kw, :markershape, :none) != :none || get(kw, :seriestype, :path) in (:scatter, :scatterhist)
         # errorbar behind markers and lines etc.
         _add_errorbar_kw(kw_list, kw)
         push!(kw_list, kw)
+    else
+        # errorbar in front of bars
+        push!(kw_list, kw)
+        _add_errorbar_kw(kw_list, kw)
     end
     _add_smooth_kw(kw_list, kw)
     return
--- a/src/pipeline.jl
+++ b/src/pipeline.jl
@@ -100,8 +100,8 @@ function _process_userrecipe(plt::Plot, kw_list::Vector{KW}, recipedata::RecipeD
     plt.n += 1
     kw[:series_plotindex] = plt.n
 
-    push!(kw_list, kw)
     _add_errorbar_kw(kw_list, kw)
+    push!(kw_list, kw)
     _add_smooth_kw(kw_list, kw)
     return
 end
--- a/src/pipeline.jl
+++ b/src/pipeline.jl
@@ -100,8 +100,15 @@ function _process_userrecipe(plt::Plot, kw_list::Vector{KW}, recipedata::RecipeD
     plt.n += 1
     kw[:series_plotindex] = plt.n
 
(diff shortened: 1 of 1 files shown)


mkborregaard:
Could you try a bar plot with that? With bars the errors should go in front.

mkborregaard:
I seem to remember there's some thing with bars and errors in the first place.

daschw:
Nice catch!

daschw:
```julia
bar(rand(10), yerror = abs.(randn(10)./4), ms = 6)
```
now produces
errorbar-bar
However with
```julia
plot(rand(10, 2), yerror = abs.(randn(10)./4), ms = 6, st = [:scatter :bar], layout = 2)
```
we still get
errorbar-both

src/pipeline.jl
```diff
@@ -100,8 +100,15 @@ function _process_userrecipe(plt::Plot, kw_list::Vector{KW}, recipedata::RecipeD
     plt.n += 1
     kw[:series_plotindex] = plt.n
 
-    push!(kw_list, kw)
-    _add_errorbar_kw(kw_list, kw)
+    if get(kw, :seriestype, :path) == :bar
```
mkborregaard: Possibly better to check if there are markers? Compare scatterhist and histogram, which both can have errorbars after @oschulz pr.

daschw: Can you show me a sensible use case of histograms with errorbars? `histogram(randn(100), yerror = 3)` does not seem to work.

mkborregaard: That's @oschulz ' domain, apparently they are commonly used in particle physics

daschw: OK, I was just curious, because I wanted to try out histogram and scatterhist examples with errorbars in order to find the right condition, but I could not find out how to do this with plots.

mkborregaard: If you can't do this intuitively, there's something wrong with the implementation. There's an example here https://github.com/JuliaPlots/Plots.jl/pull/787#issue-225267081.

daschw: The command for the "scatterhist" with errorbars there is just
```julia
plot(
    h2_norm_pdf,
    yerror = h2_err_norm,
    seriestype = :scatter,
    # xerror = 0.2,
    yscale = :log10,
)
```
So no histogram / scatterhist. To make this work intuitively for histograms too, we would probably need some special casing in the errorbar recipe.

mkborregaard: OK. It might still be nice to have a comment from @oschulz ?

mkborregaard:
Sorry for being a pain on this one (and I suggested testing for markers), but this doesn't work with recipes specifying scatter as a seriestype. Checking out your branch:
```
@recipe function f(::Type{Val{:test}}, x, y, z)
    seriestype := :scatter
    x,y
end
plot(1:10, randn(10), yerr = randn(10)/10, st = :test)
```
skaermbillede 2017-10-09 kl 09 40 30
Of course I might suggest solutions rather than pointing out problems :-p

oschulz:
Neat

daschw:
Changing the order was just my first try. Apparently it's not that straightforward.
Maybe we need to think a little more about this and update the errorbar recipe ...

daschw:
`plot(rand(10, 2), yerror = abs.(randn(10)./4), ms = 6, st = [:scatter :bar], layout = 2)` is not what we want to have either:
errorbar-subplots

src/pipeline.jl
oschulz: > @daschw: Can you show me a sensible use case of histograms with errorbars

Error bars are quite frequently uses with histograms to show statistical error estimates. If the error information is important, scatterhists with error bars are very popular in some fields (e.g. particle physics).

oschulz:
From what I understand, the problem is to get the error-bars in front for bar-plots, but in the back for scatter-plots?

mkborregaard:
just so - or in the back for scatter, in front for everything else

Issue #945, opened 2017-06-21
Plotly surface subplot overlap


Chemiseblanc:
When using Plotly a figure with a 3d surface subplot is rendered such that it overlaps the entire figure instead of being confined to the subplot area.

Code to reproduce:
```Julia
using Plots
plotly()

x = y = linspace(0,4*pi);
f = function(x,y)
    sin(x)*cos(y);
end

plt = plot(layout = @layout [a b])
surface!(plt[1],x,y,f)
contourf!(plt[2],x,y,f)
```

Issue #288, opened 2016-05-30
Need arg to control the viewing angle of 3D plots


gabrielgellner:
Currently when using a 3D plotting routine we are given the default view angle with no way of changing it.

For the name, matplotlib uses `view_init`, matlab uses `view`, Mathematica uses `ViewAngle`. I like the last one the most (that is `view_angle` or `viewangle`) as it feels the least likely to be confused with something out of context.

tbreloff:
I think `camera` is the most obvious representation, as `view` might mean a lot of things.  It would be super helpful if someone could compile links to API docs for setting this with each backend.

diegozea:
+1 to `camera`

gabrielgellner:
## Matplotlib
http://matplotlib.org/mpl_toolkits/mplot3d/api.html
with the specific api information:

> `view_init(elev=None, azim=None)`
>Set the elevation and azimuth of the axes.
>
>This can be used to rotate the axes programatically.
>
>‘elev’ stores the elevation angle in the z plane. ‘azim’ stores the azimuth angle in the x,y plane.
>
>if elev or azim are None (default), then the initial value is used which was specified in the Axes3D constructor.

## Plotly/JS
https://plot.ly/javascript/reference/#layout-scene-camera
with some examples in python
https://plot.ly/python/3d-camera-controls/

It looks like you make the json information of the form
```js
camera = {
"up": [x, y, z]
"center": [x, y, z]
"eye": [x, y, z]
}
```
where eye + center seems to control how the view angle works.


## GR
can't figure out were to find the information :(

## pgfplots?
Not sure if it is mapped in the julia package, but for raw pgfplots it is described
on pg.287 (section 4.11.1) under the "View Configuration". In brief:

>`/pgfplots/view={<azimuth>}{<elevation>}` (initially {25}{30})
>Changes both view angles of a 3D axis. The azimuth (first argument) is the horizontal angle which is rotated around the z axis. For a 3D plot, the z axis always points to the top. The elevation (second argument) is the vertical rotation around the (rotated) x axis. Positive elevation values indicate a view from above, negative a view from below. All values are measured in degree (but see trig format).

gabrielgellner:
`camera` does jive with the plotly api, might be nice to have the type like they do as well with the center/eye/up.

tbreloff:
Thanks @gabrielgellner!

gabrielgellner:
[...]

PR #1141, opened 2017-10-06
Adds camera attribute for 3D plots.

apalugniok:
This adds the camera attribute from #288 for GR, PyPlot, PGFPlots and Plotly/PlotlyJS.
```
x = 1:1:10
y = z = rand(10)
p = plot(x, y, [z, z], layout = 2)
plot!(p[1], camera = (20,50))
```
(Default view on the right)

GR:
gr

PyPlot:
pyplot

PGFPlots:
pgfplots

Plotly/PlotlyJS:
Unfortunately due to #945 the code doesn't work. 
Default:
newplot
Camera = (20,50):
camera

GLVIsualize:
Can't get GLVisualize to work on my mac 😟 . So I have not been able to implement camera functionality yet.

Issues:
GR: 
- `GR.setspace(zmin, zmax, rotation, tilt)` doesn't do anything when the input for rotation or tilt is negative.

Plotly:
- `camera = (0,90)` gives a much different view than `(0,89)` or `(0,91)`

All:
- Extreme viewing angles can make tick labels hard to read.

Notes:
I'm not too sure about GR the default elevation of 30 degrees looks different than for the other backends but I'm pretty sure the tilt argument in `GR.setspace()` is the elevation and not the theta angle from the z-axis.

EDIT: Added images for Plotly/PlotlyJS

### DIFF
--- a/src/arg_desc.jl
+++ b/src/arg_desc.jl
@@ -94,7 +94,7 @@ const _arg_desc = KW(
 :subplot_index            => "Integer.  Internal (not set by user).  Specifies the index of this subplot in the Plot's `plt.subplot` list.",
 :colorbar_title           => "String.  Title of colorbar.",
 :framestyle               => "Symbol.  Style of the axes frame. Choose from $(_allFramestyles)",
-:camera                   => "NTuple{2, Int(GR)/Real(Other)}. Sets the view angle (azimuthal, elevation) for 3D plots",
+:camera                   => "NTuple{2, Real}. Sets the view angle (azimuthal, elevation) for 3D plots",
 
 # axis args
 :guide     				  => "String. Axis guide (label).",
--- a/src/arg_desc.jl
+++ b/src/arg_desc.jl
@@ -94,6 +94,7 @@ const _arg_desc = KW(
 :subplot_index            => "Integer.  Internal (not set by user).  Specifies the index of this subplot in the Plot's `plt.subplot` list.",
 :colorbar_title           => "String.  Title of colorbar.",
 :framestyle               => "Symbol.  Style of the axes frame. Choose from $(_allFramestyles)",
+:camera                   => "NTuple{2, Int(GR)/Real(Other)}. Sets the view angle (azimuthal, elevation) for 3D plots",
 
 # axis args
 :guide     				  => "String. Axis guide (label).",
--- a/src/args.jl
+++ b/src/args.jl
@@ -533,7 +533,7 @@ add_aliases(:gridlinewidth, :gridwidth, :grid_linewidth, :grid_width, :gridlw, :
 add_aliases(:gridstyle, :grid_style, :gridlinestyle, :grid_linestyle, :grid_ls, :gridls)
 add_aliases(:framestyle, :frame_style, :frame, :axesstyle, :axes_style, :boxstyle, :box_style, :box, :borderstyle, :border_style, :border)
 add_aliases(:tick_direction, :tickdirection, :tick_dir, :tickdir, :tick_orientation, :tickorientation, :tick_or, :tickor)
-add_aliases(:camera, :cam)
+add_aliases(:camera, :cam, :view, :viewangle, :view_angle)
 
 # add all pluralized forms to the _keyAliases dict
 for arg in keys(_series_defaults)
--- a/src/args.jl
+++ b/src/args.jl
@@ -533,7 +533,7 @@ add_aliases(:gridlinewidth, :gridwidth, :grid_linewidth, :grid_width, :gridlw, :
 add_aliases(:gridstyle, :grid_style, :gridlinestyle, :grid_linestyle, :grid_ls, :gridls)
 add_aliases(:framestyle, :frame_style, :frame, :axesstyle, :axes_style, :boxstyle, :box_style, :box, :borderstyle, :border_style, :border)
 add_aliases(:tick_direction, :tickdirection, :tick_dir, :tickdir, :tick_orientation, :tickorientation, :tick_or, :tickor)
-add_aliases(:camera, :cam, :view, :viewangle, :view_angle)
+add_aliases(:camera, :cam, :viewangle, :view_angle)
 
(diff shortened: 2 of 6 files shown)


src/args.jl
```diff
@@ -532,6 +533,7 @@ add_aliases(:gridlinewidth, :gridwidth, :grid_linewidth, :grid_width, :gridlw, :
 add_aliases(:gridstyle, :grid_style, :gridlinestyle, :grid_linestyle, :grid_ls, :gridls)
 add_aliases(:framestyle, :frame_style, :frame, :axesstyle, :axes_style, :boxstyle, :box_style, :box, :borderstyle, :border_style, :border)
 add_aliases(:tick_direction, :tickdirection, :tick_dir, :tickdir, :tick_orientation, :tickorientation, :tick_or, :tickor)
+add_aliases(:camera, :cam)
```
daschw: How about `:view`, `:viewangle` and `:view_angle` as additional aliases?

src/arg_desc.jl
```diff
@@ -94,6 +94,7 @@ const _arg_desc = KW(
 :subplot_index            => "Integer.  Internal (not set by user).  Specifies the index of this subplot in the Plot's `plt.subplot` list.",
 :colorbar_title           => "String.  Title of colorbar.",
 :framestyle               => "Symbol.  Style of the axes frame. Choose from $(_allFramestyles)",
+:camera                   => "NTuple{2, Int(GR)/Real(Other)}. Sets the view angle (azimuthal, elevation) for 3D plots",
```
daschw: Could we allow `NTuple(2, Real)` in general and round to `Int` for GR?

apalugniok: Good idea I'll add that!

src/args.jl
apalugniok: I wanted to add `:view` (and others) but in the original thread tbreloff was against it. If we currently feel those are good I can add them 😄

src/arg_desc.jl
apalugniok: My initial thought would be to do `map(Int,map(round,sp[:camera]))` that way both a 2Tuple and an Array can work if they are given to `:camera` (`:camera` currently works with both a 2Tuple and Array). Not sure if there is a cleaner way to do this hence I'm posting a comment before I make a commit.

daschw: Sounds right for me :-)

src/args.jl
daschw: I think if they are not conflicting with other attribute names or aliases it's fin to have them as aliases.

src/args.jl
```diff
@@ -532,6 +533,7 @@ add_aliases(:gridlinewidth, :gridwidth, :grid_linewidth, :grid_width, :gridlw, :
 add_aliases(:gridstyle, :grid_style, :gridlinestyle, :grid_linestyle, :grid_ls, :gridls)
 add_aliases(:framestyle, :frame_style, :frame, :axesstyle, :axes_style, :boxstyle, :box_style, :box, :borderstyle, :border_style, :border)
 add_aliases(:tick_direction, :tickdirection, :tick_dir, :tickdir, :tick_orientation, :tickorientation, :tick_or, :tickor)
+add_aliases(:camera, :cam, :view, :viewangle, :view_angle)
```
mkborregaard: I'd prefer not to have `view` here, it's too easy to confuse with `Base.view`

src/backends/gr.jl
```diff
@@ -750,7 +751,7 @@ function gr_display(sp::Subplot{GRBackend}, w, h, viewport_canvas)
             isfinite(clims[1]) && (zmin = clims[1])
             isfinite(clims[2]) && (zmax = clims[2])
         end
-        GR.setspace(zmin, zmax, 35, 60)
+        GR.setspace(zmin, zmax, map(Int,map(round,sp[:camera]))...)
```
mkborregaard: Any particular reason for not `round.(Int, sp[:camera])...`? That's somewhat clearer in my view.

src/backends/plotly.jl
```diff
@@ -324,10 +325,21 @@ function plotly_layout(plt::Plot)
...
         # if any(is3d, seriesargs)
         if is3d(sp)
+            azim = sp[:camera][1] - 90 #convert azimuthal to match GR behaviour
+            theta = 90 - sp[:camera][2] #spherical coordinate angle from z axis
             d_out[:scene] = KW(
                 Symbol("xaxis$spidx") => plotly_axis(sp[:xaxis], sp),
                 Symbol("yaxis$spidx") => plotly_axis(sp[:yaxis], sp),
                 Symbol("zaxis$spidx") => plotly_axis(sp[:zaxis], sp),
+
+                #2.6 multiplier set camera eye such that whole plot can be seen
```
mkborregaard: Is 2.6 independent of the plots' `size`?

mkborregaard:
It's a bit weird with GR, but that maybe a backend issue - @jheinen do you have a view on this?

src/args.jl
daschw: Good point

src/backends/gr.jl
apalugniok: Mainly my lack off full familiarity with Julia... Your way is clearer to me will change.

src/backends/plotly.jl
apalugniok: Oops `size` does mess with the view a bit for plotly not sure how to fix this currently.

apalugniok:
I can try to make a PR for PlotReferenceImages once everyone is happy with the state of the code 😄 .

src/backends/plotly.jl
apalugniok: So I played around with the window `size` a bit more and the 2.6 works as long as the window isn't so small that it can no longer fit the plot.

mkborregaard:
Very nice

anriseth:
Thank you, this is very useful :D

juliohm:
Would also like to share my appreciation! Thanks for this very useful feature! :heart:

Issue #1087, opened 2017-09-14
Call-overloaded types do not plot as functions


ChrisRackauckas:
```julia
julia> using Plots

julia> plot(sin, -pi, pi, markershape = :circle)
INFO: Recompiling stale cache file C:\Users\Chris\.julia\lib\v0.6\GR.ji for module GR.

julia> struct MyFun end

julia> (::MyFun)(x) = sin(x)

julia> A = MyFun()
MyFun()

julia> plot(A, -pi, pi, markershape = :circle)
ERROR: No user recipe defined for MyFun
Stacktrace:
 [1] macro expansion at C:\Users\Chris\.julia\v0.6\Plots\src\series.jl:131 [inlined]
 [2] apply_recipe(::Dict{Symbol,Any}, ::Type{Plots.SliceIt}, ::MyFun, ::Float64, ::Irrational{:π}) at C:\Users\Chris\.julia\v0.6\RecipesBase\src\RecipesBase.jl:265
 [3] _process_userrecipes(::Plots.Plot{Plots.GRBackend}, ::Dict{Symbol,Any}, ::Tuple{MyFun,Float64,Irrational{:π}}) at C:\Users\Chris\.julia\v0.6\Plots\src\pipeline.jl:81
 [4] _plot!(::Plots.Plot{Plots.GRBackend}, ::Dict{Symbol,Any}, ::Tuple{MyFun,Float64,Irrational{:π}}) at C:\Users\Chris\.julia\v0.6\Plots\src\plot.jl:175
 [5] #plot#143(::Array{Any,1}, ::Function, ::MyFun, ::Vararg{Any,N} where N) at C:\Users\Chris\.julia\v0.6\Plots\src\plot.jl:56
 [6] (::RecipesBase.#kw##plot)(::Array{Any,1}, ::RecipesBase.#plot, ::MyFun, ::Float64, ::Vararg{Any,N} where N) at .\<missing>:0

julia> struct MyFun2 <: Function end

julia> (::MyFun2)(x) = sin(x)

julia> A = MyFun2()
(::MyFun2) (generic function with 1 method)

julia> plot(A, -pi, pi, markershape = :circle)
```

One solution is to change from `Function` to `Any` and just check for the single-parameter dispatch.

ChrisRackauckas:
Yeah that could work.

mkborregaard:
Feel like fixing it?

ChrisRackauckas:
[...]

PR #1148, opened 2017-10-08
WIP: Callable type support

ChrisRackauckas:
Attempts to solve #1087. Current issue is:

```julia
BoundsError: attempt to access 1-element Array{Any,1} at index [2]
macro expansion at series.jl:131 [inlined]
apply_recipe(::Dict{Symbol,Any}, ::Type{Plots.SliceIt}, ::MyFun, ::Float64, ::Irrational{:π}) at RecipesBase.jl:287
_process_userrecipes(::Plots.Plot{Plots.PlotlyBackend}, ::Dict{Symbol,Any}, ::Tuple{MyFun,Float64,Irrational{:π}}) at pipeline.jl:81
_plot!(::Plots.Plot{Plots.PlotlyBackend}, ::Dict{Symbol,Any}, ::Tuple{MyFun,Float64,Irrational{:π}}) at plot.jl:175
#plot#143(::Array{Any,1}, ::Function, ::MyFun, ::Vararg{Any,N} where N) at plot.jl:56
(::RecipesBase.#kw##plot)(::Array{Any,1}, ::RecipesBase.#plot, ::MyFun, ::Float64, ::Vararg{Any,N} where N) at <missing>:0
include_string(::String, ::String) at loading.jl:515
include_string(::String, ::String, ::Int64) at eval.jl:30
include_string(::Module, ::String, ::String, ::Int64, ::Vararg{Int64,N} where N) at eval.jl:34
(::Atom.##49#53{String,Int64,String})() at eval.jl:50
withpath(::Atom.##49#53{String,Int64,String}, ::String) at utils.jl:30
withpath(::Function, ::String) at eval.jl:38
macro expansion at eval.jl:49 [inlined]
(::Atom.##48#52{Dict{String,Any}})() at task.jl:80
```

I am not sure where this is supposed to be pointing to.

Other issue is, I don't have the other arguments for the test, so `0` may be out of bounds or the wrong type. Is there any way to know the type of the other arguments in here?

### DIFF
--- a/test/runtests.jl
+++ b/test/runtests.jl
@@ -133,6 +133,13 @@ end
     histogram([1, 0, 0, 0, 0, 0])
 end
 
+@testset "Callable Functions" begin
+    struct MyFun end
+    (::MyFun)(x) = sin(x)
+    A = MyFun()
+    plot(A, -pi, pi, markershape = :circle)
+end
+
 # tests for preprocessing recipes
 
 # @testset "recipes" begin
--- a/src/series.jl
+++ b/src/series.jl
@@ -11,7 +11,21 @@ const FuncOrFuncs{F} = Union{F, Vector{F}, Matrix{F}}
 all3D(d::KW) = trueOrAllTrue(st -> st in (:contour, :contourf, :heatmap, :surface, :wireframe, :contour3d, :image), get(d, :seriestype, :none))
 
 # unknown
-convertToAnyVector(x, d::KW) = error("No user recipe defined for $(typeof(x))")
+# in this case, check if it's actually a callable type and use as a function
+# otherwise it's unsupported and throw an error.
+function convertToAnyVector(x, d::KW)
+    try
+        x(0.0)
+    catch
+        try
+            x(0.0,0.0)
+        catch
+            error("No user recipe defined for $(typeof(x))")
+        end
+    end
+    @show "here!"
+    Any[(v...)->x(v...)]
+end
 
 # missing
 convertToAnyVector(v::Void, d::KW) = Any[nothing], nothing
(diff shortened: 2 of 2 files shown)


test/runtests.jl
```diff
@@ -133,6 +133,13 @@ end
     histogram([1, 0, 0, 0, 0, 0])
 end
 
+@testset "Callable Functions" begin
+    struct MyFun end
```
mkborregaard: Travis complains about this - " type definition not allowed inside a local scope"

mkborregaard:
You should be able to get the other arguments by `d[:x]`, `d[:y]` and `d[:z]` (I think the object itself is `d[:z]` at this point). You could use the `tryrange` framework to test for a number of values to get the right numbers? I gotta admit I'm a bit uncertain about the exact state of `d[:x]`, `d[:y]` and `d[:z]` at this point.


By the way, I think it's confusing that we still use `d` inside Plots, but `plotattributes` everywhere else. @jw3126 what was your intention here?

jw3126:
What exactly do you mean by `d` is still used in plots? That certain function arguments like
```Julia
function convertToAnyVector(x, d::KW)
    ...
end
```
are called `d` instead of `plotattributes`? Or are there still recipes left with `d`?

mkborregaard:
Yes, that's what I meant.

mkborregaard:
This may not be a big deal, just flagging it for discussion.

mkborregaard:
@ChrisRackauckas what's the status on this PR? merge?

ChrisRackauckas:
I have no idea. I think it was working? It would need a rebase.

mkborregaard:
Feel like rebasing?

ChrisRackauckas:
Nah, this is having ambiguity issues.

mkborregaard:
That's too bad because this would be nice to support 😞

ChrisRackauckas:
Looks like the ambiguities were from too much revising, but the larger issue is the first one. I think it could index to grab the first of `x` to get around that error? Maybe

mkborregaard:
Yeah OK I see that. I wonder if this isn't because we intercept too late in the pipeline. If that's true it's not trivial to make it work.

PR #1153, opened 2017-10-09
Fix marker_z for 3D plots on GR (test example 24) (fix #1152)

### LINKED ISSUE #1152: Example 24 markers stuck in xz-plane for GR
The example code from the docs gives this for GR.
```
n = 100
ts = linspace(0,8π,n)
x = ts .* map(cos,ts)
y = (0.1ts) .* map(sin,ts)
z = 1:n
plot(x,y,z,zcolor=reverse(z),m=(10,0.8,:blues,stroke(0)),leg=false,cbar=true,w=5)
plot!(zeros(n),zeros(n),1:n,w=10)
```

ref24gr

The above was made on the current Plots.jl master. I have tried the current release of Plots.jl and there is no change, the same occurs for both the master and release of GR.jl.

daschw:
gr_24

Thanks a lot @apalugniok for the awesome detective work!!!
I will run the tests locally before I merge this time, but I can only do this tonight.

### DIFF
--- a/src/backends/gr.jl
+++ b/src/backends/gr.jl
@@ -342,10 +342,6 @@ function gr_draw_markers(series::Series, x, y, clims)
     mz = normalize_zvals(series[:marker_z], clims)
     GR.setfillintstyle(GR.INTSTYLE_SOLID)
     gr_draw_markers(series, x, y, series[:markersize], mz)
-    # if hascolorbar(series[:subplot])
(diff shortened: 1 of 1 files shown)


apalugniok:
Where was the issue in the end?

daschw:
Before the markers where drawn a colorbar was already drawn in L1055, which messed up the 3d axes.

Issue #40, opened 2015-10-04
Feature Request: spy() function


joshday:
Hi Tom.  Matlab has a `spy()` function for viewing the sparsity pattern of a matrix.  I like the way Gadfly implements `spy()`, which is to make a heatmap from a matrix (https://github.com/JuliaLang/julia/issues/2121).  I didn't see a way to do this through the Plots.jl api.

tbreloff:
I wrapped the call to heatmap. Available on master for Gadfly, Immerse, PyPlot, Qwt:

<img width="656" alt="screen shot 2015-10-04 at 9 40 22 pm" src="https://cloud.githubusercontent.com/assets/933338/10272135/b45a756c-6ae0-11e5-9a9d-09c0226d48a3.png">

ptoche:
Any ideas? Thanks.

    workspace()
    SM = eye(5)
    SM[end, end] = -1
    using Gadfly
    spy(SM)

    WARNING: both Gadfly and Base export "cross"; uses of it in module Main must be qualified

Also tried the following, but no plot was produced.

    p = spy(SM)
    display(p)

Julia + Plots updated as of a few minutes ago.

mkborregaard:
Hi @ptoche - if you find an issue, it is better to open a new issue than comment on an old closed one, especially if the problem is different. We don't keep track of closed issues, so problems will not be solved. If you just have a question, the gitter channel is a good place to find us.

It does indeed look like the spy recipe is broken, and you should probably open a new issue, preferably with a screenshot detailing your problem.

However, your code would never work. You're calling Gadfly, but gadfly is no longer  a Plots backend, and also you cannot call a Plots backend with `using`. For Plots, the code would look like this:
```julia
using Plots; gr()
SM = eye(5)
SM[end, end] = -1
spy(SM)
```

mkborregaard:
`spy` isn't implemented as a heatmap anymore, BTW. You can just do a heatmap if that's what you'd like.

ptoche:
@mkborregaard, thanks for your reply. I wasn't sure it was an 'issue' as such or my misunderstanding of the plotting syntax. Apparently both.  

I had some Matlab code that used the ``spy()`` function, so I googled for ``Julia+spy`` and got here pretty quickly, which is why I left a message here. Other trails took me to [](https://stackoverflow.com/questions/33732330/using-spy-in-julia), but that didn't help much.

alanedelman:
[...]

PR #1155, opened 2017-10-09
RFC: New spy implementation

mkborregaard:


mkborregaard:
@alanedelman (https://github.com/JuliaPlots/Plots.jl/issues/40) and @dpsanders both asked for a heatmap-based version of spy, close to what's in Gadfly. This actually represents the pre-0.9.4 Plots implementation if `spy`. 
I prefer this too, so I've made a quick PR (it mab do too much unnecessary copying at the moment). This reimplements `spy` as a y-flipped heatmap, with no gridlines, transparent (white) areas for zero values (and undefined values of sparse matrices), aspect_ratio 1.

```julia
a = sprand(100, 200, 0.01)
spy(a)
```
skaermbillede 2017-10-09 kl 21 17 10

(note @daschw that the x-labels dont follow the ticks up when the y axis is flipped ).

Comments welcome.

daschw:
I like it!
I will look into the xtick labels (this is GR, right?)

mkborregaard:
yes.
The reason I make a new copy of the matrix is that I assume that matrix being passed in is often sparse - so if I want to replace all zero-elements with NaN I think it is better to just initialize a new dense matrix. Of course we could specialcase sparse matrices instead.

daschw:
Do we want both, tick marks and tick labels, up or down?

daschw:
All other backends (PyPlot, PlotlyJS, PGFPlots and GLVisualize) have them down ... So I guess this answers my question.

daschw:
fixed in #1163

mkborregaard:
Nice. I think I'll merge this unless there are any complaints about inefficient code?

daschw:
Fine with me. If anyone wants a more efficient implementation they are welcome to submit a PR 😄

PR #1164, opened 2017-10-10
fix #750 : save svg plotlyjs

### LINKED ISSUE #750: make savefig(..., "filename.svg") save self-contained SVG for plotlyjs

piever:
Superseeds #751 in a less invasive way: now PlotlyJS plots are diplayed exactly as they used to be in the Jupyter notebook, but `savefig("test.svg")` saves in svg rather than html format (as instead it does on master).

### DIFF
--- a/src/backends/plotlyjs.jl
+++ b/src/backends/plotlyjs.jl
@@ -85,7 +85,8 @@ end
 
 # ----------------------------------------------------------------
 
-function _show(io::IO, ::MIME"image/svg+xml", plt::Plot{PlotlyJSBackend})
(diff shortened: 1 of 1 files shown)


daschw:
Nice!

PR #1176, opened 2017-10-14
Don't copy input arguments - fix #1175

### LINKED ISSUE #1175: Why does Plots `copy` input arguments
Came up on Slack: https://julialang.slack.com/archives/C690QRAA3/p1507911096000451
Ref here https://github.com/JuliaPlots/Plots.jl/blob/master/src/series.jl#L72-L88

This may be wasteful, and can cause problems when passing in types that don't have `copy` defined. Solutions are to avoid the `copy` or replace with `deepcopy`.

mkborregaard:


src/series.jl
```diff
@@ -371,7 +371,7 @@ end
 end
 
 # try some intervals over which the function may be defined
-function tryrange(F::AbstractArray, vec)
```
mkborregaard: An artifact of my text editor it appears

Issue #1162, opened 2017-10-10
GR test_example 26


daschw:
When I ran the tests today I got:
![gr_ref26](https://user-images.githubusercontent.com/16589944/31405368-2116bb26-adff-11e7-96e1-084c032dbff6.png)
I did not save the new image yet, because it kind of looks like a regression.
I'm not sure it is an issue, though. This is probably caused by https://github.com/JuliaPlots/Plots.jl/pull/1122 and expected.
Any opinions?

mkborregaard:
Either set linewidth = 0 or manually increase the bar width. I'd suggest the first option, as the reference image looks terrible :-)

apalugniok:
I just ran the tests and ref 4 and ref 32 fail.

ref4:
<img width="997" alt="ref4" src="https://user-images.githubusercontent.com/22132297/31453498-49b1c26e-aeaa-11e7-8432-80154f49ba9f.png">

ref32:
<img width="1001" alt="ref32" src="https://user-images.githubusercontent.com/22132297/31453521-55a85074-aeaa-11e7-8bf0-36b2a4f902ff.png">

daschw:
I don't know what's wrong with the old ref 4. I guess that's something I missed last time. IMHO the left one is an improvement and you can replace it.
ref32 is caused by Michael's spy implementation. I think we could change the colors for matrices with unique non-zero values to only show black rectangles. I would not replace the image yet (it does not cause the travis tests to fail) but open an issue for this particular spy behavior.

PR #1183, opened 2017-10-16
spy recipe: change default color for matrices with unique values

daschw:
cf https://github.com/JuliaPlots/Plots.jl/issues/1162#issuecomment-335869576
test_gr_spy

### DIFF
--- a/src/recipes.jl
+++ b/src/recipes.jl
@@ -937,6 +937,8 @@ end
     mat = g.args[1]
     if length(unique(mat[mat .!= 0])) < 2
         legend --> nothing
+        # revert the default gradient to have a darker (almost black) color in the default theme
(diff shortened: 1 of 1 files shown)


src/recipes.jl
```diff
@@ -937,6 +937,8 @@ end
     mat = g.args[1]
     if length(unique(mat[mat .!= 0])) < 2
         legend --> nothing
+        # revert the default gradient to have a darker (almost black) color in the default theme
+        seriescolor --> cgrad([cgrad()[0], cgrad()[1]])
```
mkborregaard: why not just make it black?

daschw: I thought this might look strange with `theme(:dark)`. But we could default to the axes color if you prefer a real black in the default case.

mkborregaard: No, you're right 👍

PR #1185, opened 2017-10-16
Use ffmpeg & palette to create gifs, closes #1050

### LINKED ISSUE #1050: Quality of gifs generated by ffmpeg can be improved by using a palette
Currently FFmpeg produces really low quality animations as the colors are nearly always off. This is unfortunate especially as ImageMagicks convert does not cope that good with longer animations whereas FFmpeg does. Using a palette resolves this issue and the resulting gifs are pretty much of the same quality.

Docs for creating and using a palette can be found here:
https://ffmpeg.org/ffmpeg-filters.html#palettegen-1
https://ffmpeg.org/ffmpeg-filters.html#paletteuse

H-M-H:
First let ffmpeg generate a colorpalette so it uses the limited amount
of colors available properly and only then create the actual gif using
the generated palette. This replaces convert completely as it yields far
worse results for long gifs and is considerably slower.


mkborregaard:
Yeah I'm good with this - thanks a lot @H-M-H !

PR #1187, opened 2017-10-18
Animations: improve movie compatibility when using ffmpeg

anowacki:
Add the `-pix_fmt yuv420p` option when using ffmpeg to save
animations with the `mov` or `mp4` commands.  This improves
comptability of the movie files since many players do not
support the ffmpeg defaults.

See the following for more information:
    https://trac.ffmpeg.org/wiki/Encode/H.264 ('Compatibility')
    https://apple.stackexchange.com/questions/166553/why-wont-video-from-ffmpeg-show-in-quicktime-imovie-or-quick-preview

### DIFF
--- a/src/animation.jl
+++ b/src/animation.jl
@@ -61,18 +61,23 @@ end
 file_extension(fn) = Base.Filesystem.splitext(fn)[2][2:end]
 
 gif(anim::Animation, fn = giffn(); kw...) = buildanimation(anim.dir, fn; kw...)
-mov(anim::Animation, fn = movfn(); kw...) = buildanimation(anim.dir, fn; kw...)
-mp4(anim::Animation, fn = mp4fn(); kw...) = buildanimation(anim.dir, fn; kw...)
+mov(anim::Animation, fn = movfn(); kw...) = buildanimation(anim.dir, fn, false; kw...)
+mp4(anim::Animation, fn = mp4fn(); kw...) = buildanimation(anim.dir, fn, false; kw...)
 
 
-function buildanimation(animdir::AbstractString, fn::AbstractString;
+function buildanimation(animdir::AbstractString, fn::AbstractString,
+                        is_animated_gif::Bool=true;
                         fps::Integer = 20, loop::Integer = 0)
     fn = abspath(fn)
 
-    # generate a colorpalette first so ffmpeg does not have to guess it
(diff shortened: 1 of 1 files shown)


daschw:
Thanks a lot!
This probably conflicts with #1185.
Does it make sense to reopen this after/if we merge #1185?

mkborregaard:
We can just hold and rebase.

mkborregaard:
@anowacki would you do a rebase on master? (not a merge-master-into-PR :-)  ). I don't know anything about ffmpeg - who could review this - @H-M-H?

anowacki:
@mkborregaard: It turns out that the `-pix_fmt` option isn't helpful when creating animated GIFs, so I've introduced an if-statement to change behaviour depending on whether one has called `mov`, `mp4` or `gif`.

I'm not sure if it doesn't make sense for `mov` and `mp4` to call a different animation building function to `gif`.  I currently just check the filename, which doesn't seem terribly robust.

H-M-H:
Using `-pix_fmt yuv420p` is a good idea as far as compatibility is concerned at least for mp4 and mov but as @anowacki already noticed it is a bad idea for gifs.

Also this does not use a colorpalette for generating mov/mp4 resulting in slightly larger files (about 10KiB more than in combination with a palette for the example I used in #1050 ) which is probably even the better way as generating a colorpalette takes its time.

mkborregaard:
I don't know a lot about ffmpeg, so I'll merge this when you guys agree it is ready :-)

anowacki:
@H-M-H I think we agree that a different code path is needed for making GIF files versus movies, but please let me know if you think otherwise.  I think from your comment you were suggesting that using a colour palette for movies isn't worth the small file size reduction, which I agree with.

@mkborregaard I have updated the PR to add a third argument to `buildanimation` which specifies whether we're building a GIF or otherwise.  This is safer than testing for an extension of `.mov` or `.mp4`, since we accept the file name as an argument to `buildanimation`, and one might want to use `.m4v`, for instance.  (This doesn't extend to the `Base.show(::IO, ::MIME"text/html", ::AnimatedGif)` method, which does rely on the filename to determine how to construct the HTML for e.g. Jupyter notebooks.  Fixing that would be a different PR.)

Also rebased onto `master`.

mkborregaard:
Merging - thanks a lot @anowacki !

PR #1189, opened 2017-10-18
use fg_color for matrices with unique non-zeros in spy recipe

daschw:
cf https://github.com/JuliaPlots/Plots.jl/pull/1183#issuecomment-337289179


mkborregaard:
You're a star

PR #1190, opened 2017-10-18
Minor test_example changes (close #1162)

### LINKED ISSUE #1162: GR test_example 26

daschw:
Groups and Subplots (`test_examples(:gr, 26)`)
gr_26

Spy (`test_examples(:gr, 32)`)
gr_32

Framestyle (`test_examples(:gr, 34)`)
gr_34

### DIFF
--- a/src/examples.jl
+++ b/src/examples.jl
@@ -346,10 +346,10 @@ PlotExample("Magic grid argument",
 PlotExample("Framestyle",
     "The style of the frame/axes of a (sub)plot can be changed with the `framestyle` attribute. The default framestyle is `:axes`.",
     [:(begin
-    scatter(fill(randn(10), 7), fill(randn(10), 7),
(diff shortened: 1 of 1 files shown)


daschw:
Actually the tests should really fail here. (cf ref34 in https://github.com/JuliaPlots/PlotReferenceImages.jl/pull/22/files#diff-1a37fb28ef3065510a15ff1b337edb81) Should we maybe increase the accuracy for testing?

mkborregaard:
Yeah, something's fishy about the tests lately - since we disabled pyplot perhaps? Should we try enabling that again, if that not works let's try increasing the accuracy. Let's make those tests fail! :-D

PR #1199, opened 2017-10-20
Make revise possible

mkborregaard:
Explicitly register the `eval`-included backend files with Revise


timholy:
Cool. Once https://github.com/timholy/Revise.jl/pull/49 merges, you won't need this on 0.7. But this seems good for supporting 0.6.

mkborregaard:
Great, I wasn't aware of that PR!

PR #1208, opened 2017-10-26
Polar plots consistency

### LINKED ISSUE #1103: Inconsistent angle unit in polar plot of pgfplots backend
In polar plots as in `plot(theta,r,proj=:polar)` pgfplots expects values of  `theta` in degrees while the other backends expect it in radians.

Example:
```julia
using Plots
pyplot()
th = linspace(0,2pi,128)
r = 1 .+ sin.(th)
plot(th, r, proj=:polar)
```
In other to reproduce the same plot using `pgfplots()` one must use `plot(linspace(0,360,128), r, proj=:polar)` .

apalugniok:
Fixes #1103 

```
using Plots
plot(linspace(0, 2pi, 360), linspace(0, 3, 360), proj = :polar)
```

GR: 
polargr

PyPlot:
polarpyplot

PGFPlots:
polarpgfplots

Plotly:
polarplotly

InspectDR: Haven't done much testing yet

Remarks:

The `radialaxis` and `angularaxis` attributes in Plotly don't have a `tickvals` keyword (at least it's not in the PlotlyJS reference) so I don't think you can manually set the polar ticks for this backend. I've read somewhere that Plotly polar plots are badly underdeveloped.

The tick labels are in degrees even though the input is in radians. Getting the fractions of pi to display properly changes between backends and in my experience whenever latex is used there are issues with it not listening to `tickfont`. GR already displays the labels in degrees so I have just settled on that.

Currently a polar plot has it's angular limits locked at 0 to 2pi. I am personally not sure whether it makes sense to lift that restriction and allow limits to be set manually? In PGFPlots setting limits manually does what I expect and shows only a sector of the circle. In PyPlot the limits are mapped over the full circle. Also the default ticks calculated by `Plots` look bad on an angular axis.

Issues:
- `yflip` doesn't work
- PGFPlots has issues with negative data `linspace(0, -3, 360)` breaks but linspace(-1, -3, 360) works.
- Logarithmic scales don't work.

### DIFF
--- a/src/axes.jl
+++ b/src/axes.jl
@@ -239,8 +239,13 @@ function get_ticks(axis::Axis)
         # discrete ticks...
         axis[:continuous_values], dvals
     elseif ticks == :auto
-        # compute optimal ticks and labels
-        optimal_ticks_and_labels(axis)
+        if ispolar(axis.sps[1]) && axis[:letter] == :x
+            #force theta axis to be full circle
+            (collect(0:pi/4:7pi/4), string.(0:45:315))
+        else
+            # compute optimal ticks and labels
+            optimal_ticks_and_labels(axis)
+        end
     elseif typeof(ticks) <: Union{AVec, Int}
         # override ticks, but get the labels
         optimal_ticks_and_labels(axis, ticks)
@@ -427,7 +432,16 @@ function axis_limits(axis::Axis, should_widen::Bool = default_should_widen(axis)
     if !isfinite(amin) && !isfinite(amax)
         amin, amax = 0.0, 1.0
     end
-    if should_widen
+    if ispolar(axis.sps[1])
+        if axis[:letter] == :x
+            amin, amax = 0, 2pi
+        elseif lims == :auto
+            #widen max radius so ticks dont overlap with theta axis
+            amin, 1.1*amax
+        else
+            amin, amax
+        end
+    elseif should_widen
         widen(amin, amax)
     else
         amin, amax
--- a/src/axes.jl
+++ b/src/axes.jl
@@ -437,7 +437,7 @@ function axis_limits(axis::Axis, should_widen::Bool = default_should_widen(axis)
             amin, amax = 0, 2pi
         elseif lims == :auto
             #widen max radius so ticks dont overlap with theta axis
-            amin, 1.1*amax
+            amin, amax + 0.1 * abs(amax - amin)
         else
             amin, amax
         end
--- a/src/backends/gr.jl
+++ b/src/backends/gr.jl
@@ -214,8 +214,7 @@ function gr_polaraxes(rmin::Real, rmax::Real, sp::Subplot)
     a = α .+ 90
     sinf = sind.(a)
     cosf = cosd.(a)
-    tick = 0.5 * GR.tick(rmin, rmax)
-    n = round(Int, (rmax - rmin) / tick + 0.5)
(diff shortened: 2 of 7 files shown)


apalugniok:
That moment when you find that `Axis` does indeed has information about the subplot it belongs to 😆  Oh well, I think changes are in order.

mkborregaard:
Do you need help to advance further on this PR?

apalugniok:
I think I am actually alright right now, just don't have that much time to indulge in open source. Thanks for asking though 😄 . I can push up what I have right now just so this can be sort of updated.

apalugniok:
Something isn't quite right with the GR ticks.

apalugniok:
With regards to the issues.

1. I can't find any way for any of the backends to flip the radial axis. GR does it through `setscale()` so I don't think it can be fixed in Plots. I can find any documented function that would flip the axis, the same goes for Plotly. 

2. The issue is that when the limits are smaller than the span of the data instead of ignoring data points outside these limits GR still tries to plot them. Eg. `plot(linspace(0, 2pi, 360), linspace(-1, 0, 360), proj = :polar, ylims = (0,1))` here the data point `(0, -1)` is plotted in the position (180, 1). This can lead to really odd looking plots when you specify your own axis limits.

3. PGFPlots really doesn't like negative radial values it always interprets the origin as `0` and gets horribly confused when no positive points are passed in.

4. Still to investigate fot GR. PyPlot returns an error. Plotly doesn't have an attribute for a log scale on a `radialaxis`. PGFPlots also gives an error wonder if it's related to the origin always being 0.
Error messages

PyPlot:
```Error showing value of type Plots.Plot{Plots.PyPlotBackend}:
ERROR: PyError (ccall(@pysym(:PyObject_Call), PyPtr, (PyPtr, PyPtr, PyPtr), o, arg, C_NULL)) <type 'exceptions.RuntimeError'>
RuntimeError(u'Cannot get window extent w/o renderer',)
  File "/Users/pucefal/.julia/v0.6/Conda/deps/usr/lib/python2.7/site-packages/matplotlib/text.py", line 965, in get_window_extent
    raise RuntimeError('Cannot get window extent w/o renderer')
```

PGFPlots:
```! Package PGF Math Error: Sorry, an internal routine of the floating point unit
 got an ill-formatted floating point number `1.0000000000'. The unreadable part
 was near '1.0000000000'..

See the PGF Math package documentation for explanation.
Type  H <return>  for immediate help.
 ...                                              
                                                  
l.388 };
      
Error saving as SVG
Error showing value of type Plots.Plot{Plots.PGFPlotsBackend}:
ERROR: LaTeX error
```

mkborregaard:
You really put your hand in a hornet's nest here :-) We may not have to solve all the issues before merging IMHO

apalugniok:
Which ones do you think we should definitely fix before a merge?

mkborregaard:
Only number 2.

apalugniok:
Cool, I have a fix ready will do a few tests then push.

apalugniok:
I'm happy with the current state as well. Might be good add a note somewhere in the docs about negative radii and PGFPlots.

mkborregaard:
I don't think we need to describe present limitations in the docs, that system is too inflexible when things get fixed. Instead we'll track these things as issues. 
Could you open one or more new issues with the things you've observed here and your research? People aren't going to be tracking this PR now it's merged.

apalugniok:
Okay! I'll open the issues tomorrow 😄

PR #1213, opened 2017-10-30
Make `fill = true` fill to zero

daschw:
`fillrange = true` still fills to 1, not sure whether to change this


mkborregaard:
I see you decided to go ahead with this and to me it's also more intuitive in most cases. But it will fail when `yscale` is `:log`.

daschw:
Oh, there's always a catch ... Thanks a lot for pointing this out!
Should we revert this?
On the other hand, what would be a sensible fillrange default for log scales? Maybe the minimum value of the y data?

mkborregaard:
Let's just keep it :-)

PR #1220, opened 2017-11-02
gr: use correct aspect ratio (fixes #1219)

### LINKED ISSUE #1219: aspect_ratio depends on the existence of colorbar in GR backend
When I run:
```
heatmap(randn(10,10), aspect_ratio=1)
```
under the GR backend, the aspect ratio is not correctly set. On the other hand, if I run:
```
heatmap(randn(10,10), aspect_ratio=1, colorbar=false)
```
then it displays an image with the correct aspect ratio. Obviously, the aspect ratio specification of the current GR backend depends on the existence of the colorbar. Note that this problem does not occur under the pyplot and plotlyjs backends. My guess is that either ```function gr_colorbar``` or ```function gr_display``` in Plots.jl/src/backends/gr.jl need to be modified, but I may be wrong. Currently, I am using the master of GR.jl v0.24.0 and the master of Plots.jl v0.13.1.
Thanks for your help!
BVPs

jheinen:


daschw:
Great

BoundaryValueProblems:
@jheinen Thanks a lot!!

Issue #423, opened 2016-07-28
Ideas to enhance animations


JobLeonard:
I was just going to propose supporting a visually stable ditherer, but then I looked up some background info to get that started and realised that this might be worth a discussion of its own. Not exactly a high-priority issue, but still a fun enhancement to discuss I think.

The starting point was that I noticed  that the (beautiful) example gifs have a lot of areas that should not change from frame to frame, yet still have different pattern noise. For example, zoom in on the grey grid in the background of the Lorenz Attractor gif:

![lorentz zoom](https://user-images.githubusercontent.com/259840/32442071-58812712-c2fa-11e7-881c-91b90a789710.gif)

This noise is due to the error diffusion methods used. However, there are stable GIF encoders that do not suffer from this. Since plots usually have few colours and change very little from frame to frame, I would expect this kind of optimisation to greatly reduce size _and_ improve visual quality at the same time. Since you're using [ImageMagick with FFMPEG as a fallback](https://github.com/tbreloff/Plots.jl/blob/91aa1d718a5a2616806ff0fdb8859e0f92daa74e/src/animation.jl#[L27-L53]) I've looked at what options they provide.

For IM  there is [this gist exploring ImageMagick's GIF encoding options](https://gist.github.com/CoolOppo/f54d86abe726bb7827e6) that has some suggestions and also warns against the default settings:

> Also if you study the resulting animation further you will find that of the 107 frames in the image, 106 frames had their own own separate local color table added. That is each and every frame in the GIF animation required there own color index table. That is while each frame has less that 256 colors (due to the GIF format limitations), the whole animation is using a total of 7525 colors.
> 
> Unfortunately the GIF format does not compress color tables, so all those extra color tables could be using up to:   256 colors \* 3 byte per color \* 106 frames;   or 81,408 bytes of file space. Not a lot for a 1Gbyte video but still an appreciable amount of space, especially as we optimize the video further.
> 
> Added to this is that the animation will not GIF frame optimize very well. Not only because the background is moving (due to the camera panning upward), but also because IM used a Error Correction Dither (Hilbert Curve Dither), which produces a pseudo-random pattern of colors that is different from frame to frame. A later example will make this 'dither noise' much more visible.

Perhaps I missed it, but glancing at the code it appears that you use the default settings so the output GIF likely suffers from the same issues.

[...]

PR #1227, opened 2017-11-06
Add stats_mode=diff and diff_mode=rectangle

JobLeonard:
`stats_mode=diff` makes `palettegen` focus on the changes between frames instead of the whole frame.
Given that the whole frame (usually) consists mainly of a static white background, this is likely to improve image quality

`diff_mode=rectangle` makes the gif encoding crop to the bounding box of the changing parts. This again is likely to improve image quality because it avoids error diffusion to parts of the image that do not change. It can also greatly reduce file-size for GIFs where only a small section of the image changes between frames.

See this comment for more details: 

https://github.com/JuliaPlots/Plots.jl/issues/423#issuecomment-342173040

### DIFF
--- a/src/animation.jl
+++ b/src/animation.jl
@@ -72,9 +72,11 @@ function buildanimation(animdir::AbstractString, fn::AbstractString,
 
     if is_animated_gif
         # generate a colorpalette first so ffmpeg does not have to guess it
-        run(`ffmpeg -v 0 -i $(animdir)/%06d.png -vf palettegen -y palette.png`)
+        # use stats_mode=diff to emphasize changes and avoid bias for static white background
+        run(`ffmpeg -v 0 -i $(animdir)/%06d.png -vf palettegen=stats_mode=diff -y palette.png`)
         # then apply the palette to get better results
-        run(`ffmpeg -v 0 -framerate $fps -loop $loop -i $(animdir)/%06d.png -i palette.png -lavfi paletteuse -y $fn`)
+        # use diff_mode=rectangle to improve rendering quality and reduce file-size
+        run(`ffmpeg -v 0 -framerate $fps -loop $loop -i $(animdir)/%06d.png -i palette.png -lavfi paletteuse=diff_mode=rectangle -y $fn`)
     else
         run(`ffmpeg -v 0 -framerate $fps -loop $loop -i $(animdir)/%06d.png -pix_fmt yuv420p -y $fn`)
     end


mkborregaard:
@H-M-H and @anowacki , you're the ones who wrote the current code - would you like to comment on this? Thanks a lot!

JobLeonard:
H-M-H, anowacki: in case you missed it in the related comment thread: I have _not_ tested this on my machine; I wrote it in-line through the browser, at mkborregaard's request. So it needs to be verified.

anowacki:
I really don't know anything about video conversion or advanced ffmpeg, but I had a quick go at comparing this PR against master for both line graphs and heatmaps of random matrices, and didn't notice any difference in the quality of the output (which is good, I guess).  Files were about 1% smaller on average.  The conversion sometimes runs more quickly with this PR.  So I don't see why we shouldn't merge.

(I'm not volunteering, but perhaps this calls for an automated animation test set/script in the future which reports conversion times and file sizes for changes which touch the animation stuff?  Timings probably aren't very reliable, but the sizes should be on the same filesystem.)

mkborregaard:
Thanks!

JobLeonard:
I'm happy it didn't break anything, but the small change in size is unexpected. The linked blog page suggests there should be a difference for graphs where only a small area changes each frame. Did you try it with the Lorenz Attractor demo from the documents?

```Julia
# initialize the attractor
n = 1500
dt = 0.02
σ, ρ, β = 10., 28., 8/3
x, y, z = 1., 1., 1.

# initialize a 3D plot with 1 empty series
plt = path3d(1, xlim=(-25,25), ylim=(-25,25), zlim=(0,50),
                xlab = "x", ylab = "y", zlab = "z",
                title = "Lorenz Attractor", marker = 1)

# build an animated gif, saving every 10th frame
@gif for i=1:n
    dx = σ*(y - x)     ; x += dt * dx
    dy = x*(ρ - z) - y ; y += dt * dy
    dz = x*y - β*z     ; z += dt * dz
    push!(plt, x, y, z)
end every 10
```

mkborregaard:
I'm happy to merge this, but that code is currently causing problems (#1239), so I think it is better to hold this before that's resolved.

mkborregaard:
Could you rebase this on master please?

mkborregaard:
@JobLeonard I'll merge if yuo can rebase and fix the conflicts?

BeastyBlacksmith:
I think this is outdated, since current master has the `stats_mode=diff` flag and uses `paletteuse=dither=sierra2_4a`.

JobLeonard:
Oof, that's a relief. I can feel a bit less bad about never getting around to this then haha (sorry mkborregaard, totally my bad)

PR #1242, opened 2017-11-14
Generate a palette for each gif frame; Fixes #1239

### LINKED ISSUE #1239: FFMPEG Error in Windows
There seems to be a problem with `@gif` saving files on windows in the current master version (not sure how to check this version number, but I updated just before posting this...). It works fine on the (older) registered version (0.13.1) however, so it seems to have been broken recently (or something weird is going on for me).

MWE:
```
using Plots; gr(); 
@gif for nn=1:3
    Plots.contour((collect(1:100)*collect(1:100)').^nn)
end
```
I'm getting this error at the point of saving:
```
failed process: Process(`ffmpeg -v 0 -i 'C:\Users\user\AppData\Local\Temp\jl_59F1.tmp/%06d.png' -vf palettegen -y palette.png`, ProcessExited(1)) [1]
    in anonymous at base\<missing>
    in macro expansion at Plots\src\animation.jl:146 <inlined>
    in gif at Plots\src\animation.jl:63
    in #gif#266 at Plots\src\animation.jl:63
    in #buildanimation#269 at Plots\src\animation.jl:75
    in run at base\process.jl:651
    in pipeline_error at base\process.jl:682
```
Thanks

biggsbiggsby:
Change palette creation from png to gif to avoid png encoding bug in ffmpeg
Related to #1239

### DIFF
--- a/src/animation.jl
+++ b/src/animation.jl
@@ -67,13 +67,21 @@ mp4(anim::Animation, fn = mp4fn(); kw...) = buildanimation(anim.dir, fn, false; 
 
 function buildanimation(animdir::AbstractString, fn::AbstractString,
                         is_animated_gif::Bool=true;
-                        fps::Integer = 20, loop::Integer = 0)
+                        fps::Integer = 20, loop::Integer = 0,
+                        variable_palette::Bool=false)
     fn = abspath(fn)
 
     if is_animated_gif
-        # generate a colorpalette for each frame to improve quality
-        palette="palettegen=stats_mode=single[pal],[0:v][pal]paletteuse=new=1"
-        run(`ffmpeg -v 0 -framerate $fps -loop $loop -i $(animdir)/%06d.png -lavfi "$palette" -y $fn`)
+        if variable_palette
+            # generate a colorpalette for each frame for highest quality, but larger filesize
+            palette="palettegen=stats_mode=single[pal],[0:v][pal]paletteuse=new=1"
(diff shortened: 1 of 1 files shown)


daschw:
Great! Thanks a lot @biggsbiggsby!
@H-M-H Do you see any issues with this change?

biggsbiggsby:
I changed this request to use FFMPEGs option to create a new palette for each frame (instead of one for the entire animation), which seems to increase quality of the animation substantially on plots (e.g. both regions with solid colors and gradients) with only a slight increase in file size.

This also avoids the png palette saving bug in FFMPEG 3.4 which was the original motivation for this fix.

biggsbiggsby:
Hm, actually the quality and file size seem to be dependent on the format of the palette used... 

A single bitmap or tiff palette seems to give both the highest quality and lowest file size of those I've tried, with gif palettes being the worst for both, and jpeg in between. This seems to be linked to using lossless or lossy formats for the palette, but it's surprising these alter the file size the way they do (lossy palette = larger file).

Anyway,  since lossless formats seem to perform equally, I'm going to change this back to requesting a single palette with a file format change to BMP, which should give the best quality/filesize ratio while still being compatible with FFMPEG 3.4 on Win7 x64. 
I'll keep the variable palette code as an option in case anyone wants to enable it. 

Sorry about any confusion created by the changes here.

daschw:
Thanks a lot @biggsbiggsby!
Is this ready to merge now?

biggsbiggsby:
No problem! As far as I'm concerned it is ready now :+1:

PR #1249, opened 2017-11-16
WIP: Fix scientific formatter for PGFPlots and GR. Fixes #1241

### LINKED ISSUE #1241: PGFPlots ticks text not calculated correctly for bigger numbers
It seems like the text for the tick marks is not calculated correctly for bigger numbers when using PGFPlots as the backend (i.e. if the numbers get big for example 10^4 is shown, but the text for the ticks is not changed).

screenshot from 2017-11-13 20-25-26

apalugniok:
#1241 

polar

### DIFF
--- a/src/backends/gr.jl
+++ b/src/backends/gr.jl
@@ -222,6 +222,13 @@ function gr_polaraxes(rmin::Real, rmax::Real, sp::Subplot)
     sinf = sind.(a)
     cosf = cosd.(a)
     rtick_values, rtick_labels = get_ticks(yaxis)
+    rtick_labels = if yaxis[:formatter] == :scientific && yaxis[:ticks] == :auto
+        rtick_labels = string.(convert_sci_unicode.(rtick_labels),"\\ ")
+        # unicode × messes up superscript alignment so add space to superscript
+        replace.(rtick_labels, "{", "{ ")
+    else
+        rtick_labels
+    end
 
     #draw angular grid
     if xaxis[:grid]
@@ -882,7 +889,14 @@ function gr_display(sp::Subplot{GRBackend}, w, h, viewport_canvas)
                 xi, yi = GR.wctondc(cv, sp[:framestyle] == :origin ? 0 : xor(flip, mirror) ? ymax : ymin)
                 # @show cv dv ymin xi yi flip mirror (flip $ mirror)
                 # ensure correct dispatch in gr_text for automatic log ticks
-                dv = xaxis[:scale] in (:ln, :log10, :log2) && xaxis[:ticks] == :auto ? string(dv, "\\ ") : string(dv)
+                dv = if xaxis[:scale] in (:ln, :log10, :log2) && xaxis[:ticks] == :auto
+                    string(dv, "\\ ")
+                elseif xaxis[:formatter] == :scientific && xaxis[:ticks] == :auto
+                    # unicode × messes up superscript alignment so add space to superscript
+                    string(replace(convert_sci_unicode(dv), "{", "{ "), "\\ ")
+                else
+                    string(dv)
+                end
                 gr_text(xi, yi + (mirror ? 1 : -1) * 5e-3 * (xaxis[:tick_direction] == :out ? 1.5 : 1.0), dv)
             end
         end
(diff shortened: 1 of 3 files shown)


mkborregaard:
It'd be nice to have this working for real, like the other backends:
```julia
using Plots; pyplot() #also broken on GR
plot(
       plot(1:50000, 1:50000), 
       plot(1:50000, 1:50000, formatter = :scientific)
       )
```
skaermbillede 2017-11-17 kl 08 55 41
skaermbillede 2017-11-17 kl 08 58 24

Current behaviour of pgfplots with the same call:

apalugniok:
Right I'll look into this. So if formatter scientific is passed you should get the overall scale on the side or in each tick? And otherwise it should display the whole number?

daschw:
IMHO showing the scale on the side for a scientific formatter is nice and probably also easier to implement for PGFPlots.

apalugniok:
Yeah, I do personally prefer the scale on the side I'll see how annoying it would be to have the other backends do this.

mkborregaard:
IMHO that degree of consistency is not required

daschw:
I agree with @mkborregaard. I think the backends can differ in details like this.

apalugniok:
So after some changes the `.tex` file that is created Plots on my branch contains 
```
xticklabels = {1.00$\times 10^{3}$,1.05$\times 10^{3}$,1.10$\times 10^{3}$,1.15$\times 10^{3}$,1.20$\times 10^{3}$},
```
If I run `pdfplatex` it gives me warnings/errors saying 
```
! Missing \endcsname inserted.
<to be read again> 
\times
```
but I can just press enter/return to skip over those and I get a pdf which looks completely fine and is as I expect.
Whatever Plots does to create the `.svg` from the `.tex` throws the same error and stops. Anybody have a clue what is wrong here?

daschw:
Sorry I have no idea right now what the issue could be. Maybe @pkofod could help?

apalugniok:
https://gist.github.com/apalugniok/edf683c38458412bd26b35a5f65f332d
Here is the full `.tex` file, thanks for your help 😄

pkofod:
Alright, apparently this is sort of a PGFPlots problem, see https://tex.stackexchange.com/a/66303/36051

can be fixed by placing
 ```
\let\OrgTimes\times
\protected\def\times{%
  \ifincsname
    \string\times%
  \else
    \OrgTimes
  \fi
}
```
in the preamble

**Comment by mkborregaard** [comment]
@apalugniok do you think we should open an issue on PGFPlots?

**Comment by apalugniok** [comment]
Yeah we probably should.
It seems to be a more general issue with things like `\alpha`, `\times` etc.
```
using Plots
using LaTeXStrings
plot(rand(10), xticks = ([1],[L"\alpha"]))
```
also throws the same error complaining about the `\alpha` and the `.tex` file contains 
```
xticklabels = {$\alpha$}
```

**Comment by pkofod** [comment]
When I said pgfplots I meant the latex package

**Comment by daschw** [comment]
Can we modify the preamble from within Plots?

**Comment by apalugniok** [comment]
@daschw  I believe we can. It wouldn't solve the `\alpha` issue but at least `\times` would work which I need to make the scientific ticks.

**Comment by pkofod** [comment]
How does unicode "times" look in the labels? ×

**Comment by apalugniok** [comment]
@pkofod It gets ignored the right plot of mkborregaard's bottom plot had the labels passed as `5×10⁴`

**Comment by pkofod** [comment]
Okay, anyway that would be a hack.. The correct thing is to fix it upstream, but...

**Comment by apalugniok** [comment]
The hack worked fine for `pdflatex` but `lualatex` which is used by PGFPlots.jl gives the same error complaining about `\OrgTimes` 😟

**Comment by mkborregaard** [comment]
There was just a question about this feature on Slack

**Comment by apalugniok** [comment]
I've been giving this some thought but I currently have no idea how to make PGFPlots behave with unicode or with \times 😟

**Comment by apalugniok** [comment]
Still need to test properly but 
```
plot(0.001*rand(5), formatter = :scientific)
```
pgfformatter
GR can display the × but can't save it
grformatter

**Comment by mkborregaard** [comment]
Great! @jheinen would you have any idea why GR behaves differently with `savefig` as seen above?

**Comment by jheinen** [comment]
GR still does NOT have unicode support - you can only use [Latin-1 Supplement (Unicode block)](https://en.wikipedia.org/wiki/Latin-1_Supplement_(Unicode_block). There is no short-term solution to fix this issue, as **all** device drivers in GR have to be updated.

**Comment by apalugniok** [comment]
Yeah, all saveable formats for GR don't work well with the ×, but display looks fine (screenshot below)
cross

**Comment by mkborregaard** [comment]
OK, let's merge!

**Comment by jheinen** [comment]
I'd like to stay in "Latin-1 mode" until full GR has full Unicode support. Therefore I will have to change some things in the GR run-time, e.g. replace the Unicode multiplication sign `×` by an `x` - the filling blanks would not be necessary and we could have full support in GR, even in the saveable formats. I hope I'll find the time tomorrow ...

**Comment by jheinen** [comment]
The following example now works in GR (for all backends):
```julia
GR.textext(0.5, 0.2, "1.234 × 10^2")
```
The former alignment problems were due to missing (or wrong) font metrics information in GR's lookup tables, which were created from outdated Adobe font metrics files.

The Unicode `x` is internally converted to the Latin-1 supplement equivalent (0xd7) - the `^` is parsed by GR and displayed as an exponent.

So , for GR, the PR needs still some work ...

jheinen:
#1349 should solve the remaining problems, but has 1 (of 36) test failures !?

mkborregaard:
That's number 30, boxplot and violin, with no scientific formatting. Does it fail locally for you as well?

jheinen:
Works locally, but the X- and Y-axes titles (VoicePart, Height) are missing:
Maybe, sth has changed in StatPlots ...

mkborregaard:
Yes, and the titles were removed in #1269 . So this should be fine to merge, but I do think we need PlotReferenceImages to be updated.

daschw:
Hm, I updated the PlotReferenceImages and everything works fine locally, but the tests still fail on travis for testexample 30.
Any ideas?

Issue #1124, opened 2017-10-02
'link' subplot attribute works with pyplot() but not plotly()


ko56:
```julia
using Plots
plotly()
scatter(rand(10,6), layout=(3,2), link=:all)
```

The y axes are not identical.  But this works with pyplot().

mkborregaard:
Yes, this is a known plotly bug. It's mentioned in the meta-issue but might make sense for it to have its own issue (since this should really be a priority).

mkborregaard:
This is likely a duplicate of #850

ko56:
I've seen the same problem, but decided to report only the 'link' one.

daschw:
Closed by #1253

PR #1253, opened 2017-11-17
Plotly : always define range explicitely. Solves axis linking #1124

romainFr:
Repairs axes linking in plotly.

```julia
using Plots
plotlyjs()
plot(rand(10,2).*[1 10],layout=(1,2),link=:y)
```

newplot

### DIFF
--- a/src/backends/plotly.jl
+++ b/src/backends/plotly.jl
@@ -252,7 +252,7 @@ function plotly_axis(axis::Axis, sp::Subplot)
     end
 
     ax[:tickangle] = -axis[:rotation]
-
+    ax[:range] = axis_limits(axis)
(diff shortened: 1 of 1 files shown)


src/backends/plotly.jl
```diff
@@ -261,10 +263,10 @@ function plotly_axis(axis::Axis, sp::Subplot)
         ax[:linecolor] = rgba_string(axis[:foreground_color_axis])
 
         # lims
-        lims = axis[:lims]
-        if lims != :auto && limsType(lims) == :limits
```
mkborregaard: I'd be fine with deleting it outright rather than leaving comments in the code

mkborregaard:
Looks good!

PR #1268, opened 2017-11-30
fix axes limits for fill=true in 3d plots (fix #1267)

### LINKED ISSUE #1267: Contour fill missbehaving
Using the pyplot backend, this:
```
Plots.contour(432475.0:25.0:439525.0, 8.675175e6:25.0:8.6857e6, rand(283,422)', fill=true)
```
produces an empty plot with axes and color-axes of the right extent.  Dropping the `fill=true` produces a good plot, also dropping the x and y vectors makes it work.

daschw:
#1213 set `fillrange = 0` for `fill=true`. This extended the yaxis limits to 0 for 3D plots, which is not expected. This should fix this.

### DIFF
--- a/src/axes.jl
+++ b/src/axes.jl
@@ -355,7 +355,7 @@ function expand_extrema!(sp::Subplot, d::KW)
     if fr == nothing && d[:seriestype] == :bar
         fr = 0.0
     end
-    if fr != nothing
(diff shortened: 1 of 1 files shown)


daschw:
The tests fail because the old DataFrames syntax has been deprecated in StatPlots. This should be fixed by #1269

PR #1278, opened 2017-12-04
Examples are limited to 90 columns.

JackDevine:
Thanks @daschw for pointing out the correct place to make these
changes. I have also changed the explanation texts, which may be a
little overkill, but I do think that the style guides apply to Markdown
string too (could be wrong though).


src/examples.jl
```diff
@@ -323,7 +376,11 @@ PlotExample("Animation with subplots",
 ),
 
 PlotExample("Spy",
-    "For a matrix `mat` with unique nonzeros `spy(mat)` returns a colorless plot. If `mat` has various different nonzero values, a colorbar is added. The colorbar can be disabled with `legend = nothing`.",
+"""
+For a matrix `mat` with unique nonzeros `spy(mat)` returns a colorless plot. If `mat` has 
+various different nonzero values, a colorbar is added. The colorbar can be disabled with 
+`legend = nothing`.
+"""
```
daschw: You missed a comma here. This makes the Travis tests fail.

JackDevine: Thanks for spotting that. I didn't do a very good job of dealing with merge conflicts.

daschw:
Awesome, Thanks a lot @JackDevine !

PR #1279, opened 2017-12-04
New PlotTheme type

daschw:
This is required by https://github.com/JuliaPlots/PlotThemes.jl/pull/21 and vice versa.

### DIFF
--- a/src/args.jl
+++ b/src/args.jl
@@ -417,6 +417,7 @@ const _all_defaults = KW[
 ]
 
 const _initial_defaults = deepcopy(_all_defaults)
+const _initial_axis_defaults = deepcopy(_axis_defaults)
 
 # to be able to reset font sizes to initial values
 const _initial_fontsizes = Dict(:titlefont  => _subplot_defaults[:titlefontsize],
@@ -614,7 +615,10 @@ function default(d::KW, k::Symbol)
     get(d, k, default(k))
 end
 
-reset_defaults() = foreach(merge!, _all_defaults, _initial_defaults)
+function reset_defaults()
+    foreach(merge!, _all_defaults, _initial_defaults)
+    merge!(_axis_defaults, _initial_axis_defaults)
+end
 
 # -----------------------------------------------------------------------------
 
--- a/src/themes.jl
+++ b/src/themes.jl
@@ -4,7 +4,7 @@
 Specify the colour theme for plots.
 """
 function theme(s::Symbol; kw...)
-    # Reset to defaults to overwrite active team
+    # Reset to defaults to overwrite active theme
     reset_defaults()
     thm = PlotThemes._themes[s]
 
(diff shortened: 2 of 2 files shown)


daschw:
I also added a recipe to quickly preview a theme:
`Plots.showtheme(:solarized)`:
shotheme_solarized

`Plots.showtheme(:solarized, :protanopic)`:
shotheme_solarized_protanopic

mkborregaard:
That's great! If one were to change other things than colors in the themes, how would that play together with Plots' default system?

daschw:
I'm not sure I understand your question correctly. Is your question related to the new PlotTheme type or to showtheme? In the new PlotTheme type the default of any Plots attribute can be changed and showtheme shows the figures above with the defaults specified by the theme:
```julia
using Plots
add_theme(:mytheme, PlotThemes.PlotTheme(framestyle = :none, linestyle = :dot))
Plots.showtheme(:mytheme)
```
mytheme

mkborregaard:
To the new PlotTheme type and how it relates to defaults set with `default()` (I think they should override the theme) and with those set by the `PLOTS_DEFAULTS` environmental variable. Right now there's some tricky behavior:
```
Plots.showtheme(:mytheme, color = :redsblues)
```
I'd hoped this would change the color gradient for the bottom row and leave the top row intact, but:
skaermbillede 2017-12-14 kl 11 15 56

daschw:
Ah, I see.
- I think the top row is expected: Just try `plot(rand(10, 4), color = :redsblues)` without any theme.
- The second row is an issue with the `showtheme` recipe. I explicitely use the theme's gradient colors here, to be able to modify them for different visions (protanopic, etc...). In praxis, `theme(:dark); heatmap(rand(10, 10), color = :redsblues)` works as expected. But there's definitely room for improvement in the `showtheme` recipe ...
- `theme()` in Plots uses `default()` so default value specified by `PLOTS_DEFAULTS` or a call to `default()` earlier which is also specified by a theme is overwritten by the theme's default. This has also been this way with the old PlotTheme type. Calling `default()` after calling `theme()` overwrites the theme's default values again.

daschw:
Maybe `theme()` really should ignore the attributes set in `PLOTS_DEFAULTS`. I think this could be implemented quite easily.

mkborregaard:
Ah OK. I think it's actually unfortunate that `color` covers both gradient and single color, whereas `palette` is distinct. Maybe we should split `color` into `color` and `gradient`? Maybe not.
Also look at the weird gridlines I get in showtheme when specifying color???

Great about the interaction with defaults.

Not sure about your last suggestion - maybe we should replace `PLOTS_DEFAULTS` by the `theme` functionality altogether.

daschw:
The grid line colors issue in 3D plots really is weird.

PR #1281, opened 2017-12-05
Fix the display of an grayscale image for image-nonsupported backends

hayatoikoma:
Gray-scale images were not plottable with the PlotlyJS backend due to the scope of some misplaced variables. This PR corrects this issue.


daschw:
Well spotted! Thanks a lot!

PR #1282, opened 2017-12-05
Update iterator properly in _subplot_setup

### LINKED ISSUE #1266: UndefRefError with `seriestype` and `scale`
Using the `scale` attribute together with `seriestype` results in the following error

```
UndefRefError: access to undefined reference

Stacktrace:
 [1] _subplot_setup(::Plots.Plot{Plots.PyPlotBackend}, ::Dict{Symbol,Any}, ::Array{Dict{Symbol,Any},1}) at $HOME/.julia/v0.6/Plots/src/pipeline.jl:267
 [2] _plot!(::Plots.Plot{Plots.PyPlotBackend}, ::Dict{Symbol,Any}, ::Tuple{StepRangeLen{Float64,Base.TwicePrecision{Float64},Base.TwicePrecision{Float64}},##15#16}) at $HOME/.julia/v0.6/Plots/src/plot.jl:203
 [3] #plot#145(::Array{Any,1}, ::Function, ::StepRangeLen{Float64,Base.TwicePrecision{Float64},Base.TwicePrecision{Float64}}, ::Vararg{Any,N} where N) at $HOME/.julia/v0.6/Plots/src/plot.jl:56
 [4] (::RecipesBase.#kw##plot)(::Array{Any,1}, ::RecipesBase.#plot, ::StepRangeLen{Float64,Base.TwicePrecision{Float64},Base.TwicePrecision{Float64}}, ::Function, ::Vararg{Function,N} where N) at ./<missing>:0
 [5] include_string(::String, ::String) at ./loading.jl:522
```
Each attribute individually works as expected.

Minimal example

`plot(linspace(1,3,10), x->x^2, scale=:log10, seriestype=:scatter)` or even

`plot(linspace(1,3,10), x->x^2, scale=:log10, seriestype=:path)`

Happens with all backends I tried (`PyPlot`,`GR`,`Plotly`).
```
julia 0.6.1
Plots.jl 0.13.1
```

wkearn:
This seems to fix #1266 

I'm still not entirely sure why this doesn't work on master and why changing 

```julia
for (k,v) in kw
```
to
```julia
for (k,v) in collect(kw)
```
does work. I guess it's something about what happens if you update an iterator while you are iterating over it. So there may be a better way to handle this, but using `collect` seems to fix the problems referenced in the issue.

### DIFF
--- a/src/pipeline.jl
+++ b/src/pipeline.jl
@@ -264,7 +264,7 @@ function _subplot_setup(plt::Plot, d::KW, kw_list::Vector{KW})
 
         # extract subplot/axis attributes from kw and add to sp_attr
         attr = KW()
-        for (k,v) in kw
+        for (k,v) in collect(kw)
             if haskey(_subplot_defaults, k) || haskey(_axis_defaults_byletter, k)
                 attr[k] = pop!(kw, k)
             end


mkborregaard:
Awesome! So I think what's going on here is as @wkearn says that `pop!(kw, k)` modifies `kw` several times within the loop while it's being iterated over, while `collect` realises kw into an anonymous Dict outside the loop.
Good to see you @wkearn - you were the first Julia person I ever interacted with when I first heard about the language and went to check the Shapefile capability :-)

PR #1284, opened 2017-12-06
change author section

tbreloff:
Hey guys... you're doing really awesome work!

I figured you wouldn't change the wording of this author line yourselves, which I appreciate, so I did it.  I'm sure you can understand I'd like to keep my name on it still, but "creator" seems like the right title now.  You have my blessing to change anything else about it... you could name names, or link to the commit rankings, or whatever.

Also, whatever changes you decide on should probably be made in kind elsewhere (I think there's something in the docs as well?)  Anyways all the best!


mkborregaard:
Great, thanks Tom!

PR #1286, opened 2017-12-06
say how to get the list of attributes in `?plot`

mforets:


src/plot.jl
```diff
@@ -42,7 +42,8 @@ The main plot command.  Use `plot` to create a new plot object, and `plot!` to a
 ```
 
 There are lots of ways to pass in data, and lots of keyword arguments... just try it and it will likely work as expected.
-When you pass in matrices, it splits by columns.  See the documentation for more info.
+When you pass in matrices, it splits by columns. To see the list of available attributes, use the `plotattr([attr])`
+function, where `attr` is the symbol `:Series:`, `:Subplot:`, `:Plot` or `:Axis`.
```
mkborregaard: Thanks, that's a nice idea. Might it be worth adding "Pass any attribute to `plotattr` as a String to look up it's docstring; e.g. `plotattr("seriestype")`"?

mforets: sure, good to know that too.

mkborregaard:
Thanks!

Issue #1297, opened 2017-12-08
`abline` shouldn't update limits


goretkin:
I think the intent of `abline`, like `hline` and `vline` is to plot a line that spans the plot, but that doesn't update the plot limits beyond what's "necessary". 

```julia
using Plots
@gif for c  = ["plot()", "plot(x->10*sin(x))", "Plots.abline!(10,0)"]
  eval(parse(c))
  title!(c)
end
```
![out](https://user-images.githubusercontent.com/581043/33768449-1806bc52-dbf4-11e7-8092-72cadcadcd4b.gif)

In this case, I think the limits should stay the same after the sinusoid is plotted.

In any case, is there a way to force the limits to stay the same? Or better yet, give a list of invisible points that should be used to update the limits (and if this list is empty, then don't update the limits?) I'm interested in drawing lines that extend beyond the limits, so that when the graph is panned or zoomed, the line appears to extend forever.

mkborregaard:
You can't just set `xlims` and `ylims`?

goretkin:
like `Plots.abline!(1, 0, xlims=xlims(current()), ylims=ylims(current()) )` ? Yes, that's one way to force the limits to stay the same. If there's an abbreviation, I'd like to know.

I opened the issue because I think the current behavior of `abline!` could be improved, especially for lines with high slope.

mkborregaard:
Cool, yeah makes sense. BTW, `xlims()` is sufficient, you don't need `xlims(current())`.

mkborregaard:
#1300

goretkin:
Thanks! And good to know there's a shortcut for getting the plot limits.

In https://github.com/JuliaPlots/Plots.jl/pull/1300#issuecomment-350681081, it was brought up that `vline` and `hline` don't update either (whether panning or adding new data). I was thinking about this too, and whether there's some mechanism for plotting "infinite" objects, like what one would like `hline`, `vline`, and `abline` to be.

matplotlib can do infinite `vline` and `hline` (but not something like `abline`, I think) because it has a separate coordinate system attached to the plot ("axes" here https://matplotlib.org/users/transforms_tutorial.html). So it can plot a line from x_axis=0 to x_axis=1 and it will be an hline. But, again, this doesn't really seem to generalize beyond `hline` and `vline`.

PR #1300, opened 2017-12-11
RFC: abline! does not change axis limits; fix: #1297

mkborregaard:
This leaves the axis limits in place, by only calculating the part of the line that's visible when calling the function. It's slightly hacky, though, as the line will not change when another series updates the axis limts (true for the existing implementation too, but maybe more tricky here)

### DIFF
--- a/src/recipes.jl
+++ b/src/recipes.jl
@@ -985,9 +985,11 @@ end
 
 # -------------------------------------------------
 
-"Adds a+bx... straight line over the current plot"
+"Adds a+bx... straight line over the current plot, without changing the axis limits"
 function abline!(plt::Plot, a, b; kw...)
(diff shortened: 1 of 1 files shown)


mkborregaard:
julia3mumak

daschw:
Looks great!
I don't think that it's that hacky. `hline` and `vline` don't update with either when adding series.
This would be great to have, though.

mkborregaard:
OK, thanks!

goretkin:
```
plot(rand(10))
Plots.abline!(500, 1)
```
gives 
ERROR: ArgumentError: interval must be given as (min, max)

mkborregaard:
That happens when the line falls outside the plot window. I've changed it to just skip the plot call in that case (#1326)

PR #1304, opened 2017-12-12
plotly: bar_width and stroke_width support for bar plots

### LINKED ISSUE #1303: Plotly backend silently ignores bar_width
`bar_width` appears in the supported arguments in plotly.jl:

https://github.com/JuliaPlots/Plots.jl/blob/f2e2ffa17033145d06166db217b39ce3a45a02a1/src/backends/plotly.jl#L43 

but it appears to get ignored.  I believe the `StatsBase.Histogram` recipe has gaps between bars as a result.  

# Example
```
julia> bar([1, 2, 3], bar_width = [.1, .5, 1])
```
screen shot 2017-12-12 at 1 43 44 pm

# The Fix
The javascript part is here (https://plot.ly/javascript/bar-charts/#customizing-individual-bar-widths).

I think we just need `d_out[:width] = ...` around the section below?  I don't understand the Plots internals well enough to write what the RHS is, but I'm willing to submit a PR if someone can give me a pointer.

https://github.com/JuliaPlots/Plots.jl/blob/f2e2ffa17033145d06166db217b39ce3a45a02a1/src/backends/plotly.jl#L537-L544

joshday:
Fixes #1303 and uses `stroke_width` to put border on bars for plotly.

```julia
histogram(randn(1000))
```

screen shot 2017-12-12 at 3 16 57 pm

### DIFF
--- a/src/backends/plotly.jl
+++ b/src/backends/plotly.jl
@@ -542,7 +542,6 @@ function plotly_series(plt::Plot, series::Series)
             y, x, "h"
         end
         d_out[:width] = series[:bar_width]
-        info(series[:strokewidth])
         d_out[:marker] = KW(:color => rgba_string(series[:fillcolor]),
                             :width => series[:strokewidth])
 
--- a/src/backends/plotly.jl
+++ b/src/backends/plotly.jl
@@ -541,7 +541,10 @@ function plotly_series(plt::Plot, series::Series)
         else
             y, x, "h"
         end
-        d_out[:marker] = KW(:color => rgba_string(series[:fillcolor]))
+        d_out[:width] = series[:bar_width]
+        info(series[:strokewidth])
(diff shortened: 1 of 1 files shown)


joshday:
Also:

```julia
bar([1,2,3], bar_width = [.1, .5, 1])
```

screen shot 2017-12-12 at 3 23 32 pm

mkborregaard:
Awesome @joshday , thanks a bunch!

mkborregaard:
@joshday sorry, I merged this too fast - in fact it breaks plotly (#1312). There is no key in series called `strokewidth` (the relevant to use here would be `:linewidth` but just setting width does not appear to work). I should have noticed that the figures you've posted above are actually from GR, not plotly!

joshday:
oh yikes, sorry.  I'm not sure how I forgot to change the backend to check.

PR #1307, opened 2017-12-13
Add linestyle argument to the legend

### LINKED ISSUE #1305: PyPlot backend - legend does not reflect linestyles specified for plot
When I plot using PyPlot backend the legend defaults to all sold line style even though I have specified a different style for the plot (dash)

wkearn:
This fixes #1305


daschw:
Thanks a lot!

PR #1308, opened 2017-12-14
add line_z and marker_z example

rdeits:
From discussion with @mkborregaard on gitter

### DIFF
--- a/src/examples.jl
+++ b/src/examples.jl
@@ -416,6 +416,22 @@ attribute. The default framestyle is `:axes`.
     end)]
 ),
 
+PlotExample("Lines and markers with varying colors",
+"""
+You can use the `line_z` and `marker_z` properties to associate a color with
+each line segment or marker in the plot. 
+"""
+    [:(begin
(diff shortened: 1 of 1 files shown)


rdeits:
Looks like I need to update the reference images as well?

mkborregaard:
Great! Yes that's right, you do that by cloning PlotReferenceImages.jl, running Plots' tests manually in an interactive REPL session, then PR the result back to  PlotReferenceImages.jl referring to this.

rdeits:
Ok, done: https://github.com/JuliaPlots/PlotReferenceImages.jl/pull/23

rdeits:
BTW: one inconvenient part of this process was having to manually install the test requirements for Plots in order to run it interactively and get the reference image popups. Perhaps an environment variable would be more use-friendly than checking `isinteractive()` ?

mkborregaard:
Yeah, good point, this could definitely be a lot smoother. If you've got a concrete idea a PR would def be welcome.

PR #1316, opened 2017-12-15
RFC: Invisible data used to calculate extent

goretkin:
I think it would be beneficial to (optionally) specify data that should be used to calculate the extent of a series, separately from the plot data.

Here's an attempt, and how it fixes (in a limited sense) handling `hline` and `vline` after the extents update.


goretkin:
fixed silly typo. tests should pass now.

goretkin:
Passing `x_extent_data = []` doesn't work but `x_extent_data = Float64[]` does, due to the check here: https://github.com/goretkin/Plots.jl/blob/45e0aaacb4c0806a34784f56ce98d445d985240f/src/axes.jl#L337

Not sure how to fix that.

mkborregaard:
Thanks for making a PR - I'm not convinced adding a new set of keywords is the best way to handle this type of behaviour, though. We try to keep the overall number of keywords down, since using a keyword is often not the most intuitive.
Could you expand a bit more on exactly the use case, the problem as you see it, and why this might be the best way to solve it?

goretkin:
I'm not sure if this is the best way to solve it. I understand wanting to keep keywords down.

The problem is drawing objects that appear to have infinite extent, such as `hline`, `vline`, `abline`, as users pan or add new plots.

Currently, it's not possible to draw an `hline` that stays an `hline` when the user pans. This PR makes it possible to draw an hline that extends beyond the limits without affecting the limits and without needing to take the limits away from `:auto`, so that they keep updating in the face of new series.

The PR arbitrarily extends the objects by 10 times the size of the current extent, in both directions. So if the user plots data too far, or pans too far, the illusion of infinity will be destroyed. I think this can be handled later with some callbacks "I need to be redrawn if the view port moves beyond this rectangle". In order to not need to redraw `hline`s, etc on every single user pan, it makes sense to draw them beyond the current limits, to avoid redrawing so often.

In general, I think it makes sense to optionally separate the data of a series, and the extent it has. Perhaps this could accommodate  stroke widths not getting clipped, as they are here: https://github.com/JuliaPlots/PlotReferenceImages.jl/blob/master/Plots/gr/0.12.3/ref3.png

mkborregaard:
So, I am sympathetic to the issue, but I don't see this as the correct approach to solving it. Most of the time, a user can either actively use `lims` or just call `hline!` etc. again at the end. I do think that we need a way to have more interactive feedback and updating of Plots, where e.g. specifying an `hline` series would lead that series to be redrawn every time the plot limits changed, but that should be implemented like that. 
So I vote for closing the PR. @daschw ?

goretkin:
I think there are two separate things here.

1.  Should plots have a mechanism for specifying the extent of a series separate from the data that backs that series

2. Is this a good way to implement that?

I hear that your answer to 2 is no. I'm kind of reading that the answer to 1 is also no, but I'm not sure.

Here's why my answer to 1 is yes:

 * Manually setting `lims` takes it away from `:auto`, which means you should only do this at the end, otherwise you have to manage the limits yourself. In an interactive plotting situation, there is no "end", really. And I personally don't enjoy a workflow which requires me to manually manage the limits.

* Backends already implement clipping logic. It seems unnecessary to re-implement this clipping logic. For things like `hline`, it's straightforward enough. Even for `abline`, it's not so easy to get this logic right. For more complicated unbounded shapes, it gets worse.

* The only way to benefit from a backend's clipping feature while keeping the limits at `: auto` is with a mechanism like what this PR tries to provide, as far as I can tell.

* When `Plots` does get some features for redrawing these "infinite" objects, it would be wasteful (and I bet it would have a noticeable performance impact) to redraw them for every single pixel that a user pans by, which is what would be needed if these objects are plotted only up against the view limits and not any further.

Of course, feel free to close the PR. But I would like to have something like this feature in Plots.

mkborregaard:
I'm not completely sure what you mean with 2 and 3 above? 4 depends on the implementation. With regards to 1, is what you mean that you want to be able to prevent some series from affecting the lims?

daschw:
I just checked out this branch and tried it a little, and I personally prefer this proposed behaviour of `hline`. Although I share your worries about adding new keywords, @mkborregaard, I actually think that @goretkin 's approach is quite a nice solution to make hline update when adding new series.

One major issue with this is that the `10*span` approach is (probably a reasonable but) not a very generic solution:
```julia
plot(rand(10))
hline!([0.5])
plot!(rand(200))
```
only shows the hline from 1 to 99.

goretkin:
@daschw  yes, `10` was pretty arbitrary! You're kind to suggest it's reasonable, but it's not ;).

@mkborregaard By "clipping logic" I mean how do you cut away parts of a shape so that they don't bleed out onto the axes. I assumed all backends implement this, but I realize that perhaps that's not true. In any case, *most* backends do, I would guess. Please let me know if I can clarify.

> 4 depends on implementation

I'm not sure I understand. If you only ever draw the objects up until the borders, then you certainly have to redraw them every time you pan, even by a pixel.

> With regards to 1, is what you mean that you want to be able to prevent some series from affecting the lims?

Not at the granularity of "series" but instead of data within the series. But that's not what I was talking about. I was responding to your suggestion:

> Most of the time, a user can either actively use lims or just call hline! etc. again at the end.

I was arguing that during interactive use, where you're adding series to plots, there isn't really an "end". It would be annoying to keep calling e.g. `hline`.

daschw:
An alternative solution is provided in #1445. Still, thank you very much for the PR @goretkin !

PR #1317, opened 2017-12-15
3D marker_z handling. Fixes #1205

### LINKED ISSUE #1205: marker_z does not work properly for 3D scatter plot
Specifying marker colors/shading via marker_z works for 2D scatter plot, but not 3D scatter plot. See, e.g.,
```
scatter(1:10,1:10,marker_z=0.9:-0.1:0.0) # correct with 2D scatter!
scatter(1:10,1:10,1:10,marker_z=0.9:-0.1:0.0) # incorrect with 3D scatter!
```
In 3D, the color bar indicates the range of the z-coordinates, not the marker colors/shading. I want to use colors/shading of the markers according to an input array specified by ```marker_z```, not the z-coordinates of the marker positions. ```surface``` plot with ```fill_z``` does exactly this (although for 3D surface objects, not for 3D scattered points) as we discussed at #640, #641. 
I think this functionality is quite important.
Thanks!
BVPs

apalugniok:
#1205 
Not sure if I have all the seriestypes which use z values to color themselves in `z_colored_series`.
```
using Plots
scatter(1:10,1:10,1:10,marker_z=0.9:-0.1:0.0, markersize = 10)
```
GR
marker_z
PyPlot
markerz

### DIFF
--- a/src/utils.jl
+++ b/src/utils.jl
@@ -553,8 +553,9 @@ zlims(sp_idx::Int = 1) = zlims(current(), sp_idx)
 
 function get_clims(sp::Subplot)
     zmin, zmax = Inf, -Inf
+    z_colored_series = (:contour, :contour3d, :heatmap, :histogram2d, :surface)
     for series in series_list(sp)
-        for vals in (series[:z], series[:line_z], series[:marker_z], series[:fill_z])
+        for vals in (series[:seriestype] in z_colored_series ? series[:z] : nothing, [:line_z], series[:marker_z], series[:fill_z])
             if (typeof(vals) <: AbstractSurface) && (eltype(vals.surf) <: Real)
                 zmin, zmax = _update_clims(zmin, zmax, ignorenan_extrema(vals.surf)...)
             elseif (vals != nothing) && (eltype(vals) <: Real)


mkborregaard:
it's slightly puzzling to me why this works :-O

apalugniok:
They issue I saw is that `get_clims()` always considered the z values of the series when making the color bar and the logic would always pick the widest range, even when the z values of the series had nothing to do with the colors. For the case in the issue a z range of 1-10 and marker_z range of 0-0.9 gave clims as 0-10 which then made the markers have the wrong colors. Something like `scatter(1:10,1:10,1:10,marker_z=90:-10:0.0, markersize = 10)` will actually provide the expected result on current master as the marker_z range is larger than the z values range.

The change makes `get_clims()` only consider the z values of a series for clims if the seriestype is one whose colors depend in z.

mkborregaard:
nice detective work!

apalugniok:
My only worry is do I have all the seriestypes whose color depends on z in `z_colored_series`

src/utils.jl
```diff
@@ -553,8 +553,9 @@ zlims(sp_idx::Int = 1) = zlims(current(), sp_idx)
 
 function get_clims(sp::Subplot)
     zmin, zmax = Inf, -Inf
+    z_colored_series = (:contour, :contour3d, :heatmap, :histogram2d, :surface)
```
mkborregaard: Could you just use `like_surface(series[:seriestype])`?

src/utils.jl
```diff
@@ -553,8 +553,9 @@ zlims(sp_idx::Int = 1) = zlims(current(), sp_idx)
...
     for series in series_list(sp)
-        for vals in (series[:z], series[:line_z], series[:marker_z], series[:fill_z])
+        for vals in (series[:seriestype] in z_colored_series ? series[:z] : nothing, [:line_z], series[:marker_z], series[:fill_z])
```
mkborregaard: why the `nothing`?

apalugniok: The `? :` needs to return something when the statement is false and the `nothing` definitely skips the limits logic.

src/utils.jl
apalugniok: `like_surface` gives true for `wireframe` and `image` which aren't colored by z so I don't think so.

mkborregaard:
OK, cool - I think you do have all the relevant seriestypes.

daschw:
Nice work @apalugniok !

BoundaryValueProblems:
@apalugniok , thank you very much! I'm very happy because I can now use ```scatter``` with correct colors. Happy Holidays!

PR #1320, opened 2017-12-16
Fix mismatched legend color when linecolor is given. Fixes #1319

### LINKED ISSUE #1319: PyPlot legend doesn't follow line color.
```
using Plots
pyplot()
plot(rand(10), linecolor = :red)
```
legend
I think it's an easy fix should just be changing `:fillcolor` to `:linecolor` in https://github.com/JuliaPlots/Plots.jl/blob/61be113714c5fa0e6f7e654c23b4cfed75e2eb2e/src/backends/pyplot.jl#L1237
I'll open a PR 😄

apalugniok:
#1319 
```
using Plots
pyplot()
plot(rand(10), linecolor = :red)
```
legend


daschw:
Great!

PR #1327, opened 2017-12-18
Don't return false from abline!

mkborregaard:


mkborregaard:
Sorry, amending previous pr

PR #1331, opened 2017-12-19
RFC: Subplot labeling. Fixes #998

### LINKED ISSUE #998: Plot Numbering and Labeling
For publications you usually need label plots like `(a)` `(b)` `(c)`. It would be nice if there was a way these kinds of annotations could automatically be added when using subplots/layouts.

apalugniok:
#998 
I opted for `ann = (:top_left, :auto)` instead of `ann = :top` as it makes the processing easier. 
```
using Plots
plot(rand(10,8), ann = (:top_left, :auto), layout = grid(2,2))
```

autolabels

```
plot(rand(10,8), 
ann = [([:top_left, :top_right], ["test", "(b)"]) (:top_center, :auto) ([5, 9], 0.5, ["middle", "stuff"])],
layout = grid(2,2))
```

customlabels

Note:
- Polar plot annotations don't behave well. The interpretation of annotation coordinates differs between backends.
- 3D plots also don't work.

### DIFF
--- a/src/components.jl
+++ b/src/components.jl
@@ -557,6 +557,7 @@ function process_annotation(sp::Subplot, positions::Union{AVec{Symbol},Symbol}, 
     positions, labs = makevec(positions), makevec(labs)
     for i in 1:max(length(positions), length(labs))
         pos, lab = _cycle(positions, i), _cycle(labs, i)
+        pos = get(_positionAliases, pos, pos)
         if lab == :auto
             alphabet = "abcdefghijklmnopqrstuvwxyz"
             push!(anns, (pos, text(string("(", alphabet[sp[:subplot_index]], ")"), font)))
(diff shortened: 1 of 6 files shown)


src/components.jl
```diff
@@ -535,6 +536,53 @@ annotations(anns::AVec) = anns
...
+            push!(anns, (pos, isa(lab, PlotText) ? lab : text(lab, font)))
+        end
+    end
+    anns
+end
+
+# Give each annotation coordinates based on specified position
+function locate_annotation(sp::Subplot, pos::Symbol, lab::PlotText)
+    position_multiplier = Dict{Symbol, Tuple{Float64,Float64}}(
+        :top_left => (0.1, 0.9),
```
daschw: I'd use versions without underscores (`:topleft`, ...) here to be consistent with the position options for the `legend` attribute.
The Plots way would probably be to allow both versions, as well as `:t`, `:tl`, `:tr`, ... for annotations and legendpositions. This could be achieved by introducing a `_positionAliases` `Dict` and do some preprocessing in args.jl.

apalugniok:
Forgot to mention Plotly cannot deal with annotations when multiple subplots are present.
brokenplotly

daschw:
IMO we can merge this and open an issue for plotly.

abehersan:
Hi! I just realized this convenience feature existed from the discourse posts! Is it documented in the API guide at all and I just missed it entirely? I could not find anything...

BeastyBlacksmith:
I don't think it is documented

PR #1340, opened 2018-01-06
Add a check for ticks==nothing or false

### LINKED ISSUE #1336: Not plotting ticks with framestyle=:origin
```julia
scatter(randn(10),randn(10),framestyle=:origin,ticks=nothing)
```

errors because of this line: https://github.com/JuliaPlots/Plots.jl/blob/a57745ed64094879b4c166587b58cb0e2277d7e2/src/axes.jl#L555

which tries to call `length(nothing)`. 

```julia
scatter(randn(10),randn(10),framestyle=:origin,ticks=[])
```

does what I want:

plotsframestyle

wkearn:
Fixes #1336


daschw:
Thanks!

PR #1341, opened 2018-01-06
RFC: Error if linecolor is a vector

### LINKED ISSUE #1325: Plotly/PlotlyJS behavior when linecolor is a vector
I haven't fixed `rgba_string` here: https://github.com/JuliaPlots/Plots.jl/blob/09c5e8a375dbd924347dac42a3c6067745d62276/src/backends/plotly.jl#L637

because I'm not entirely sure what

```julia
plot(randn(10,2),color=[:blue,:red])
```

should do. This will currently still cause an error for Plotly/PlotlyJS. 

GR just outputs two blue lines:

plots1306-4

If I try to apply my previous fix (#1324) to this line, you end up with

plots1306-3

Note that

1. This is different from what GR gives
2. The colors here are not blue and red, nor are they the default colors. I'm not entirely sure where they're coming from.

wkearn:
I don't think this is ready to merge yet, but I wanted to push something up and get everyone's take on it. This does fix #1325, but it will break code that currently works. 

A few questions:

- Is this the best place to throw this error? I'm still not completely sure how the whole pipeline works, but this looks like the first place that we definitely know `seriestype==:path` and `linecolor<:Vector`.
- Is that the right condition? Is there ever a situation with `seriestype==:path`, but it is okay for linecolor to be a vector?
- Perhaps a smoother transition is required, say warning when linecolor is a vector for a few releases then start erroring.

I'm also happy to change the wording of the error message.

### DIFF
--- a/src/pipeline.jl
+++ b/src/pipeline.jl
@@ -394,6 +394,10 @@ function _process_seriesrecipe(plt::Plot, d::KW)
         d[:fillrange] = nothing
     end
 
+    if (d[:seriestype] == :path) && typeof(d[:linecolor]) <: Vector
+        error("1D column vectors not supported for color specification in line plots. Perhaps you meant to use a 2D row vector?")
+    end
+
     # if it's natively supported, finalize processing and pass along to the backend, otherwise recurse
     if is_seriestype_supported(st)
         sp = _prepare_subplot(plt, d)


mkborregaard:
Great. There are ways to color line segments with a Vector, by using `line_z` - something like this along with setting a custom palette might make it possible to allow the behaviour rather than erroring on it.
E.g. 
```julia
plot(cumsum(randn(10)), line_z= rand(1:10,10), c = :Spectral, lw = 3)
```
skaermbillede 2018-01-08 kl 15 25 25

daschw:
Closing, because linecolor can now be a vector with #1467.

PR #1353, opened 2018-01-13
Check if any of the iterator args are zero

### LINKED ISSUE #1329: Plotting empty Shape gives DivideError
This happens under the Plotly backend too. Maybe the behavior could be a no-op, or maybe add an empty series, I'm not sure.

```
julia> plot(Shape([],[]))
ERROR: DivideError: integer division error
Stacktrace:
 [1] fld at ./int.jl:191 [inlined]
 [2] mod at ./int.jl:171 [inlined]
 [3] mod1 at ./operators.jl:649 [inlined]
 [4] _cycle(::Array{Float64,1}, ::Int64)
```

wkearn:
Fixes #1329

### DIFF
--- a/src/utils.jl
+++ b/src/utils.jl
@@ -199,7 +199,7 @@ allnan(istart::Int, iend::Int, args::Tuple) = all(i -> anynan(i, args), istart:i
 
 function Base.start(itr::SegmentsIterator)
     nextidx = 1
-    if anynan(1, itr.args)
+    if !any(iszero,map(length,itr.args)) && anynan(1, itr.args)
         _, nextidx = next(itr, 1)
     end
     nextidx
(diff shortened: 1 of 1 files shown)


src/utils.jl
```diff
@@ -199,7 +199,7 @@ allnan(istart::Int, iend::Int, args::Tuple) = all(i -> anynan(i, args), istart:i
 
 function Base.start(itr::SegmentsIterator)
     nextidx = 1
-    if anynan(1, itr.args)
+    if !any(iszero,map(length,itr.args)) && anynan(1, itr.args)
```
daschw: `!any(isempty, itr.args)` would be clearer IMO.

wkearn: Good point!

daschw:
Great, LGTM!

PR #1354, opened 2018-01-13
Fix axis limits not cropping image. Fixes #1296

### LINKED ISSUE #1296: limiting the axis doesn't crop the plotted image
When plotting an image, and then applying limits on the axis, the image doesn't get affected. 
```julia
h1 = plot(img, title="full")
h2 = plot(img, ylim=(200,300), xlim=(100,200), title="cropped?")
plot(h1, h2, aspect_ratio=1)
```
bad
But if we treat the image as a matrix of numbers, then it works (but we loose the original colors).
```julia
bw = Gray.(img)
m = Float64.(bw)
h1 = heatmap(m, title="full")
h2 = heatmap(m, ylim=(200,300), xlim=(100,200), title="cropped!")
plot(h1, h2, aspect_ratio=1)
```
good

apalugniok:
#1296 
```
using Plots, TestImages
img = testimage("cameraman")
plot(plot(img, lims = (0,300)), plot(img))
```
cropped
PyPlot cropped images as expected 
Not sure if Plotly is meant to support images. It had weird behavior where it would only crop the x axis and ignore the limits for y.

### DIFF
--- a/src/backends/gr.jl
+++ b/src/backends/gr.jl
@@ -1208,7 +1208,7 @@ function gr_display(sp::Subplot{GRBackend}, w, h, viewport_canvas)
                                         round(Int, green(c) * 255) << 8  +
                                         round(Int,   red(c) * 255) ), z)
             end
-            GR.drawimage(xmin, xmax, ymax, ymin, w, h, rgba)
+            GR.drawimage(0, w, h, 0, w, h, rgba)
         end
 
         # this is all we need to add the series_annotations text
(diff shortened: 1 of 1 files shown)


daschw:
Nice work!

daschw:
Maybe we were to fast here. When I run the tests locally I get:
gr_ref6
To reproduce: `Plots.test_examples(:gr, 6)`
Any ideas @apalugniok ?

apalugniok:
At a brief glance it looks like the limits aren't calculated correctly. The image's size is `(300,500)` so an x-extent of `(0,500)` but the limits get set to `(0,300)` and vice versa for y. I'll look into it this afternoon.

Issue #1196, opened 2017-10-19
plot(img, ylims=(-3,3)) produces unexpected direction


dlfivefifty:
The attached is weird to me that the negative y axis is above the positive y axis. 

<img width="739" alt="untitled" src="https://user-images.githubusercontent.com/1229737/31791618-0f27bcf8-b511-11e7-809d-c0a2d28f6416.png">

Even weirder is that reversing the order of arguments produces something not at all predictable:

<img width="809" alt="untitled 2" src="https://user-images.githubusercontent.com/1229737/31791655-352601b2-b511-11e7-98d2-069f2445546b.png">

In my actual usage case, I'm trying to put a phase portrait of a complex-valued function and so I really want the positive complex plane on top.

apalugniok:
```
plot(img)
```
![light](https://user-images.githubusercontent.com/22132297/31792994-d40699a0-b515-11e7-8d05-35cdb2d745a5.png)
I'd hazard a guess that the img[1,1] element is actually the top left corner of the image. So if the ticks follow the indices of `img` the first case make sense.

Not sure how to get the behaviour you'd like apart from the tedious way of specifying your own ticks 😟

dlfivefifty:
But that behaviour makes it useless for combining background images with plots on top.

dlfivefifty:
I think the `plot(img)` behaviour is fine, but when you specify `xlims` or `ylims` it should do what you tell it to do: `ylims = (-3,3)` means -3 on the bottom and 3 at the top, while `ylims = (700, 0)` should give back the same behaviour as the default.

mkborregaard:
That's how images are defined, though, starting at the upper left corner. Whereas heatmaps start at the lower left corner. The latter is standard everywhere, except for pyplot that transposes matrices in heatmap...
Could you use `yflip = true` on the plots you put on top?

mkborregaard:
You could also perhaps use `twinx()` to put your plot on top on an alternative y axis?

apalugniok:
I tried `yflip = true` and it seems broken but `xflip` works fine.

dlfivefifty:
I know that’s how images are defined, but specifying ylims is telling it to throw out the original coordinate system and use the specified coordinate system.

As it is now, I can’t see any reasonable usage for ylims. Explain to me how you would draw a half circle in the positive y axis on top of an image in the current convention?

dlfivefifty:
[...]

PR #1362, opened 2018-01-18
Fix image axis limits and add image movement and resizing.

apalugniok:
This should fix the incorrect image limits mentioned in #1354.
I've also added some functionality for moving and resizing the image by specifying x,y coordinate vectors. You can now manually unflip the yaxis and along with the move/resize functionality it should be easier to achieve the use case discussed in #1196. 

I'm not particularly happy with the two recipes I added for (x,y,img) as they are basically copies of the recipe that gets used when just the image is given, but I couldn't think of a neat way to make those work when x,y data is also given.

```
using Plots, TestImages
img = testimage("lighthouse")
plot(img)
```
img

```
plot(linspace(0,10,768), linspace(0,5,512), img)
```
imgresize

```
plot(linspace(0,10,768), linspace(0,5,512), img, xlims = (0,5))
```
imgresizecrop
```
plot(linspace(0,10,768), linspace(0,5,512), img[end:-1:1,:], yflip = false)
plot!(x -> sin(x)+2, linewidth = 10)
```
imgoverlay

Issues:
- I've experienced no issue with GR and GLVisualize.
- Resizing with PyPlot as in the second command changes the aspect ratio, this can be fixed by setting aspect ratio manually.
- Plotly is slow with plotting images.

### DIFF
--- a/src/backends/glvisualize.jl
+++ b/src/backends/glvisualize.jl
@@ -1215,7 +1215,7 @@ end
 
 function gl_image(img, kw_args)
     rect = kw_args[:primitive]
-    kw_args[:primitive] = GeometryTypes.SimpleRectangle{Float32}(rect.x, rect.y, rect.h, rect.w) # seems to be flipped
+    kw_args[:primitive] = GeometryTypes.SimpleRectangle{Float32}(rect.x, rect.y, rect.w, rect.h)
     visualize(img, Style(:default), kw_args)
 end
 
--- a/src/backends/gr.jl
+++ b/src/backends/gr.jl
@@ -1199,6 +1199,7 @@ function gr_display(sp::Subplot{GRBackend}, w, h, viewport_canvas)
         elseif st == :image
             z = transpose_z(series, series[:z].surf, true)'
             w, h = size(z)
+            xmin, xmax = ignorenan_extrema(series[:x]); ymin, ymax = ignorenan_extrema(series[:y])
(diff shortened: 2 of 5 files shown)


apalugniok:
Yeah I agree that is a bit weird. To be honest I'm not too sure how to interpret that kind of call since something like `heatmap([1, 10, 11], [1, 10, 11], rand(10, 10))` does one thing on GR another on Plotly and errors on PyPlot.

I mostly wanted something that would allow the images to occupy different limits on the axis since setting `lims` now crops the image. Otherwise the image would just have a fixed size and plotting over it would likely mean modifying data by hand and then creating custom ticks.

I forgot to mention that for Plotly setting `cbar = true` doesn't bring back the colorbar that used to be there but I don't think a colorbar would ever be useful for an image? But if you were to plot a contour ` f(x,y) = x^2+y^2;  contour!(1:50,1:50,f, cbar = true)` you have to set the `cbar = true` for a colorbar to appear, so I'm not 100% sure on the `cbar --> false` setting.

daschw:
> Yeah I agree that is a bit weird. To be honest I'm not too sure how to interpret that kind of call since something like heatmap([1, 10, 11], [1, 10, 11], rand(10, 10)) does one thing on GR another on Plotly and errors on PyPlot.

Good point! I'd keep `cbar --> false`.

apalugniok:
Great I'll give this a final check-over today and then go ahead and merge 😄

Issue #1375, opened 2018-01-31
Animation :every frequency must be constant


grahamas:
This works:
```
anim = @animation for i in 1:20
 plot(1:5)
end every 2
```
But this doesn't:
```
f = 2
anim = @animation for i in 1:20
 plot(1:5)
end every f
```

daschw:
fixed by #1376

PR #1376, opened 2018-01-31
Allow animation frame saving frequency to be a variable.

grahamas:
see #1375


daschw:
Thanks a lot @grahamas !

PR #1395, opened 2018-02-15
WIP: Add :native option to ticks attribute (Fixes #1382)

### LINKED ISSUE #1382: Zooming in PlotlyJS does not update the tick marks
When I zoom in a PlotlyJS.jl plot in a notebook, the axis gains new tick marks (`1.5, 1.6`, etc.)

```julia
using PlotlyJS
plot(scatter(;x=1:4, y=[10, 15, 13, 17]))
```

screenshot from 2018-02-06 07-03-38

This does not happen if I use Plots.jl to do the same

```julia
using Plots
plotlyjs()
plot(1:4, [10, 15, 13, 17])
```

screenshot from 2018-02-06 07-06-08

Zooming further, the x axis gets completely blank.

apalugniok:
#1382 

Proof of concept at this stage.

```
using Plots; plotly()
plot(rand(10), ticks = :native)
```
native_ticks

```
using Plots; plotly()
xs = [ "x$i" for i in 1:10 ]
ys = [ "y$i" for i in 1:4 ]
z = rand(4, 10)
plot(heatmap(xs, ys, z), heatmap(xs, ys, z, ticks=:native))
```
discrete-native-ticks

### DIFF
--- a/src/axes.jl
+++ b/src/axes.jl
@@ -234,6 +234,9 @@ function get_ticks(axis::Axis)
     ticks = _transform_ticks(axis[:ticks])
     ticks in (nothing, false) && return nothing
 
+    # treat :native ticks as :auto
+    ticks = ticks == :native ? :auto : ticks
+
     dvals = axis[:discrete_values]
     cv, dv = if !isempty(dvals) && ticks == :auto
         # discrete ticks...
--- a/src/backends/gr.jl
+++ b/src/backends/gr.jl
@@ -649,7 +649,7 @@ function _update_min_padding!(sp::Subplot{GRBackend})
     end
     # Add margin for x and y ticks
     xticks, yticks = axis_drawing_info(sp)[1:2]
-    if !(xticks in (nothing, false))
+    if !(xticks in (nothing, false, :none))
         flip, mirror = gr_set_xticks_font(sp)
         l = gr_get_ticks_size(xticks, 2)
         if mirror
@@ -658,7 +658,7 @@ function _update_min_padding!(sp::Subplot{GRBackend})
             bottompad += 1mm + gr_plot_size[2] * l * px
         end
     end
-    if !(yticks in (nothing, false))
+    if !(yticks in (nothing, false, :none))
         flip, mirror = gr_set_yticks_font(sp)
         l = gr_get_ticks_size(yticks, 1)
         if mirror
--- a/src/backends/gr.jl
+++ b/src/backends/gr.jl
@@ -222,7 +222,7 @@ function gr_polaraxes(rmin::Real, rmax::Real, sp::Subplot)
     sinf = sind.(a)
     cosf = cosd.(a)
     rtick_values, rtick_labels = get_ticks(yaxis)
-    if yaxis[:formatter] == :scientific && yaxis[:ticks] == :auto
+    if yaxis[:formatter] == :scientific && yaxis[:ticks] in (:auto, :native)
         rtick_labels = convert_sci_unicode(rtick_labels)
     end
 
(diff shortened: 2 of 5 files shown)


daschw:
Looks great @apalugniok !

diegozea:
I've found that the categorical axis of heatmaps became numeric with the :native option using the plotly backend.

apalugniok:
@diegozea Can you give an MWE?

diegozea:
Sure! Sorry for the late reply, here is the MWE:

```
using Plots; plotly()
xs = [ "x$i" for in in 1:10 ]
xs = [ "x$i" for i in 1:10 ]
z = rand(4, 10)
plot(heatmap(xs, ys, z), heatmap(xs, ys, z, ticks=:native))
```

newplot

diegozea:
It's useful in my case. I have a similar plot with more than 1,000 species and 10,000 genes (both are categorical axis with their respective species and gene names) and the count of gene repetitions as the colour. Since there are so many genes, I zoom the heatmap to look for interesting gene/species combinations.

apalugniok:
I currently only have a hack 
```
using Plots; plotly()
xs = [ "x$i" for i in 1:10 ]
ys = [ "y$i" for i in 1:4 ]
z = rand(4, 10)
p = heatmap(xs, ys, z, ticks = :native)
p[1].series_list[1][:x] = xs
p[1].series_list[1][:y] = ys
p
```
I does align the heatmap wrong initially but pressing autoscale fixes it.

diegozea:
Great! I've noticed that the hack works with `plotly()` but It doesn't work with `plotlyjs()`.

apalugniok:
My plotlyjs is currently broken so unfortunately I can't check that 😟. 

My main worry with the categorical axes is that by the time the `Plot` object is created the only information about the `xs` and `ys` given to `plot` would be in `ticklabels` (I think) since eg `series[:x]` contains numeric values which are needed to draw the heatmap on other backends. So the only quick solution I see is passing the appropriate `ticklabels` to the `:x` and `:y` in the Plotly dictionary which feels like asking for trouble. @daschw any ideas when this could go wrong or produce something unexpected?

daschw:
Hm, for `ticks = :native` you could check if the axis `discrete_values` are empty in the backend code and otherwise use them as `x` resp. `y` values in the backend object:
```julia
julia> plt = plot(["x$i" for i in 1:10], rand(10))

julia> plt[1].attr[:xaxis][:discrete_values]
10-element Array{Any,1}:
 "x1"
 "x2"
 "x3"
 "x4"
 "x5"
 "x6"
 "x7"
 "x8"
 "x9"
 "x10"

julia> plt[1].attr[:yaxis][:discrete_values]
0-element Array{Any,1}
```
That's more ore less what you were proposing, right? It feels a little hacky but I think it should be fine.

apalugniok:
I don't think PlotlyJS will be working yet. I haven't touched since I can't even get it to display a simple plot. 

The `discrete_values` seem to be working quite well apart from horizontal bar charts which are broken on master so I think it's unrelated, I'll open an issue for it.

daschw:
Could you also check the categorical axis case for PyPlot @apalugniok ?

apalugniok:
I don't think matplotlib has the same functionlity that Plotly does. Using the first example from https://matplotlib.org/gallery/lines_bars_and_markers/categorical_variables.html in standard matplotlib gives
matplotlib-categorical
So changes to `:native` ticks and categorical axes for PyPlot would add consistency but no extra functionality.

daschw:
I agree, but still the consistency would be nice to have, if it's not to much trouble to implement, no? Otherwise `:native` ticks and categorical axes in Plots would return numerical ticks in PyPlot, if I'm not mistaken.

apalugniok:
Seems like PyPlot can't interpret categorical ticks very well natively.
```
using PyPlot
xs = [ "x$i" for i in 1:10 ]
plot(xs, rand(10))
```
categorical-axes-pyplot
I think we should leave PyPlot with numerical ticks. I've had no luck getting Plots to display the right ticks for categorical axes. The behavior with `x10` being the second tick did show up as well
``` 
using Plots; pyplot()
xs = [ "x$i" for i in 1:10 ]
 p = plot(xs, rand(10), ticks = :native)
p[1][1][:x] = xs 
p
```
categorical-axes-plots-pyplot

daschw:
In that case go for the numerical ticks and feel free to merge, whenever you're ready!

PR #1401, opened 2018-02-19
Prevent addition of curly braces around style string. (Fixes #1396)

### LINKED ISSUE #1396: PGFPlots backend produces extra braces in .tex output leading to compile failures
Here's a minimal example:
```
using Plots
pgfplots()  
using LaTeXStrings  
plt = plot()  
xticks!([0.1,0.2],[L"\alpha", L"\beta"])  
savefig(plt, "minimal_bug.tex")  
```
gives an error upon running the fifth line. If I change `L"\alpha"` and `L"\beta"` to `L"a"` and `L"b"` it works fine.
Upon examining the .tex output I see the following:
```
\begin{tikzpicture}[]
\begin{axis}[height = {101.6mm}, ylabel = {}, xmin = {0.0}, xmax = {1.0}, ymax = {1.0}, xlabel = {}, {unbounded coords=jump, scaled x ticks = false, xticklabel style={rotate = 0}, xmajorgrids = true, xtick = {0.1,0.2}, xticklabels = {$\alpha$,$\beta$}, xtick align = inside, axis lines* = left, scaled y ticks = false, yticklabel style={rotate = 0}, ymajorgrids = true, ytick = {0.0,0.2,0.4,0.6000000000000001,0.8,1.0}, yticklabels = {0.0,0.2,0.4,0.6,0.8,1.0}, ytick align = inside, axis lines* = left,     xshift = 0.0mm,
    yshift = 0.0mm,
    axis background/.style={fill={rgb,1:red,1.00000000;green,1.00000000;blue,1.00000000}}
}, ymin = {0.0}, width = {226.822mm}]\end{axis}

\end{tikzpicture}
```
In the above, there is an extra pair of braces from `{unbounded coords = jump` to `1.00000000}}
}` which causes the above error. Removing this set of braces gives a .tex file that compiles properly.
I suspect the problem is in the usage of one of the `push!` methods in the pgf backend [file](https://github.com/JuliaPlots/Plots.jl/blob/master/src/backends/pgfplots.jl).

apalugniok:
#1396
```
using Plots
pgfplots()  
using LaTeXStrings  
plt = plot()  
xticks!([0.1,0.2],[L"\alpha", L"\beta"])   
``` 
This seems to fix the annoying '\endcsname' issue that PGFPlots had.

### DIFF
--- a/src/backends/pgfplots.jl
+++ b/src/backends/pgfplots.jl
@@ -450,7 +450,7 @@ function _update_plot_object(plt::Plot{PGFPlotsBackend})
         end
         @label colorbar_end
 
-        o = axisf(; style = style, kw...)
+        o = axisf(; style = join(style, ","), kw...)
 
         # add the series object to the PGFPlots.Axis
         for series in series_list(sp)


daschw:
Nice!

Issue #1409, opened 2018-02-24
Show columnames in legend of a grouped plot of a DataFrame


crstnbr:
This has been discussed in a Slack thread.

Example from [StatPlots.jl github page](https://github.com/JuliaPlots/StatPlots.jl) where the legend entries are fine because the entries in the DataFrame are nice strings:
```julia
using RDatasets
school = RDatasets.dataset("mlmRev","Hsb82")
@df school density(:MAch, group = :Sx)
```
![good](https://user-images.githubusercontent.com/187980/36630216-1ffa9706-1963-11e8-8c1c-9b82ecb49bd1.png)

If one has numerical data though, let's say something like
```julia
df = DataFrame(L = vcat([fill(i, 10) for i in [8,10]]...), T = vcat([fill(t,5) for t in [0.2, 0.1, 0.2, 0.1]]...), R = vcat([collect(1:5) for _ in 1:4]...), F = [1,2,3,4,5,5,4,3,2,1,1,1,2,2,3,3,3,3,3,3])
```
one gets a legend just containing numbers
```julia
@df df plot(:R, :F, group = (:L, :T), legend=:topleft, marker=:circle, frame=:box, grid=false)
```
![bad](https://user-images.githubusercontent.com/187980/36630408-2d1eceae-1966-11e8-9249-b90b54b9750d.png)

It would be nice to have a dedicated option, which naively would be called `showcolumnnamesinlegend` , that would lead to this much nicer legend.
![goodagain](https://user-images.githubusercontent.com/187980/36630431-90d79d90-1966-11e8-9fa1-74ca4f3aa4ca.png)

piever:
Just to make sure: this issue only concerns StatPlots and `@df` or is there something you'd like to add to Plots as well to address this?

piever:
After some more thinking, the best is to probably allow passing a NamedTuple to `group`, in which case the names are shown. For example:

```julia
plot(a, group = @NT(a = [1,1,2,2], b = [1,2,1,2]))
```

should display `a = 1 b = 1` legends. Then in StatPlots you could do:

`@df df plot(:R, :F, group = @NT(L = :L, T = :T), legend=:topleft, marker=:circle, frame=:box, grid=false)`

and probably I could also add the `{}` from Query/JuliaDBMeta where `{a, b}` is short-hand for `@NT(a = a, b = b)` as more often then not the name of the column is the name you want to appear.

piever:
[...]

PR #1429, opened 2018-03-12
Partially address #1409: group with named tuples

piever:
This allows grouping with NamedTuples to give a name to the entries in the legend. For example:

```julia
julia> using Plots

julia> using NamedTuples

julia> plot(rand(10), group = @NT(a = rand(Bool,10), b = rand(Bool, 10)))
```
test

In #1409 this would allow:

`@df df plot(:R, :F, group = @NT(L = :L, T = :T), legend=:topleft, marker=:circle, frame=:box, grid=false)`


crstnbr:
Maybe put a comma, i.e. `a = false, b = true` (I forgot this in #1409). Otherwise this looks great!

PR #1430, opened 2018-03-14
add option to supress saved message

yakir12:
Added option `show_msg::Bool` to `buildanimation` to show/suppress the "Saved animation to..." message.


daschw:
Thanks @yakir12

PR #1436, opened 2018-03-15
fix fill magic argument

daschw:
`fill = ([-1, -1], 0.4, :red)` now works correctly. (Originally `[-1, -1]` was interpreted as vector of colors.)

Furthermore, it's now possible to on Plotly(JS) to do
```julia
plot(rand(10), fillrange = -1)
plot(rand(10), fillrange = (-1, 1))
plot(rand(10), fillrange = (-1, 1 + rand(10))
```
(Originally only `AbstractArray`s, `0`, or `Tuple`s of `AbstractArray`s were supported.)

### DIFF
--- a/src/args.jl
+++ b/src/args.jl
@@ -751,6 +751,10 @@ function processFillArg(d::KW, arg)
     elseif allAlphas(arg)
         d[:fillalpha] = arg
 
+    # fillrange provided as vector or number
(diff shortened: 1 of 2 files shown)


staticfloat:
I thought this might fix https://github.com/JuliaPlots/StatPlots.jl/issues/133 as well, but unfortunately it doesn't look like it does, despite the symptom appearing very similar (area fills not starting at the proper locations)

PR #1437, opened 2018-03-16
RFC: use bar recipe in plotly

daschw:
As reported in https://github.com/JuliaPlots/StatPlots.jl/issues/133 `groupedbar` does not work properly with Plotly(JS). The reason for this is that fillrange is not working with Plotly's native bar implementation. Using the bar recipe of Plots instead would fix this issue and allow stacked `groupedbar` plots:

plotly_bar_recipe

However, as you can see, this would be a regression in terms of the values shown when hovering (`x` values at the border of the bar) and maybe also legend entry symbols. Here's the current behaviour for comparison:

plotly_bar_native

Personally, I think having `fillrange` and `groupedbar` work is preferable, but maybe the majority has a different opinion (I guess the plotly(js) backends are often chosen because of the interactivity and all this hovering stuff). Any thoughts?

### DIFF
--- a/src/backends/plotly.jl
+++ b/src/backends/plotly.jl
@@ -49,7 +49,7 @@ const _plotly_attr = merge_with_base_supported([
   ])
 
 const _plotly_seriestype = [
-    :path, :scatter, :bar, :pie, :heatmap,
+    :path, :scatter, :pie, :heatmap,
     :contour, :surface, :wireframe, :path3d, :scatter3d, :shape, :scattergl,
 ]
 const _plotly_style = [:auto, :solid, :dash, :dot, :dashdot]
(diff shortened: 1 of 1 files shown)


piever:
Nice!

Personally my opinion is that this is a step in the right direction and we should use recipes as much as possible: having the interfaces for the various backends be at different levels of complexity means that all backend codes are different and therefore hard to maintain. I think in the long term we should aim to only use very basic functions from the backend (say draw a line or a fill a shape) so that all new features can be added directly in the Plots code by coding them once. This issue is not specific to `bar` and PlotlyJS: as soon as we use some high-level function from some backends but not others we're vulnerable to this kind of bugs...

daschw:
I completely agree!

PR #1438, opened 2018-03-16
allow ticks = :all and ticks = n::Int for categorical axes

daschw:
```julia
using Plots

x = ["x$i" for i in 1:50]
y = rand(50)

plot(
    plot(x, y), # xticks = :auto
    plot(x, y, xticks = :all),
    plot(x, y, xticks = 9),
    layout = (3, 1), legend = false
)
```
categorical_ticks

### DIFF
--- a/src/axes.jl
+++ b/src/axes.jl
@@ -238,10 +238,18 @@ function get_ticks(axis::Axis)
     ticks = ticks == :native ? :auto : ticks
 
     dvals = axis[:discrete_values]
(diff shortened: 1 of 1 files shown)


piever:
Is there a way for this to work with integer valued axis as well? Could `bar(1:3, rand(3), ticks = :all)` give `1:3` as ticks?

daschw:
Yes, I think that should be possible.

PR #1443, opened 2018-03-18
Partial attempt to fix labeling to atomatically scientific notation

fedxa:
At present unless you specify explicitly `:scientific` formatting for the labels (hmm, have yet to find how to do this...), you are going to get fixed notation for any numbers - really inconvenient!  A the same time, `Showoff.jl` provides `:auto` formatter, which guesses properly the style to use.

Here is a simple fix that selects :auto label formatter instead of :plain, making default behaviour as expected from a well behaved plotter in the following situation:

```
x=1e-12:1e-13:1e-10;Plots.plot(x,1./x)
```

Note, that I had to fix in a crude way the unicode translator for the GR backend -- definitely a better approach is possible.

### DIFF
--- a/src/axes.jl
+++ b/src/axes.jl
@@ -212,7 +212,7 @@ function optimal_ticks_and_labels(axis::Axis, ticks = nothing)
         formatter = axis[:formatter]
         if formatter == :auto
             # the default behavior is to make strings of the scaled values and then apply the labelfunc
-            map(labelfunc(scale, backend()), Showoff.showoff(scaled_ticks, :plain))
+            map(labelfunc(scale, backend()), Showoff.showoff(scaled_ticks, :auto))
         elseif formatter == :scientific
             Showoff.showoff(unscaled_ticks, :scientific)
         else
--- a/src/axes.jl
+++ b/src/axes.jl
@@ -213,11 +213,19 @@ function optimal_ticks_and_labels(axis::Axis, ticks = nothing)
         if formatter == :auto
             # the default behavior is to make strings of the scaled values and then apply the labelfunc
             map(labelfunc(scale, backend()), Showoff.showoff(scaled_ticks, :auto))
+        elseif formatter == :plain
+            # Leave the numbers in plain format
+            map(labelfunc(scale, backend()), Showoff.showoff(scaled_ticks, :plain))
         elseif formatter == :scientific
             Showoff.showoff(unscaled_ticks, :scientific)
         else
             # there was an override for the formatter... use that on the unscaled ticks
             map(formatter, unscaled_ticks)
+            # if the formatter left us with numbers, still apply the default formatter
+            # However it leave us with the problem of unicode number decoding by the backend
+            # if eltype(unscaled_ticks) <: Number
+            #     Showoff.showoff(unscaled_ticks, :auto)
+            # end
         end
     else
         # no finite ticks to show...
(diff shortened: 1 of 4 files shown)


daschw:
Thanks @fedxa ! I don't think this is ready yet. With this branch checked out I get the following figures for `x=1e-12:1e-13:1e-10;Plots.plot(x,1./x)` on different backends: 
With `pyplot()` I get
sci_pyplot

which is nice! It also seems to work with `plotlyjs()`. However, `plotly()` produces
sci_plotly

`pgfplots()`:
sci_pgfplots

`gr()` prints
```julia
missing '}'
missing '}'
missing '}'
missing '}'
missing '}'
missing '}'
missing '}'
missing '}'
missing '}'
```
in the REPL and shows
sci_gr

fedxa:
Ok, I believe I fixed plotly backend -- it was actually the bug/feature unconnected to my pull request, as far as utf-8 was not working in it anyway (or at least depended on the default charset settings of the browser)

For GR I don't know what to do -- this is actually the backend that I use, and it was the first one I fixed, even before submitting the pull request... Can you give more details on the configuration you have?

daschw:
I think, all the issues are unrelated to your PR, but if we want to make this the default it should work on all backends.

I had an old version of GR, works now with 0.30.0, sorry!

Still, the `pgfplots()` result is weird. `plot(x, 1 ./ x, formatter = :scientific)` produces the correct result (with `-` in the exponent for the `xaxis`, though. @apalugniok If I remember correctly, you worked on the scientific formatter for `pgfplots()`. Do you have an idea what the issue could be here?

`glvisualize()` also seems to work fine.

However, it looks like this change is breaking the `formatter` attribute:
```julia
julia> plot(x, 1 ./ x, formatter = x -> 2x)
ERROR: MethodError: no method matching contains(::Float64, ::String)
Closest candidates are:
  contains(::Function, ::Any, ::Any) at reduce.jl:664
  contains(::AbstractString, ::AbstractString) at strings/search.jl:378
```

fedxa:
Thank you, a cleaner checking for the formatter fixed both pgfplots and arbitrary formatter! Please, check if I am missing something.

There is one issue, but I believe it is a bit beyond the scope of the current patch.  Your suggested`formatter=x->2*x` scales the axis, but the output of it is not processed further through the pretty printing logic -- so that instead of nice '10^10' you get '1e10' as the label. There are two solutions to this -- one is to decide that if the user provided the formatter function, then just leave it like this -- it was the user's problem how to format. Other -- still check the output from the formatter function, and if it is still numeric (not ready to display text), run the `:auto` formatter on top.

daschw:
Nice! It would be also great to have the option `formatter = :plain` to get back the old behaviour if someone prefers that.

> There are two solutions to this -- one is to decide that if the user provided the formatter function, then just leave it like this -- it was the user's problem how to format. Other -- still check the output from the formatter function, and if it is still numeric (not ready to display text), run the :auto formatter on top.

I prefer the second option.

fedxa:
> It would be also great to have the option formatter = :plain to get back the old behaviour if someone prefers that.

That is easy, I'can add it

> >    There are two solutions to this -- one is to decide that if the user provided the formatter function, then just leave it like this -- it was the user's problem how to format. Other -- still check the output from the formatter function, and if it is still numeric (not ready to display text), run the :auto formatter on top.

> I prefer the second option.

This one is a bit more tricky. I can pass on an additional attribute (calling it say `:numberlabels`) in the Axis telling whether the backend is required to treat the result as a unicode number, or leave it as it is. Otherwise it is impossible for the backend to guess whether or not to try this exercise.  Do you thing this is in the style of the Plots design?

daschw:
Ok, I can see the difficulty with "the second option" and I think this is ready to merge, right?

daschw:
Thanks again @fedxa!

fedxa:
Great -- sorry for disappearance, was out of internet reach for the last couple of weeks.

daschw:
🙂 no worries

Issue #1374, opened 2018-01-31
Drawing a span


cstjean:
Is there a Plots.jl [equivalent](https://stackoverflow.com/questions/8270981/in-a-matplotlib-plot-can-i-highlight-specific-x-value-ranges/8271438#8271438) of `axvspan`? I'd like to highlight a region of my plot. `vline!` looked promising at first, but its `width` parameter is in pixel, whereas I would like it in x-coordinates (as in matplotlib)

piever:
As of now, I'm afraid that, to replicate the python example, you'd have to do:

```juilia
plt = plot(0:9)
x1, x2 = 3, 6
y1, y2 = ylims(plt)
plot!(plt, [x1, x2, x2, x1], [y1, y1, y2, y2], color="red", alpha=0.5, seriestype = :shape)
```
Maybe it'd be nice to have a `:rectangle` as well as a `vspan` and `hspan` seriestype, which could be implemented using `:shape`.

cstjean:
Thank you, that works! I used `linealpha=0` to take out the hard edges.

piever:
Glad it worked for you. Better to keep the issue open though as it's a good candidate for a series recipe  potentially called `vspan` (and `hspan`). The recipe could probably already have the transparency default to `0.5` so that the new syntax would be:
```julia
plt = plot(0:9)
plot!([3, 6], color="red", seriestype = :vspan)
```

cstjean:
> Better to keep the issue

Agreed; your solution is perfect for what I need, but the fact that the rectangles are not infinite in the y-axis shows up when panning/zooming-out.

daschw:
You can now use `vspan` and `hspan`. See #1445

PR #1445, opened 2018-03-19
Infinite objects (fix #1422)

### LINKED ISSUE #1297: `abline` shouldn't update limits

### LINKED ISSUE #1422: Plotting a polygon with infinite vertices does not plot the polygon
e.g.
```
plot([1, 2, 2, 1], [Inf, Inf, 6, 5], seriestype=:shape, ylim=(-10,10))
```

It should clip the filled polygon to the viewing area.

daschw:
This supercedes #1316 and implements `hline`, `vline` and `Plots.abline` that update automatically with changing axis limits. This is done via a new seriestype `:straightline` which is characterized by two points. The final points to be connected by the `:straightline` are calculated by the new `Plots.straightline_data(series)` in the backend code based on the axes limits of all series.
Similarly, the function `Plots.shape_data(series)` is introduced and called in the backend code to replace infinite values with values far outside the axes limits.
Below you find all the plots produced by executing the following code snippet:
```julia
using Plots
plot(Shape([-Inf, Inf, Inf, -Inf, -Inf], [-Inf, -Inf, 0, 0, -Inf]))
scatter!(randn(10), randn(10))
hline!([0.5])
Plots.abline!(-0.3, 1)
vline!(-5:2.5:5)
ylims!((-5, 5))
```
infinte_1
infinte_2
infinte_3
infinte_4
infinte_5
infinte_6

**NOTE** that these are not really infinite objects: When zooming out the plot area e.g. in Plotly, these objects do not get updated. Maybe it would make sense to have the lines go 10 or 100 times the axes lenghts beyond the axes limits, so people could zoom for a while before reaching the lines' limits?

### DIFF
--- a/src/Plots.jl
+++ b/src/Plots.jl
@@ -187,6 +187,8 @@ include("output.jl")
 @shorthands sticks
 @shorthands hline
 @shorthands vline
+@shorthands hspan
+@shorthands vspan
 @shorthands ohlc
 @shorthands contour
 @shorthands contourf
--- a/src/axes.jl
+++ b/src/axes.jl
@@ -287,8 +287,8 @@ end
 
 
 function expand_extrema!(ex::Extrema, v::Number)
-    ex.emin = NaNMath.min(v, ex.emin)
-    ex.emax = NaNMath.max(v, ex.emax)
+    ex.emin = isfinite(v) ? min(v, ex.emin) : ex.emin
+    ex.emax = isfinite(v) ? max(v, ex.emax) : ex.emax
     ex
 end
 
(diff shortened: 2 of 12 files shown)


daschw:
Plotly still does not correctly update, because series are added individually - not the entire plot is updated. However, I extended the data 100 times beyond the axes limits and the behaviour with this definitely is an improvement to the status quo. GR, PyPlot, PGFPlots, etc. update correctly.

piever:
See also: #1374

daschw:
I added a `hspan` and `vspan` recipe now:
```julia
hspan(1:10)
plot!(randn(20))
vspan!([5, 11])
```
hvspan

goretkin:
Really cool! Is there a way to draw an infinite ray (only extends infinitely in one direction), or does that require a new series type?

Issue #1447, opened 2018-03-21
@gif in JupyterLab


OliverEvans96:
Currently, `@gif` works fine in the standard notebook, but fails to display in JupyterLab.

See [here](https://github.com/JuliaPlots/Plots.jl/blob/master/src/animation.jl#L96):
```
# write out html to view the gif... note the rand call which is a hack so the image doesn't get cached
function Base.show(io::IO, ::MIME"text/html", agif::AnimatedGif)
    ext = file_extension(agif.filename)
    write(io, if ext == "gif"
        "<img src=\"$(relpath(agif.filename))?$(rand())>\" />"
    elseif ext in ("mov", "mp4")
        "<video controls><source src=\"$(relpath(agif.filename))?$(rand())>\" type=\"video/$ext\"></video>"
    else
        error("Cannot show animation with extension $ext: $agif")
    end)
end
```

It seems to be the `rand() ` trick to prevent caching that's causing the issue.
If `"<img src=\"$(relpath(agif.filename))?$(rand())>\" />"` is changed to `"<img src=\"$(relpath(agif.filename))\" />"`, then the gif displays fine.

I'd love to learn why this is acting differently in JupyterLab. Maybe @gnestor or @ian-r-rose can shed some light on the situation?

Thanks!
Oliver

OliverEvans96:
Also, pinging @tbreloff since you wrote the hack :)

ian-r-rose:
@OliverEvans96 If you use the browser inspector to look at the HTML for the produced output, does anything look wrong? Sometimes the HTML sanitizer can do unexpected things...

OliverEvans96:
Ah, good catch - it is an HTML sanitization issue. Without removing the `rand()`, the `img` tag has `src="http://localhost:8888/files/academic/research/droplet/odes/tmp.gif%3F0.9281699715131861%3E"`, whereas in the classic notebook, the same output refers to `src="tmp.gif?0.9281699715131861>"`.

If I copy and paste `http://localhost:8888/files/academic/research/droplet/odes/tmp.gif%3F0.9281699715131861%3E` into my browser, it gives a 404, but changing `%3F0` to `?` shows the gif.

Is there a recommended fix for this type of thing?

ian-r-rose:
Ah, I now don't think it is a sanitization issue. I think it's related to jupyterlab/jupyterlab#3851.

OliverEvans96:
[...]

PR #1448, opened 2018-03-21
Remove cache-busting from gif animation. Should be done by the frontend.

OliverEvans96:
Solves #1447 on the Plots.jl end.

### DIFF
--- a/src/animation.jl
+++ b/src/animation.jl
@@ -93,13 +93,13 @@ end
 
 
 
-# write out html to view the gif... note the rand call which is a hack so the image doesn't get cached
+# write out html to view the gif
 function Base.show(io::IO, ::MIME"text/html", agif::AnimatedGif)
     ext = file_extension(agif.filename)
     write(io, if ext == "gif"
-        "<img src=\"$(relpath(agif.filename))?$(rand())>\" />"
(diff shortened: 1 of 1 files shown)


daschw:
Thanks a lot for this @OliverEvans96 !
I just checked out this branch and it seems to work in Jupyter notebook too. Can anyone confirm that? In that case I don't see a reason for keeping the old hack.

OliverEvans96:
I'm actually seeing it not working without doing cache busting in Plots.jl. In the classic notebook, if you do somethings like:
```julia
@gif for i=1:3
    plot(rand(4),rand(4))
end
```
it produces the plot as expected.

If you then try to run
```julia
@gif for i=1:3
    heatmap(rand(4,4)
end
```
the first line plot appears, again, not the heatmap. Upon inspecting the HTML, I see `` with no cache busting.

I think that this is an outstanding issue with the classic notebook jupyter/notebook#1369

daschw:
Yes, I can confirm the issue you described and I agree that it's probably best to wait for a fix on the jupyter side.

mkborregaard:
Can this be merged now?

mkborregaard:
Thanks @OliverEvans96

mkborregaard:
We'll see if people have problems

PR #1471, opened 2018-04-04
Remove SVG support for plotly, which is not there

hycakir:
plotly backend does not support SVG output. What it shows is the HTML output of Plotly backend.
This commit makes the request for SVG output to return an error as in PNG.


daschw:
Thanks @hycakir!

PR #1476, opened 2018-04-07
allow to set the theme in PLOTS_DEFAULTS

daschw:


mkborregaard:
Awesome. Should we think about what to do when a PLOTS_DEFAULTS entry collides with one in the theme?

daschw:
The theme entries are overwritten by the `PLOTS_DEFAULTS` entries.
However, once the user calls `theme(thm)`, all `PLOTS_DEFAULTS` entries that are also specified by `thm` are overwritten.
I think that makes sense.

mkborregaard:
Cool.

mkborregaard:
But does that mean that mean that which Plots defaults will be overridden here depends on the order in which they are located in the PLOTS_DEFAULT Dict here? Because of the call to `theme()`?

daschw:
No, first `theme(PLOTS_DEFAULTS[:theme])` is called and then all the other elements from `PLOTS_DEFAULTS` are applied.

mkborregaard:
billede

ehh.

mkborregaard:
Sorry :joy:

daschw:
> However, once the user calls theme(thm), all PLOTS_DEFAULTS entries that are also specified by thm are overwritten.

The question, if this makes sense. (?) But that's not really related to this PR.

mkborregaard:
I think it does

PR #1467, opened 2018-04-02
Segments: vector arguments, line_z and fill_z for GR, PyPlot, Plotly(JS) and PGFPlots

daschw:
**Easter Special** :slightly_smiling_face: 

This allows to pass vectors to the color, alpha, linewidth and linestyle arguments. (https://github.com/JuliaPlots/Plots.jl/pull/1341). Plots then cycles through these vectors.
Furthermore, line_z and fill_z is now completely implemented for paths and shapes on gr, pyplot, plotly(js) and pgfplots. This also means that fillrange now works with PGFPlots. (https://github.com/JuliaPlots/Plots.jl/issues/1216)

A little demonstration:
```julia
using Plots

y1 = cumsum(randn(100))
y2 = randn(20)
shapes = [Shape([x, x+1, x+1, x, x], [0, 0, 1, 1, 0]) for x in 0:2:10]

plot(
    plot(y1, line_z = y1, fill_z = y1, fill = 0, fillalpha = [0.2, 0.5, 1.0]),
    plot(y2, linecolor = [:purple, :green], linealpha = [0.1, 0.7, 1.0], linewidth = 1:10, linestyle = [:solid, :dash, :dot, :dashdot]),
    plot(shapes, fill_z = 1:6),
    plot(shapes, fillcolor = [:purple, :green], fillalpha = 0.3:0.3:1, linestyle = [:dash, :dot], linewidth = 1:6)
    )
```

**GR** works perfectly
segments_gr

**PyPlot** can be quite slow with a lot of different line segments (The issue with the line not shown in the first plot below is already solved)
segments_pyplot

**PGFPlots** has (unrelated) colorbar issues
segments_pgfplots

**Plotly(JS)** also has some unrelated colorbar/legend issues with subplots (unfortunately I could not save as png).

One nice bonus is that with fill_z working for shapes now we can implement a heatmap recipe for these backends, which could fix https://github.com/JuliaPlots/Plots.jl/issues/1462 or https://github.com/JuliaPlots/Plots.jl/issues/1287 and provide heatmap for pgfplots.

Issue #1287, opened 2017-12-06
feature request: heatmap with varying alpha


kleinschmidt:
I'd like to plot a matrix as a heatmap, using another matrix as an alpha mask (to represent weights).  When I try that I get the following error:

```
julia> heatmap(X, weights)
MethodError: no method matching plot_color(::ColorTypes.RGBA{Float64}, ::Array{Float64,1})
Closest candidates are:
  plot_color(::ColorTypes.Colorant) at /home/dave/.julia/v0.6/PlotUtils/src/colors.jl:9
  plot_color(::ColorTypes.Colorant, ::Number) at /home/dave/.julia/v0.6/PlotUtils/src/colors.jl:19
  plot_color(::Any, ::Void) at /home/dave/.julia/v0.6/PlotUtils/src/colors.jl:15
  ...

Stacktrace:
 [1] getSeriesRGBColor(::Symbol, ::Array{Float64,1}, ::Plots.Subplot{Plots.PlotlyJSBackend}, ::Int64) at /home/dave/.julia/v0.6/Plots/src/args.jl:1298
 [2] _add_defaults!(::Dict{Symbol,Any}, ::Plots.Plot{Plots.PlotlyJSBackend}, ::Plots.Subplot{Plots.PlotlyJSBackend}, ::Int64) at /home/dave/.julia/v0.6/Plots/src/args.jl:1350
 [3] _plot!(::Plots.Plot{Plots.PlotlyJSBackend}, ::Dict{Symbol,Any}, ::Tuple{Array{Int64,2}}) at /home/dave/.julia/v0.6/Plots/src/plot.jl:223
 [4] (::RecipesBase.#kw##plot)(::Array{Any,1}, ::RecipesBase.#plot, ::Array{Int64,2}) at ./<missing>:0
 [5] #heatmap#461(::Array{Any,1}, ::Function, ::Array{Int64,2}, ::Vararg{Array{Int64,2},N} where N) at /home/dave/.julia/v0.6/RecipesBase/src/RecipesBase.jl:381
 [6] (::Plots.#kw##heatmap)(::Array{Any,1}, ::Plots.#heatmap, ::Array{Int64,2}, ::Vararg{Array{Int64,2},N} where N) at ./<missing>:0
 [7] include_string(::String, ::String) at ./loading.jl:515
```

It looks like the colors are being generated with an alpha channel already in the heatmap recipe.

kleinschmidt:
[...]

PR #1477, opened 2018-04-07
RFC: WIP: Heatmap recipe

daschw:
This is a follow-up on #1467 and implements a heatmap recipe.
For now I've kept the native heatmap implementation in the backends and created a new `:heatmapr` seriestype to compare the results and decide on which backend we want to choose which implementation.

### Comparison

Here's the output of
```julia
using Plots
y = (1:21) .* (1:21)'
plot(
    plot(y, seriestype = :heatmap, title = "Native Heatmap"),
    plot(y, seriestype = :heatmapr, title = "Heatmap Recipe"),
    plot(y, seriestype = :heatmap, scale = :log10),
    plot(y, seriestype = :heatmapr, scale = :log10),
    lw = 1,
    layout = (2, 2),
    )
```
for several backends:

**GR**
heatmap_recipe_gr

This would fix the logscale problem for heatmaps on GR.

**PyPlot**
heatmap_recipe_pyplot

Almost identical output

**PGFPlots**
heatmap_recipe_pgfplots

Finally, a heatmap implementation

**Plotly(JS)**
heatmap_recipe_plotly

There's a regression for Plotl(JS) in the interactive view:
Native:
heatmap_plotly_interactive_old

Recipe:
heatmap_plotly_interactive_new

### More control

Furthermore, this would allow us something like (#1287):
```julia
plot(y, seriestype = :heatmapr,  fα = 0.2:0.2:0.8)
```
(I'm still thinking about allowing to pass a matrix as `fillalpha`, if that's possible)
Here's the output on different backends:

**GR**
heatmap_advanced_gr

**PyPlot**
heatmap_advanced_pyplot

**PGFPlots**
heatmap_advanced_pgfplots

**Plotly(JS)**
heatmap_advanced_plotly

### Benchmarks

There comes the downside:
Here's the output of
```julia
using BenchmarkTools
@benchmark display(plot(y, seriestype = :heatmap))
@benchmark display(plot(y, seriestype = :heatmapr))
```
for several backends:

**GR**
Native:
```julia
BenchmarkTools.Trial:
  memory estimate:  5.55 MiB
  allocs estimate:  221728
  --------------
  minimum time:     72.282 ms (0.00% GC)
  median time:      142.958 ms (0.00% GC)
  mean time:        150.112 ms (1.54% GC)
  maximum time:     217.139 ms (0.00% GC)
  --------------
  samples:          34
  evals/sample:     1
```
Recipe:
```julia
BenchmarkTools.Trial:
  memory estimate:  8.26 MiB
  allocs estimate:  306245
  --------------
  minimum time:     61.424 ms (0.00% GC)
  median time:      82.728 ms (0.00% GC)
  mean time:        89.293 ms (3.18% GC)
  maximum time:     214.373 ms (2.98% GC)
  --------------
  samples:          56
  evals/sample:     1
```

**PyPlot**
Native:
```julia
BenchmarkTools.Trial:
  memory estimate:  2.44 MiB
  allocs estimate:  88155
  --------------
  minimum time:     407.102 ms (0.00% GC)
  median time:      503.468 ms (0.00% GC)
  mean time:        517.391 ms (0.17% GC)
  maximum time:     694.881 ms (0.00% GC)
  --------------
  samples:          10
  evals/sample:     1
```
Recipe:
```julia
BenchmarkTools.Trial:
  memory estimate:  8.46 MiB
  allocs estimate:  268626
  --------------
  minimum time:     1.318 s (0.00% GC)
  median time:      1.376 s (0.00% GC)
  mean time:        1.377 s (0.25% GC)
  maximum time:     1.439 s (0.95% GC)
  --------------
  samples:          4
  evals/sample:     1
```

**PGFPlots**
Recipe:
```julia
BenchmarkTools.Trial:
  memory estimate:  14.83 MiB
  allocs estimate:  307429
  --------------
  minimum time:     10.571 s (0.00% GC)
  median time:      10.571 s (0.00% GC)
  mean time:        10.571 s (0.00% GC)
  maximum time:     10.571 s (0.00% GC)
  --------------
  samples:          1
  evals/sample:     1
```

**PlotlyJS**
Native:
```julia
BenchmarkTools.Trial:
  memory estimate:  4.43 MiB
  allocs estimate:  162473
  --------------
  minimum time:     145.293 ms (0.00% GC)
  median time:      194.977 ms (0.00% GC)
  mean time:        195.806 ms (1.22% GC)
  maximum time:     225.369 ms (5.02% GC)
  --------------
  samples:          26
  evals/sample:     1
```
Recipe:
```julia
BenchmarkTools.Trial:
  memory estimate:  105.84 MiB
  allocs estimate:  2623521
  --------------
  minimum time:     8.184 s (0.49% GC)
  median time:      8.184 s (0.49% GC)
  mean time:        8.184 s (0.49% GC)
  maximum time:     8.184 s (0.49% GC)
  --------------
  samples:          1
  evals/sample:     1
```

**Plotly**
Native:
```julia
BenchmarkTools.Trial:
  memory estimate:  3.76 MiB
  allocs estimate:  148744
  --------------
  minimum time:     222.058 ms (0.00% GC)
  median time:      318.871 ms (0.00% GC)
  mean time:        317.100 ms (0.93% GC)
  maximum time:     384.451 ms (0.00% GC)
  --------------
  samples:          16
  evals/sample:     1
```
Recipe:
```julia
BenchmarkTools.Trial:
  memory estimate:  8.97 MiB
  allocs estimate:  277455
  --------------
  minimum time:     256.525 ms (0.00% GC)
  median time:      393.302 ms (0.00% GC)
  mean time:        382.719 ms (1.61% GC)
  maximum time:     425.111 ms (0.00% GC)
  --------------
  samples:          14
  evals/sample:     1
```
### Conclusion

The recipe is really slow for PGFPlots and PlotlyJS. For all other backends there's still a regression.

I'd vote for implementing the recipe for GR (quite fast) and PGFPlots (no alternative implementation) and keeping the native implementation for Plotly(JS) (regression in interactivity and extremly slow for PlotlyJS). I'm not sure about PyPlot, but I think having control about individual alphas is not as desirable as being almost three times as fast in general. Furthermore, for bigger matrices this regression in speed might be more significant.
An alternative would be to keep the native implementation and the `:heatmapr` recipe (maybe with a better name) and let the user decide. This would give more control, but may be to confusing.

I'm really excited to hear about your opinions!

### DIFF
--- a/src/Plots.jl
+++ b/src/Plots.jl
@@ -183,6 +183,7 @@ include("output.jl")
 @shorthands histogram2d
 @shorthands density
 @shorthands heatmap
+@shorthands plots_heatmap
 @shorthands hexbin
 @shorthands sticks
 @shorthands hline
--- a/src/backends/plotly.jl
+++ b/src/backends/plotly.jl
@@ -669,6 +669,8 @@ function plotly_series_shapes(plt::Plot, series::Series)
         d_outs[i] = d_out
     end
     if series[:fill_z] != nothing
+        push!(d_outs, plotly_colorbar_hack(series, base_d, :fill))
+    elseif series[:line_z] != nothing
         push!(d_outs, plotly_colorbar_hack(series, base_d, :line))
     end
     d_outs
--- a/src/recipes.jl
+++ b/src/recipes.jl
@@ -390,8 +390,8 @@ end
 @deps bar shape
 
 # ---------------------------------------------------------------------------
-# Heatmap
-@recipe function f(::Type{Val{:heatmapr}}, x, y, z)
+# Plots Heatmap
+@recipe function f(::Type{Val{:plots_heatmap}}, x, y, z)
     xe, ye = heatmap_edges(x), heatmap_edges(y)
     m, n = size(z.surf)
     x_pts, y_pts = fill(NaN, 6 * m * n), fill(NaN, 6 * m * n)
@@ -415,7 +415,7 @@ end
     label := ""
     ()
 end
-@deps heatmapr shape
+@deps plots_heatmap shape
 
 # ---------------------------------------------------------------------------
 # Histograms
--- a/src/recipes.jl
+++ b/src/recipes.jl
@@ -407,8 +407,7 @@ end
     end
     ensure_gradient!(plotattributes, :fillcolor, :fillalpha)
     fill_z := fz
-    linewidth := 0
-    linecolor := invisible()
+    line_z --> fz
     x := x_pts
     y := y_pts
     z := nothing
--- a/src/recipes.jl
+++ b/src/recipes.jl
@@ -407,7 +407,7 @@ end
     end
     ensure_gradient!(plotattributes, :fillcolor, :fillalpha)
     fill_z := fz
-    line_z --> fz
+    line_z := fz
     x := x_pts
     y := y_pts
     z := nothing
--- a/src/recipes.jl
+++ b/src/recipes.jl
@@ -389,6 +389,34 @@ end
 end
 @deps bar shape
 
+# ---------------------------------------------------------------------------
+# Heatmap
+@recipe function f(::Type{Val{:heatmapr}}, x, y, z)
+    xe, ye = heatmap_edges(x), heatmap_edges(y)
+    m, n = size(z.surf)
+    x_pts, y_pts = fill(NaN, 6 * m * n), fill(NaN, 6 * m * n)
+    fz = zeros(m * n)
+    for i in 1:m # y
+        for j in 1:n # x
+            k = (j - 1) * m + i
+            inds = (6 * (k - 1) + 1):(6 * k - 1)
+            x_pts[inds] .= [xe[j], xe[j + 1], xe[j + 1], xe[j], xe[j]]
+            y_pts[inds] .= [ye[i], ye[i], ye[i + 1], ye[i + 1], ye[i]]
+            fz[k] = z.surf[i, j]
+        end
+    end
+    ensure_gradient!(plotattributes, :fillcolor, :fillalpha)
+    fill_z := fz
+    linewidth := 0
+    linecolor := invisible()
+    x := x_pts
+    y := y_pts
+    z := nothing
+    seriestype := :shape
+    label := ""
+    ()
+end
+@deps heatmapr shape
 
 # ---------------------------------------------------------------------------
 # Histograms
--- a/src/series.jl
+++ b/src/series.jl
@@ -8,7 +8,7 @@
 
 const FuncOrFuncs{F} = Union{F, Vector{F}, Matrix{F}}
 
-all3D(d::KW) = trueOrAllTrue(st -> st in (:contour, :contourf, :heatmap, :surface, :wireframe, :contour3d, :image, :heatmapr), get(d, :seriestype, :none))
+all3D(d::KW) = trueOrAllTrue(st -> st in (:contour, :contourf, :heatmap, :surface, :wireframe, :contour3d, :image, :plots_heatmap), get(d, :seriestype, :none))
 
 # unknown
 convertToAnyVector(x, d::KW) = error("No user recipe defined for $(typeof(x))")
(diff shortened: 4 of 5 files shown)


jheinen:
I will try to implement a "native" log-scaled heatmap for GR ASAP

daschw:
In my opinion, using the recipe for GR is fine and this would already fix the log-scale issue.

daschw:
@mkborregaard @piever @apalugniok What's your opinion on this?

piever:
Personally, I really like the combination of an efficient backend interfaced with Plots at a low-level, recipes being handled in Plots as much as possible. I'd say to switch to recipe in the GR and PyPlot case. PyPlot is not the correct choice anyway when very high performance is crucial. I also think we should be happy to add new features (say the transparency here) only for GR/PyPlot as they are the ones where this low-level interface is possible and we can add new features by only coding them once in the Plots code.

For Plotly and PGFPlots it's always hard to come up with a good solution  because interfacing with them at a low-level may be not ideal performance-wise (and also because of the hovering functionality in Plotly). I think we have to accept that those are the cases where the Plots interface is mainly a syntactic convenience and it's difficult to offer a lot of new functionality.

daschw:
Agreed, I'd still include the PGFPlots recipe, because there we have no other heatmap implementation.

mkborregaard:
I like the idea of having a recipe, and I think @piever 's comments make a lot of sense. One note on pgfplots, though - pgfplots have native `surface` plots, which are simply not yet implemented in the Plots backend, but should be easy to do. And a canonical way of making a heatmap in pgfplots is simply to draw a flat `surface` seen directly from above!
So it would be easy to add support for native surface and heatmap in pgfplots without the recipe.

daschw:
Great, thanks for all your comments!

apalugniok:
Well those numbers don't look great... 😆. I don't see a big issue with keeping the recipe as something like `alpha_heatmap`. It would be out of the way for normal usage and if someone was desperate for a varying alpha heatmap it would be available.

mkborregaard:
I plot large heatmaps all the time, at least.

mkborregaard:
+1 for plots_heatmap

piever:
@daschw do you have any intuition as to why the Plots implementation should be so much slower than native `heatmap`?

daschw:
In the recipe, a vector of coordinates for the individual rectangles, seperated by `NaN`s, is created. In the backend code this series is then split into separate segments for each rectangle, which is then plotted basically as a filled line plot (or `patch` in PyPlot). I guess, on PyPlot, one, somehow optimized, heatmap call is just more efficient than thousands of plot calls ... on GR it's (currently) plotting an image of RGBs vs. thousands of filled line plots.

piever:
I'm still trying to wrap my head around what is the best level at which to interact with the backend. I think `heatmap` is too high-level, but the fill line does not seem sufficient. Maybe an intermediate way would be to have the interaction at the level of "plot `Array` of RGBA values" and Plots would take care of computing that `Array` from the data and the color gradient + `alpha` parameters. Do you know which backends offer native support to plot `Array` of RGBA?

piever:
I see what you mean as we are not plotting squares of the same size in the logscale case. Then I agree, it has to be something more sophisticated than that (but not by much). For example it should be an `Array{RGBA}` of size `(M, N)` and two `Vector{Float64}` of length `M` and `N` respectively describing the sizes of the rectangles. Do you think the various backends have a simple and efficient way (more efficient than plotting `MxN` shapes) of plotting this thing? We could also look at what @jheinen uses to implement the logscale `heatmap` and hook into the same machinery.

jheinen:
FYI: The **native** log-scaled heatmap for GR is WIP:
```
using Plots
y = (1:21) .* (1:21)'
heatmap(y,scale = :log10)
```
screen shot 2018-04-11 at 15 45 53

jheinen:
```julia
y = (1:201) .* (1:201)'
@time heatmap(y,scale = :log10)
0.003181 seconds (42.02 k allocations: 1.079 MiB)
```

jheinen:
Does someone have a typical use case for varying alpha? Constant alpha for all `heatmap` cells would be easy to implement. Just wondering, how important that would be ....

daschw:
OK, I changed the recipe name to `:plots_heatmap` now and left the current backend implementations untouched.

Issue #1470, opened 2018-04-04
Surface using hdf5 backend


goldnd:
It appears that surfaces don't work with the hdf5 backend.  This example fails:
```julia
using HDF5
using Plots
hdf5()
x = y = linspace(-5, 5, 40)
zs = zeros(0,40)
n = 100
i = 1
f(x,y) = sin(x + 10sin(i)) + cos(y)

# create a plot with 3 subplots and a custom layout
l = @layout [a{0.7w} b; c{0.2h}]
p = plot(x, y, f, st = [:surface, :contourf], layout=l)

Plots.hdf5plot_write(p, "temp.hdf5");
```
with this message:
ERROR: MethodError: no method matching write(::HDF5.HDF5Group, ::String, ::Plots.Surface{Array{Float64,2}})
Closest candidates are:
  write(::Union{HDF5.HDF5File, HDF5.HDF5Group}, ::String, ::Any, ::String, ::Any, ::Any...) at /Users/dect/.julia/v0.6/HDF5/src/HDF5.jl:1576
  write(::AbstractString, ::Any...) at io.jl:152
  write(::IO, ::Any...) at io.jl:286

ma-laforge:
Thanks for letting me know.  I will try to take a look at this in the following few days.

ma-laforge:
@goldnd : Please give PR https://github.com/JuliaPlots/Plots.jl/pull/1480 a try with your configuration.

See comments in PR for more details.

PR #1480, opened 2018-04-09
Add support for Surface plots.

ma-laforge:
Should work... but unable to verify.

The example provided in https://github.com/JuliaPlots/Plots.jl/issues/1470 does not run with the current version of GR.  I also cannot get PyPlot to work at the moment.

I have confirmed that I can re-load the plot by using:
```
#[...] Code from Issue 1470
Plots.hdf5plot_write(p, "temp.hdf5");

#"Replay" saved .hdf5 file
inspectdr() #Does not actually support Surface plots, but...
#gr() #Current version does not appear to plot provided surface example.
pread = Plots.hdf5plot_read("temp.hdf5") #Seems to return valid object
```
The added functions are very unlikely to break any current code.

@goldnd: Do you mind testing out the solution with your configuration?

### DIFF
--- a/src/backends/hdf5.jl
+++ b/src/backends/hdf5.jl
@@ -167,6 +167,7 @@ function _initialize_backend(::HDF5Backend)
 #                "PLOTTEXT" => PlotText,
                 "COLORGRADIENT" => ColorGradient,
                 "AXIS" => Axis,
+                "SURFACE" => Surface,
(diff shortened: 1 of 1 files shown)


ma-laforge:
Fixes issue https://github.com/JuliaPlots/Plots.jl/issues/1470.

Please merge when appropriate.

PR #1501, opened 2018-04-27
replace Int64 by Int

daschw:
@BoundaryValueProblems could you check if this would fix the 32Bit issues?

### DIFF
--- a/src/series.jl
+++ b/src/series.jl
@@ -627,7 +627,7 @@ group_as_matrix(t) = false
     else
         g = args[1]
         if length(g.args) == 1
-            x = zeros(Int64, lengthGroup)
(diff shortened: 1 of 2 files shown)


BoundaryValueProblems:
@daschw , it worked for both gr and plotlyjs backends with @jheinen 's MWE as well as my original complicated plotting function! However, we need to be careful. If I replace ```Int``` by ```Integer``` in line 261-273 of ```utils.jl```, the type mismatch occurs again, and does not work. So, somehow, we cannot replace ```Int``` by ```Integers``` for those lines involving ```_cycle``` functions.

daschw:
OK, thanks for checking!

PR #1508, opened 2018-05-04
Replace Nullable{T} with Union{T, Nothing}

mkborregaard:
As Nullable is deprecated on 0.7


daschw:
I think that's already fixed by some other update-to-0.7 PR

PR #1510, opened 2018-05-05
Make series of shapes and segments toggle together in plotly[js] (fixes #1488)

### LINKED ISSUE #1488: Series with gaps (segments) don't toggle correctly when legend is clicked in plotly and plotlyjs
Usually, a series' visibility can be toggled by clicking on its entry in the legend with plotly and plotlyjs backends. When a series consists of segments (gaps indicated by `NaN`), only the first segment's visibility is toggled. The same holds for `Shape` series, which I think are internally converted to line segments.

Example:
```julia
using Plots
plotly() # or plotlyjs()
plot([1, 2, NaN, 3, 4], [1, 2, NaN, 3, 4])
```

This is not the case if the plot is directly created in plots.js:
https://codepen.io/anon/pen/NMKbyp?editors=0010#0

chfin:
As series of segments or shapes generate one trace per segment/shape in the plotly (and plotlyjs) backend, only the first trace is shown in the legend to avoid duplicate entries. As a result, the visibility of only the first trace is toggled when the corresponding entry in the legend is clicked.

This can be fixed by adding all traces to the same `legendgroup`. This attribute is otherwise unused by Plots.jl and AFAIK cannot be used manually by the user. However, plotly.js adds a gap between legend groups by default, which can be avoided by setting `layout.legend.tracegroupgap` to 0 by default (this attribute is similarly unused).

I ran the plotlyjs backend tests once for master (to update the reference images) and once for this fix, and there seem to be no differences (after setting `tracegroupgap` to 0).

### DIFF
--- a/src/backends/plotly.jl
+++ b/src/backends/plotly.jl
@@ -388,6 +388,7 @@ function plotly_layout(plt::Plot)
                 :bgcolor  => rgba_string(sp[:background_color_legend]),
                 :bordercolor => rgba_string(sp[:foreground_color_legend]),
                 :font     => plotly_font(legendfont(sp)),
+                :tracegroupgap => 0,
                 :x => xpos,
                 :y => ypos
             )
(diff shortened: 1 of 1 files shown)


mkborregaard:
The conflict needs to be resolved of course - but I don't see how that arises?

daschw:
I introduced it in #1511

daschw:
Thanks @chfin!

Issue #1502, opened 2018-04-27
savefig(foo, "bar.svg") works, svg(foo, "bar") fails


t37w:
The behavior described in the subject, is this the intention?
Calling 'using Plots; gr();' ... 'foo=plot(x,y,z); svg(foo,"Try_save_small_svg_v005_plot(x,y,z))")'  fails with 'ERROR: UndefVarError: svg not defined'.
Looking at the sources, I assume it's just because 'svg' is not exported in 'Plots.jl'.
Seems similar for the other formats one can use as output.

PR #1521, opened 2018-05-10
Enable the output shortcuts 'svg', 'pdf', 'ps', 'eps'.

t37w:
Hello,
to solve issue #1502 , I extended the export list of Plots.
With
```
using Plots; gr()
x=rand(101); y=rand(101)+10; z=rand(101)+100;
bar=plot3d(x,y,z)
svg("3D_test_shortcut_svg")
pdf("3D_test_shortcut_pdf")
ps("3D_test_shortcut_ps")
```
I tested 3 of them, assuming eps works with another backend as it is similar.
Verified with 'xpdf' and 'display'.

Cheers,
Thomas

### DIFF
--- a/src/Plots.jl
+++ b/src/Plots.jl
@@ -62,8 +62,13 @@ export
     ylims,
     zlims,
 
+    # output.jl
     savefig,
     png,
+    svg,
+    pdf,
+    ps,
+    eps,
     gui,
     inline,
     closeall,


mkborregaard:
Another possibility is to change the documentation. IMHO these names are so common outside the Plots ecosystem that we should not export them.

t37w:
Good point.  Especially pdf is used in in statistics.  I prefer a specific command to save to a format instead of recognition via extension. Thus instead of exporting svg, pdf, ... using save_svg, save_pdf, or similar ... for export is cheap to define in julia. The plot operation itself and the write to disk is much more expensive.

t37w:
I just want to know whether longer function names would be an option.
But this pull request can be closed as the very short names a no good choice.

daschw:
To be honest, I don't really see the point of longer-named functions if we already have `savefig`.

t37w:
Savefig depends on the extension of the filename.  A function saves in the related format and adds the extension.

daschw:
I get the difference, I just don't get why typing
```julia
save_svg("path/to/my/file")
```
would be so much more convenient than typing
```julia
savefig("path/to/my/file.svg")
```
to justify exporting 3-5 extra functions. But if it's really what people want, I won't be in the way.

mkborregaard:
I agree.

t37w:
It is more or less what one may prefer.  And in the first case you have the format info in the beginning of the line and recognize it faster while the latter has the format at the end.
Adding such additional functions is very cheap in julia.
As this discussion is stored here, the info is not lost and easy to repeat.
So I suggest to put this additional functions on hold.

Issue #1520, opened 2018-05-10
seriescolor in recipe is ignored 


yakir12:
@mkborregaard found out that the following doesn't plot in red (this is from my question on [Discourse](https://discourse.julialang.org/t/consistent-plotting-attributes/10822?u=yakir12)):
```julia
@recipe function f(::Type{Val{:home}}, x,y,z)
       seriescolor := :red
       markershape := :star5
       seriestype := :scatter
       x,y
  end

plot(rand(5), st = :home)
```

mkborregaard:
I also tried the more specific `markercolor`, which fails with error message:
```
Error showing value of type Plots.Plot{Plots.GRBackend}:
ERROR: MethodError: no method matching alpha(::Symbol)
Closest candidates are:
  alpha(::ColorTypes.AGray32) at /Users/michael/.julia/v0.6/ColorTypes/src/traits.jl:11
  alpha(::ColorTypes.ARGB32) at /Users/michael/.julia/v0.6/ColorTypes/src/traits.jl:10
  alpha(::ColorTypes.RGB24) at /Users/michael/.julia/v0.6/ColorTypes/src/traits.jl:9
  ...
```

daschw:
The issue is the order in the pipeline in https://github.com/JuliaPlots/Plots.jl/blob/master/src/plot.jl#L219-L236: At first the defaults are applied including matching `:auto` linecolors etc. with the corresponding seriescolors. Then the seriesrecipe is applied. If this changes e.g. the seriescolor, linecolor still holds the default seriescolor from before.
We could try reversing the order, but it might not be that straightforward and I'm not sure if this could cause other issues.
Otherwise, series recipes really should specify the most specific argument.
@mkborregaard's example from above should be fixed by https://github.com/JuliaPlots/Plots.jl/pull/1522
However, there's still an issue, for instance, with
```julia
gr()
@recipe function f(::Type{Val{:test}}, x,y,z)
       markercolor := 3
       seriestype := :scatter
       x,y,z
end
plot(1:5, st = :test)
```
`markercolor = 3` would be replaced by the 3rd color of the default palette in `_add_defaults!`, which is however called before the series recipe is processed. Hence gr.jl and finally `alpha(c)` gets `markercolor = 3`, which it cannot handle:
```julia
plot(1:5, st = :test)
Error showing value of type Plots.Plot{Plots.GRBackend}:
ERROR: MethodError: no method matching plot_color(::Int64)
Closest candidates are:
[...]

PR #1523, opened 2018-05-11
update series attributes after processing series recipes fix (#1520)

daschw:
see discussion in #1520.
This reverses the order in the pipeline: first series recipe processing, then updating of markercolor etc. via seriescolor.
I really hope this does not mess anything else up. It would be great if someone could test or review this.

### DIFF
--- a/src/args.jl
+++ b/src/args.jl
@@ -1473,26 +1473,19 @@ function _replace_linewidth(d::KW)
 end
 
 function _add_defaults!(d::KW, plt::Plot, sp::Subplot, commandIndex::Int)
-    pkg = plt.backend
-    globalIndex = d[:series_plotindex]
-
     # add default values to our dictionary, being careful not to delete what we just added!
     for (k,v) in _series_defaults
         slice_arg!(d, d, k, v, commandIndex, false)
     end
 
-    # this is how many series belong to this subplot
-    # plotIndex = count(series -> series.d[:subplot] === sp && series.d[:primary], plt.series_list)
-    plotIndex = 0
-    for series in sp.series_list
-        if series[:primary]
-            plotIndex += 1
-        end
-    end
-    # plotIndex = count(series -> series[:primary], sp.series_list)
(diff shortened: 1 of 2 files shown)


daschw:
Thanks @mkborregaard! Just to provide some context:

`_add_defaults!` added all default values for attributes not specified by the user to `d` and "post-processed" some series args, e.g. getting markercolor from seriescolor if not defined. This function is called before series recipes are processed.

I removed all the arg processing part from `_add_defaults!` and put them in a new function `_update_series_attributes!` which is called in the recursive function `_process_seriesrecipe` after the series recipes are processed.

src/args.jl
```diff
@@ -1562,3 +1555,19 @@ function _add_defaults!(d::KW, plt::Plot, sp::Subplot, commandIndex::Int)
...
     d
 end
+
+function _series_index(d, sp)
+    idx = 0
+    for series in series_list(sp)
+        if series[:primary]
+            idx += 1
+        end
+        if series == d
```
mkborregaard: Not entirely sure what this does?

mkborregaard:
I think that edit makes a lot of sense.

src/args.jl
daschw: This determines the "primary index" of the series in the subplot.
A short explanation, how it ended up here:
This piece of code was actually part of `_add_defaults!` and at first I did not plan to remove the series processing of this function, but I also needed the series index in `_update_series_attributes!`. So I put this code in a separate function `series_index`, that would be called by both `_add_defaults!` and `_update_series_attributes!`. It turned out that series processing in `_add_defaults!` was no longer necessary, so I removed all that. ... Hence, the separate `series_index` function.

daschw:
Thanks for taking a look @mkborregaard!

mkborregaard:
Awesome work!

src/args.jl
mkborregaard: Missing the :mind_blown:  reaction

PR #1528, opened 2018-05-16
Contour plots with x,y in grid (i.e. multi-dimensional array) form

jdeldre:
These small changes allow for x, y in grid form, which is allowed by the PyPlot backend for making contour plots on deformed grids. (Might need some additional checks for other backends, which might not allow this.)


daschw:
Thanks @jdeldre!

PR #1536, opened 2018-05-21
restructure some show methods

### LINKED ISSUE #1529: Plots shouldn't lie about output data
If I ask for `text/html` data to be written to my buffer I better get `text/html` data back.

https://github.com/JuliaPlots/Plots.jl/blob/12e5937a91da9595c5b9ac070d42698047d999f1/src/output.jl#L179-L196

fredrikekre:
First commit forwards `Base.showable` to `_show` instead, since that is what Plots uses.
Second commit removes some `show` method that don't return data in the requested mime type (fix #1529).
Third commit remove `Base.show` methods for plotly and plotlyjs and thus unifies all Plot types show by going through the same `Base.show` -> `prepare_output` -> `Plots._show`. path.

### DIFF
--- a/src/output.jl
+++ b/src/output.jl
@@ -195,10 +195,11 @@ function Base.show(io::IO, ::MIME"text/html", plt::Plot)
     end
 end
 
-function _show(io::IO, m, plt::Plot{B}) where B
-    # Base.show_backtrace(STDOUT, backtrace())
-    warn("_show is not defined for this backend. m=", string(m))
+# delegate mimewritable (showable on julia 0.7) to _show instead
+function Base.mimewritable(m::M, plt::P) where {M<:MIME, P<:Plot}
+    return method_exists(_show, Tuple{IO, M, P})
 end
+
 function _display(plt::Plot)
     warn("_display is not defined for this backend.")
 end
--- a/src/output.jl
+++ b/src/output.jl
@@ -157,17 +157,6 @@ end
 
 # ---------------------------------------------------------
 
-const _mimeformats = Dict(
-    "application/eps"         => "eps",
-    "image/eps"               => "eps",
-    "application/pdf"         => "pdf",
-    "image/png"               => "png",
-    "application/postscript"  => "ps",
-    "image/svg+xml"           => "svg",
-    "text/plain"              => "txt",
-    "application/x-tex"       => "tex",
-)
-
 const _best_html_output_type = KW(
     :pyplot => :png,
     :unicodeplots => :txt,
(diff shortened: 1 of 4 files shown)


src/output.jl
```diff
@@ -304,11 +290,6 @@ end
             out
         end
 
-        # default text/plain passes to html... handles Interact issues
```
fredrikekre: Unclear what issue this fixed and if it is still needed, introduced in 16a2e34e950a020746cddf9ed964275a9f131ea3

daschw:
OK, so this messes up plotting from Juno for all backends except plotly and plotlyjs.

fredrikekre:
Yea, I think this PR is the faulty one. I will comment again when I have pushed a fix.

fredrikekre:
> except plotly and plotlyjs.

Wait, what? Those work on this branch?

daschw:
Well, plotly always works, because it does not use Juno's plot pane, but opens a browser window.
Let me check plotlyjs again ...

daschw:
... but all other backends (except for unicodeplots) work in IJulia ...

fredrikekre:
I can't test, but I don't think so. #1537 should be good though (but feel free to test it properly), and then I can rebase ans test this branch again.

daschw:
I can test it today in the evening on my linux machine. Somehow I'm struggling with building MbedTLS and IJulia on Windows right now.

daschw:
In Atom for unicodeplots it produces something weird in the Plot pane. Should we make it use the Console/REPL by default?
All the other backends work fine in REPL and Atom. Can someone else please test this PR on IJulia? @mkborregaard, @piever, @apalugniok ? IJulia is not working anywhere for me.

fredrikekre:
> In Atom for unicodeplots it produces something weird in the Plot pane

Yea, apparently we try to show colors with ANSI color codes, which does not quite work. It would be fixed by
```diff
diff --git a/src/output.jl b/src/output.jl
index cebfe92..52d3a30 100644
--- a/src/output.jl
+++ b/src/output.jl
@@ -178,7 +178,7 @@ function _show(io::IO, ::MIME"text/html", plt::Plot)
         # info("writing svg to html output")
         show(io, MIME("image/svg+xml"), plt)
     elseif output_type == :txt
-        show(io, MIME("text/plain"), plt)
+        show(IOContext(io; :color => false), MIME("text/plain"), plt)
     else
         error("only png or svg allowed. got: $(repr(output_type))")
     end
```
but it on Julia v0.6 `print_with_color` (which UnicodePlots uses) ignores that. Perhaps we can revisit this in Julia v0.7.

mkborregaard:
False alarm on working in Atom, it worked exactly once.
There's a 0.7 branch already which should be merged into master when we make that jump if we feel like being proactive :-) Time for dropping 0.6 support on master will soon be up anyways I guess.

fredrikekre:
> False alarm on working in Atom, it worked exactly once.

What do you mean? This branch doesn't work for you in Atom?

mkborregaard:
Sorry, I was unclear. Plots doesn't work well for me in Atom, for some reason. I checked out this branch, and suddenly everything worked. Only once, though. This PR is not responsible for the problem.

daschw:
Also a big thank you from my side for your amazing PRs @fredrikekre and sorry that I kept you waiting so long for https://github.com/JuliaPlots/Plots.jl/pull/1537!

daschw:
Thanks for testing on IJulia @mkborregaard !

fredrikekre:
No worries, I don't use Plots so doesn't matter to me :smile:

PR #1537, opened 2018-05-21
only try the pdf -> png convert sometimes

fredrikekre:
Only try to convert `pdf` -> `png` if the backend supports writing to pdf in the first place.


fredrikekre:
This branch should be ok to merge as well, perhaps you can test when you get some time over @daschw ?

mkborregaard:
Should we just merge this?

daschw:
Oh sorry, I completely overlooked this. I just tested now and it's fine from REPL and Atom, my IJulia install is currently broken. But I think this PR is OK! Thanks @fredrikekre !

PR #1544, opened 2018-05-30
PyPlot: fix bg_legend = invisible()

daschw:
Setting `bg_legend = invisible()` on pyplot did not inherit the subplots background color but set the color of the legend frame to be really `invisible()`. When using a dark theme in Atom for example this shows a dark legend box with the default Plots theme, which is not expected.

This PR fixes the issue by explicitly setting the framealpha of the legend frame in matplotlib as the alpha of the legend background color. This inherits the subplots background color. The downside, however, is that you cannot set the facealpha and the edgealpha separately in matplotlib. Thus, setting `bg_legend = invisible()` also removes the legend frame.

I still think this is an improvement to the current behaviour, but other opinions are very welcome.

Thanks to @davidzentlermunro for reporting on Gitter!


mkborregaard:
I think it is an improvement as well 👍

davidzentlermunro:
This looks great - thanks!

Issue #833, opened 2017-05-16
Axis failure with log `xscale`


mkborregaard:
If I try
```julia
using Plots
histogram(abs(randn(100)), xscale = :log10)
```
I get
```
INFO: binning = auto
Error showing value of type Plots.Plot{Plots.GRBackend}:
ERROR: DomainError:
 in nan_dom_err at ./math.jl:196 [inlined]
 in log10(::Float64) at ./math.jl:202
 in (::Plots.##114#115{Symbol})(::Float64) at /Users/michael/.julia/v0.5/Plots/src/axes.jl:148
 in optimal_ticks_and_labels(::Plots.Axis, ::Void) at /Users/michael/.julia/v0.5/Plots/src/axes.jl:186
 in get_ticks(::Plots.Axis) at /Users/michael/.julia/v0.5/Plots/src/axes.jl:229
 in tick_padding(::Plots.Axis) at /Users/michael/.julia/v0.5/Plots/src/backends.jl:71
 in _update_min_padding!(::Plots.Subplot{Plots.GRBackend}) at /Users/michael/.julia/v0.5/Plots/src/backends.jl:104
 in _collect(::Array{Plots.AbstractLayout,2}, ::Base.Generator{Array{Plots.AbstractLayout,2},Plots.#_update_min_padding!}, ::Base.EltypeUnknown, ::Base.HasShape) at ./array.jl:320
 in _update_min_padding!(::Plots.GridLayout) at /Users/michael/.julia/v0.5/Plots/src/layouts.jl:302
 in prepare_output(::Plots.Plot{Plots.GRBackend}) at /Users/michael/.julia/v0.5/Plots/src/plot.jl:254
 in display at /Users/michael/.julia/v0.5/Plots/src/output.jl:133 [inlined]
 in gui at /Users/michael/.julia/v0.5/Plots/src/output.jl:123 [inlined]
 in display(::Base.REPL.REPLDisplay{Base.REPL.LineEditREPL}, ::MIME{Symbol("text/plain")}, ::Plots.Plot{Plots.GRBackend}) at /Users/michael/.julia/v0.5/Plots/src/output.jl:138
 in display(::Base.REPL.REPLDisplay{Base.REPL.LineEditREPL}, ::Plots.Plot{Plots.GRBackend}) at ./REPL.jl:135
 in display(::Plots.Plot{Plots.GRBackend}) at ./multimedia.jl:143
[...]

PR #1559, opened 2018-06-14
Log-scale heatmap edge computation

### LINKED ISSUE #920: y-axis log scale bug

yha:
Heat map edge computation in log scale.
Fixes #920 for PyPlot and Plotly{,JS}.
GR seems not to support it natively as far as I can tell (or any heatmap where edges are not equally spaced in linear scale), so I added a warning instead.

```
x = logspace(-1,4,6)
y = logspace(-1,1,5)
z = log10.(x.*y')
p1 = heatmap(log10.(x),log10.(y),z',
             title="(log.(x),log.(y),z) :identity")
p2 = heatmap(x,y,z',
             xscale=:log10,yscale=:log10,
             title="(x,y,z) :log10")
x, y = [0.001,1,5,100], 1:2:10
z = log10.(x.*y')
p3 = heatmap(log10.(x),log10.(y),z')
p4 = heatmap(x,y,z',
             xscale=:log10,yscale=:log10)
plot(p1,p2,p3,p4,colorbar=false)

```

heatmap-log

### DIFF
--- a/src/backends/gr.jl
+++ b/src/backends/gr.jl
@@ -718,6 +718,9 @@ function gr_display(sp::Subplot{GRBackend}, w, h, viewport_canvas)
         end
         if st == :heatmap
             outside_ticks = true
+            for ax in (sp[:xaxis], sp[:yaxis])
+                ax[:scale] != :identity && warn("GR: heatmap with $(ax[:scale]) scale not supported.")
+            end
             x, y = heatmap_edges(series[:x], sp[:xaxis][:scale]), heatmap_edges(series[:y], sp[:yaxis][:scale])
             xy_lims = x[1], x[end], y[1], y[end]
             expand_extrema!(sp[:xaxis], x)
--- a/src/backends/plotly.jl
+++ b/src/backends/plotly.jl
@@ -587,6 +587,8 @@ function plotly_series(plt::Plot, series::Series)
         return plotly_series_segments(series, d_out, x, y, z)
 
     elseif st == :heatmap
+        x = heatmap_edges(x, sp[:xaxis][:scale])
+        y = heatmap_edges(y, sp[:yaxis][:scale])
         d_out[:type] = "heatmap"
         d_out[:x], d_out[:y], d_out[:z] = x, y, z
         d_out[:colorscale] = plotly_colorscale(series[:fillcolor], series[:fillalpha])
--- a/src/utils.jl
+++ b/src/utils.jl
@@ -358,19 +358,17 @@ const _scale_base = Dict{Symbol, Real}(
     :ln => e,
 )
 
-"create an (n+1) list of the outsides of heatmap rectangles"
-function heatmap_edges(v::AVec, scale::Symbol = :identity)
+function _heatmap_edges(v::AVec)
   vmin, vmax = ignorenan_extrema(v)
   extra_min = extra_max = 0.5 * (vmax-vmin) / (length(v)-1)
-  if scale in _logScales
-      vmin > 0 || error("The axis values must be positive for a $scale scale")
-      while vmin - extra_min <= 0
-          extra_min /= _scale_base[scale]
-      end
-  end
   vcat(vmin-extra_min, 0.5 * (v[1:end-1] + v[2:end]), vmax+extra_max)
 end
 
+"create an (n+1) list of the outsides of heatmap rectangles"
+function heatmap_edges(v::AVec, scale::Symbol = :identity)
+  f, invf = scalefunc(scale), invscalefunc(scale)
+  map(invf, _heatmap_edges(map(f,v)))
+end
 
 function calc_r_extrema(x, y)
     xmin, xmax = ignorenan_extrema(x)


daschw:
Here is Plots.jl current version on the left and the version of this PR on the right:
```julia
using Plots

x = y = 1:10
z = x * y'
xlog10 = log10.(x)

plot(
    heatmap(x, y, z, xscale = :log10, title = "Current"),
    heatmap(xlog10, y, z, xformatter = x -> "10^$x", title = "PR")
)
```

**GR**
heatmap_logscale_gr

**PyPlot**
heatmap_logscale_pyplot

To be honest, I don't really see an improvement, right?

daschw:
I don't really see the issue with the current version. Is it that the x axis is expanded too far on the left?

mkborregaard:
I think the issue is that the bins aren't regularly spaced. I've had the same concern for histograms on log scale here https://github.com/JuliaPlots/Plots.jl/issues/833#issuecomment-302072070
The PR doesn't look like it's working for pyplot, though.

daschw:
But shouldn't they be irregularly spaced? 
```julia
plot(1:10, xscale = :log10)
```
doesn't give a straight line either:
gr_xscale

mkborregaard:
I think that's an open discussion. I'd naturally expect `histogram(randn(1000), xscale = :log10)` to give the same-looking histogram as `histogram(log10.(randn(1000))` (although I could see the argument for different behaviours), also just because the uneven-looking histogram doesn't feel like it would have any practical usage.

daschw:
Histogram is more tricky, but in my opinion that's a different issue because it includes some preprocessing of data and should be dealt with in the histogram recipe(s). I think that the current Plots heatmap logsacale implementation is "correct", although I see your point about the lack of a practical use case.

mkborregaard:
Fair enough.

yha:
@daschw The issue as I see it is precisely the fact that the left and right subplots in your example don't look the same (the GR example is broken in both subplots, but this is not addressed by this PR). I think that is very surprising behavior, and honestly I fail to see how it would make sense, or why someone would prefer the current inconsistent behavior. Can you explain in what sense you think the current implementation is "correct"?
I suppose you could argue that when the data is not (logarithmically) evenly spaced, the choice of edges is somewhat arbitrary anyway, so it doesn't matter much. However,
- Taking the arithmetic average of two data points as the edge between them does not really seem to make sense in the context of a logarithmic scale. For example, if consecutive data points are 10^-3 and 10^3, the arithmetic average is ≈10^2.7.
- More importantly, I think we can at least agree that in the case of logarithmically evenly spaced data (as in `logspace`) plotted in logarithmic scale, the edges should be equally spaced, with the given (x,y) data centered (logarithmically!) in each bin. This is a special case of the proposed behavior. I believe this is probably the most common use case, and data like this `1:10` example would be very unusual in log scale.

Consider the following:
```
using Plots
pyplot()

x = logspace(-5,5,6)
y = 1:10
z = log10.(x)*y'
xlog10 = log10.(x)

p1 = heatmap(x, y, z', xscale = :log10, title = "Current")
scatter!(p1, x, y',label="")
p2 = heatmap(xlog10, y, z', xformatter = x -> "\$10^{$x}\$", title = "PR")
scatter!(p2,xlog10,y',label="")
plot(p1,p2)
```

log-bug-2

(the extra whitespace is caused by the scatter plots, which is perhaps another issue)

yha:
@mkborregaard 
> The PR doesn't look like it's working for pyplot, though.
What isn't working for pyplot?

daschw:
OK, I get it now. Thanks for explaining again @yha!

I guess my (probably rather aesthetic) issue with my pyplot example in https://github.com/JuliaPlots/Plots.jl/pull/1559#issuecomment-397386249 is that the left (current) version somehow looks more "regular" (sorry, I don't know how to express myself clearly here) than the right version, where the two outer x box lengths don't really match the pattern of the interiors.

But that's not really the fault of your changes, which I think is the correct implementation now, but its the result of Plots setting the left outer box extension `extra_min` to the same value of the right outer box extension `extra_max` in https://github.com/JuliaPlots/Plots.jl/pull/1559/files#diff-de08f51fe93b9406046593271d4e89c3R363.

I think the choice of the two outer heatmap edges really is arbitrary, but what do you think of
```julia
extra_min = (v[2] - v[1]) / 2
extra_max = (v[end] - v[end - 1]) / 2
```
instead?

yha:
I think that would be an improvement.
Should I change it in this PR, or is it preferred to have one issue per PR?

daschw:
No, please, go ahead and change it here.

PR #1573, opened 2018-06-24
Plotly: fix log scale with no ticks

yha:
Plotly backend currently does not set the axis scale unless ticks are present.
Fixed by setting the axis scale unconditionally.
```
using Plots; plotlyjs()
plot(1:10,1:10,xscale=:log10,xticks=nothing)
```
Before:
plot-noticks-bug
After:
plot-noticks-fix


daschw:
Thanks a lot @yha !

PR #1575, opened 2018-06-26
Fix axis flip on Plotly (Fixes #1556)

### LINKED ISSUE #1556: yflip=true does not respect axis limits in PlotlyJS
```julia
using Plots; plotlyjs();
plot(0:0.1:π, sin, ylim=(0.2,0.8))
```
plotly1

But after the flip 
```julia
yflip!()
```

plotly2

Notice here the the clipping is gone.

apalugniok:
#1556 
```
using Plots; plotlyjs();
plot(0:0.1:π, sin, ylim=(0.2,0.8))
yflip!()
```
yflip

### DIFF
--- a/src/backends/plotly.jl
+++ b/src/backends/plotly.jl
@@ -278,7 +278,7 @@ function plotly_axis(plt::Plot, axis::Axis, sp::Subplot)
 
         # flip
         if axis[:flip]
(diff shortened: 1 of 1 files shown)


daschw:
Great to see you back!

PR #1579, opened 2018-06-29
WIP: interim version for Julia v0.7

RalphAS:
This branch is intended to allow Plots users to work on Julia v0.7-beta.

It currently passes rudimentary checks on Linux x86_64 with GR, PyPlot, and UnicodePlots backends.  It uses undocumented internals to deal with backends, so may not be the right way to handle them in the long run.

Because of the large number of dependencies for the Plots test suite, a full release of Plots for v0.7 may be delayed for some time, but this may be useful until then.

It may need a development version of RecipesBase (julia0.7 branch).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,7 +2,7 @@
 language: julia
 os:
   - linux
-  # - osx
+  - osx
 julia:
 #  - 0.7
   - nightly
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,7 +4,7 @@ os:
   - linux
   # - osx
 julia:
-  - 0.6
+  - 0.7
 matrix:
   allow_failures:
     - julia: nightly
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,10 +4,11 @@ os:
   - linux
   # - osx
 julia:
-  - 0.7
-matrix:
-  allow_failures:
-    - julia: nightly
+#  - 0.7
+  - nightly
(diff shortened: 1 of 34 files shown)


daschw:
Thanks a lot for this @RalphAS! However, actually I'd prefer releasing a last 0.6 version now and then switching to 0.7-beta on master. Do you think you could make a 0.7-only version of this that I could merge into master after the next release?

RalphAS:
Sure. Just to clarify, do you mean I should remove the code blocks which are conditional on version below 0.7? And generally clean up to avoid reference to `Compat`? (I already switched `REQUIRE` to "julia 0.7" since `Compat` doesn't handle everything.)

I mistakenly thought this would show up immediately as a branch which users could fetch with `Pkg` to try. Since it is not yet so accessible, I suggest you let me know when you are ready to work with it; I will rebase it on master at that point.

daschw:
Awesome, thanks!

> Just to clarify, do you mean I should remove the code blocks which are conditional on version below 0.7? And generally clean up to avoid reference to Compat? 

Exactly :slightly_smiling_face: 
I just released 0.17.3 as the last Plots version to support julia 0.6. Development on master should now target julia 0.7. So you can go ahead, whenever you are ready.

daschw:
This has been done in a seperate PR now. Still, thanks a lot @RalphAS for your contribution!

PR #1585, opened 2018-07-01
Switch to 0.7 on master

daschw:


daschw:
The test will fail until https://github.com/JuliaImages/ImageMetadata.jl/issues/25 is fixed.

REQUIRE
```diff
@@ -1,4 +1,4 @@
-julia 0.6
+julia 0.7
```
fredrikekre: Set to `julia 0.7-alpha`.

REQUIRE
```diff
@@ -14,3 +14,4 @@ NaNMath
 Requires
 Contour
 GR 0.31.0
+Dates
```
fredrikekre: `Dates` should not be in `REQUIRE`.

daschw:
OK, thanks a lot for the hints @fredrikekre !

fredrikekre:
I have some additional fixes that I will push soonish

fredrikekre:
Ok, you can push your changes and I can rebase on top of that if I have something to add.

PR #1593, opened 2018-07-04
small additional fixes for 0.7

fredrikekre:
Turns out our changes where essentially identical @daschw , but here is some more things.


daschw:
Thanks!

fredrikekre:
Ah, I made this PR against the `julia0.7` branch, but you merged that one to master first it looks like, so these changes did only end up in that branch. I can open a new one against master

PR #1596, opened 2018-07-11
Minorgrid and minor ticks

wfrgra:
Adds minor grids and ticks. Works on GR and glvisualize.


daschw:
Thanks, looks good! I'll probably need a little time to review this.
Maybe it would make more sense if the logic to determine minor tick positions from axis limits (and possibly major tick positions) would live in https://github.com/JuliaPlots/PlotUtils.jl/blob/master/src/ticks.jl where `optimize_ticks` is defined. What do you think?

src/axes.jl
```diff
@@ -562,12 +591,16 @@ function axis_drawing_info(sp::Subplot)
...
+    xminorticks = get_minor_ticks(xaxis,xticks)
+    yminorticks = get_minor_ticks(yaxis,yticks)
     xaxis_segs = Segments(2)
     yaxis_segs = Segments(2)
     xtick_segs = Segments(2)
     ytick_segs = Segments(2)
     xgrid_segs = Segments(2)
     ygrid_segs = Segments(2)
+    xminorgrid_segs = Segments(2)
+    yminorgrid_segs = Segments(2)
```
daschw: Wouldn't it be better to push the minor grid segments to these variables, also return them and add some code in `backends/gr.jl` and `backends/glvisualize.jl` to draw them similar to the grid segments drawing code? Then we could choose a lower alpha value or a thiner linewidth value there for minor grid lines and wouldn't have to draw major grid lines twice.

wfrgra: Yes, that's a neater solution, I got a bit lazy there. Drawing the major grids multiple times darkens them and gives them some contrast but it is a bit of a hack. It'll take a bit longer to add in all the options for the minor grids but I should get it done this weekend

daschw: Awesome!

daschw:
Great, thanks @wfrgra!

daschw:
Just tried it locally, really nice work @wfrgra!

mkborregaard:
Awesome!

PR #1601, opened 2018-07-19
RFC: use Requires.jl for backend dependencies

daschw:
This is an alternative to #1598 and uses Requires.jl for conditional backend dependencies. Thanks to @piever we can keep the old behaviour of switching backends without calling `import Backend`. This has only been tested for Plotly, GR and PyPlot so far:

#### This PR
```julia
julia> @time using Plots
┌ Warning: `readstring(s::IO)` is deprecated, use `read(s, String)` instead.
│   caller = #open#302(::Base.Iterators.Pairs{Union{},Union{},Tuple{},NamedTuple{(),Tuple{}}}, ::Function, ::typeof(readstring), ::String, ::Vararg{String,N} where N) at iostream.jl:369
└ @ Base ./iostream.jl:369
  7.470374 seconds (11.54 M allocations: 642.697 MiB, 5.32% gc time)

julia> @time plot(rand(10))
 17.980940 seconds (47.46 M allocations: 2.338 GiB, 8.24% gc time)

julia> @time plot(rand(10))
  0.001128 seconds (2.59 k allocations: 151.516 KiB)
```
#### #1598 
```julia
julia> @time using Plots
  5.295788 seconds (11.44 M allocations: 630.732 MiB, 7.73% gc time)

julia> @time plot(rand(10))
 17.101824 seconds (45.39 M allocations: 2.244 GiB, 8.01% gc time)

julia> @time plot(rand(10))
  0.001249 seconds (2.91 k allocations: 178.938 KiB)
```
#### Julia 0.6
```julia
julia> @time using Plots
 22.140182 seconds (6.36 M allocations: 359.689 MiB, 0.92% gc time)

julia> @time plot(rand(10))
 10.615325 seconds (8.89 M allocations: 468.252 MiB, 2.30% gc time)

julia> @time plot(rand(10))
  0.441130 seconds (110.43 k allocations: 5.644 MiB)
```
#### This PR
```julia
julia> @time using Plots
┌ Warning: `readstring(s::IO)` is deprecated, use `read(s, String)` instead.
│   caller = #open#302(::Base.Iterators.Pairs{Union{},Union{},Tuple{},NamedTuple{(),Tuple{}}}, ::Function, ::typeof(readstring), ::String, ::Vararg{String,N} where N) at iostream.jl:369
└ @ Base ./iostream.jl:369
  7.348829 seconds (11.53 M allocations: 641.992 MiB, 5.32% gc time)

julia> @time pyplot()
 11.928570 seconds (30.14 M allocations: 1.464 GiB, 6.03% gc time)
Plots.PyPlotBackend()

julia> @time plot(rand(10))
 15.173691 seconds (34.04 M allocations: 1.676 GiB, 6.71% gc time)

julia> @time plot(rand(10))
  0.001223 seconds (2.61 k allocations: 152.047 KiB)
```
#### #1598 
```julia
julia> @time using PlotsPyPlot
 14.890743 seconds (35.46 M allocations: 1.777 GiB, 6.46% gc time)

julia> @time plot(rand(10))
┌ Warning: `Array{T}(d::NTuple{N, Int}) where {T, N}` is deprecated, use `Array{T}(undef, d)` instead.
│   caller = plot_color(::Array{RGB{FixedPointNumbers.Normed{UInt8,8}},1}) at colors.jl:24
└ @ PlotUtils ~/.julia/packages/PlotUtils/xQ9v/src/colors.jl:24
┌ Warning: `a::Number + b::AbstractArray` is deprecated, use `a .+ b` instead.
│   caller = get_zvalues(::Int64) at color_utils.jl:104
└ @ PlotUtils ~/.julia/packages/PlotUtils/xQ9v/src/color_utils.jl:104
 16.935470 seconds (36.21 M allocations: 1.778 GiB, 6.88% gc time)

julia> @time plot(rand(10))
  0.011821 seconds (4.23 k allocations: 336.875 KiB)
```
#### Julia 0.6
```julia
julia> @time using Plots
 22.113781 seconds (6.36 M allocations: 359.460 MiB, 1.05% gc time)

julia> @time pyplot()
  7.606616 seconds (4.68 M allocations: 248.842 MiB, 1.88% gc time)
Plots.PyPlotBackend()

julia> @time plot(rand(10))
  9.572784 seconds (6.94 M allocations: 369.290 MiB, 3.41% gc time)

julia> @time plot(rand(10))
  0.000804 seconds (2.78 k allocations: 142.813 KiB)
```
I'd go for this version because it does not affect user behaviour. 
Also there seems to be quite a regression in the time and memory usage of the first `plot(rand(10))` call from 0.6 to 0.7.


daschw:
At least the performance regression in the first `plot(rand(10))` call on 0.7 is not caused by my changes in the refactor or this PR.
Here are the benchmarks for current master on 0.7:
```julia
julia> @time using Plots
  6.430114 seconds (11.28 M allocations: 622.684 MiB, 7.32% gc time)

julia> @time plot(rand(10))
 18.316891 seconds (37.64 M allocations: 1.854 GiB, 7.56% gc time)

julia> @time plot(rand(10))
  2.840519 seconds (4.15 M allocations: 209.818 MiB, 3.20% gc time)

julia> @time plot(rand(10))
  0.003291 seconds (2.58 k allocations: 150.766 KiB)
```
(There's also a difference between 2nd and 3rd call, which is weird.)

daschw:
I'd like to go ahead with this implementation instead of the reorg in #1598. Are there any objections @piever @mkborregaard @apalugniok ?

apalugniok:
I'm fine with this 😄

mkborregaard:
First of all, sorry for being away when such important things are being discussed. I'm just now resurfacing, in fact I'm at JuliaCon at the moment :-)

So, sorry for not commenting before the merge. I do whole-heartedly agree with it, though. 
 
Amazing work, @daschw , this really does solve some of the big issues we've had, and I suspect that the startup time issue is something that could now be addressed with SnoopCompile? (BTW timings need to be `@time display(plot(rand(10))` as most of the actual processing happens in `display` which I don't think is timed with the current approach).

Should we push a version 0.18.0? And - when is the time to start talking in ernest about Plots 1.0?

daschw:
To be honest I'd really love to push 0.18.0. I've been waiting for quite some time now. But the tests still fail due to issues in the JuliaImages land - ImageFiltering.jl, I think, but maybe more. I'm not sure if we can tag a new release with failing tests. I opened some small PRs for some JuliaImages packages recently, but I think there are some non-trivial issues that I do not feel compenent enough to resolve and I'm sure Tim Holy has a lot of other stuff to do right now, so I'm not really sure how to proceed.

Considering all the complaining on Slack about Plots, it does not really seem like we are ready for a 1.0 release.

I really regret now that I did not apply for JuliaCon this year, but I wish you a lot of fun @mkborregaard!

mkborregaard:
Yeah, you should have been here!
Just disable the image example and uncomment the image stack from the test deps imho. No reason to delay 0.7 compatibility over a failing test dep.

daschw:
Unfortunately it's not about one example, VisualRegressionTests requires Images.

mkborregaard:
OK - my suggestion is to run a release with a comment that our test suite is broken by test dependencies but that it does appear to work and is 0.7 compatible - that would make it up to the metadata guys

mkborregaard:
do you think there should be two releases - a 0.7 one and a requires one? I do

mkborregaard:
it's very late now but i could do the housekeeping in the morning if you agree

daschw:
Why two releases?

mkborregaard:
I guess for clarity? Don't do too many important changes at once?

daschw:
OK, yes, if you volunteer to do it, great!
Currently GR, PyPlot and Plotly should work on 0.7 - I'm not sure about the other backends. PyPlot also has an issue in Atom (displaying twice in Plot pane and gui window).
But I still think we should try to release, otherwise we will get a lot of issues in the upcoming days...

mkborregaard:
https://github.com/JuliaLang/METADATA.jl/pull/16449

src/init.jl
```diff
@@ -0,0 +1,123 @@
+function __init__()
+    include(joinpath(@__DIR__, "backends", "plotly.jl"))
+    include(joinpath(@__DIR__, "backends", "gr.jl"))
+    include(joinpath(@__DIR__, "backends", "web.jl"))
+
+    @require GLVisualize = "4086de5b-f4b6-55f3-abb0-b8c73827585f" include(joinpath(@__DIR__, "backends", "gr.jl"))
```
CarloLucibello: ..., "backends", "glvisualize.jl"))

mkborregaard: Oh :-O Care to make a PR?

CarloLucibello: sure, #1660

PR #1615, opened 2018-08-03
Fix display on Juno 0.7

pfitzseb:
This PR makes Plots.jl compatible with Juno 0.7 on Julia 0.7. The `PLOTS_USE_ATOM_PLOTPANE` environment variable isn't necessary anymore since you can just disable the plot pane from within Juno now.

image


daschw:
Awesome, thanks!

PR #1622, opened 2018-08-09
Change dep to julia 0.7 and add Dates to Require

mkborregaard:


daschw:
Are you sure dates belongs in require?

mkborregaard:
Oh was that too fast? I'm still learning the new package manager (struggled a lot with it this morning, and e.g. ended up committing directly to master by accident). It was just my julia session that told me that this was a problem.

daschw:
AFAIK standard library packages like Dates don't belong to REQUIRE. Maybe @fredrikekre can confirm?

mkborregaard:
I'll take it out. I should have read the manual before acting.

fredrikekre:
Dates should not be in there, now this will reference the old Dates package from Julia 0.3 or something.

daschw:
Thanks!

mkborregaard:
Thanks. There was a warning when loading Plots on a fresh install that was a bit hard to interpret. I'll take it out.

PR #1630, opened 2018-08-09
1.0 fixes WIP RFC

mkborregaard:
See #1631

### DIFF
--- a/src/backends/plotly.jl
+++ b/src/backends/plotly.jl
@@ -76,7 +76,7 @@ end
 const _plotly_js_path = joinpath(dirname(@__FILE__), "..", "..", "deps", "plotly-latest.min.js")
 const _plotly_js_path_remote = "https://cdn.plot.ly/plotly-latest.min.js"
 
-_js_code = open(readstring, _plotly_js_path, "r")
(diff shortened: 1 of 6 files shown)


daschw:
There's still the
```julia
ERROR: MethodError: no method matching Array{RGBA{Float64},N} where N(::Tuple{Int64})
```
error, right?

daschw:
I'm currently having a look.

mkborregaard:
I don't think that's a Plots error.

daschw:
Yep, you're right it's PlotUtils

mkborregaard:
I fixed it then screwed up in git and lost my changes. Classic.

daschw:
I'm on it.

daschw:
https://github.com/JuliaPlots/PlotUtils.jl/pull/42
With this the next error is from ShowOff.jl

daschw:
There already is an open PR:
https://github.com/JuliaGraphics/Showoff.jl/pull/12

mkborregaard:
Thanks!

mkborregaard:
The showoff PR got merged and I've just plotted a histogram on 1.0! Let me try to test so we can push a release today, also with the new requires framework.

PR #1636, opened 2018-08-11
More fixes for Julia v1.0

oschulz:
* Use `length(eachindex(...))` instead of `length(linearindices(...))`
* Fix  _stepbins_path (new Julia v0.7/v1.0 iteration API)


mkborregaard:
💖

PR #1643, opened 2018-08-13
Make sure only one colorbar is drawn with GR

### LINKED ISSUE #1642: Multiple plots with line_zs add many colorbars
Here is the problematic example

```julia
using Plots
gr()

ts = range(0, stop=2π, length=50)
zs = 1:10
xs_list = [r .* cos.(ts) for r in zs]
ys_list = [r .* sin.(ts) for r in zs]

# This is what I want to plot:
xs = vcat((vcat(NaN, xs) for xs in xs_list)...)
ys = vcat((vcat(NaN, ys) for ys in ys_list)...)
plt1 = plot(xs, ys, line_zs=repeat(zs, inner=length(ts) + 1), label="")

# It looks like constructing the same plot with plot! produces
# "overdubbed" colorbar:
plt2 = plot()
for (xs, ys, z) in zip(xs_list, ys_list, zs)
    plot!(plt2, xs, ys, line_zs=z, label="")
end

# Trying to make it work by explicitly turning off colorbar.
plt3 = plot()
for (xs, ys, z) in zip(xs_list, ys_list, zs)
    plot!(plt3, xs, ys, line_zs=z, label="", colorbar=:none)
end
plot!(plt3, colorbar=:right)

# This does not work as well:
# plt3 = plot()
# for (i, (xs, ys, z)) in enumerate(zip(xs_list, ys_list, zs))
#     plot!(plt3, xs, ys, line_zs=z, label="",
#           colorbar=i == length(xs_list) ? :right : :none)
# end

# Calling plot once does not fix the issue.
plt4 = plot(xs_list, ys_list, line_zs=reshape(zs, (1, :)), label="")

plt = plot(plt1, plt2, plt3, plt4)
savefig(plt, "fig.png")
```

which outputs

fig

Checked with:

````
(v1.0) pkg> status -m
    Status `~/.julia/environments/v1.0/Manifest.toml`
  ...
  [28b8d3ca] GR v0.32.2
  ...
  [ccf2f8ad] PlotThemes v0.3.0
  [995b91a9] PlotUtils v0.5.1
  [91a5bcdd] Plots v0.19.0
  ...
  [3cdcf5f2] RecipesBase v0.5.0
  ...
````

I can plot what I want with the "NaN trick" but it's very hard to use when I want to define plot recipes for nested structure:

```julia
using RecipesBase

struct Circle
    xs
    ys
    z
end

struct CircleList
    circles
end

@recipe function f(c::Circle)
    line_zs := c.z
    label := ""
    (c.xs, c.ys)
end

@recipe function f(data::CircleList)
    for c in data.circles
        @series begin
            c
        end
    end
end

ts = range(0, stop=2π, length=50)
zs = 1:10
circles = CircleList([Circle(
    r .* cos.(ts),
    r .* sin.(ts),
    r,
) for r in zs])

using Plots
plot(circles)
```

tkf:
This patch fixes #1642 but I haven't tested it with anything else.  Probably better to wait for CI/test suite to be working?  Also, I have no idea if `GR.savestate()` and `GR.restorestate()` are necessary there (I couldn't find any doc).

### DIFF
--- a/src/backends/gr.jl
+++ b/src/backends/gr.jl
@@ -1227,16 +1227,18 @@ function gr_display(sp::Subplot{GRBackend}, w, h, viewport_canvas)
             gr_text(GR.wctondc(xi, yi)..., str)
         end
 
-        # draw the colorbar
-        if cmap && st != :contour # special colorbar with steps is drawn for contours
-            gr_set_line(1, :solid, yaxis[:foreground_color_axis])
-            gr_set_transparency(1)
-            gr_colorbar(sp, clims)
-        end
-
         GR.restorestate()
     end
 
+    # draw the colorbar
+    GR.savestate()
+    # special colorbar with steps is drawn for contours
+    if cmap && any(series[:seriestype] != :contour for series in series_list(sp))
+        gr_set_line(1, :solid, yaxis[:foreground_color_axis])
+        gr_set_transparency(1)
+        gr_colorbar(sp, clims)
+    end
+    GR.restorestate()
 
     # add the legend
     if sp[:legend] != :none


mkborregaard:
Thanks! Yes, let's wait for the test suite to work.

src/backends/gr.jl
```diff
@@ -1227,16 +1227,18 @@ function gr_display(sp::Subplot{GRBackend}, w, h, viewport_canvas)
...
-            gr_set_line(1, :solid, yaxis[:foreground_color_axis])
-            gr_set_transparency(1)
-            gr_colorbar(sp, clims)
-        end
-
         GR.restorestate()
     end
 
+    # draw the colorbar
+    GR.savestate()
```
mkborregaard: @jheinen is the `savestate` and `restorestate` necessary here?

jheinen: The `savestate()` ... `restorestate()` functions ensure, that the attributes are reset to the old settings after drawing the colorbar. They don't produce much overhead ...

mkborregaard: Great, thanks!

PR #1644, opened 2018-08-13
Don't broadcast `plot_color`; fix #1641

### LINKED ISSUE #1641: another 0.7 / 1.0 thing
I use plots with Plotly().  This warning arises:

```
┌ Warning: broadcast will default to iterating over its arguments in the future. Wrap arguments of
│ type `x::PlotUtils.ColorGradient` with `Ref(x)` to ensure they broadcast as "scalar" elements.
│   caller = ip:0x0
└ @ Core :-1
```

It is triggered by this call:
```
    pyplot()
    clibrary(:misc)  # collection of color palettes
    img = reshape(digit_data, digit_dims...)'
    pldigit = plot(img, seriestype=:heatmap, color=:grays,  
        showaxis=false, legend=false, yflip=true, size=(400,400)) # right_margin=6mm, bottom_margin=6mm
    display(pldigit)
```
clibrary is a holdover from PlotlyJS.  It has lots of problems--maybe a long time till we see it again.
I tried wrapping the color named argument with Ref() but that did nothing.  So, I think the warning must from from within Plots.  Perhaps?

mkborregaard:
On 1.0 you can no longer pass a singleton argument to broadcast without escaping with `Ref`. But - `plot_color` has `Vector` methods, so it should be fine to not broadcast here. The example from the PR that introduced the broadcast here still works on this PR https://github.com/JuliaPlots/Plots.jl/pull/1467#issue-178940271

### DIFF
--- a/src/args.jl
+++ b/src/args.jl
@@ -1569,13 +1569,13 @@ function _update_series_attributes!(d::KW, plt::Plot, sp::Subplot)
     for s in (:line, :marker, :fill)
         csym, asym = Symbol(s,:color), Symbol(s,:alpha)
         d[csym] = if d[csym] == :auto
-            plot_color.(if has_black_border_for_default(d[:seriestype]) && s == :line
+            plot_color(if has_black_border_for_default(d[:seriestype]) && s == :line
                 sp[:foreground_color_subplot]
             else
                 d[:seriescolor]
             end)
         elseif d[csym] == :match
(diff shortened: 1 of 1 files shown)


mkborregaard:
I'm going to go ahead and merge this, as it totally breaks specifying color gradients on 1.0 - we can go over and perhaps edit later @daschw

PR #1657, opened 2018-08-15
Implement colorbar_title for GR

tkf:
Mostly a bit tweaked version of ylabel code.  Is this the right way to do it?

### DIFF
--- a/src/backends/gr.jl
+++ b/src/backends/gr.jl
@@ -487,7 +487,7 @@ function gr_colorbar(sp::Subplot, clims)
     GR.axes(0, ztick, xmax, clims[1], 0, 1, 0.005)
 
     gr_set_font(guidefont(sp[:yaxis]))
-    GR.settextalign(GR.TEXT_HALIGN_CENTER, GR.TEXT_VALIGN_BOTTOM)
+    GR.settextalign(GR.TEXT_HALIGN_CENTER, GR.TEXT_VALIGN_TOP)
     GR.setcharup(-1, 0)
     gr_text(viewport_plotarea[2] + gr_colorbar_ratio,
             gr_view_ycenter(), sp[:colorbar_title])
--- a/src/backends/gr.jl
+++ b/src/backends/gr.jl
@@ -27,7 +27,7 @@ const _gr_attr = merge_with_base_supported([
     :guidefontfamily, :guidefontsize, :guidefonthalign, :guidefontvalign,
     :guidefontrotation, :guidefontcolor,
     :grid, :gridalpha, :gridstyle, :gridlinewidth,
-    :legend, :legendtitle, :colorbar,
+    :legend, :legendtitle, :colorbar, :colorbar_title,
     :fill_z, :line_z, :marker_z, :levels,
     :ribbon, :quiver,
     :orientation,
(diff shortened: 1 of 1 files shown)


src/backends/gr.jl
```diff
@@ -436,7 +438,7 @@ function gr_viewport_from_bbox(sp::Subplot{GRBackend}, bb::BoundingBox, w, h, vi
         viewport[4] = 0.5 * (vp[3] + vp[4] + extent)
     end
     if hascolorbar(sp)
-        viewport[2] -= 0.1
+        viewport[2] -= gr_colorbar_ratio
```
tkf: I wanted to make it clear that 0.1 here means the space for colorbar so I defined it as a `const`.  `gr_colorbar_ratio` is used in the code I added to `gr_colorbar` as well.  Are there any other places this value is used?

src/backends/gr.jl
```diff
@@ -657,6 +666,9 @@ function _update_min_padding!(sp::Subplot{GRBackend})
     if sp[:yaxis][:guide] != ""
         leftpad += 4mm
     end
+    if sp[:colorbar_title] != ""
+        rightpad += 4mm
```
tkf: Is it required?  I added it in the patch just to follow how other guides work (and ask this question) but it seems that this includes extra unnecessary space in the plot.

mkborregaard: If it doesn't look right then don't include it - you might post an example?

mkborregaard:
LGTM!

src/backends/gr.jl
jheinen: I don't really understand why the right padding should be increased by a fixed amount. I think it depends on the title's text extent.

jheinen: Ok, now I see. The title is displayed vertically, so it only depends on the text height.

tkf: @jheinen Yes, I thought it makes sense but incrementing the right padding introduces extra space which is a bit unexpected.

@mkborregaard So maybe `gr_colorbar_ratio` is too large?  But what I tried was to set the bottom of `colorbar_title` at the right edge of the figure (= `viewport_plotarea[2] + gr_colorbar_ratio`):

https://github.com/JuliaPlots/Plots.jl/blob/2acd088226bde5a5ce58358369dfa804a9da8731/src/backends/gr.jl#L489-L493

So if `gr_colorbar_ratio` is too large, I think there has to be a space between the colorbar and title, not between the right edge of the figure and the title.  But this is not the case and I'm puzzled.

tkf: OK. So I think I found the source of my confusion.  I was thinking that `viewport_plotarea[2] + gr_colorbar_ratio` would be the edge of entire image (i.e., anything right to it would not be rendered in (say) png file).  But that's not the case, right?  My current understanding is that it is the right edge of area for the "subplots/axes" (not sure what's the technical word... plot area?) and the guides (title/xlabel/ylabel) have to be located outside of it.  If this is correct, using `TEXT_VALIGN_BOTTOM` was wrong since that would put the colorbar title inside the edge.  So I changed it to `TEXT_VALIGN_TOP`:

https://github.com/JuliaPlots/Plots.jl/blob/b56238089cf3985da11574497659045ec881b896/src/backends/gr.jl#L489-L493

With this change (and still keeping `rightpad += 4mm`) I now have:

fig

I think this is the right direction.  Previous two examples seem to have colorbar title that is too close to the colorbar tick labels.

mkborregaard: Yes, this should be correct.

daschw:
Awesome work @tkf !

Issue #1661, opened 2018-08-15
Attributes from PLOTS_DEFAULTS dict are avoided


lnacquaroli:
I'm trying to use `PLOTS_DEFAULTS` from `startup.jl` but it doesn't seems to work. It's called and the variables are set, but the plot command doesn't take them into account (I tried the `startup.jl` file with other commands and works fine). REPL gives with same results:

`julia> PLOTS_DEFAULTS = Dict(:linewidth=>5)`
`Dict{Symbol,Int64} with 1 entry:`
`  :linewidth => 5`

Looks it worked but when plotting the line is still with linewidth = 1. Doing 

`using Plots`
`gr(linewidth=5)` 

works though. I'm on Julia 1.0 and Plots v0.19.2

mkborregaard:
Darn. Must be the startup.jl file is implemented differently on 0.7

lnacquaroli:
Maybe I expressed myself incorrectly before, but calling `PLOTS_DEFAULTS = Dict(...)` from REPL also doesn’t work for me (neither from the `startup.jl`). So I guess is not just the startup file in this case, right?

mkborregaard:
huh

pfitzseb:
Will be fixed by https://github.com/JuliaPlots/Plots.jl/pull/1667.

lnacquaroli:
This is solved now for me. Maybe wanna close it. Thanks.

daschw:
Thanks!

PR #1667, opened 2018-08-16
WIP: Actually use PlotsDisplay()

pfitzseb:
instead of overloading display for REPLDisplay.

This unfortunately doesn't work yet because the `@init` block apparently is never executed.

### DIFF
--- a/src/Plots.jl
+++ b/src/Plots.jl
@@ -294,19 +294,4 @@ end
 
 const CURRENT_BACKEND = CurrentBackend(:none)
 
-# for compatibility with Requires.jl:
-@init begin
-    if isdefined(Main, :PLOTS_DEFAULTS)
-        if haskey(Main.PLOTS_DEFAULTS, :theme)
-            theme(Main.PLOTS_DEFAULTS[:theme])
-        end
-        for (k,v) in Main.PLOTS_DEFAULTS
-            k == :theme || default(k, v)
-        end
-    end
-    pushdisplay(PlotsDisplay())
-end
-
-# ---------------------------------------------------------
-
 end # module
--- a/src/Plots.jl
+++ b/src/Plots.jl
@@ -304,6 +304,7 @@ const CURRENT_BACKEND = CurrentBackend(:none)
             k == :theme || default(k, v)
         end
     end
+    pushdisplay(PlotsDisplay())
 end
 
 # ---------------------------------------------------------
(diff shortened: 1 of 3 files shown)


mkborregaard:
👍 
The non-executed `init` block is probably also the cause of https://github.com/JuliaPlots/Plots.jl/issues/1661
Could it be related to https://github.com/JuliaPlots/Plots.jl/pull/1601 ?

pfitzseb:
Turns out that Plots.jl now actually has a `__init__` function, so the `@init` macro doesn't make any sense any more. This works mostly fine now, I think.

pfitzseb:
Note: This only works with GR if you overwrite `GR.isatom` with `GR.isatom() = false`.

mkborregaard:
Nice!

mkborregaard:
It works fine in the REPL with GR for me - it doesn't work in Atom, but then again I don't think Plots master works for me in Atom either - it's always been flaky.

pfitzseb:
In Atom you might need https://github.com/JunoLab/Atom.jl/commit/234890134f24f1a8e2bb7d3a3ee281f653c1b7cf for Plots to appear in the Plot Pane and possibly also overwrite GR's `isatom` method with `isatom() = false`.

Ping @jheinen for coordination with your efforts to get GR working in Atom.

daschw:
It works perfectly on my system :tada: thanks a lot @pfitzseb

mkborregaard:
So we don't need `isatom`?

daschw:
Oh, was I too fast? I didn't need to overwrite anything.

mkborregaard:
I need to overwrite `GR.isatom` for this to work - which @jheinen says can be done automatically if we somehow define an `isplots` function in GR (I guess it could just evaluate to `false` in GR and to `true` in Plots).

daschw:
So, should we revert this until there is an `isplots` in GR?

mkborregaard:
No let me make a quick PR and then just wait with doing a release?

mkborregaard:
Wait :mind_blown: Now it works on master of GR and Plots with no problem even on my machine. I wonder if my machine was in some weird state when I tried it? But at least I now have branches here and on GR that would solve the problem if there was one...

EricForgy:
This is still breaking plots in VS Code. Any thoughts?

PR #1686, opened 2018-08-21
Fix NamedTuple integration

piever:
On Julia 0.7 and above NamedTuples leave in Base, so we should switch to supporting those.


piever:
Tests are failing but I think it's unrelated to this PR.

mkborregaard:
Yes they're unrelated.

mkborregaard:
Should we wait for Travis to work again before merging?

piever:
I don't think we need to, it's a very circumscribed change.

PR #1689, opened 2018-08-23
Fix CI

tkf:
I just blindly replace old API to the new one.  Not sure if all the upstream packages are ready but I think it's handy if we can check Travis to see how much of them are ready.

### DIFF
--- a/test/imgcomp.jl
+++ b/test/imgcomp.jl
@@ -19,7 +19,7 @@ end
 
 using Plots
 # using StatPlots
-using PlotReferenceImages
+import PlotReferenceImages
 using Random
 using Test
 


test/travis_commands.jl
```diff
@@ -1,34 +1,29 @@
...
+# need this to use Conda
+# ENV["PYTHON"] = ""
 
-# Pkg.add("StatPlots")
-Pkg.checkout("PlotUtils")
-Pkg.checkout("RecipesBase")
+Pkg.add([
+    PackageSpec("ImageMagick"),
+    # PackageSpec("GR"),
+    PackageSpec(url="https://github.com/tkf/PlotReferenceImages.jl", rev="pkg3"),
```
tkf: Revert a5855929eec3dae71d4099173e93a25e3258adf7 once https://github.com/JuliaPlots/PlotReferenceImages.jl/pull/30 is merged

test/add_packages.jl
```diff
@@ -0,0 +1,24 @@
+using Pkg
+
+# need this to use Conda
+# ENV["PYTHON"] = ""
+
+Pkg.add([
+    PackageSpec(url="https://github.com/tkf/PlotReferenceImages.jl", rev="pkg3"),
```
tkf: TODO for me: Revert 17480637d721ccbeccc28cacaba7a0103cf5dfee once JuliaPlots/PlotReferenceImages.jl#30 is merged

tkf:
Oops, I didn't mean to go this deep into tweaking your testing setup.  I thought it was just a quick small fix :)  I hope I didn't mess up anything.

The biggest change is that non-registered test requirements are now `Pkg.add`ed in `test/runtests.jl`.  This is for adding those packages inside the testing environment instead of the default global environment.  I *guess* this is the easiest choice if you want to use `Pkg.test` but avoid checking in `Manifest.toml`.  (see also https://github.com/JuliaDiff/ForwardDiff.jl/pull/332#issuecomment-408435332)

Anyway, I think now the CI is reporting meaningful errors and I will not to try fixing them.

mkborregaard:
I'm not really sure what the purpose is of this PR - could you explain? I think our existing setup quite closely followed the recommended default travis script for pacakges.

tkf:
I fixed the test suite and CI setup so that they work with Julia 1.0 / Pkg3.  I don't think the current setup in master branch is compatible with Julia 1.0.  I should have edited the title since it does not describe this PR well any more.

tkf:
Some more info: I don't think installing non-registered packages in `travis_commands.jl` as you were doing is a valid approach if you want to use `Pkg.test`:

> The tests are run by generating a temporary environment with only pkg and its (recursive) dependencies (recursively) in it. If a manifest exist, the versions in that manifest is used, otherwise a feasible set of package are resolved and installed.
>
> --- https://docs.julialang.org/en/latest/stdlib/Pkg/#Pkg.test

I can think of three choices:

1. Add non-registered packages in `runtests.jl` as I did here.
2. Add `Project.toml` *and* `Manifest.toml` (which is already [supported by the default script](https://github.com/travis-ci/travis-build/blob/386d788aa49869a2028f23296d587ddead3023bb/lib/travis/build/script/julia.rb#L76)).  `Manifest.toml` can contain `repo-url` pointing to a non-registered git repository.  The downside is that `Manifest.toml` has to be updated manually to reflect upstream changes.
3. Skip `Pkg.test` and directly invoke `julia test/runtests.jl`.  The downside is that you have to install everything in `test/REQUIRE` manually.

Maybe there are more solutions. But I thought the first choice was reasonable.

daschw:
Nice work @tkf! I actually have a local branch starting a quite similar implementation, but you are one step ahead.
We still need to finalize updating VisualRegressionTests.jl, but in the future it will optionally depend on Gtk. If the tests are run interactively (i.e. locally by a user) we want a Gtk dependency, so that the user can compare possible changes in the plot examples in a GUI window.
My first suggestion would be to also add Gtk to the dependencies and add something like
```julia
if isinteractive()
    using Gtk
end
```
This, however, also requires the Gtk dependency for CI runs, where it is not actually used. So if you have a better idea, I'm open to suggestions.
Otherwise, I'd ask you to add something similar to my code snippet above somewhere in e.g. runtests.jl, and we can revisit, once https://github.com/JuliaPlots/VisualRegressionTests.jl/pull/15/commits/1cba46290fa61a0b8ad047493997810d943091d3 is released.

tkf:
Why doesn't `@eval Main import Gtk` (or maybe `@static if isinteractive()`?) work?

test/add_packages.jl
```diff
@@ -0,0 +1,24 @@
...
+
+Pkg.add([
+    PackageSpec(url="https://github.com/JuliaPlots/PlotReferenceImages.jl.git"),
+    # PackageSpec(url="https://github.com/JuliaStats/KernelDensity.jl.git"),
+    PackageSpec(name="PlotUtils", rev="master"),
+    PackageSpec(name="RecipesBase", rev="master"),
+    # PackageSpec(name="Blink", rev="master"),
+    # PackageSpec(name="Rsvg", rev="master"),
+    # PackageSpec(name="PlotlyJS", rev="master"),
+    PackageSpec(name="VisualRegressionTests", rev="master"),
```
tkf: TODO: Comment out above (revert a8e3eccb40c2f9704baab323779f347fe39b9fe8)

tkf:
So one of the problem with `if isinteractive()` is that it's not `true` in the subprocess launched by `Pkg.test`.  It means that the user have to manually `run("test/runtests.jl")` if s/he wants to test `Gtk` related components (which may not be so bad?).   Another option would be to check (say) `ENV["PLOTS_TEST_GUI"]` and install `Gtk` in `test/add_packages.jl`.

test/runtests.jl
```diff
@@ -1,5 +1,6 @@
 module PlotsTests
 
+include("add_packages.jl")
```
tkf: I feel bad about unconditionally installing packages via runtests.jl which could ruin users' environments.  Something like

```julia
if isinteractive()
    @info """
    Skip installing unregistered packages.  To run Plots.jl test suite
    in interactive session, you need to install them manually by, e.g.,
    `include("test/add_packages.jl")`.  Alternatively, you can run the test
    with `Pkg.test("Plots")`.
    """
else
    include("add_packages.jl")
end
```

would be nice.  However, this still is not enough when the test is invoked from shell like `julia test/runtests.jl`.  Checking `Base.HOME_PROJECT` and `Base.ACTIVE_PROJECT` may solve this problem but I guess we need to be careful not to exclude their values when invoked via `Pkg.test`.

mkborregaard:
>  It means that the user have to manually run("test/runtests.jl") if s/he wants to test Gtk related components (which may not be so bad?).

Yes, that's the idea. Advanced users and people making PRs can run the tests manually to overwrite images with new versions, people just running `test` to know if this works on julia-someversion will just get an exception (the GTK window is only launched upon test errors)

src/Plots.jl
```diff
@@ -116,7 +116,7 @@ ignorenan_minimum(x) = Base.minimum(x)
 ignorenan_maximum(x::AbstractArray{F}) where {F<:AbstractFloat} = NaNMath.maximum(x)
 ignorenan_maximum(x) = Base.maximum(x)
 ignorenan_mean(x::AbstractArray{F}) where {F<:AbstractFloat} = NaNMath.mean(x)
-ignorenan_mean(x) = Base.mean(x)
```
mkborregaard: Maybe `Statistics.mean` to be explicit?

test/imgcomp.jl
```diff
@@ -15,6 +19,8 @@ using VisualRegressionTests
 
 using Plots
 # using StatPlots
+using PlotReferenceImages
```
mkborregaard: Maybe just `import PlotReferenceImages`?  You're just using it for `pathof`, right?

mkborregaard:
You're on fire @tkf , great with all the contribs lately!

src/Plots.jl
tkf: Makes sense. Applied in 6d0d8ea90aefb3044c3d9778cfb6e7ff2c5ef11a

PR #1511, opened 2018-05-05
fix marker arguments (fix #1503 fix #1505)

daschw:
#1467 Introduced some issues with marker arguments (cf. #1503, #1505). Thank you @dpsanders and @cstjean for reporting!

```julia
plotly() # or plotlyjs()
scatter([1], [2])
```
marker_plotly

```julia
gr()
scatter([1,2,3], alpha=0.1)
```
markeralpha_gr

Furthermore, other marker arguments should also work fine on GR, Plotly(JS), PyPlot and PGFPlots:
```julia
n = 20; x = 1:n
plot(
    scatter(x.^2, markeralpha = [0.8, 0.2], markercolor = [:orange, :red, :purple], markersize = [5, 8]),
    scatter(x.^2, markeralpha = linspace(1, 0, n), marker_z = sqrt.(x), markersize = 8),
    markerstrokewidth = 0,
    markershape = [:circle, :diamond, :dtriangle, :star5, :square]
    )
```
marker_arguments

PR #1691, opened 2018-08-24
Performance improvements

daschw:
This improves the performance of large scatter plots with `marker_z`. See https://github.com/JuliaPlots/Plots.jl/pull/1511#discussion_r212554048.
Consider
```julia
n = 100000
x, y = randn(n), randn(n)
z = sqrt.(x.^2 + y.^2)
```
This changes the output of the second
```julia
@time display(scatter(x, y, mz = z))
```
call from
```
44.086065 seconds (3.41 M allocations: 231.762 MiB, 0.19% gc time)
```
to
```
2.927478 seconds (2.21 M allocations: 207.285 MiB, 15.22% gc time)
```

cc @mkborregaard

### DIFF
--- a/src/backends/gr.jl
+++ b/src/backends/gr.jl
@@ -338,7 +338,11 @@ end
 
 
 # draw the markers, one at a time
-function gr_draw_markers(series::Series, x, y, msize, mz)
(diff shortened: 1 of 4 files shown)


mkborregaard:
Awesome!

Issue #1694, opened 2018-08-25
backends/gr.jl needs :match_dimensions added


JeffFessler:
As discussed in #273 and #196, the option `transpose=true` should help those of us who like f(x,y) not f(y,x).  (I work with a lot of 3D images and strongly prefer f(x,y,z) to f(y,x,z)...)
However, the `transpose=true` option currently throws a warning with the GR backend:
`Warning: Keyword argument match_dimensions not supported with Plots.GRBackend`

It can be fixed simply by adding the keyword `:match_dimensions` to backends/gr.jl in the long list of supported keywords there.

Here's a MWE that generates the warning:  
`using Plots # uses gr() by default`
`x=-10:10; y=-3:3; f=x.^2 .+ 0*y'`
`heatmap(x, y, f, transpose=true)`

I'm using Plots v0.19.3 with Julia 0.7.0

mkborregaard:
Absolutely. In the mood for a quick PR? :-p

JeffFessler:
Sure.  It wasn't so quick because it's my first PR (!) so hopefully I did it correctly...
The CI tests failed it seems but the failures didn't seem related to my tiny modification...

mkborregaard:
Yay! welcome to the wonderful world of collaborative open source development!
Do you know how to rebase your PR on current master?

JeffFessler:
I did my one-line edit and PR in the browser.  The rebase instructions seem to be for someone who has a local copy of the repo on their host:
https://github.com/edx/edx-platform/wiki/How-to-Rebase-a-Pull-Request
Is there a way to rebase a PR on the website in the browser?
Maybe it would be easier to make my one-line change as a new PR on the current master?

daschw:
Are you talking about #1696 ? Why would this need to be rebased, @mkborregaard ?

mkborregaard:
Yes, exactly, it was just to run the tests. But I think it's fine. Thank you @JeffFessler !

daschw:
@JeffFessler, honestly I don't know how to do this in the browser. To do this locally, you can open julia and run
```julia
(v1.0) pkg> dev Plots
```
this clones the current master of Plots in your julia dev directory. Head over to this directory and add your fork as a remote:
```julia
julia> cd(joinpath(homedir(), ".julia", "dev", "Plots"))
shell> git remote add forked git@github.com:JeffFessler/Plots.jl.git
```
Fetch origin and your fork and checkout the branch you want to rebase:
[...]

Issue #273, opened 2016-05-24
Axes swapped in heatmap


oschulz:
Plots.jl 0.6.2 mirrors the axes for heatmap with the PyPlot backend:

```
Plots.heatmap(0:0.1:1, 0:0.1:1, (x, y) -> e^-x)
```

![exp-heatmap-mirrored](https://cloud.githubusercontent.com/assets/546147/15498209/f2e72c2e-219e-11e6-8d4b-fc8ac81dc786.png)

The bright area should be on the left, not on the bottom. This is a new problem, with the Plots.jl version (sorry, forgot which one) I used before, everything was fine. Interestingly, `histogram2d` still plots correctly:

```
# Random numbers distributed according to e^-x:
X = -log(rand(100000))
# Uniform random numbers:
Y = rand(100000)
Plots.histogram2d(X, Y)
```

![exp-histogram2d](https://cloud.githubusercontent.com/assets/546147/15498267/3ca475ce-219f-11e6-9614-c7d79a058b03.png)

tbreloff:
This is actually a feature, not a bug.  I was convinced to change it to follow conventions in which the y-axis should match the rows of the z-matrix.  See: https://github.com/tbreloff/Plots.jl/issues/196#issuecomment-216725551

You can switch the behavior with `transpose = true`

oschulz:
> This is actually a feature, not a bug.

It results in a lot of inconsistency, though, if x and y data have to be passed to heatmap in a different order than for all other plots (including histogram2d, which should be the same, modulo auto-binning of the data). I would imagine that it will surprise many users.

oschulz:
> in a different order than for all other plots
Ok maybe not exactly in a different order - it's matrix data, after all. Still, it's confusing when using a generation function, like in my example - so shouldn't x and y be passed in a different order too, then, if z is a function?

tbreloff:
I don't disagree... which is why I fought it.  Just like anything, there can only be one built-in default, and someone's always going to disagree with the choice.  My solution is to tie it to a keyword (`transpose` is an alias for `match_dimensions`) so that you can set your own personal default value if you want.

oschulz:
Sure, transpose works. But what I mean is, couldn't Plots.jl just feed x and y into z in a different order if `transpose` is set to false? I would assume that if z is given as a function, the user would always assume that the x-axis is the first, and the y-axis the second argument. If z is a function, the user will never see the resulting matrix, after all.

[...]

Issue #196, opened 2016-04-22
spy shows the matrix transposed


dpsanders:
An upper-triangular matrix is shown as lower-triangular by `spy`.

tbreloff:
I agree... there needs to be a transpose somewhere, but I need to figure out the correct place to add it.

dpsanders:
Yes, I see -- there seems to be a similar flipping issue with `heatmap`, the usual problem of how to treat "axes" in a matrix, which is certainly tricky since in different situations you might want different solutions. (I guess this is what the `yflip` keyword argument is about.)

In the case of `spy`, though, it seems pretty unambiguous that you would want the image on screen to correspond to the layout of the matrix as shown by Julia, so for now you could just add the transpose in the definition of `spy`:

```julia
function spy{T<:Real}(z::AMat{T}; kw...)
  heatmap(map(zi->float(zi!=0), z)'; leg=false, yflip=true, kw...)
end
```

tbreloff:
Yes I did that just now, but just as you said it might be better to do the transpose for all heatmaps.  And maybe also contours, surfaces, etc?  I'm wondering if my `Plots.Surface` should really be a subtype of AbstractMatrix and make it effectively a "transpose view".  Certainly adding a transpose to spy will fix this particular problem, but I want to make sure I have the correct behavior generally.

Note: you can always pass in a transposed matrix to the current function to get it working now...

tbreloff:
And if it's not clear... I'd love to have some other eyes looking at the current behavior of heatmap, contour, etc to figure out what needs to be transposed/flipped (and under what circumstances).  Help appreciated!

dpsanders:
Hmm I see. I guess this is a problem that all plotting libraries have previously battled with. Is there any kind of consensus out there? I certainly remember being frustrated with PyPlot's take in the past, but don't really remember details. Opposite conventions for `imshow` and `pcolor`?

tbreloff:
For reference: 

http://stackoverflow.com/questions/21166679/when-to-use-imshow-over-pcolormesh
http://thomas-cokelaer.info/blog/2014/05/matplotlib-difference-between-pcolor-pcolormesh-and-imshow/

tbreloff:
From the [matplotlib docs](http://matplotlib.org/api/pyplot_api.html#matplotlib.pyplot.pcolor):

[...]

PR #1696, opened 2018-08-25
Add :match_dimensions to address #1694

JeffFessler:
see also #273 and #196, for f(x,y) vs f(y,x), and transpose=true


mkborregaard:
Great, thanks!

PR #1698, opened 2018-08-26
WIP: Fix `-->` ignored in series recipes (Fixes #1639) 

### LINKED ISSUE #1639: `-->` ignored in series recipes
I noticed this as I was testing through all of StatPlots (it makes the ea_histogram recipe fail), but I also get it now even on very old (~1 year) versions of RecipesBase and Plots, and running on 0.6.3. If all users experience this, it is a big problem for the recipe ecosystem, so it is a priority. It is not 0.7-related.

```julia
@recipe function g(::Type{Val{:testhist}}, x, y, z)
        markercolor := :red
        seriestype := :scatter
        ()
end

@recipe function f(::Type{Val{:testhist2}}, x, y, z)
        markercolor --> :red
         seriestype := :scatter
         ()
end

plot(rand(10,2), st = [:testhist :testhist2], layout = 2)
```

skaermbillede 2018-08-12 kl 16 55 26

apalugniok:
#1639 
Quick fix, still need to tidy a bit. I haven't used recipes much so I'm not quite sure what some odd cases could break this. So help with testing would be appreciated.

```
@recipe function g(::Type{Val{:testhist}}, x, y, z)
        markercolor := :red
        seriestype := :scatter
        ()
end

@recipe function f(::Type{Val{:testhist2}}, x, y, z)
        markercolor --> :red
         seriestype := :scatter
         ()
end

plot(rand(10,2), st = [:testhist :testhist2], layout = 2)
```

image

### DIFF
--- a/src/args.jl
+++ b/src/args.jl
@@ -1536,7 +1536,13 @@ function _replace_linewidth(d::KW)
     end
 end
 
-function _add_defaults!(d::KW, plt::Plot, sp::Subplot, commandIndex::Int)
+function _slice_kw!(d::KW, commandIndex::Int)
+    for (k,v) in d
+        slice_arg!(d, d, k, v, commandIndex, false)
+    end
+end
+
+function _add_defaults!(d::KW, commandIndex::Int)
     # add default values to our dictionary, being careful not to delete what we just added!
     for (k,v) in _series_defaults
         slice_arg!(d, d, k, v, commandIndex, false)
--- a/src/args.jl
+++ b/src/args.jl
@@ -1536,7 +1536,7 @@ function _replace_linewidth(d::KW)
     end
 end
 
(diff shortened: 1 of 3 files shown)


mkborregaard:
Great @apalugniok , I'll have a look. Would you mind rebasing this on current master, so testing works?

apalugniok:
@mkborregaard should normal user defined series recipes have access to `plotattributes`?

mkborregaard:
yes, they do have that, and it is used in some cases (like in the Spy recipe)

apalugniok:
I've been giving this issue some thought and I cant think of any way that would fix this without either having a `plotattributes` which doesn't contain the default values in a series recipe or somehow having to change the way recipes are handled 😟

mkborregaard:
Me neither.

daschw:
I think that's superseded by #2373

PR #1704, opened 2018-08-28
Reinstate scatter-based spy implementation; fix #1682

### LINKED ISSUE #1682: `spy` function not working on large sparse matrices
I'm not sure it has been mentioned here but I could not find a similar previous issue. The `spy` function segment-faults when the size is too big for it (on my system).
MWE that does not work for me (v0.7): 

```julia
using SparseArrays, LinearAlgebra, Plots
n = 10000 # or anything more
A = sprand(n,n,10/n) + sparse(Diagonal(rand(n))) ;
spy(A)
```

The MWE above gives me the following error:

```
signal (11): Segmentation fault: 11
in expression starting at no file:0
```

mkborregaard:
Turns out to be much faster for very big matrices. And - there's always `heatmap` for those that prefer that style.

### DIFF
--- a/src/recipes.jl
+++ b/src/recipes.jl
@@ -1042,7 +1042,7 @@ end
         markershape := :circle
     end
     if plotattributes[:markersize] == default(:markersize)
-        markersize := 1
+        markersize := 1.5
     end
     markerstrokewidth := 0
     marker_z := zs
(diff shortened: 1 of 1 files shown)


mkborregaard:
@briochemc

mkborregaard:
```julia
using SparseArrays, Plots
n = 1000_000
a = sprand(n, n, 1/100n)
spy(a)
@time display(spy(a))
# 7.722475 seconds (8.27 M allocations: 420.252 MiB, 3.92% gc time)
```
skaermbillede 2018-08-28 kl 10 02 41

PR #1729, opened 2018-09-06
gr: use new DPI setting for SVG output

jheinen:


src/backends/gr.jl
```diff
@@ -544,6 +544,9 @@ function gr_display(plt::Plot, fmt="")
 
     _gr_thickness_scaling[1] = plt[:thickness_scaling]
     dpi_factor = plt[:dpi] / Plots.DPI
+    if fmt == "svg" then
```
mkborregaard: there's a stray `then` here which breaks this code.

mkborregaard:
Hi @jheinen I can see you already merged this, but there's an issue with the Atom integration, where the size of the plot no longer adapts to the plot pane in Atom. @pfitzseb is that best to fix from the Atom side do you think?

src/backends/gr.jl
mkborregaard: I've pushed a fix for this now.

jheinen:
The output is ok both in Jupyter and Atom. The `width` and `height` attributes in the SVG header don't change, only the `viewBox` (and all internal coordinates - of course) have a higher resolution (x4)

mkborregaard:
This is how `heatmap(5,5)` looks in my Atom on master
skaermbillede 2018-09-07 kl 15 51 04

jheinen:
Not for me - are you really using the 0.34.0 run-time?

mkborregaard:
probably not, though I did the ENV+build thing

mkborregaard:
Let me try to checkout GR master - maybe we should up the `REQUIRE` again?

mkborregaard:
It worked on the new master :heart: So OK, let's update the GR require version and push a Release once the GR metadata one gets merged.

mkborregaard:
Or, no, there's still the concern that @daschw mentioned?

daschw:
> The width and height attributes in the SVG header don't change, only the viewBox (and all internal coordinates - of course) have a higher resolution (x4)

I guess that is alright for SVGs then, so my concern should be no problem with the new GR run-time.

Issue #1721, opened 2018-09-05
Can't display multiple plots (errors with `plotlyjs()`)


kylecarbon:
Hi! First off, I wanted to say thanks for `Plots.jl` :) It's been a joy to use and I've particularly enjoyed `RecipesBase.jl`. However, I am running into some errors when I try plotting multiple instances.

#### Ideal behaviour
This works when I use PlotlyJS directly.
```
using PlotlyJS
x=1:10; plot(x)
plot(x)
```
I get two separate windows, each displaying a plot. In my use-case, this is ideal, as I'm typically viewing multiple plots at a time when analysing data. If there's a better way, I'm happy to hear it :)

#### Behaviour with `Plots` and `pyplot()` as the backend
```
using Plots
pyplot()
x=1:10; plot(x)
plot(x)
```
Only one window is ever displayed.

#### Behaviour with `Plots` and `plotlyjs()` as the backend
```
using Plots
plotlyjs()
x = 1:10
plot(x)
plot(x)

ERROR: type SyncPlot has no field view
Stacktrace:
 [1] getproperty(::Any, ::Symbol) at ./sysimg.jl:18
 [2] _create_backend_figure(::Plots.Plot{Plots.PlotlyJSBackend}) at /home/kyle/.julia/packages/Plots/AYsQt/src/backends/plotlyjs.jl:15
 [3] #invokelatest#1 at ./essentials.jl:686 [inlined]
 [4] invokelatest at ./essentials.jl:685 [inlined]
 [5] _plot_setup(::Plots.Plot{Plots.PlotlyJSBackend}, ::Dict{Symbol,Any}, ::Array{Dict{Symbol,Any},1}) at /home/kyle/.julia/packages/Plots/AYsQt/src/pipeline.jl:215
 [6] _plot!(::Plots.Plot{Plots.PlotlyJSBackend}, ::Dict{Symbol,Any}, ::Tuple{UnitRange{Int64}}) at ./logging.jl:313
 [7] #plot#132(::Base.Iterators.Pairs{Union{},Union{},Tuple{},NamedTuple{(),Tuple{}}}, ::Function, ::UnitRange{Int64}) at /home/kyle/.julia/packages/Plots/AYsQt/src/plot.jl:57
 [8] plot(::UnitRange{Int64}) at /home/kyle/.julia/packages/Plots/AYsQt/src/plot.jl:51
 [9] top-level scope at none:0
```
Bug?

#### Extra info
```
[...]

PR #1730, opened 2018-09-06
Fix PlotlyJS backend and prepare for PlotDocs update

daschw:
I had to move all the supported attribute dicts out of the (`@required`) backend code so that they are available for PlotDocs to build the html tables.


sglyon:
hmm I can't get this to render with plotlyjs backend from REPL. Are you able to get it to work there, or perhaps you are testing in a different environment?

daschw:
You're right, I did not test thoroughly in the REPL. It works perfectly in Atom, but there we only display a png. In the REPL for me only the first plot works, unfortunately (as reported in https://github.com/JuliaPlots/Plots.jl/issues/1721#issuecomment-418620996).

BoundaryValueProblems:
@daschw and @sglyon , I am wondering what is the status of the ```plotlyjs``` backend of ```Plots.jl```. With today's update of ```PlotlyJS.jl``` (v0.11.2), now ```PlotlyJS.jl``` genuine functions work under Juno/Atom. Yet, as the backend of ```Plots.jl``` (v0.20.5), it still doesn't work under Juno/Atom. Under REPL using Terminal, it doesn't work either. In both case, it generates the following error: ```ERROR: type SyncPlot has no field view```, especially when I try to plot something twice. Thanks!

daschw:
I did not proceed here. PlotlyJS still is not working from the REPL.
Still, I will go ahead and merge the changes from here in a new clean PR.
We will have to fix plotlyjs in a separate PR.

PR #1731, opened 2018-09-07
rudimentary missings support

mkborregaard:
Only works for the positional args, but that should cover very many cases

### DIFF
--- a/src/series.jl
+++ b/src/series.jl
@@ -21,9 +21,10 @@ convertToAnyVector(n::Integer, plotattributes::KW) = Any[zeros(0) for i in 1:n],
 
 # numeric vector
 convertToAnyVector(v::AVec{T}, plotattributes::KW) where {T<:Number} = Any[v], nothing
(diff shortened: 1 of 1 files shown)


src/series.jl
```diff
@@ -21,18 +21,25 @@ convertToAnyVector(n::Integer, plotattributes::KW) = Any[zeros(0) for i in 1:n],
...
 function convertToAnyVector(v::AMat, plotattributes::KW)
+    v = handlemissingsinmatrix(v)
     if all3D(plotattributes)
         Any[Surface(v)]
     else
         Any[v[:,i] for i in 1:size(v,2)]
     end, nothing
 end
 
+handlemissingsinmatrix(v::AMat) = v
```
daschw: Why not just `handlemissings`?

mkborregaard: no reason, I guess just to signal that it's a very local function just for use above. Happy to rename.

mkborregaard: done

daschw: Great, I would have been ok with `handlemissindsinmatrix` too though. I was just curious and wondering, if you were also planning to implement something like `handlemissingsinvector` or similar too, and in that case I would have suggested dispatching on the types.

PR #1741, opened 2018-09-13
Add option to change guide positions

Godisemo:
This PR adds the option to change guide position in the PyPlot backend
```julia
plot(rand(10), yguide="rand", xguide="index", yguide_position=:right, xguide_position=:top)
```
test

### DIFF
--- a/src/arg_desc.jl
+++ b/src/arg_desc.jl
@@ -111,6 +111,7 @@ const _arg_desc = KW(
 
 # axis args
 :guide     				  => "String. Axis guide (label).",
+:guide_position		  => "Symbol. Position of axis guides: :top, :bottom, :left or :right",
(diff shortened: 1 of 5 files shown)


mkborregaard:
Great! We try to avoid adding keywords that only map to a single backend - would you mind having a look at implementing this for GR, plotly(js) and pgfplots as well?

Godisemo:
Sure, I'll see if I can figure out how to do it in other backends.

Godisemo:
Looking at Plotly I think the way to do this is to add another axis which only has the label, and then change the `side` argument. I'm not sure how to deal with multiple axis though.

For PGFPlots I'm struggling to install the backend for Julia v0.7 and v1.0.

mkborregaard:
Awesome!

Godisemo:
After taking a closer look at Plotly I now believe this is easier to achieve using annotations, but this is not a task I will be able to perform at the moment. I think it is better left for someone more familiar with that backend.

mkborregaard:
Awesome, thanks for the effort! The only thing left then is to document the new arg in arg_desc.jl

mkborregaard:
Thanks @Godisemo !

PR #1763, opened 2018-09-26
Changes to get glvisualize backend running on v0.7

lwabeke:
Only tested basic line plot, scatter, heatmap and contour

Now seems to work both in Repl and from IJulia.

### DIFF
--- a/REQUIRE
+++ b/REQUIRE
@@ -14,3 +14,4 @@ NaNMath
 Requires
 Contour
 GR 0.34.0
+GLWindow
--- a/src/backends/glvisualize.jl
+++ b/src/backends/glvisualize.jl
@@ -1,3 +1,12 @@
+using GLWindow
+using GeometryTypes
+import GeometryTypes.Vec3f0, GeometryTypes.Point2f0, GeometryTypes.Vec2f0, GeometryTypes.Point
+using Reactive
+using GLAbstraction
+using GLVisualize
+using FixedPointNumbers
+using FileIO
+using Contour
 #=
 TODO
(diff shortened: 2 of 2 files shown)


mkborregaard:
It shouldn't add a dep on GLWindow. All those packages would have to be eval-imported when activating the backend.

SimonDanisch:
I've seen you opened commits at GLWindow/GLVisualize/GLAbstraction... Did you see that I deprecated GLVisualize in favor of [Makie](https://github.com/JuliaPlots/Makie.jl) ?
Sorry if this isn't very visible...
Also Makie has it's own version of [GLAbstraction & GLVisualize](https://github.com/JuliaPlots/Makie.jl/tree/master/src/glbackend) in it. So maybe the easiest way to get the GLVisualize backend up to date would be - to use a Makie backend :P

mkborregaard:
@SimonDanisch is it time to deprecate the glvisualize backend to Plots altogether?

SimonDanisch:
>is it time to deprecate the glvisualize backend to Plots altogether?

Let's do such things at Vizcon :)

lwabeke:
I missed the Makie notice initially (only saw it after having made all the changes), but also don't see thre Makie backend for the Plots library.

daschw:
[Makie](https://github.com/JuliaPlots/Makie.jl) is a different package, not a Plots backend. It will most likely be the Plots successor and also support different backends. Plots will no longer support the glvisualize backend.

mkborregaard:
That backend has now been formally deprecated. Sorry @lwabeke

PR #1765, opened 2018-09-26
honour `html_output_format` in Juno

pfitzseb:
Requires https://github.com/JunoLab/Atom.jl/commit/a939fcf908aa5afb6b2ef4465430e4341502a78f.

### DIFF
--- a/src/output.jl
+++ b/src/output.jl
@@ -278,7 +278,7 @@ end
 
 function _showjuno(io::IO, m::MIME"image/svg+xml", plt)
   if Symbol(plt.attr[:html_output_format]) ≠ :svg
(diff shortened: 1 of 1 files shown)


pfitzseb:
Yeah, I think that's the best we can do.
For reference: https://github.com/JuliaLang/METADATA.jl/pull/18437

pfitzseb:
Ok, new Atom.jl version is out, so this should be good to go.

mkborregaard:
Does the atom plotting depend on this, i.e. do we need to tag a new version immediately?

pfitzseb:
No, the new Atom version doesn't introduce any problems without this patch.

Would still be nice to get this out so people can plot many data points in Atom tho :)

Issue #1074, opened 2017-09-05
plotlyjs() in atom: window scaling issue


tkind:
Using plotlyjs() with Plots.jl within Atom leads to the screenshot attached: The plot does not scale to the window size. In contrast, using PlotlyJS.jl works fine. Has someone else faced that issue? 
Note: Manually resizing the window leads to the correct sizing temporaneously. After plotting it again the problem persists. 
![screen shot 2017-09-05 at 15 30 24](https://user-images.githubusercontent.com/17366230/30081610-bd686e3e-927f-11e7-844e-26ca03f15c0d.png)

mkborregaard:
Can you post your `versioninfo()` and `Pkg.status.(["Plots", "PlotlyJS", "Juno"])`? It behaves differently on my machine.

tkind:
Thank you for the quick reply! I get the following two results:
`Julia Version 0.6.0`
`Commit 903644385b (2017-06-19 13:05 UTC)`
`Platform Info:`
  `OS: macOS (x86_64-apple-darwin13.4.0)`
  `CPU: Intel(R) Core(TM) i7-7920HQ CPU @ 3.10GHz`
  `WORD_SIZE: 64`
  `BLAS: libopenblas (USE64BITINT DYNAMIC_ARCH NO_AFFINITY Prescott)`
  `LAPACK: libopenblas64_`
  `LIBM: libopenlibm`
 ` LLVM: libLLVM-3.9.1 (ORCJIT, broadwell)`
` - Plots                         0.12.4`
` - PlotlyJS                      0.6.4`
` - Juno                          0.3.0`
I apologise if I am I missing something very obvious. Again, thank you very much for your help.

mkborregaard:
Ha, that is in fact identical to my system. I'll have a look again :-)

mkborregaard:
I have this behaviour too. Interactively resizing the window makes the plotlyjs plot fit to it, but every new plot call resets the size again. There must be some miscommunication between Plots and Juno. Tagging @pfitzseb for the possibility that this is Juno side (not likely since this is only in PlotlyJS).

tkind:
Fantastic, I am not alone - the behaviour you describe fits exactly to my problem:) I do not know how Plots and Juno communicate, but if it is Juno, I wonder what the package PlotlyJS.jl does differently compared to Plots.jl+plotlyjs(). Using simply
`using PlotlyJS`
`plot(rand(10))` 
[...]

PR #1775, opened 2018-10-03
Allow plotly plots to auto-resize to fill the window.

NHDaly:
Fixes https://github.com/JuliaPlots/Plots.jl/issues/1074.

I think @kylecarbon hit the nail on the head with the commit he found in this comment:
https://github.com/JuliaPlots/Plots.jl/issues/1074#issuecomment-418738280

I just basically reimplemented that same commit for the Plots repo. That said, would it be possible to simply delegate to the Plotly implementation of these functions rather than copy/pasting like this, so we don't get left behind again?

----------------------------------
The changes: 

 - Remove w and h from plotly_layout
 - Add window.onresize function to produced javascript, calling
    Plotly.Plots.resize (copied from this commit):
    https://github.com/sglyon/PlotlyBase.jl/commit/ab0e5c4f9291275f9685c58a26fbfeb184f661a3

### DIFF
--- a/src/backends/plotly.jl
+++ b/src/backends/plotly.jl
@@ -309,7 +309,6 @@ function plotly_layout(plt::Plot)
     plotattributes_out = KW()
 
     w, h = plt[:size]
-    plotattributes_out[:width], plotattributes_out[:height] = w, h
     plotattributes_out[:paper_bgcolor] = rgba_string(plt[:background_color_outside])
     plotattributes_out[:margin] = KW(:l=>0, :b=>20, :r=>0, :t=>20)
 
@@ -892,27 +891,41 @@ end
 
 function html_body(plt::Plot{PlotlyBackend}, style = nothing)
     if style == nothing
-        w, h = plt[:size]
-        style = "width:$(w)px;height:$(h)px;"
+        style = "width:100%;height:100%;"
     end
     uuid = UUIDs.uuid4()
-    html = """
-        <div id=\"$(uuid)\" style=\"$(style)\"></div>
-        <script>
-        PLOT = document.getElementById('$(uuid)');
-        Plotly.plot(PLOT, $(plotly_series_json(plt)), $(plotly_layout_json(plt)));
-        </script>
     """
-    html
-end
-
-function js_body(plt::Plot{PlotlyBackend}, uuid)
-    js = """
-          PLOT = document.getElementById('$(uuid)');
-          Plotly.plot(PLOT, $(plotly_series_json(plt)), $(plotly_layout_json(plt)));
+    <div id=\"$(uuid)\" style=\"$(style)\"></div>
+    <script>
+        $(js_body(plt, uuid))
+     </script>
     """
 end
 
+function js_body(plt::Plot{PlotlyBackend}, divid)
+    """
+    gd = (function() {
+      var WIDTH_IN_PERCENT_OF_PARENT = 100;
+      var HEIGHT_IN_PERCENT_OF_PARENT = 100;
+      var gd = Plotly.d3.select('body')
+        .append('div').attr("id", "$(divid)")
+        .style({
+          width: WIDTH_IN_PERCENT_OF_PARENT + '%',
+          'margin-left': (100 - WIDTH_IN_PERCENT_OF_PARENT) / 2 + '%',
+          height: HEIGHT_IN_PERCENT_OF_PARENT + 'vh',
+          'margin-top': (100 - HEIGHT_IN_PERCENT_OF_PARENT) / 2 + 'vh'
+        })
+        .node();
+      var data = $(plotly_series_json(plt));
+      var layout = $(plotly_layout_json(plt));
+      Plotly.newPlot(gd, data, layout);
+      window.onresize = function() {
+        Plotly.Plots.resize(gd);
+      };
+      return gd;
+    })();
+    """
+end
 
 # ----------------------------------------------------------------
 


NHDaly:
Also, per the documentation: https://plot.ly/javascript/responsive-fluid-layout/, we can set `{responsive: true}` instead of adding a custom `onresize` function. I've tested that that works as well, so whichever is probably fine.

mkborregaard:
I don't know very much about plotlyjs, so I'll leave it to those capable to review this. But just making sure, what is the behaviour here when Plots opens e.g. a blink window? Is the `size` attribute always ignored, or is it just on plotting to an existing window?

NHDaly:
Ah, sorry, i don't fully understand your question. How do you get Plots to open a blink window?
(I would _love_ to have it do that, but i didn't see that as an option, so i've rolled that myself with the following code:
```julia
function displayPlotInWindow(p)
    global plotWindow
    if (!@isdefined(plotWindow) || !Blink.active(plotWindow))
        # Initialize the plot window the first time this is called.
        plotWindow = Blink.Window(Blink.@d(:width=>900, :height=>750), async=false)
        Blink.@js plotWindow Plotly = require($(Plots._plotly_js_path));
    end
    # Convert the plot to web format (HTML+SVGJS)
    buf = IOBuffer()
    Base.invokelatest(show, buf, MIME("text/html"), p);
    plothtml = String(take!(buf))

    # Send to plotWindow, _asynchronously_, since we don't care about clobbering these.
    Blink.body!(plotWindow, plothtml, fade=false, async=true);
end
```
)

> Is the size attribute always ignored, or is it just on plotting to an existing window?

That said, I think your answer is that yeah, no, the size is just completely ignored after this PR. But as far as I can tell, it wasn't used anyway, since from what I can see, the plotly() backend is _always_ plotting to an existing window: either a Juno tab, or just opening a new tab in the broswer locally. Is that wrong?

NHDaly:
> `        plotWindow = Blink.Window(Blink.@d(:width=>900, :height=>750), async=false)`

Oh, but I could be using the size here, and maybe i'll start doing that! :D

asinghvi17:
Bump @NHDaly

NHDaly:
Hi, sorry for my delay in return. This is not much of a priority for me, these days.. :/


Regarding the size thing, @mkborregaard  I never quite followed your question. You asked:

> But just making sure, what is the behaviour here when Plots opens e.g. a blink window?

But I'm missing something about the premise of your question: is it possible to get Plots to open a `Blink` window? By default, i think plotly current renders to an _existing_ window, or if there isn't one it opens a new tab in your browser? In both cases, the new behavior would just fill the entire window.

But so, yes, the `size` attribute on the plot is currently 100% unused.

I'll look at rebasing now.

NHDaly:
So yeah, i just verified that for new tab plots, it just fills the whole screen:
```julia
julia> using Plots
[ Info: Precompiling Plots [91a5bcdd-55d7-5caf-9e0b-520d859cae80]

julia> plotly()
[ Info: For saving to png with the Plotly backend ORCA has to be installed.
Plots.PlotlyBackend()

julia> plot(sin, 0:0.1:2π)
```
Screen Shot 2019-12-31 at 1 43 35 PM

Which, admittedly, is a little awkward..

The old behavior was to stick to the size of the plot (which as also sort of awkward):
Screen Shot 2019-12-31 at 1 45 50 PM


------------------

So maybe this needs to be controlled with an argument somehow? Like maybe by setting `:size` to like `"stretch"` or something? I'm not sure.

But anyway, this is the current state of things! :) If someone wants to make more changes, it might be best to take over the PR, since I don't have a lot of time to work on this right now.

Happy new year! :) 🎉 Cheers!

vladimmi:
Any chance to this finally happen?

mkborregaard:
Thanks for bumping @vladimmi . The issue being I think that the plotly backend doesn't really have a maintainer, so something like this easily gets lost.
Maybe we should just merge this? @sglyon would you mind having a look?

vladimmi:
Probably worth to note: I've used this patch locally and made changes to plotly backend to test whether that would help. It definitely helps but something is wrong with sizing - plot seem to be too big in height and Plots pane in Juno gets permanent vertical scroll bar.

sglyon:
@mkborregaard let's see if @vladimmi is able to come up with a fix!

Happy to review after your investigation @vladimmi

vladimmi:
Here is what I mean with permanent scroll: as you can see, plot content is just a bit higher than it's container and plot pane always has vertical scroll bar:

image

There reason of it: src/backend/plotly.jl:851 (if you watch code through "changed files" in PR) - it uses `100vh` as height specifier with `vh` being `viewport height` units. Those units are calculated relative to viewport size and ignore any styles on parent items. `body` by default has margin that adds to content height - and that way it become a bit bigger than a container.

That can be fixed by setting `margin: 0` on `body` - for example, on line 847 when we already have selected `body` element. I would also add `overflow: hidden` there to prevent scrollbars flashing on pane resize. Something like that:

```jl
var gd = Plotly.d3.select('body')
    .style({
        margin: 0,
        overflow: 'hidden'
    })
    .append('div').attr("id", "$(divid)")
    # ...etc
```

As side effect that also removes spacing between pane borders and plot:

image

Also noticed that current code generates unneeded element. Line 834 adds `div` element with set ID and `script` element right after it. But script itself generates new `div` with same ID (which is technically incorrect because ID should be unique) at line 847 and uses it, so first one can be removed.

vladimmi:
With such changes I'm a bit worried that in that way code assumes "exclusive" ownership of provided container and changes styles of `body` that wasnt created by plotly but provided by outer GUI. Generally, whole PR code assumes that code has a `body` that was made specifically for it and attempt to fill it, not some abstract parent element. In theory if plots output with plotly backend would be put not into `iframe` but some generic `div` on some page, such plot would go up to top `body`, append to there and cover whole page with existing elements...

vladimmi:
What is worse: if plot would be done to fill only parent element where it was placed, it wouldnt actually fill plots pane in Juno - it seems that pane `iframe` lacks needed styles on `html` element and it doesnt fill whole pane itself...

vladimmi:
As an idea that can both make plots fill Juno panes and doesnt break other possible use-cases (Jupiter sheets or wherever else Julia and Plots can be used): is it possible for plots to add some keyword argument that would be meaningful for plotly(js) backend? And by default plots would have fixed size (like without PR) but with something explicit like `fillViewport = true` it would fill whole available window.

NHDaly:
sorry that I disappeared. Unfortunately, I don't have time to contribute to this anymore; please feel free to take over this PR, and/or clone it to a new one. If you don't have edit access, i'll gladly approve any PRs to merge into this branch if you just want to push up a few commits.

Thanks in advance and sorry for the inconvenience! ❤️

BeastyBlacksmith:
> As an idea that can both make plots fill Juno panes and doesnt break other possible use-cases (Jupiter sheets or wherever else Julia and Plots can be used): is it possible for plots to add some keyword argument that would be meaningful for plotly(js) backend? And by default plots would have fixed size (like without PR) but with something explicit like `fillViewport = true` it would fill whole available window.

That would be possible using `extra_kwargs` #2648

PR #1777, opened 2018-10-04
use remote urls plotly

SimonDanisch:


SimonDanisch:
>When you say optional, is the default that it will not embed the massive script?

I preserved the default (downloading on build without ENV), assuming that it became the default for a reason! If no one objects, I surely can flip the default ;)

mkborregaard:
The reason was that originally Plots had no backend dependencies, so it installed a local within-Plots plotly to have plotting functionality even for users that didn't have a backend installed. By popular demand we recently decided to depend on GR and make that the default backend, so there is no reason anymore that plotly should be initialised on Plots launch.

SimonDanisch:
>works on both master and this branch in IJulia. So I'm not sure what you mean with "IJulia doesn't work"

Weird, for me it says in the js console `Plotly` not defined on both versions - so it seems like the import didn't get through. Where did you reproduce this?

daschw:
I agree with @mkborregaard that we can avoid initialising plotly by default.
@SimonDanisch I ran into this "`Plotly` not defined" error on IJulia also the first time I tried, but after building Plots it worked. If by "where" you mean on which operating system, it's linux :)

SimonDanisch:
@daschw this is on current master:
image

SimonDanisch:
Ok, made it the default to require plotly from the cdn! I also fixed IJulia and removed the madness, that plotly would insert 16mb of js into a string and then display that at precompile time!

jlperla:
Is this working?  If so, it would be great to have a solution to this issue tagged.

SimonDanisch:
Since Plotly seemed to be bugged by platform issues, I was waiting for someone to test this ;) But might as well wrap this up

daschw:
I can test again today in the evening and report back.

SimonDanisch:
@daschw, so you can now reproduce the failure on master, but my fix in here doesn't actually fix Plotly for you!? :( I think we should really figure this one out - there should be no good reason why requirejs is so flaky!

SimonDanisch:
Maybe there is a difference in the requirejs version?

daschw:
> @daschw, so you can now reproduce the failure on master, but my fix in here doesn't actually fix Plotly for you!?

Exactly, for `plotly()`, I now get the `Plotly nod defined` error in IJulia on both, `master` and `sd-remote` on  my archlinux system with julia 1.0.1 using Firefox.
On my windows 10 system with julia 1.0.0 using Firefox I get an empty `Out[*]` cell using `plotly()` in IJulia, also both, on `master` and `sd-remote`.

Anything I can do to help figuring this out? How do I check my `requirejs` version?

jlperla:
I just want to point out that this issue (the massive jupyter notebook sizes) is affecting people **not using plotly**.

In fact, I have never personally seen the plotly backend work on v1.0.  So if it is possible to decouple integrating this fix soon from the debugging of the actual plotly backend, it would be very helpful for GR users.

daschw:
I think @jlperla makes a good point. Let's figure this out somewhere else!

PR #1782, opened 2018-10-08
Add support for annotations and "Shape"s.

ma-laforge:
Also build plot attributes list after series are read in
to avoid having the internal mechanisms overwrite data.


daschw:
Thanks a lot!

PR #1787, opened 2018-10-13
GR: use correct attribute for axis color

greimel:
```julia
using Plots
using StatPlots, DataFrames

test = DataFrame(
    x = rand(20),
    y = rand(20),
    grp = rand(["g 1"; "g 2"], 20)
 )

gr()
@df test scatter(:x, :y, group=:grp, xlab="x", ylab="y",
                  foreground_color_axis=:blue,
                  foreground_color_border=:green)
```
Before:

scatter_gr_before

After:
scatter_gr_after

Now it matches behavior for `pyplot()`, I can't reproduce right now, because my `PyCall.jl` doesn't build.

This will be helpful for JuliaPlots/PlotThemes.jl/pull/30

As an aside, the attribute names could be changed so something more intuitive, IMO.

### DIFF
--- a/src/backends/gr.jl
+++ b/src/backends/gr.jl
@@ -855,12 +855,12 @@ function gr_display(sp::Subplot{GRBackend}, w, h, viewport_canvas)
 
         # axis lines
         if xaxis[:showaxis]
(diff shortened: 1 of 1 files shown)


daschw:
Thanks @greimel !

PR #1788, opened 2018-10-13
[WIP] Treat major and minor ticks differently

greimel:
I am working on a theme that resembles the default theme of `ggplot2` (JuliaPlots/PlotThemes.jl/pull/30).

The goal is this:
line_gg

Note that `ggplot2` doesn't show ticks for minor grid lines. This is not currently possible in plots.

The first commit partly solves the problem for `gr()`.

scatter_gr_minor

I compute the minor grids separately and assign them the color of the border. This would in principle solve my problem. However, I do not like this fix yet. I would like to add a new attribute `foreground_color_minor_ticks` or similar to set this color. I tried to do this, but got an error that I do not understand. Unfortunately I know too little about `Plots.jl`.

Could someone look at my last commit and guide me a bit in proceeding here?

PS: I propose to introduce an attribute `foreground_color_major_ticks` as well. It's not yet clear to me what the exact purpose of `foreground_color_axis` and `foreground_color_border` are (see my previous PR #1787)


greimel:
```
plot(rand(10), rand(10), foreground_color_tick = :green, foreground_color_minortick=:red, minorgrid=true)
```
gr-tick-colors

As laid out in #1794 I could delete the `foreground_color_axis` attribute and replace it by  `foreground_color_tick` in all backends--if you think that this is a good idea.

src/args.jl
```diff
@@ -1283,6 +1285,8 @@ const _match_map2 = KW(
     :fontfamily_subplot       => :fontfamily,
     :tickfontfamily           => :fontfamily_subplot,
     :guidefontfamily          => :fontfamily_subplot,
+    :foreground_color_tick                => :foreground_color_subplot,
+    :foreground_color_minortick           => :foreground_color_subplot
```
daschw: I would add these to `_match_map` instead and let them default to `:foreground_color_axis` (which then defaults to `:foreground_color_subplot`). Then you should be able to have different tick colors for different axis.

mkborregaard: hey @greimel would you mind doing this minor change so we can get this merged? Thanks!

greimel:
Sorry for not finishing this PR. I've not used Plots.jl for a year or two. Please feel free to take the code or close the PR.

BeastyBlacksmith:
It was decided in the last maintainer call (https://github.com/JuliaPlots/Plots.jl/issues/3538) that rebasing this PR is not worth the effort and that we'd need a reimplementation of this on the current state of Plots.

isentropic:
Totally missed this maintainers' call, is there a reminder thread or something? @BeastyBlacksmith

BeastyBlacksmith:
You should have got an email. If not hit me on zulip.

PR #1805, opened 2018-10-23
function recipe: taking axis scales into account

### LINKED ISSUE #1604: Plotting functions with xaxis=:log
Plotting functions with a logarithmic x axis fails, as in this example:

    h(x) = x
    plot(h,xaxis=:log)

It tries to evaluate the log at negative arguments.

yha:
Take axis scales into account for choosing default range and sampling grid in the function recipes.

Before:
```
plot(x->exp10(10sin(x)),m=true,yscale=:log10)
```
image
after:
image

Also fixes #1604, as the default range for, e.g., `:log2` scale, is now `[2^-5,2^5]`.

### DIFF
--- a/src/series.jl
+++ b/src/series.jl
@@ -386,8 +386,9 @@ end
     xmin, xmax = try
         axis_limits(plt[1][:xaxis])
     catch
-        xm = tryrange(f, [-5,-1,0,0.01])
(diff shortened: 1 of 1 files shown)


daschw:
Thanks @yha!

Issue #1781, opened 2018-10-05
GR3 error (gr3.c, l. 153): init failed 


JonathanAnderson:
I just had an issue trying to use GR directly (https://github.com/jheinen/GR.jl/issues/159), which I was able to fix by calling `surface` with `accelerate=false`.

I tried doing this here and tried reading through `backend/gr.jl` and couldn't find an obvious way to add this attribute so I thought I would register an issue.

As far as timing, as soon as I figured out the `GR.jl` issue, I immediately tried it here. So the setup is the same. I am `ssh -Y`-ing from a mac into a centos server running julia. Xsession is configured and works. 

```
               _
   _       _ _(_)_     |  Documentation: https://docs.julialang.org
  (_)     | (_) (_)    |
   _ _   _| |_  __ _   |  Type "?" for help, "]?" for Pkg help.
  | | | | | | |/ _` |  |
  | | |_| | | | (_| |  |  Version 1.0.1 (2018-09-29)
 _/ |\__'_|_|_|\__'_|  |  Official https://julialang.org/ release
|__/                   |

(v1.0) pkg> test Plots
   Testing Plots
 Resolving package versions...
    Status `/tmp/tmpllPfii/Manifest.toml`
...
  [28b8d3ca] GR v0.34.1
...
  [91a5bcdd] Plots v0.20.4
...
Test Summary: | Pass  Total
GR            |   35     35
Test Summary: | Pass  Total
UnicodePlots  |    3      3
Test Summary: | Pass  Total
Axes          |    6      6
Test Summary: |
NoFail        | No tests
   Testing Plots tests passed

julia> using Plots

julia> gr()
Plots.GRBackend()

julia> surface(rand(100,40))
Error showing value of type Plots.Plot{Plots.GRBackend}:
ERROR: GR3 error (gr3.c, l. 153): init failed
Stacktrace:
 [1] _check_error() at /home/janders/.julia/packages/GR/joQgG/src/gr3.jl:73
 [2] surface(::UnitRange{Int64}, ::UnitRange{Int64}, ::Array{Float64,1}, ::Int64) at /home/janders/.julia/packages/GR/joQgG/src/gr3.jl:379
 [3] gr_display(::Plots.Subplot{Plots.GRBackend}, ::Measures.Length{:mm,Float64}, ::Measures.Length{:mm,Float64}, ::Array{Float64,1}) at /home/janders/.julia/packages/Plots/UtVfv/src/backends/gr.jl:1088
[...]

PR #1808, opened 2018-10-24
Fallback to non gr3 surface call

JonathanAnderson:
Taking an idea from @FlorianRhiem in this issue https://github.com/jheinen/GR.jl/issues/159, I figured I would fall back to GR if GR3 had an error. With this change, I was able to solve my issue in https://github.com/JuliaPlots/Plots.jl/issues/1781


daschw:
Thanks @JonathanAnderson !

PR #1851, opened 2018-11-22
Feature Request: Add `showlegend` option to allow a series to opt out of being in the legend

NHDaly:
We wanted this when we had three subplots all sharing a legend, and didn't want to have the same keys show up in the shared legend three times. So we have one of the series on one of the subplots set `showlegend = true`, and the others set `showlegend = false`.

Does this seem like a useful option to include? I named it `showlegend` to match the _plotly_ setting with the same name, but I actually think maybe something like `addtolegend` or `include_in_legend` or something _might_ be better?

Anyway, I just wanted to push this up to get your feedback on it, since we found it useful locally in our work! :)

### DIFF
--- a/src/args.jl
+++ b/src/args.jl
@@ -279,6 +279,7 @@ const _series_defaults = KW(
     :series_annotations => nothing,       # a list of annotations which apply to the coordinates of this series
     :primary            => true,     # when true, this "counts" as a series for color selection, etc.  the main use is to allow
                                      #     one logical series to be broken up (path and markers, for example)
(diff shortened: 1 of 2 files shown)


mkborregaard:
This functionality already exists but may not be sufficiently documented.
Plots has a `primary` option - setting that to `false` will exclude it from the legend and also opt it out of the default color cycle (it will have the same color as the latest preceding series.
If you just want to opt out of the legend (rarer I think) you set `label = ""` - which admittedly may not be intuitive if you don't know it.

mkborregaard:
I'd welcome a PR to improve the documentation.

PR #1854, opened 2018-11-28
Fix #1799

### LINKED ISSUE #1799: 'c' argument issue in pyplot
With this code:
```julia
using Plots
pyplot()
p=plot(rand(2), rand(2), rand(2), seriestype=:path)
plot!(p, [1.0], [0.5], [1.0], seriestype=:scatter)
```
I get this warning:
```
'c' argument looks like a single numeric RGB or RGBA sequence, which should be avoided as value-mapping will have precedence in case its length matches with 'x' & 'y'.  Please use a 2-D array with a single row if you really want to specify the same RGB or RGBA value for all points.
```
It doesn't seem to happen with ``gr()``.  Any insights on what the issue is or how to avoid it without switching to another backend?

Thanks!

mauro3:
This may fix #1799.  Although, note, I have no deeper understanding of this code...

### DIFF
--- a/src/backends/pyplot.jl
+++ b/src/backends/pyplot.jl
@@ -222,17 +222,6 @@ function fix_xy_lengths!(plt::Plot{PyPlotBackend}, series::Series)
     end
 end
 
-# total hack due to PyPlot bug (see issue #145).
-# hack: duplicate the color vector when the total rgba fields is the same as the series length
-function py_color_fix(c, x)
-    if (typeof(c) <: AbstractArray && length(c)*4 == length(x)) ||
-                    (typeof(c) <: Tuple && length(x) == 4)
-        vcat(c, c)
-    else
-        c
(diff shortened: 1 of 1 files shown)


mauro3:
No, this does not work

mkborregaard:
Damn, I was already running the tests (Travis only checks on GR)

mauro3:
Ok, maybe this works. (Also added `"off"->false` to remove other deprecations)

mkborregaard:
Oooh, it's an actual fix now! :-D Let me see if it works

mauro3:
Yep, I had to roll out emacs...

mkborregaard:
It works :heart:

mkborregaard:
Thank you @mauro3 !

daschw:
That's great, thanks @mauro3 !

Issue #1830, opened 2018-11-02
wireframe plot does not work with pyplot backend in Julia v1.0.1


BoundaryValueProblems:
A very simple wireframe plot under pyplot backend produces the following error. The environment is: Julia v1.0.1; Plots v0.21.0; PyPlot v2.6.3, matplotlib v3.0.1. Under gr backend, this works without any error. The situation is the same on Mac OS X 10.11.6 and Windows 10.
```Julia
julia> wireframe(randn(10,10))
Error showing value of type Plots.Plot{Plots.PyPlotBackend}:
ERROR: PyError ($(Expr(:escape, :(ccall(#= /Users/xxx/.julia/packages/PyCall/0jMpb/src/pyfncall.jl:44 =# @pysym(:PyObject_Call), PyPtr, (PyPtr, PyPtr, PyPtr), o, pyargsptr, kw))))) <class 'AttributeError'>
AttributeError('Unknown property vmin')
  File "/Users/xxx/.julia/conda/3/lib/python3.7/site-packages/mpl_toolkits/mplot3d/axes3d.py", line 1879, in plot_wireframe
    linec = art3d.Line3DCollection(lines, *args, **kwargs)
  File "/Users/xxx/.julia/conda/3/lib/python3.7/site-packages/matplotlib/collections.py", line 1224, in __init__
    **kwargs)
  File "/Users/xxx/.julia/conda/3/lib/python3.7/site-packages/matplotlib/collections.py", line 166, in __init__
    self.update(kwargs)
  File "/Users/xxx/.julia/conda/3/lib/python3.7/site-packages/matplotlib/artist.py", line 916, in update
    ret = [_update_property(self, k, v) for k, v in props.items()]
  File "/Users/xxx/.julia/conda/3/lib/python3.7/site-packages/matplotlib/artist.py", line 916, in <listcomp>
    ret = [_update_property(self, k, v) for k, v in props.items()]
  File "/Users/xxx/.julia/conda/3/lib/python3.7/site-packages/matplotlib/artist.py", line 912, in _update_property
    raise AttributeError('Unknown property %s' % k)

Stacktrace:
 [1] pyerr_check at /Users/xxx/.julia/packages/PyCall/0jMpb/src/exception.jl:60 [inlined]
 [2] pyerr_check at /Users/xxx/.julia/packages/PyCall/0jMpb/src/exception.jl:64 [inlined]
[...]

PR #1857, opened 2018-12-02
vmin, vmax in KW were suppressed for :wireframe

BoundaryValueProblems:
This is a potential solution to the wireframe plot under the pyplot backend #1830 . I just suppressed ```vmin, vmax``` in the ```extrakw``` only if ```st == :wireframe```.


mkborregaard:
Hey looks good. I never use wireframe plots but - do you never have the "wires" in color? It seems like if you do this might break that?

mkborregaard:
OK, thanks!

PR #1874, opened 2018-12-27
insert PlotDisplay directly after REPLDisplay

pfitzseb:
This should fix the issue mentioned in https://github.com/JuliaPlots/Plots.jl/pull/1667#issuecomment-450092480.

I'm still planning to open a PR to Base for fixing this properly, but this is a perfectly fine short term solution I think.

### DIFF
--- a/src/init.jl
+++ b/src/init.jl
@@ -1,3 +1,5 @@
+using REPL
+
 const use_local_dependencies = Ref(false)
 
 function __init__()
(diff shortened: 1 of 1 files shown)


mkborregaard:
Awesome :heart:

EricForgy:
Looks like a typo?

https://github.com/JuliaPlots/Plots.jl/blob/master/src/init.jl#L19

`PlotDisplay()` should be `PlotsDisplay()` I think.

Issue #1793, opened 2018-10-16
UndefVarError using PlotlyJS backend


paulduf:
Hello

I'm new to Julia and I'm trying to plot offline using PlotlyJS module.

Basic plot gives me this error. Surely I'm missing some important setup part but I can't debug this (at least not without spending many hours)

```
UndefVarError: _use_remote not defined

Stacktrace:
 [1] _show(::Base.GenericIOBuffer{Array{UInt8,1}}, ::MIME{Symbol("text/html")}, ::Plots.Plot{Plots.PlotlyJSBackend}) at /home/pol/.julia/packages/Plots/Ufx0i/src/backends/plotlyjs.jl:54
 [2] show(::Base.GenericIOBuffer{Array{UInt8,1}}, ::MIME{Symbol("text/html")}, ::Plots.Plot{Plots.PlotlyJSBackend}) at /home/pol/.julia/packages/Plots/Ufx0i/src/output.jl:201
 [3] #sprint#325(::Nothing, ::Int64, ::Function, ::Function, ::MIME{Symbol("text/html")}, ::Vararg{Any,N} where N) at ./strings/io.jl:101
 [4] sprint(::Function, ::MIME{Symbol("text/html")}, ::Vararg{Any,N} where N) at ./strings/io.jl:97
 [5] display_dict(::Plots.Plot{Plots.PlotlyJSBackend}) at /home/pol/.julia/packages/Plots/Ufx0i/src/init.jl:76
 [6] #invokelatest#1 at ./essentials.jl:697 [inlined]
 [7] invokelatest at ./essentials.jl:696 [inlined]
 [8] execute_request(::ZMQ.Socket, ::IJulia.Msg) at /home/pol/.julia/packages/IJulia/0cLgR/src/execute_request.jl:256
 [9] #invokelatest#1 at ./essentials.jl:697 [inlined]
 [10] invokelatest at ./essentials.jl:696 [inlined]
 [11] eventloop(::ZMQ.Socket) at /home/pol/.julia/packages/IJulia/0cLgR/src/eventloop.jl:8
 [12] (::getfield(IJulia, Symbol("##12#15")))() at ./task.jl:259
```

mkborregaard:
Could you post the code to reproduce the error? What packages are you `using` etc.

pnavaro:
```julia
using Plots
plotlyjs()
plot(Plots.fakedata(50,5),w=3)
```

fpmenninger:
[...]

PR #1877, opened 2019-01-02
Partial fix for plotlyjs backend, for IJulia

lwabeke:
Tries to address #1793 and https://github.com/sglyon/PlotlyJS.jl/issues/221

IJulia and REPL plotting seems to work again, however from REPL plot! creates a new empty window in addition to updating existing window (this behaviour seems to be similar to master), but only if run on a separate line, running plot(rand(10));plot!(rand(5)) doesn't create the extra window


mkborregaard:
There are some conflicts - could you rebase your pr on current master?

mkborregaard:
Thanks. It looks sensible - what's "partial" about the fix?

lwabeke:
"Partial" in terms of cutting away branches, which I can only assume had been required in certain scenarios. 
Maybe it should then be only partially tested: From REPL, from localhost using IJulia notebook() and later minimally from Juno. 
My guess is the use_remote part that I removed was for JuliaBox or some other setup of using IJulia remotely. Also is there other IDEs where plotting has been integrated, maybe it was for one of them?

Initially I added the partial also due to the fact of a blank plot window still showing up when calling plot!, however it seems that is a seperate bug also in the mainline.

mkborregaard:
superseded by #1885 . But thanks @lwabeke !

Issue #1024, opened 2017-08-19
How to plot contourf with GR?


juliohm:
When I specify the seriestype `:contourf` with PyPlot and PlotlyJS, I get a nice colored contour map:

![pyplot](https://user-images.githubusercontent.com/3345261/29488472-de827a64-84c0-11e7-8f6d-d14179152a81.png)

With GR, the contours don't show up:

![gr](https://user-images.githubusercontent.com/3345261/29488500-42b0feb6-84c1-11e7-8ecd-c282d2f43f90.png)

but I know that GR is capable of drawing these colored contours, see the illustration in the homepage: https://gr-framework.org

juliohm:
cc: @jheinen

mkborregaard:
This looks weird, is it screenshots or based on savefig?

juliohm:
@mkborregaard these are screenshots, and I think I have read other threads for which I don't have the URLs unfortunately where people discuss the difference between heatmaps, contour plots and etc. in GR and other backends. It would be nice to have GR working here, I am trying to rely on it the most in my package and tutorials.

mkborregaard:
Can you post an MWE to replicate? It makes it easier to troubleshoot.

juliohm:
I have one tutorial that is working with the master branch, but it may break soon, as I am trying to transition the problem definitions to a separate package: https://github.com/juliohm/GeoStats.jl/blob/master/examples/EstimationProblems.ipynb

jheinen:
I'm currently busy, but I'll check this ASAP. Kind of a color quantization in GR might help ...

jheinen:
@juliohm : where is `GeoDataFrame` defined?
```julia
# inform the coordinates of the data
geodata = GeoDataFrame(df, [:x,:y])
ERROR: UndefVarError: GeoDataFrame not defined
```

juliohm:
Hi Josef, it should be in GeoStats.jl master branch, I can check it
tomorrow...

jheinen:
I'm already on `GeoStats` master - will try another MWE ...

juliohm:
Sorry, I may have corrupted the master branch with the transitions to
GeoStatsBase.jl. Any plot recipe with contourf reproduces the issue though.
Thank you again for looking into this, I'm going to bed, but will double
check tomorrow morning where the GeoDataFrame ended up.

jheinen:
[...]

PR #1880, opened 2019-01-06
Added support for discrete contour plots

jheinen:
requires GR 0.36.0

### DIFF
--- a/REQUIRE
+++ b/REQUIRE
@@ -13,4 +13,4 @@ JSON
 NaNMath
 Requires
 Contour
-GR 0.35.0
+GR 0.36.0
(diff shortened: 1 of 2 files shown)


mkborregaard:
Awesome. Closes #1024

mkborregaard:
@jheinen this PR broke Plots Travis tests because it makes test example 22 look different. Are you set up so you can put a new image in PlotReferenceImages for this test? Otherwise I might do it, but I think actually Travis will error on pngs generated on mac...

juliohm:
I've came across this inconsistency between plots generated in different platforms multiple times while using GR. I thought the issue had been solved. :cry:

mkborregaard:
It's got nothing to do with Plots, it's just that different platforms save different pngs.

juliohm:
Got it, thanks for clarifying. I've heard loading/saving images across platforms is a nightmare.

PR #1883, opened 2019-01-07
Document @shorthands functions

sswatson:
Add documentation to the `@shorthands` functions (`scatter`, `path`, `boxplot`, `quiver`, etc). Other changes: 

* Make `font` accept keyword arguments in addition to the usual unordered argument interface. 
* Fix font rotation on the pyplot backend (there was previously a radian/degree discrepancy between PyPlot and GR). 
* Fix typos in argument descriptions (`_arg_desc`). 

(This is my first pull request---not sure whether this should have been split into separate PRs.)


mkborregaard:
Hi, thanks a lot. Yes, it would be great if you could split this into separate atomic PR's. Would you be willing to do that?

PR #1892, opened 2019-01-16
changed behaviour of Vector{<:SVector} as input argument; close #1886

### LINKED ISSUE #1886: static array length should not change data interpretation
Currently, plotting an array of 2-element StaticVectors treats the elements as 2d points. 3-element StaticVectors are treated as 3d points. Arrays of 4+-element static arrays, or other types of arrays, are instead treated as multiple line plots.

I believe the purpose of StaticArrays is to specialize code and data representation for a certain size, not to change semantics. For example `SVector{2}([1,2]) == [1,2]` is true; the objects are supposed to "mean" the same thing.

I know this would be a (possibly major) breaking change, but I think it would make the API and mental model much cleaner. An extra small motivation is that ideally Plots should not depend on StaticArrays, since I don't believe it truly needs them (e.g. for performance). I don't know what the impact would be on load time and compile time, but removing dependencies can only help.

mkborregaard:
No longer treats Vectors of length-2 or length-3 StaticVectors as a list of points. This is by request from @JeffBezanson to not change the behaviour of StaticVectors outside the domain of an AbstractVector (#1886)

### DIFF
--- a/src/components.jl
+++ b/src/components.jl
@@ -1,12 +1,12 @@
 
 
-const P2 = StaticArrays.SVector{2,Float64}
-const P3 = StaticArrays.SVector{3,Float64}
+const P2 = NTuple{2,Float64}
+const P3 = NTuple{3,Float64}
 
-nanpush!(a::AbstractVector{P2}, b) = (push!(a, P2(NaN,NaN)); push!(a, b))
-nanappend!(a::AbstractVector{P2}, b) = (push!(a, P2(NaN,NaN)); append!(a, b))
-nanpush!(a::AbstractVector{P3}, b) = (push!(a, P3(NaN,NaN,NaN)); push!(a, b))
-nanappend!(a::AbstractVector{P3}, b) = (push!(a, P3(NaN,NaN,NaN)); append!(a, b))
+nanpush!(a::AbstractVector{P2}, b) = (push!(a, (NaN,NaN)); push!(a, b))
+nanappend!(a::AbstractVector{P2}, b) = (push!(a, (NaN,NaN)); append!(a, b))
+nanpush!(a::AbstractVector{P3}, b) = (push!(a, (NaN,NaN,NaN)); push!(a, b))
+nanappend!(a::AbstractVector{P3}, b) = (push!(a, (NaN,NaN,NaN)); append!(a, b))
 compute_angle(v::P2) = (angle = atan(v[2], v[1]); angle < 0 ? 2π - angle : angle)
 
 # -------------------------------------------------------------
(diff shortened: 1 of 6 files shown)


src/components.jl
```diff
@@ -736,17 +736,17 @@ end
...
 function (bc::BezierCurve)(t::Real)
-    p = zero(P2)
+    p = zeros(2)
     n = length(bc.control_points)-1
     for i in 0:n
-        p += bc.control_points[i+1] * binomial(n, i) * (1-t)^(n-i) * t^i
+        @. p += bc.control_points[i+1] * binomial(n, i) * (1-t)^(n-i) * t^i
     end
-    p
+    (p...)
```
daschw: Could the travis error
```
ERROR: LoadError: LoadError: syntax: "..." expression outside call
Stacktrace:
 [1] include at ./boot.jl:326 [inlined]
 [2] include_relative(::Module, ::String) at ./loading.jl:1038
 [3] include at ./sysimg.jl:29 [inlined]
 [4] include(::String) at /home/travis/build/JuliaPlots/Plots.jl/src/Plots.jl:1
 [5] top-level scope at /home/travis/build/JuliaPlots/Plots.jl/src/Plots.jl:166
 [6] include at ./boot.jl:326 [inlined]
 [7] include_relative(::Module, ::String) at ./loading.jl:1038
 [8] include(::Module, ::String) at ./sysimg.jl:29
 [9] top-level scope at none:2
 [10] eval at ./boot.jl:328 [inlined]
 [11] eval(::Expr) at ./client.jl:419
 [12] top-level scope at ./none:3
in expression starting at /home/travis/build/JuliaPlots/Plots.jl/src/components.jl:743
in expression starting at /home/travis/build/JuliaPlots/Plots.jl/src/Plots.jl:166
```
be related to this?

daschw:
Great!
Out of curiosity: have you checked, if this improves compilation/load time?

mkborregaard:
Nope, this was just to put it into practice. Could be fun to see

mkborregaard:
It does not change the load time of Plots at all, it's roughly 5 seconds regardless. So, maybe we should keep the StaticArrays dep and simply drop the `unzip` functions as requested?

daschw:
I'm for this change, I just had the feeling you were not really convinced.
To be clear: I don't have a strong opinion on SVector recipes, I just think that ideally they should live in StaticArrays.jl.

daschw:
Is it just me, or are there no changes in this PR anymore?

daschw:
I think I'll merge this and add a GeometryTypes dependency, as you suggested in https://github.com/JuliaPlots/Plots.jl/issues/1886#issuecomment-460059400 if that's fine with you @mkborregaard.

mkborregaard:
I suggest closing this, and then making a new PR using GeometryTypes.Point instead of P2, P3 and the StaticVectors

daschw:
superseded by #1986

PR #1894, opened 2019-01-16
Fix annotation font rotation in PyPlot backend

sswatson:
The font rotation value for annotations appears to be in degrees, but the pyplot backend includes a radian-to-degree conversion. This PR removes that conversion.


mkborregaard:
Sweet!

PR #1895, opened 2019-01-16
Support keyword arguments for font options

sswatson:
Allow keyword arguments for font options. For example, `font(pointsize = 14, rotation = 45)` is equivalent to `font(14, 45.0)`.


daschw:
Thanks a lot @sswatson

PR #1903, opened 2019-01-24
gr: clamp rgb values

jheinen:


mkborregaard:
For greyscale values I solved this centrally in #1901 and #1900 . I guess the same could be done for rgb's - do you not think it's more sustainable to do it in the common pipeline?

jheinen:
> For greyscale values I solved this centrally in #1901 and #1900 . I guess the same could be done for rgb's - do you not think it's more sustainable to do it in the common pipeline?

Yes, I'd prefer a central solution.

mkborregaard:
Hi @jheinen just to understand - why did you add the volume code to this PR? I thought we'd decided to supersede this with central functionality?

jheinen:
@mkborregaard : GR provides bindings for C, Python, Ruby and Julia. I have no idea how to add this functionality with central (Julia) code.

mkborregaard:
No I meant should the rgb clamping not be done here: https://github.com/mkborregaard/Plots.jl/blob/877811979295f9b8200f5a7d0fc160fa7be71f41/src/series.jl#L351-L365 rather than in Plots/backends/gr.jl?

mkborregaard:
I'm just confused why rgb clamping and the volume functionality are in the same PR. Is it just because you made it on your own master branch?

jheinen:
yes - I think that is/was the problem - my fault.

PR #1919, opened 2019-02-18
GR: contour colorbar fixes

yha:
Unifies colorbar logic for heatmap and contour, which removes code duplication and fixes some bugs  (as far as I can tell, not reported yet): colorbar is now only from the last plot, as in heatmaps, and colorbar title works for contours too. Also, use `linecolor` rather than `fillcolor` for unfilled contour color, (as PyPlot backend already does).

```
r = 0:0.1:10

p1 = contour(r, r, *, fill=true)
plot!(rand(10), legend=false, title="plot over contour")

p2 = heatmap(r, r, *, c=:viridis, alpha=0.75)
contour!(r, r, *, c=:grays, title="contour over heatmap")

p3 = contour(r, r, *, lc=:grays, fc=:viridis,
             title="lc=:grays, fc=:viridis")

p4 = contour(r, r, *, lc=:grays, fc=:viridis, fill=true,
                colorbar_title="colorbar title",
                title="lc/fc, colorbar title")

plot( p1, p2, p3, p4 )
```
Before:
before

After:
after

The bottom-right plot still has wrong line colors, but at least now there's a warning.

### DIFF
--- a/src/backends/gr.jl
+++ b/src/backends/gr.jl
@@ -478,11 +478,8 @@ end
 
 const _gr_gradient_alpha = ones(256)
 
-function gr_set_gradient(c; ignore_colorant = true)
-    @show c
-    gradient(g::ColorGradient) = g
-    gradient(c::Colorant) = ignore_colorant ? cgrad() : ColorGradient([c,c])
-    grad = gradient(c)
+function gr_set_gradient(c)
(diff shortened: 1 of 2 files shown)


mkborregaard:
This is great, thanks @yha !

PR #1923, opened 2019-02-21
plotlyjs: Fix custom tick labels do not work when supplied as a tuple

### LINKED ISSUE #1918: plotlyjs: custom tick labels do not work when supplied as a tuple
```julia
using Plots; plotlyjs()
plot(rand(10), xticks=((1, 10), ("A", "B"))) # works with gr
plot(rand(10), xticks=([1, 10], ["A", "B"])) # works with gr and plotly
```

I believe this comes down to how `AVec` and `TickArgs` are defined:

https://github.com/JuliaPlots/Plots.jl/blob/f2e98acc65c417fce2b8338a3300c9a790f40008/src/types.jl#L5-L8

Two possible fixes:

* Define `AVec` as:
```julia
const AVec{T} = Union{AbstractVector{T}, NTuple{N,T}} where N
```

* Or make 

https://github.com/JuliaPlots/Plots.jl/blob/f905cb957721be871d3fa9c86eea2bdbaefd69a3/src/utils.jl#L434

more flexible, e.g.

```julia
ticksType(ticks::Tuple{T,S}) where {T<:Union{AVec,Tuple},S<:Union{AVec,Tuple}}  = :ticks_and_labels
```

🎉 Both of these now work:

```julia
plot(rand(10), xticks=((1, 10), ("A", "B")))
plot(rand(10), xticks=([1, 10], ("A", "B")))
```

I'm leaning more towards the `AVec` solution, but would be keen to hear what others think. Cheers

harryscholes:
Fixes #1918

### DIFF
--- a/src/utils.jl
+++ b/src/utils.jl
@@ -431,7 +431,7 @@ isvertical(series::Series) = isvertical(series.plotattributes)
 
 ticksType(ticks::AVec{T}) where {T<:Real}                      = :ticks
 ticksType(ticks::AVec{T}) where {T<:AbstractString}            = :labels
-ticksType(ticks::Tuple{T,S}) where {T<:AVec,S<:AVec}  = :ticks_and_labels
+ticksType(ticks::Tuple{T,S}) where {T<:Union{AVec,Tuple},S<:Union{AVec,Tuple}} = :ticks_and_labels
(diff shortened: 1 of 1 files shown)


mkborregaard:
Thank you @harryscholes !

PR #1925, opened 2019-02-23
Add LaTeX output for pyplot backend.

### LINKED ISSUE #1198: Add saving as TeX when using PyPlot
`matplotlib` supports exporting `pgf` commands, which can then be seamlessly `\input` in LaTeX documents. This functionality is already supported by `PyPlot.jl`.

It would be cool if `Plots.jl` also supported this using the `PyPlot` backend, as an alternative to the `PGFPlots` backend (potentially more mature, but also less flexible).

H-M-H:
Actually matplotlib supports outputting LaTeX via pgf:
https://matplotlib.org/users/pgf.html

This fixes #1198.

### DIFF
--- a/src/backends/pyplot.jl
+++ b/src/backends/pyplot.jl
@@ -1332,7 +1332,8 @@ const _pyplot_mimeformats = Dict(
     "application/pdf"         => "pdf",
     "image/png"               => "png",
     "application/postscript"  => "ps",
-    "image/svg+xml"           => "svg"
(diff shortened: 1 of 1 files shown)


mkborregaard:
Thanks!

tzxiong:
Hi, in the recent Plots.jl versions this function seems to be broken. 

Both `savefig(fig,"name.tex")` and `savefig(fig,"name.pgf")` produce errors under the `pyplot()` backend

PR #1936, opened 2019-03-03
Add option :tex_output_standalone for standalone documentclass

balazsmezei:
Add option :tex_output_standalone to set the 'include_preamble' argument in the PGFPlots backend. It allows creating a .tex file with a standalone document class, so it can be compiled on its own.


daschw:
Thanks a lot @balazsmezei!

PR #1941, opened 2019-03-05
Display in svg format in Juno if :html_output_format is :auto

balazsmezei:
I think Juno should display a plot if :html_output_format is set to :auto and the backend supports :svg.

However, I'm a bit unsure on how :html_output_format is/should be handled. Should it be used only when a MIME of text/html is requested, or even when just a display(plt) is called? (Is it format or *html* format?)

Also not sure what would happen if :html_output_format was set to :auto and a backend didn't support :svg (I don't think such a backend exists currently)? Would the error be caught and a different MIME tried automatically?

### DIFF
--- a/src/output.jl
+++ b/src/output.jl
@@ -250,7 +250,7 @@ function showjuno(io::IO, m, plt)
 end
 
 function _showjuno(io::IO, m::MIME"image/svg+xml", plt)
-  if Symbol(plt.attr[:html_output_format]) ≠ :svg
+  if !(plt.attr[:html_output_format] in (:auto, :svg))
     throw(MethodError(show, (typeof(m), typeof(plt))))
   else
     _show(io, m, plt)


daschw:
I'm really not an expert on this stuff. @pfitzseb would you mind taking a quick look on this?

pfitzseb:
Hm, I'm not sure if this is a good idea -- `png` is usually a saner default, especially since `svg`s might get arbitrarily big and crash the editor and things like that.

daschw:
So, I think we should stick with the current version then.
@balazsmezei you can still use `default(fmt = :svg)` or add `PLOTS_DEFAULTS = Dict(:fmt => :svg)` to your julia startup file to get your desired behavior.

daschw:
Nevertheless, thanks a lot for your contribution!

PR #1943, opened 2019-03-07
support plotly mime type

SimonDanisch:
... so displays supporting to directly show plotly json don't need to know about plots ;)


SimonDanisch:
Let me check if it works before merge :D Somehow, package install takes ages right now, so I may not be able to test this today...

SimonDanisch:
Ok this seems to work :) Although I had to add ORCA.jl manually, after checking this out...

Helveg:
Does this mean we can save a plot as its Plotly JSON object? That would be really nice! How to use it?

PR #1969, opened 2019-03-28
Some more fixes for contours and colorbars (mostly GR)

yha:
This PR has various loosely-related fixes that I accumulated recently for drawing my own plots. I can try to split it to different PRs if necessary.

1. The `levels` keyword, when an integer, counts contour lines that are actually rendered, i.e., without min/max values. If specified as vector for filled contours, fill only between the levels specified rather than full data range. This is consistent with how the PyPlot backend already behaves, as well as matplotlib (although the matplotlib docs seem to be saying something different). This also removes the special case for `levels==1`.
2. GR: proper contour/contourf colorbar with lines or filled areas aligned correctly with the ticks. PyPlot and Plotly already do this. 
3. GR: the last drawn contour, contourf, and heatmap are merged to a single colorbar. Plotly already does this (but messes up the ticks). (Rendering the last series of each type is just copying the current odd heuristic. The ideal would be to support multiple colorbars and arbitrarily assigning series to different ones)
4. new `:colorbar_entry` keyword to exclude a series from the colorbar.
5. Support for non-gradient colors in contours.

Examples (on GR):
```
gr(size=(600,300))
x = 0:0.01:2
p1 = heatmap(x,x,+, c=:viridis)
contour!(x,x,+, lc="white", lw=3, ls=:dot, levels=3)
p2 = contourf(x,x,+, levels=3)
plot(p1,p2)
```
contour-levels
Note that `levels=3` now looks consistent between `contour` and `contourf`

```
heatmap(x,x,*, c=:gray)
contourf!(x,x,*, c=:magma, levels=[2.0,2.5,2.75,3.5,4.0])
contour!(x,x,*, c="blue",levels=5)
```
oddplot

Regarding contourf colorbar, pyplot backend currently renders areas of the same size even when levels are not evenly spaced. This is the default matplotlib behavior, although there's a colorbar option `scaling` that can be set to `"proportional"` to override this (see https://matplotlib.org/api/_as_gen/matplotlib.pyplot.colorbar.html). In matplotlib, if you add extra series to a "uniform" colorbar, it does some funky non-linear scaling to align stuff with the ticks. Not sure if this odd behavior is worth replicating, maybe pyplot behavior should be proportional colorbars as implemented here for GR? (it should only take passing the option to matplotlib).


yha:
Should be OK now.
There's some logic that I removed from the following lines in `args.jl` when transforming the `hascolorbar` function into `colorbar_style`:
```
    if st == :contour
        hascbar = (isscalar(series[:levels]) ? (series[:levels] > 1) : (length(series[:levels]) > 1)) && (length(unique(Array(series[:z]))) > 1)
    end
```
So now a colorbar is shown even if there's one level. I don't see why it should be automatically removed in this case. It can still be manually removed if preferred.
Also removed:
```
    # no colorbar if we are creating a surface LightSource
    if xor(st == :surface, series[:fill_z] != nothing)
        hascbar = true
    end
```
Not sure what a "surface LightSource" means, but that `xor` doesn't look like it was doing anything (`hascbar` is already `true` when it's relevant).

mkborregaard:
Awesome @yha

juliohm:
@yha identified this PR as a possible cause to this new GR warning:

```julia
 Warning: Overwriting colorbar entry
└ @ Plots ~/.julia/packages/Plots/4OMIA/src/backends/gr.jl:438
```

Could you please double check?

yha:
> Could you please double check?

What do you mean? The warning definitely comes from this PR.
Is it showing up when it shouldn't?

juliohm:
It is showing up in the notebook of StratiGraphics.jl. I set the colorbar
to false by default and get the warnings. Could you please check that repo?
The recipe is under StratiGraphics/src/strata.jl if I remember at the very
end of the file.

juliohm:
It is showing all the time @yha could you please take a look?

Below is a MWE:

```julia
using GeoStats
using Plots

o = PointSetData(Dict(:z=>rand(1000)), rand(2,1000))

p = partition(o, UniformPartitioner(10))

plot(p)
```

yha:
Should be fixed by #2043.

PR #1970, opened 2019-03-28
Update `yaxis!` description

SebRollen:
Just fixing a small typo. The description for `yaxis!` used to read "Specify **x** axis attributes for an existing plot" (emphasis added), changed to "Specify y axis attributes for an existing plot"


daschw:
Thanks @SebRollen !

PR #1978, opened 2019-04-01
gr: use correct workstation type after savefig()

jheinen:
requires GR#master - GR will be tagged soon


mkborregaard:
can you update the GR require version then too?

jheinen:
> can you update the GR require version then too?

Sure.

PR #2011, opened 2019-04-27
Make legend=:outertopright work for GR

8uurg:
Plot showing legend being positioned outside of the plotarea

As the plotarea's bounding box had to be changed depending on the legend's size, the code for computing the legend's size had to be moved upwards.
Constants were taken from [GR's own plotting functions](https://github.com/jheinen/GR.jl/blob/master/src/jlgr.jl#L525) where the legend is positioned outside of the plotarea (positions 11, 12 and 13).

### DIFF
--- a/src/backends/gr.jl
+++ b/src/backends/gr.jl
@@ -528,7 +528,12 @@ function gr_legend_pos(s::Symbol,w,h)
         str = "topright"
     end
     if occursin("right", str)
-        xpos = viewport_plotarea[2] - 0.05 - w
+        if occursin("outer", str)
+            # As per https://github.com/jheinen/GR.jl/blob/master/src/jlgr.jl#L525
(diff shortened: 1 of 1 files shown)


daschw:
This is so awesome! We've had so many questions about how to position the legend outside with GR and could only provide hacky workarounds. Thanks @8uurg!

mkborregaard:
Yes, great! 
What's the behaviour like when you have multiple subplots?

daschw:
legend_out

It's what you would expect for GR, IMO. Plotly(JS) is the only backend with a shared legend.

mkborregaard:
Yes that looks great

Jo-Fleck:
Thanks a lot for this feature @8uurg ! Was truly missed.

Jo-Fleck:
btw: I cannot find this in the manual yet. Would be helpful to update.

PR #2016, opened 2019-04-30
Add attribute draw_arrow for axes.

### LINKED ISSUE #1348: Axis arrow tips
This is an enhancement request (unless this feature already exists).
Would it be possible to allow for arrow tips at the end of the axis lines?
In particular, I am using the PGFPlots backend, where the relevant native latex command is "axis lines = middle"

Thanks!

H-M-H:
The draw_arrow attribute enables drawing an arrow at the end of an axis.
Currently supported only by the PGFPlots backend.

Resolves #1348 for the PGFPlots backend.


H-M-H:
I hope I did not miss anything. If anyone knows how to support other backends help is appreciated. I already looked into PyPlot but it seems there is no standard way to do this.

surelynottrue:
I am not sure if I am missing something, but draw_arrows draws arrows only in the first quadrant. framestyle=:origin doesn't give me arrows in the negative axes.

H-M-H:
That's standard behavior though, afaik the arrows are supposed to show in which direction values increase. If you have arrows in both directions this information is lost.

surelynottrue:
For a parabola centred at the origin with framestyle=:origin, it looks weird with the arrows in the first quadrant. Values increase in both directions in this case, both the quadrants are equally important.

PR #2022, opened 2019-05-08
use Project.toml + clean up test deps

SimonDanisch:


mkborregaard:
Yay :heart:
do we want to register with TagBot?

SimonDanisch:
If it passes, we should use Registrator & Tagbot ;)

daschw:
I think we should register PlotReferenceImages.

SimonDanisch:
https://github.com/JuliaPlots/PlotReferenceImages.jl/issues/40

test/imgcomp.jl
```diff
@@ -31,6 +12,33 @@ default(size=(500,300))
...
+        else
+            rm(tarfile, force = true)
+        end
+    else
+        @info "using reference images for version $version (already downloaded)"
+    end
+    refpath
+end
+
+const ref_image_dir = download_reference()
```
daschw: With theses changes we should probably also change http://docs.juliaplots.org/latest/contributing/#visualregressiontests, which should be updated anyway.
I'm just commenting, not saying that you should do this @SimonDanisch, thanks for your efforts!

SimonDanisch:
Yay it's green :) Should we merge? Maybe squash before, I guess^^

PR #2031, opened 2019-05-17
Make sure show return nothing.

fredrikekre:
Currently:
```
julia> p = plot(rand(2), rand(2));

julia> x = show(IOBuffer(), MIME("image/png"), p);

julia> x
"png"                         <-- ??
```


daschw:
Thanks!

PR #2049, opened 2019-06-03
Fix hiding of axis on PyPlot

### LINKED ISSUE #2034: pyplot : showaxis=false gives a bug
Using the pyplot backend, using the option `showaxis=false` or `axis=:hide` provokes an error:
```julia
pyplot()
plot(rand(3),rand(3),showaxis=false)
```
(with Plots v0.25.1, PyPlot v2.8.1)

yha:
Fixes #2034


daschw:
Thanks!

PR #2061, opened 2019-06-12
don't let aspect_ratio impact subplot size

daschw:
As discussed with @mkborregaard on slack.
Changes
```julia
plot(
    plot(Plots.partialcircle(0, 2π, 100)),
    plot(rand(10)),
    plot(Plots.partialcircle(0, 2π, 100)),
    aspect_ratio = 1, layout = @layout([a [b; c]]))
```
from
aspect_old
to
aspect_new

... still needs to be tested on all backends.


mkborregaard:
Oh wow, nice implementation! This is really great.

daschw:
Now tested on GR, PyPlot, PlotlyJS and PGFPlots.

Issue #2053, opened 2019-06-06
Histogram2d's default colors with transparent background considered harmful


yha:
`histogram2d` renders empty bins as transparent, which contrasts strongly with nearly-empty bins in colormaps where small is dark, including the default colormap:
```
x = Float64[]
let y = 0.0
    for _=1:500
        push!(x, y += randn() - 0.1y)
    end
end
histogram2d([x[2:end]], [x[1:end-1]])
```
![tmp](https://user-images.githubusercontent.com/4170948/59071019-a635f680-88c5-11e9-8e7d-86699c72d704.png)
I would say this is a misleading representation of the data. This is much better:
```
histogram2d([x[2:end]], [x[1:end-1]], background="black")
```
![tmp](https://user-images.githubusercontent.com/4170948/59070985-830b4700-88c5-11e9-9646-50e31e3066d4.png)

It's worse when the distribution has spikes surrounded by zeros:
```
using Distributions
x = let n = 100_000
    rand(Poisson(1),n) .* exp.(rand(n).*(2π*im)) .+ 0.1randn(n)
end
histogram2d(x)
```
![tmp](https://user-images.githubusercontent.com/4170948/59071089-e301ed80-88c5-11e9-94e2-3f2edc7b9eef.png)
The spike in the middle is barely visible. The distribution really looks like this:
![tmp](https://user-images.githubusercontent.com/4170948/59071125-0a58ba80-88c6-11e9-83cb-851c1fa2be99.png)

I think the best solution is to simply render empty bins as zeros rather than transparent. This is what matplotlib does. The distinction between 0 and 1 should be usually negligible, having to do with noise and sample size.
Alternately, the default colormap for 2d histograms should have white at 0 (I don't think there's currently any such colormap, though some are close), or the default background for it should be black.

mkborregaard:
I disagree with this. For most uses the absence or presence of something is meaningful, and shouldn't be obscured by being the same black color. You can easily use a color scale that is light at the lowest values, if you need this - e.g. `color = :Blues_r`.

yha:
> For most uses the absence or presence of something is meaningful
Can you give an example of such a use case?
In any case, I think the bigger problem is that the scale is reversed relative to the default white background, so the higher values are almost invisible next to zero.
Even if the distinction between 0 and 1 is meaningful, surely it's never as meaningful as the distinction between 0 and a lot?

mkborregaard:
[...]

PR #2062, opened 2019-06-13
Series attribute to show empty bins as 0.

yha:
ref #2053


src/args.jl
```diff
@@ -574,6 +575,7 @@ add_aliases(:xerror, :xerr, :xerrorbar)
 add_aliases(:yerror, :yerr, :yerrorbar, :err, :errorbar)
 add_aliases(:quiver, :velocity, :quiver2d, :gradient, :vectorfield)
 add_aliases(:normalize, :norm, :normed, :normalized)
+add_aliases(:show_empty_bins, :showemptybins, :showempty, :show_empty, :show_zeros, :showzeros)
```
mkborregaard: Do we need the `show_zeros` etc aliases? It seems like they might be appropriated for some other uses.

yha: I added these based on @daschw's suggested name for the attribute, but I don't mind removing them, or doing other name-bikesheddin

yha:
> In any case, new examples have to be added at the end of the `_examples` `Dict` for this process to work. Could you update this PR accordingly?

Do you mean it needs to be moved to the end of the `_examples` array? So it can be placed in its desired position only after updating the reference images?

daschw:
> Do you mean it needs to be moved to the end of the _examples array?

Exactly, array - not dict (my bad). The reference images, e.g. in https://github.com/JuliaPlots/PlotReferenceImages.jl/tree/master/Plots/gr/0.25.0 are named by their index in that array. So, if you add a new reference image in between the Plots test suite will compare wrong images and fail.

PR #2063, opened 2019-06-13
Fix SegmentsIterator for more than one initial NaN

yha:
Bug in `iterate(::SegmentsIterator)` caused the pyplot backend to error in `Base.display` in some cases with multiple `missing`s or `NaN`s, such as
```
pyplot()
plot(2:2:8,[NaN,NaN,1,2])
```


mkborregaard:
Nice!

PR #2064, opened 2019-06-14
use PlotReferenceImages Julia Package for tests

daschw:
With this the PlotReferenceImages can be updated again by using the following steps:
```julia
julia> ]

pkg> add PlotReferenceImages

pkg> dev PlotReferenceImages

pkg> test Plots
```
If reference images differ, they can be overwritten and. With a remote fork of PlotReferenceImages, the user can add the new files, commit, push and submit a PR. Currently, I think we have to wait for the new image to be in a tagged release of PlotReferenceImages.jl until the Plots tests will pass again. I'm not sure if I can depend on the master branch of a Package in the Plots test dependencies.

**Do not merge** This requires https://github.com/JuliaRegistries/General/pull/1389


mkborregaard:
nice

asinghvi17:
@daschw, you can't depend on a master version in Project.toml, but you _can_ do two things:
1.  In the CI script, add `PlotReferenceImages#master` before the test phase
2.  In the README, mention that if you're using `Plots#master` you will need `PlotReferenceImages#master`.
3.  Don't _explicitly_ depend on PlotReferenceImages; instead, download a tarball of the reference images to use for comparison.  You can download tarballs of source for master through the Github API, and whenever you release you'll just need to change two things instead of one.

PR #2067, opened 2019-06-16
Limit number of automatic bins

yha:
Caps number of automatic bins at 10,000, which prevents long waits or GR crash for very long-tailed inputs.


daschw:
Thanks @yha! Could you provide a MWE of an issue that is fixed by this?

yha:
> Could you provide a MWE of an issue that is fixed by this?

```
Random.seed!(0)
x = exp.(exp.(randn(200)))
y = rand(length(x))

Plots._auto_binning_nbins((x,y),1) # == 4154028703978
```
I stumbled across this problem with a similar example, after accidentally `exp`ing my data one time too many.
This one makes GR crash (in `cairoplugin.dll`) before this PR:
```
x = [1,1,1,1,10] .* 1e6
y = 0.1:0.1:0.5
histogram2d(x,y)
```

PR #2081, opened 2019-06-30
Fix representation of zero-valued bins on step-histograms with log-y axis

oschulz:
Zero-values bins are at minus infinity on y-logscale, so they shouldn't be drawn.

Before:

```julia
julia> stephist(vcat(randn(10^5).-10, randn(10^5).+10), normalize = true, bins = 500, yscale = :log10)
```

loghist-sep-peaks-before

After:

loghist-sep-peaks-after

### DIFF
--- a/src/recipes.jl
+++ b/src/recipes.jl
@@ -545,6 +545,8 @@ function _stepbins_path(edge, weights, baseline::Real, xscale::Symbol, yscale::S
             if !isnan(last_w)
                 push!(x, a)
                 push!(y, baseline)
+                push!(x, NaN)
+                push!(y, NaN)
             end
         else
             if isnan(last_w)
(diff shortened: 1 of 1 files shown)


oschulz:
Regarding CI failing on Travis: I get a different of about 0.001 for the three failed image comparisons when I run the tests locally (on Linux). So I think they're actually fine.

How do I make Plots store the results of the test run as image files, for a visual comparison? I know there's a way, but I forgot how.

yha:
> How do I make Plots store the results of the test run as image files, for a visual comparison? I know there's a way, but I forgot how.

This should be easier once #2064 is merged: see https://github.com/JuliaPlots/Plots.jl/pull/2062#pullrequestreview-250207517

daschw:
The reasons for the tests failing is a change in Showoff v0.3.0 (introduced by myself :sweat:). There's already a fix, we just have to wait until v0.3.1 gets registered (https://github.com/JuliaRegistries/General/pull/1671).

oschulz:
Thanks,  @daschw !

PR #2082, opened 2019-06-30
Support arbitrary element type (esp. missing) in tuple/Point recipes

yha:
This simplifies the recipes for vectors of tuples or `GeometryTypes.Point`s to not care about length or element type. Currently, the behavior of `plot(Point.(v))` (say `v isa Vector{<:SArray}`) is completely changed in a surprising way if `v` contains `missing`s (subsequent recipes reinterpret each `Point` as the y vector of one series). 
Now both `missing`s and strings work in tuples, the same as `Number`s:
```
plot( (plot(x, marker=true) for x in 
    ( [(1,20),(3,21),(2,22)], 
      [(1,20),(missing,21),(2,22)],
      [("a",20),(missing,21),("c",22)],
      [(1,2,3),(4,5,6),(2,4,6)] ))..., legend=false )
```
tmp

### DIFF
--- a/src/series.jl
+++ b/src/series.jl
@@ -553,33 +553,14 @@ end
 # # Lists of tuples and GeometryTypes.Points
 # # --------------------------------------------------------------------
 #
-# # if we get an unhandled tuple, just splat it in
-@recipe f(tup::Tuple) = tup
 
-#
-# # (x,y) tuples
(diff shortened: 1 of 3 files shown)


oschulz:
CI on Travis on Linux fails for the same three reference images as in #2081. I think we should visually check and then update the ref images.

oschulz:
Same for  #2081 I guess, @daschw (wait for new Showoff)?

PR #2095, opened 2019-07-09
Allow BinaryProvider installation of `ffmpeg`

ericphanson:
[VideoIO](https://github.com/JuliaIO/VideoIO.jl) recently switched to building `ffmpeg` and all its dependencies via BinaryProvider, and I thought it would be helpful for Plots' users to have a BinaryProvider option as well (since lacking `ffmpeg` can be frustrating for those wanting to make an animation or gif).

I based the changes to `build.jl` off of [VideoIO.jl/deps/build.jl](https://github.com/JuliaIO/VideoIO.jl/blob/master/deps/build.jl).

I think one concern here will be keeping the links (e.g. "https://github.com/JuliaIO/FFMPEGBuilder/releases/download/v4.1.0/build_FFMPEG.v4.1.0.jl") up to date.

The other concern I thought might arise would be those who already have `ffmpeg` might not want to download all the binaries for `ffmpeg` and its dependencies. I tried to work around that by first making an environmental variable "PLOTS_INSTALL_FFMPEG" which can be set to "true" or "false" to control whether or not Plots installs its own ffmpeg via BinaryProvider. If the variable is not set, the default behavior is to then install ffmpeg if and only if it cannot find a system version (using `Sys.which`, as recommended [on Discourse](https://discourse.julialang.org/t/how-to-check-if-a-program-exists-in-a-system-using-julia/19604/12)).

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -32,6 +32,7 @@ StatsBase = "2913bbd2-ae8a-5f71-8c99-4fb6c76f3a91"
 UUIDs = "cf7118a7-6976-5b1a-9a39-7adc72f591a4"
 
 [compat]
+BinaryProvider = "≥ 0.3.0"
 FixedPointNumbers = "≥ 0.3.0"
 GR = "≥ 0.31.0"
 PlotThemes = "≥ 0.1.3"
--- a/Project.toml
+++ b/Project.toml
@@ -41,7 +41,6 @@ StatsBase = "≥ 0.14.0"
 julia = "≥ 1.0.0"
 
 [extras]
-BinaryProvider = "b99e7846-7c00-51b0-8f62-c81ae34c0232"
 FileIO = "5789e2e9-d7fb-5bc7-8068-2c6fae9b9549"
 ImageMagick = "6218d12a-5da1-5696-b52f-db25d2ecc6d1"
 Images = "916415d5-f1e6-5110-898d-aaa5f9f070e0"
@@ -55,4 +54,4 @@ UnicodePlots = "b8865327-cd53-5732-bb35-84acbb429228"
 VisualRegressionTests = "34922c18-7c2a-561c-bac1-01e79b2c4c92"
 
 [targets]
-test = ["BinaryProvider", "Pkg", "Test", "Random", "StatsPlots", "VisualRegressionTests", "LaTeXStrings", "Images", "ImageMagick", "RDatasets", "FileIO", "UnicodePlots"]
+test = ["Pkg", "Test", "Random", "StatsPlots", "VisualRegressionTests", "LaTeXStrings", "Images", "ImageMagick", "RDatasets", "FileIO", "UnicodePlots"]
--- a/Project.toml
+++ b/Project.toml
@@ -5,12 +5,14 @@ version = "0.25.3"
 
 [deps]
 Base64 = "2a0f44e3-6c83-55bd-87e4-b1978d98bd5f"
+BinaryProvider = "b99e7846-7c00-51b0-8f62-c81ae34c0232"
 Contour = "d38c429a-6771-53c6-b99e-75d170b6e991"
 Dates = "ade2ca70-3891-5945-98fb-dc099432e06a"
 FixedPointNumbers = "53c48c17-4a7d-5ca2-90c5-79b7896eea93"
 GR = "28b8d3ca-fb5f-59d9-8090-bfdbd6d07a71"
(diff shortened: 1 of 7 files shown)


jlperla:
Since the package seems to run `ffmpeg` in the shell (see https://github.com/JuliaPlots/Plots.jl/blob/0735a3475c3747bbac92925cdc58668bbb74929b/src/animation.jl#L79  ) why not `ffmpeg` and check the return code to see if it is installed in the path during `build`?

Project.toml
```diff
@@ -50,7 +52,6 @@ StatsPlots = "f3b207a7-027a-5e70-b257-86293d7955fd"
 Test = "8dfed614-e22c-5e08-85e1-65c5234f0b40"
 UnicodePlots = "b8865327-cd53-5732-bb35-84acbb429228"
 VisualRegressionTests = "34922c18-7c2a-561c-bac1-01e79b2c4c92"
-BinaryProvider = "b99e7846-7c00-51b0-8f62-c81ae34c0232"
 
 [targets]
-test = ["BinaryProvider", "Pkg", "Test", "Random", "StatsPlots", "VisualRegressionTests", "LaTeXStrings", "Images", "ImageMagick", "RDatasets", "FileIO", "UnicodePlots"]
```
daschw: BinaryProvider is still required in the tests to download the reference images. That's why they are failing on travis.

deps/build.jl
```diff
@@ -1,4 +1,80 @@
 
+using BinaryProvider # requires BinaryProvider 0.3.0 or later
```
daschw: Could you add this requirement in the [compat] section of Project.toml, please?

SimonDanisch:
I just created https://github.com/JuliaIO/FFMPEG.jl for the purpose, that Packages like plots don't need to mess with their build.jl just to get the ffmpeg binary!

daschw:
Nice @SimonDanisch! This means we could depend on FFMPEG.jl and replace our 
```
run(`ffmpeg ...`)
```
commands with

```julia
FFMPEG.exe("...")
```
or
```julia
FFMPEG.run(FFMPEG.ffmpeg, "...")
```
?

ericphanson:
FFMPEG.jl seems like a good idea! Would also reduce my concern about needing to update the binary provider links everywhere. I suppose this way though Plots would depend on FFMPEG.jl which would mean a slightly slower first time install / build step (even for those with system ffmpeg).

deps/build.jl
ericphanson: Done!

Project.toml
ericphanson: I don't mind making the change, but are you sure this is the case? I was under the impression that this section was for test-only dependencies, and dependencies of Plots were available already. For example, `GeometryTypes`  is used in the tests and is a dependency of Plots, but is not declared in `[targets]`. Additionally, the specific Travis error is:
```
curl: (77) Error reading ca cert file /etc/pki/tls/certs/ca-bundle.crt - mbedTLS: (-0x3E00) PK - Read/write of file failed
ERROR: LoadError: LoadError: failed process: Process(`curl -g -L -f -o /home/travis/build/JuliaPlots/Plots.jl/test/reference_images/reference_images.zip https://github.com/JuliaPlots/PlotReferenceImages.jl/archive/v0.0.1.tar.gz`, ProcessExited(77)) [77]
```
which seems like it might not be BinaryProvider-related.

daschw: I think the tests use BinaryProvider's `download` and that's why they are failing with the removal of BinaryProvider from the test dependencies. I'm not sure, though.
Anyway I'd prefer using FFMPEG.jl instead. Then Plots does not have to deal with keeping the links up-to-date as you mentioned in your original post.

asinghvi17:
With JuliaIO/FFMPEG.jl#3 you could also do:
```julia
ffmpeg`-version`
```
which will expand to 
```julia
withenv(FFMPEG.execenv) do
    Base.run(`$(FFMPEG.ffmpeg) -version)
end
```

asinghvi17:
>  I tried to work around that by first making an environmental variable "PLOTS_INSTALL_FFMPEG" which can be set to "true" or "false" to control whether or not Plots installs its own ffmpeg via BinaryProvider.

I think that we could probably include this as a feature in FFMPEG.jl, so that users can choose there itself.

ericphanson:
Closing in favor of https://github.com/JuliaPlots/Plots.jl/pull/2099

daschw:
Thanks a lot for getting this started, though, @ericphanson !

ericphanson:
Thanks!

Issue #2030, opened 2019-05-16
Add a binaryprovider to plots so that ffmpeg works "out of the box"


logankilpatrick:
Feature request per a conversation in the plotting slack channel. We need a binary provider so that ffmpeg can be downloaded automatically (I think).  This will allow the @gif macro work work by default shown here: (2nd example) http://docs.juliaplots.org/latest/examples/gr/

According to @SimonDanisch :+1: 
"that's done :slightly_smiling_face: VideoIO comes with a binaryprovider provided binary of ffmpeg"

Please let me know if you need any other details.

PR #2099, opened 2019-07-13
Use FFMPEG.jl for video encoding

asinghvi17:
Use FFMPEG.jl for FFMPEG things.


asinghvi17:
@ericphanson @daschw this should supercede #2095 - I've used FFMPEG.jl here and all the tests seem to be working, but there aren't any animation tests.  I'll add some.

test/runtests.jl
```diff
@@ -93,3 +93,20 @@ end
         @test isequal(collect(zip(Plots.unzip(Point.(z))...)), z)
     end
 end
+
+@testset "Recording animations" begin
+    
+    for i in (gr, plotlyjs, pyplot, plotly)
```
daschw: Plots does not test PyPlot and Plotly(JS) in the test suite.

asinghvi17:
Cool, thanks for the info @daschw!  I put them there to make sure FFMPEG was working properly (which it isn't right now) but will change that shortly.  The problem is in FFMPEG, I'll quickly change that to another API and revert back.

daschw:
Awesome!

asinghvi17:
Perfect, it works now - I'll remove the check.

ericphanson:
I think this closes https://github.com/JuliaPlots/Plots.jl/issues/2030

asinghvi17:
Good catch, thanks!

logankilpatrick:
Thanks for this @asinghvi17!

Issue #2123, opened 2019-08-06
Histogram for strings? (categorical)


Moelf:
At the moment the following produces an error:
```
using Random
histogram(randstring("abc",12))
```

not even this:
```
using Random, StatsBase
cm = countmap(randstring("ABC", 20) |> collect)
bar(keys(cm) |> collect, values(cm) |> collect)
```

daschw:
Apparently Plots can't handle `Char` input. This should be fixed in #2124, until then you can convert to `String` first:
```julia
bar(string.(collect(keys(cm))), collect(values(cm)))
```

Moelf:
I wonder if we can just make `histogram` works with categorial Arrays, if so, where do we do it. I'm happy to make a PR

daschw:
The relevant code is here
https://github.com/JuliaPlots/Plots.jl/blob/79777adc079cb4811c03d333a348ab03e898103e/src/recipes.jl#L667-L700
and above.
Adding histograms for categorical Arrays is probably just a matter of implementing a method of `_make_hist` for categorical Arrays.
Right, @oschulz?

oschulz:
I think so - however, currently, everything is based on `StatsBase.Histogram`. I don't know if that supports categorical axes (I think not). If not, maybe it would be good to start there and add support for such histograms to `StatsBase` - such a histogram would also be useful in data analysis, not just plotting, I'd guess.

yha:
`StatsBase` already has `countmap` for categorical data. It's not the same as a histogram (there are no bins), and I don't think they should be conflated in `StatsBase`.
Maybe the two two can share the name `histogram` in Plots (and disambiguated based on data type), but I think it would be better to separate them in Plots as well. We could have a general recipe for plotting `AbstractDict{<:Any,<:Real}` as bar plots, which I think is a useful and reasonable default anyway, and then you can do
```
using StatsBase
plot(countmap(data))
```

oschulz:
Yes, that sound very reasonable.

mkborregaard:
I agree it makes sense, people may also want to print the count of numbers.

mkborregaard:
Maybe have a new series recipe called `:counts` that would run the `countmap` automatically and plot the Dict?

PR #2124, opened 2019-08-06
handle chars

daschw:
Plots cannot (no longer?) deal with `Char`s as input. This came up in https://github.com/JuliaPlots/Plots.jl/issues/2123.
This change should fix the problem.


yha:
Wouldn't that be better handled by a type recipe?
```
@recipe f(::Type{<:AbstractChar}, ::AbstractChar) = (string, string)
```

daschw:
Good point. Feel free to change, if that works as well.

PR #2133, opened 2019-08-08
Recipe for plotting dicts

yha:
See #2123.

### DIFF
--- a/src/series.jl
+++ b/src/series.jl
@@ -352,12 +352,9 @@ end
     end
 end
 
-# Dicts with Real values are plotted as bars
+# Dicts: each entry is a data point (x,y)=(key,value)
 
-@recipe function f(d::AbstractDict{T,<:Real}) where T
-    seriestype --> :bar
-    x, y = collect(keys(d)), collect(values(d))
-end
+@recipe f(d::AbstractDict) = collect(keys(d)), collect(values(d))
 
 # function without range... use the current range of the x-axis
 
--- a/src/series.jl
+++ b/src/series.jl
@@ -352,7 +352,12 @@ end
     end
 end
 
+# Dicts with Real values are plotted as bars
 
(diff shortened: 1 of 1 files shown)


src/series.jl
```diff
@@ -352,7 +352,12 @@ end
     end
 end
 
+# Dicts with Real values are plotted as bars
 
+@recipe function f(d::AbstractDict{T,<:Real}) where T
+    seriestype --> :bar
+    x, y = collect(keys(d)), collect(values(d))
```
mkborregaard: I'd prefer sorting by key, unless we're dealing with an `OrderedDict`.

mkborregaard:
I think this makes sense when T is a String, but I think a Dict(Number, Number) could be many other things than counts?
Maybe just 
```julia
@recipe function f(d::AbstractDict{T,<:Real}) where T
    xs = keys(d) |> collect |> sort
    xs, [d[x] for x in xs]
end
```
and let the user choose to display this as a bar?

mkborregaard:
See also https://github.com/JuliaPlots/Plots.jl/issues/2123#issuecomment-519598784

yha:
I'm not sure sorting automatically is a good idea. An `AbstractDict` may or may not have a specific ordering depending on the subtype (if it's not specifically an `OrderedDict` it might still be ordered). 
If we make a new `counts` recipe it can do `plot(sort(countmap(...)))`.

mkborregaard:
That's also fine with me

yha:
I suppose if we don't default to a bar plot and just extract the keys and values from the dict, we might as well do this for all types of dict, so you can plot, e.g.,
```
plot(OrderedDict("a"=>Date(2000),"b"=>Date(2001)))
```
?

mkborregaard:
I think so

mkborregaard:
I've often thought there are other places it would be nice to accept Dicts. Like for colours of groups.

Issue #1907, opened 2019-01-29
Errorbars are not rotated in horizontal bar plot


johannes-fischer:
I wanted to attach error bars to a horizontal bar plot, however they seem to be anchored to where the bars would end if the orientation was vertical. Example:
`bar(rand(3), fill=:orange, orientation=:h, xerror=fill(.2,3), xlims=(0,4),ylims=(0,4))`
I've tried GR, PyPlot and Plotly backends.

bvolkmer:
As an addition: If the error type is not compatible with the data type of the other axis, it will fail.

For example: xdata::String and yerror::Float32 will result in:
- Pyplot and plotlyjs will plot no error bars without warning
- gr will fail with `MethodError: no method matching isfinite(::String)`

Issue #1412, opened 2018-02-25
horizontal layout issues


piever:
On GR, I've found several issues with the horizontal layout, not sure what's going on:

1. `scatter(rand(10), 10*rand(10), dir = :h)`

Here the points are displayed according to the non rotated system and go out of bounds (they don't appear)

2. `density(randn(10000), orientation = :h)` 

For some reason, I get a straight vertical line, which also shows in:

`marginalhist(randn(10000), randn(10000), density = :true)`

And finally [#129](https://github.com/JuliaPlots/StatPlots.jl/issues/129) with `groupedbar`

sethaxen:
This seems to still be the case.

sethaxen:
Just verified that this is also broken for UnicodePlots and Plotly.

andreaskoher:
Thumbs up for this feature :+1: 

I would love to have a horizontal violin plot:

```julia
using StatsPlots
y, z = repeat([1,2,3],outer=100), randn(300)
violin(y,z; orientation = :horizontal)
```
![vertical_violine](https://user-images.githubusercontent.com/18704101/158792285-f0ad5423-b21d-4dc0-bc65-8abd1f4cdfde.png)


```julia
violin(y,z; orientation = :horizontal)
```
![horizontal_violine](https://user-images.githubusercontent.com/18704101/158792315-424b13e2-4a59-4dd8-b16f-a25217c55b25.png)

tpgillam:
For boxplots I added the feature here:

https://github.com/JuliaPlots/StatsPlots.jl/commit/be7f9c03110242707915aaaae8e1e316b179a3d5

Maybe we can just use the same approach for `violin`?

Issue #1410, opened 2018-02-24
Horizontal bars broken on Plotly.


apalugniok:
```
using Plots; plotly()
xs = [ "x$i" for i in 1:10 ]
p = bar(xs, -5:4, orientation = :h)
```
![newplot](https://user-images.githubusercontent.com/22132297/36631095-c79bd122-1969-11e8-9715-0b4498616caa.png)
```
julia> p[1][1][:x]
10-element Array{String,1}:
 "x1" 
 "x2" 
 "x3" 
 "x4" 
 "x5" 
 "x6" 
 "x7" 
 "x8" 
 "x9" 
 "x10"

julia> p[1][1][:y]
10-element Array{Float64,1}:
 0.5
 1.5
 2.5
 3.5
 4.5
 5.5
 6.5
 7.5
 8.5
 9.5
```
There is also a disagreement with `axis[:discrete_values]`
```
julia> p[1][:yaxis][:discrete_values]
10-element Array{Any,1}:
 "x1" 
 "x2" 
 "x3" 
 "x4" 
 "x5" 
 "x6" 
 "x7" 
 "x8" 
 "x9" 
 "x10"

julia> p[1][:xaxis][:discrete_values]
0-element Array{Any,1}
```
which is making `ticks = :native` output both x and y as the array `["x1", "x2", ... , "x10"] and gives 
![newplot 1](https://user-images.githubusercontent.com/22132297/36631285-227db526-196d-11e8-8f9e-2ecc8b339705.png)

wkearn:
This was a fun one. It happens in [here](
https://github.com/JuliaPlots/Plots.jl/blob/dbe5b7830738d1b8f745ad98b701dec01c8d6fb6/src/axes.jl#L319-L348) because the x-axis does not get properly updated when the orientation is horizontal. It affects Plotly/PlotlyJS because they rely on native support for `:bar` whereas the others (GR, PyPlot, and PGFPlots, at least) use the `step` recipe.

A fix is on the way!

PR #2134, opened 2019-08-08
[WIP] Generic support for permuting axes (generalizes `orientation=:horizontal`)

yha:
I was trying to add support for `orientation=:horizontal` to the `dendrogram` recipe from `StatsPlots`, and some code in `axes.jl` that also reads the `orientation` attribute was getting in the way. Trying to see how other recipes implement this keyword, I discovered most of them are currently broken (see #1412, #1907, and maybe also #1410).

So instead, I decided to implement a more general solution to changing plot orientation that doesn't need re-implementing for each recipe/plot type. This PR adds functions `permuteaxes!, swapxy!, swapxz!, swapyz!` that can be called on existing plots to swap the axes around. 
So now a horizontal dendrogram can be plotted like this, without any support from the `dendrogram` recipe itself:
```
using Plots, StatsPlots, Clustering, LinearAlgebra
X = rand(10,10)
D = [norm(x.-y) for x in eachcol(X), y in eachcol(X)]
hc = hclust(D)
plt = plot(hc)
swapxy!()
plot!(xflip=true, ymirror=true)
```
tmp

I'm not sure if we also want a keyword interface, and what should be done with the current `orientation` keyword. Maybe something like:
```
histogram(foo, axis_permutation=[:y,:x])
```
but then what to make of
```
histogram(foo, axis_permutation=[:y,:x], xticks=[-5,0,5])
```
should the permutation be applied before or after the ticks?

- [x] Permute axes of existing plot
- [ ] Keyword argument support similar to or extending current `orientation`?
- [ ] Clean up `orientation` cruft
- [ ] Docs

### DIFF
--- a/src/Plots.jl
+++ b/src/Plots.jl
@@ -72,6 +72,11 @@ export
     xgrid!,
     ygrid!,
 
+    permuteaxes!,
+    swapxy!,
+    swapxz!,
+    swapyz!,
+
     xlims,
     ylims,
     zlims,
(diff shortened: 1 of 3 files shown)


mkborregaard:
It's a great idea to have a general framework. Could all of this be hidden for the user, so they would still just use the old `orientation` framework?
IMHO `xticks` etc should refer to the x axis AFTER the rotation.

yha:
> IMHO `xticks` etc should refer to the x axis AFTER the rotation.

That's certainly more intuitive in this case. But I'm wondering what the general semantics for this might be, and if it's doable. Should we apply all recipes first, then axis permutations, and then explicitly provided keywords? Is there even currently any way to identify explicit keywords at this stage? (maybe this needs something along the lines of https://github.com/JuliaPlots/Plots.jl/pull/1698#issuecomment-420973969 implemented first)

> Could all of this be hidden for the user, so they would still just use the old `orientation` framework?

We can keep the `orientation` keyword, but I think it should be an additional alias for specific plot types where it makes sense to think of them as vertical vs. horizontal and it's clear which is which, rather than the only interface. For example, you can use this to plot a violin plot "sideways", but calling it "horizontal" is weird, I think. Something like `swapxy!` or `axis_permutation` is clearer IMO.

daschw:
I'm not sure about the original intention, but with `orientation` being a series attribute, I think this should be handled per series, so something like
```julia
using Plots
bar(1:3, orientation = :horizontal)
bar!(1:3)
```
is possible:
horizontal

So I think ideally, the `orientation` keyword would be checked after a series recipe is processed and `x` and `y` values would be swapped then in case. This is however not as general (3D) as this PRs implementation.

If I understand it correctly this PR rather acts on a subplot level (transposing all series of the subplot) so it's not exactly the same approach as `orientation`, right?
I think it can make sense to have both, `orientation` and `permuteaxis!` as a subplot concept. In the latter case, however, it could also make sense that `xticks` refer to the axis with the `x` data, so possibly the vertical axis after swapping, no?

mkborregaard:
I'm not sure what your'e suggesting in terms of subplots @daschw ? But I agree completely with you that `orientation` is and should be a series attribute.

kescobo:
Bump! This looks great. What still needs doing?

yha:
The discussion here convinced me that it's more work than I initially thought. First, that it's better to support this at the series level rather than the (sub)plot level. And also that the interaction with `x*` and `y*` attributes is a bit tricky: swapping axes should probably happen after recipe-supplied keywords but before user-supplied keywords. I *think* implementing something like https://github.com/JuliaPlots/Plots.jl/pull/1698#issuecomment-420973969 first is the way to go. I will look at it again later this week and try to come up with a new plan and checklist.

PR #2135, opened 2019-08-08
Replace low-level Char handling with a type recipe.

yha:
See #2124


asinghvi17:
Is there a test which covers the char transformations?

yha:
No. Maybe testing should wait for #2064?
I though we could test something like `plot(["a","b"]) == plot(['a','b'])` without adding reference images, but there isn't currently a `==` defined for plots, and I'm not sure it's actually easy to implement correctly.

PR #2148, opened 2019-08-17
use === nothing

ChrisRackauckas:
Should decrease compile times

### DIFF
--- a/src/args.jl
+++ b/src/args.jl
@@ -685,15 +685,15 @@ function processLineArg(plotattributes::KW, arg)
         plotattributes[:linestyle] = arg
 
     elseif typeof(arg) <: Stroke
-        arg.width == nothing || (plotattributes[:linewidth] = arg.width)
-        arg.color == nothing || (plotattributes[:linecolor] = arg.color == :auto ? :auto : plot_color(arg.color))
-        arg.alpha == nothing || (plotattributes[:linealpha] = arg.alpha)
-        arg.style == nothing || (plotattributes[:linestyle] = arg.style)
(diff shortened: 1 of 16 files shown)


ChrisRackauckas:
From

```
julia> @time using Plots
  5.215902 seconds (10.59 M allocations: 580.846 MiB, 5.43% gc time)

julia> @time plot(rand(4,4))
 16.619736 seconds (49.28 M allocations: 2.408 GiB, 8.68% gc time)
```

to

```
  julia> @time using Plots
    4.865916 seconds (10.48 M allocations: 575.247 MiB, 4.54% gc time)
  
  julia> @time plot(rand(4,4))
   15.923668 seconds (49.26 M allocations: 2.407 GiB, 8.65% gc time)
```

yha:
I don't see performance gains here (Julia 1.1.1, Windows 10).
Before:
```
julia> @time using Plots
  3.741940 seconds (7.43 M allocations: 434.046 MiB, 5.01% gc time)

julia> @time plot(rand(4,4))
 13.814570 seconds (49.98 M allocations: 2.436 GiB, 9.78% gc time)
```
After:
```
julia> @time using Plots
  3.788572 seconds (7.44 M allocations: 434.568 MiB, 4.92% gc time)

julia> @time plot(rand(4,4))
 13.691652 seconds (50.00 M allocations: 2.437 GiB, 9.97% gc time)
```
But it seems harmless.

ChrisRackauckas:
Yeah, it's at least harmless and the right direction. There must be other stuff going on, but since we've seen a lot of libraries get a speedup I thought we might as well do it here as well.

PR #2157, opened 2019-08-24
A real "first time to plot" analysis

ChrisRackauckas:
After seeing a bunch of nonsense like [this thread](https://discourse.julialang.org/t/roadmap-for-a-faster-time-to-first-plot/22956/142), I decided it was a good idea to document why "first time to plot" is a thing. tl;dr, it's because the original author was not careful with types, nothing more nothing less. 

(I have to do this on my desktop to not segfault, but @vtjnash saw it. Will post the profile when I am back from travelling.)

First of all, there is no code loading going on in like a year. GR is baked into the system since that PR was done, so let's stop assuming that code loading nonsense is still the issue since we ruled that out a long time ago... and it doesn't even exist anymore. Get it out of your head. Now let's continue with a clear mind.

If you look at the profile using `ProfileView.view(C=true)`, we can distinguish between Julia type inference time and LLVM compile time. That gigantic bump is Julia type inference time, while the sliver between the bumps is LLVM time. Conclusion: it's our fault in type inference and there's no one else to blame. This is actually quite peculiar since most of the time you see LLVM taking a lot of the time, but that's not the case specifically for Plots.jl.

And the reason is because Plots.jl's choices with types are a mess. To start, almost everything is stored in a `KW = Dict{Symbol.Any}`. All of the args are handled in this structure, but this item does not show up in the Julia sysimage, so all operations with it require being compiled. Changing this to `KW = Dict{Any,Any}` is actually a little helpful (Requires the similar first time to plot RecipesBase branch). Basically, we want type inference to give up quickly and just go to `Any`. However, this library throws `KW` into small structs, and then has some abstract types over fields. To make it give up, we can just kick to `Any` again. Etc. 

Usually that can be helpful, but here we kick to Any too much and then we start getting type inference nightmares. We know from around the ecosystem that function-barriers cause a lot of compile time increases:

https://github.com/JuliaDiffEq/OrdinaryDiffEq.jl/pull/841
https://github.com/JuliaDiffEq/StochasticDiffEq.jl/pull/207

and this is precisely where we start running into a wall. With every function call, we hit a barrier and so this overhead cost of knowing that it's a dynamic dispatch function barrier will be hit (to be honest, I'm not sure why this extra cost exists, @vtjnash?). But similarly, putting everything into one super long function will cause superlinear compile time growth, so we need to split functions.

So as a proof of concept for this idea, I split out the series types and legends handling branches in the `gr.jl` file. If you comment those portions out you see a massive compile time decrease. If you split them out to functions, you don't get that same decrease because you're now hitting the secondary inference compile time issue.

So given how these two play against each other, the only real way to solve this is to make Plots not drop so many types. That a giant feat because everything is stored in bad dictionaries, so someone would have to

1) Set the argument types underlying the attributes.
2) Converting to a standard form the moment the arguments are seen like that.
3) Pushing this type information down.

Currently this is very difficult, the biggest offender being Measures.jl. Measures just has so much type information on every value, like Units, that it's very difficult to keep the types aligned with them. It's just way too much type information, so it should just be made a struct with a value and symbol, using the symbol to branch behavior, instead of holding each new value with a possibly new type. 

When you dig into the pipeline, you see that 7 seconds of compilation is on the type handling of the plot.jl `_plot!` function, and then 13 is in the gr_display function (at least with GR). Essentially all of that time could be cut if the types were more inferred, given that the entire issue is that inference is just going insane. Oh, and this splitting shows why it's so hard to fix this issue: 7 seconds comes from mostly 3 function calls, while the other 13 comes from a few different branches, but that's still quite a few hotspots to hit, so just trying to "be type inference" and declare types would still be a long process (and you still lose inference every time you dig into one of these types without information...)

In conclusion, don't use Measures.jl and a bunch of non-inferred partially abstractly typed dictionaries and then wonder why it takes forever to compile. I can tell you why: type inference gone amok. The profiles show it, and it's hard to get around since the original tools like function barriers end up contributing to the problem. This PR shows that you can get very minor gains by avoiding that issue, but you actually need to address it if you want to go anywhere. It really has nothing to do with "tradeoffs" or whatever nonsense, it's just about writing code that's easy on inference. We could fix this in Plots, Makie, GR, etc., does matter which. This makes it clear it's not the method but it's the implementation.

### DIFF
--- a/src/backends/gr.jl
+++ b/src/backends/gr.jl
@@ -9,7 +9,7 @@ export GR
 
 # --------------------------------------------------------------------------------------
 
-const gr_linetype = KW(
+const gr_linetype = Dict{Symbol,Int}(
     :auto => 1,
     :solid => 1,
     :dash => 2,
@@ -18,7 +18,7 @@ const gr_linetype = KW(
     :dashdotdot => -1
 )
 
-const gr_markertype = KW(
+const gr_markertype = Dict{Symbol,Int}(
     :auto => 1,
     :none => -1,
     :circle => -1,
@@ -45,19 +45,19 @@ const gr_markertype = KW(
     :hline => -31
 )
 
-const gr_halign = KW(
+const gr_halign = Dict{Symbol,Int}(
     :left => 1,
     :hcenter => 2,
     :right => 3
 )
 
-const gr_valign = KW(
+const gr_valign = Dict{Symbol,Int}(
     :top => 1,
     :vcenter => 3,
     :bottom => 5
 )
 
-const gr_font_family = Dict(
+const gr_font_family = Dict{String,Int}(
     "times" => 1,
     "helvetica" => 5,
     "courier" => 9,
@@ -84,7 +84,7 @@ gr_set_transparency(c, α) = gr_set_transparency(α)
 gr_set_transparency(c::Colorant, ::Nothing) = gr_set_transparency(c)
 gr_set_transparency(c::Colorant) = GR.settransparency(alpha(c))
 
-const _gr_arrow_map = Dict(
+const _gr_arrow_map = Dict{Symbol,Int}(
     :simple => 1,
     :hollow => 3,
     :filled => 4,
@@ -1580,7 +1580,7 @@ end
 
 # ----------------------------------------------------------------
 
-const _gr_mimeformats = Dict(
+const _gr_mimeformats = Dict{String,String}(
     "application/pdf"         => "pdf",
     "image/png"               => "png",
     "application/postscript"  => "ps",
--- a/src/backends/gr.jl
+++ b/src/backends/gr.jl
@@ -751,7 +751,7 @@ end
 
 @noinline function gr_xaxis_height(sp)
     xaxis = sp[:xaxis]
-    xticks, yticks = axis_drawing_info(sp)[1:2]
+    xticks::Tuple{Array{Float64,1},Array{String,1}}, yticks::Tuple{Array{Float64,1},Array{String,1}} = axis_drawing_info(sp)[1:2]
     gr_set_font(tickfont(xaxis))
     h = (xticks in (nothing, false, :none) ? 0 : last(gr_get_ticks_size(xticks, xaxis[:rotation])))
     if xaxis[:guide] != ""
@@ -763,7 +763,7 @@ end
 
 @noinline function gr_yaxis_width(sp)
     yaxis = sp[:yaxis]
-    xticks, yticks = axis_drawing_info(sp)[1:2]
+    xticks::Tuple{Array{Float64,1},Array{String,1}}, yticks::Tuple{Array{Float64,1},Array{String,1}} = axis_drawing_info(sp)[1:2]
     gr_set_font(tickfont(yaxis))
     w = (xticks in (nothing, false, :none) ? 0 : first(gr_get_ticks_size(yticks, yaxis[:rotation])))
     if yaxis[:guide] != ""
@@ -774,7 +774,7 @@ end
 end
 
 function _update_min_padding!(sp::Subplot{GRBackend})
-    dpi = sp.plt[:thickness_scaling]
+    dpi::Float64 = sp.plt[:thickness_scaling]
     if !haskey(ENV, "GKSwstype")
         if isijulia()
             ENV["GKSwstype"] = "svg"
@@ -785,6 +785,9 @@ function _update_min_padding!(sp::Subplot{GRBackend})
     toppad::AbsoluteLength    = 2mm  + sp[:top_margin]
     rightpad::AbsoluteLength  = 2mm  + sp[:right_margin]
     bottompad::AbsoluteLength = 2mm  + sp[:bottom_margin]
+    l::Float64 = 0.0
+    h::AbsoluteLength = 0mm
+    w::AbsoluteLength = 0mm
     # Add margin for title
     if sp[:title] != ""
         gr_set_font(titlefont(sp))
@@ -793,10 +796,10 @@ function _update_min_padding!(sp::Subplot{GRBackend})
         toppad += h
     end
     # Add margin for x and y ticks
-    xticks, yticks = axis_drawing_info(sp)[1:2]
+    xticks::Tuple{Array{Float64,1},Array{String,1}}, yticks::Tuple{Array{Float64,1},Array{String,1}} = axis_drawing_info(sp)[1:2]
     if !(xticks in (nothing, false, :none))
         flip, mirror = gr_set_xticks_font(sp)
-        l = 0.01 + last(gr_get_ticks_size(xticks, sp[:xaxis][:rotation]))
+        l = 0.01 + last(gr_get_ticks_size(xticks, Float64(sp[:xaxis][:rotation])))
         h = 1mm + gr_plot_size[2] * l * px
         if mirror
             toppad += h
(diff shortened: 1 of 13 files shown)


src/plot.jl
```diff
@@ -155,17 +154,18 @@ function plot!(plt::Plot, args...; kw...)
...
+    _plot!(plt, plotattributes, Any[args...])
 end
 
 # -------------------------------------------------------------------------------
 
 # this is the core plotting function.  recursively apply recipes to build
 # a list of series KW dicts.
 # note: at entry, we only have those preprocessed args which were passed in... no default values yet
-function _plot!(plt::Plot, plotattributes::KW, args::Tuple)
+function _plot!(plt::Plot{T}, plotattributes::KW, args::Vector{Any}) where {T}
```
ChrisRackauckas: This is the main function

src/plot.jl
```diff
@@ -202,7 +204,11 @@ function _plot!(plt::Plot, plotattributes::KW, args::Tuple)
     # --------------------------------
     # Plot/Subplot/Layout setup
     # --------------------------------
+
+    # 2.5 seconds
     _plot_setup(plt, plotattributes, kw_list)
+
+    # 6 seconds
```
ChrisRackauckas: You get to 6 seconds by the time it's here due to inference issues in the argument handling chain. Profile it to see.

src/plot.jl
```diff
@@ -233,24 +239,24 @@ function _plot!(plt::Plot, plotattributes::KW, args::Tuple)
...
+    # 7 seconds
     current(plt)
 
     # do we want to force display?
     # if plt[:show]
     #     gui(plt)
     # end
     _do_plot_show(plt, plt[:show])
-
     plt
```
ChrisRackauckas: The rest of the 13 is in the display

src/plot.jl
```diff
@@ -233,24 +239,24 @@ function _plot!(plt::Plot, plotattributes::KW, args::Tuple)
...
 end
 
 
 # we're getting ready to display/output.  prep for layout calcs, then update
 # the plot object after
-function prepare_output(plt::Plot)
+@noinline function prepare_output(plt::Plot)
```
ChrisRackauckas: The prepare_output portion is actually quite trivial for compile time

src/backends/gr.jl
```diff
@@ -837,7 +842,504 @@ function _update_min_padding!(sp::Subplot{GRBackend})
     if sp[:colorbar_title] != ""
         rightpad += 4mm
     end
-    sp.minpad = Tuple(dpi * [leftpad, toppad, rightpad, bottompad])
+    sp.minpad = Tuple{AbsoluteLength,AbsoluteLength,AbsoluteLength,AbsoluteLength}(dpi * [leftpad, toppad, rightpad, bottompad])
+end
+
+@noinline function viewport_calcs(viewport_plotarea,viewport_size,data_lims,ratio)
```
ChrisRackauckas: Splitting out functions because the `(function () ... end)()` trick didn't make a noticable change. However, these still cause inference time because it's hit with values that are uninferred, and the problem is largely due to these being a function barrier. @vtjnash should `@nospecialize` do anything about that? I'm kind of confused why it doesn't.

src/backends/gr.jl
```diff
@@ -837,7 +842,504 @@ function _update_min_padding!(sp::Subplot{GRBackend})
...
+    end
+end
+
+@noinline function calc_volume(sp,y)
+    sp[:legend] = :none
+    GR.gr3.clear()
+    dmin, dmax = GR.gr3.volume(y.v, 0)
 end
 
 function gr_display(sp::Subplot{GRBackend}, w, h, viewport_canvas)
```
ChrisRackauckas: This function is 13 seconds, so it needs inference help for this to be fixed.

src/backends/gr.jl
```diff
@@ -851,56 +1353,18 @@ function gr_display(sp::Subplot{GRBackend}, w, h, viewport_canvas)
     xy_lims = data_lims
 
     ratio = sp[:aspect_ratio]
```
ChrisRackauckas: every single value here comes out uninferred. Womp womp.

src/backends/gr.jl
```diff
@@ -993,173 +1458,22 @@ function gr_display(sp::Subplot{GRBackend}, w, h, viewport_canvas)
...
-
+        is3d_calcs(sp,zaxis,xmin,xmax,ymin,ymax,viewport_plotarea)
     elseif ispolar(sp)
         r = gr_set_viewport_polar()
         #rmin, rmax = GR.adjustrange(ignorenan_minimum(r), ignorenan_maximum(r))
         rmin, rmax = axis_limits(sp, :y)
         gr_polaraxes(rmin, rmax, sp)
-
     elseif draw_axes
-        if xmax > xmin && ymax > ymin
```
ChrisRackauckas: If you comment these things out you get down do 8 seconds! But the issue is that adding in the function barrier doesn't get you to there.

Tokazama:
As I approach assisting on this I want to ensure I have a good plan of attack. Would it make sense to just have mutable attribute types instead of the `_default_xxx` dictionaries? For example, something like `AxisAttributes` which can be initialized with the default values by `AxisAttributes()`.

ChrisRackauckas:
Indeed that would be a good idea. Concretely typed structs for the defaults would help a lot.

Tokazama:
I've spent some more time digging through this and begun to work on defining the defaults. It's fairly clear why this has been addressed yet. The lack of type definitions is pretty thoroughly baked into the package so there will have to be quite a few decisions made.

* Numerous aliases for plot attributes: I can think of several ways to define concrete structs and still use the many different aliases for attributes but it's kind of ridiculous. Do we really need 9 different ways to identify "foreground color"? I doubt this will be a popular opinion, but we should have one name for each attribute. I think it overcomplicates learning the interface for users and would require rewriting them all in some new way (which wouldn't be trivial given the sheer number of aliases).
*  Turn attribute `ticks` into  `tickvalues` and `ticklabels`: currently ticks can be several things whether it has labels or not. Separating them would just apply lables if `tickslabels` was not an empty vector.
* `String` and/or `Symbol` are the acceptable types on multiple attributes. Here's a break down of several and what I think makes sense.
    * titles of any sort should be strings
    * positional arguments such as `:topleft` should always be symbols or there should be a some simple types to define these (e.g., subtypes of `AbstractPosition` like `TopLeft`). Unless we can define type aliases for  a position it's obnoxious to refer to the same thing by multiple Symbol arguments.
    * Fonts should be strings (unless there's some formal subtype of AbstractSring for a specific fons).
* Marker shapes: I haven't gotten to these yet but it seems like these would have to be defined by the GeometryTypes package at some point so it makes sense to have specific type definitions here.
* `arg_desc.jl` should be merged into the formal struct attribute definitions (which I've begun doing already).
* No more `:auto` values for attributes. Shouldn't this just be the default arguments anyway?
* How much of a role do extra user defined arguments (i.e. `_plot_defaults[:extra_kwargs]` ) actually play a role?

ChrisRackauckas:
>How much of a role do extra user defined arguments (i.e. _plot_defaults[:extra_kwargs] ) actually play a role?

User-defined arguments in type recipes are quite important. But the rest I agree with you. Accepting Symbol vs String: just do a convert in the first `plot!` command if you want to handle both, but it shouldn't permeate into every single data structure below that! `:auto` is not a way to handle defaults: default arguments are. Etc. It would take quite a bit of time to really unravel this, but it's all possible.

mauro3:
Are you aware of `Base.@kwdef`?  Probably useful here.

utotch:
I really hope your improvement :)
The slow plot problem is the most critical problem for me (and also for all the people who try to see the performance of Julia)

Issue #2182, opened 2019-09-11
gr doesn't plot additional plots with array data


olegshtch:
I'm trying to draw multiple series of different data on the single plot and provide ordinate data and approprivate labels as array.

It works without array:
```julia
gr()
plot([1,2],[1.5,2.5], labels = :o)
scatter!([1,2],[1,2], labels = :s)
```
or with plotly:
```julia
plotly()
plot([1,2],[[1.5,2.5]], labels = [:o])
scatter!([1,2],[[1,2]], labels = [:s])
```
But doesn't work with gr:
```julia
gr()
plot([1,2],[[1.5,2.5]], labels = [:o])
scatter!([1,2],[[1,2]], labels = [:s])
```
and produces error:
```
Error showing value of type Plots.Plot{Plots.GRBackend}:
ERROR: BoundsError: attempt to access 1-element Array{Symbol,1} at index [2]
Stacktrace:
 [1] getindex(::Array{Symbol,1}, ::Int64) at .\array.jl:728
 [2] gr_display(::Plots.Subplot{Plots.GRBackend}, ::Measures.Length{:mm,Float64}, ::Measures.Length{:mm,Float64}, ::Array{Float64,1}) at C:\Users\oleg_schelicalnov\.julia\packages\Plots\h3o4c\src\backends\gr.jl:894
 [3] gr_display(::Plots.Plot{Plots.GRBackend}, ::String) at C:\Users\oleg_schelicalnov\.julia\packages\Plots\h3o4c\src\backends\gr.jl:685
 [4] _show(::IOContext{Base64.Base64EncodePipe}, ::MIME{Symbol("image/png")}, ::Plots.Plot{Plots.GRBackend}) at C:\Users\oleg_schelicalnov\.julia\packages\Plots\h3o4c\src\backends\gr.jl:1601
 [5] _showjuno at C:\Users\oleg_schelicalnov\.julia\packages\Plots\h3o4c\src\output.jl:261 [inlined]
 [6] showjuno(::IOContext{Base64.Base64EncodePipe}, ::MIME{Symbol("image/png")}, ::Plots.Plot{Plots.GRBackend}) at C:\Users\oleg_schelicalnov\.julia\packages\Plots\h3o4c\src\output.jl:245
[...]

Issue #1550, opened 2018-06-07
[PyPlot] Accepting column vectors for labels


fbruetting:
That code doesn’t work in PyPlot due to the string vector being a column vector – in GR it works fine:
```julia
using Plots

param = [10; 20]
a = [1 2; 3 4]

plot(a, lab = "Parameter = " .* string.(param))
```

For PyPlot, I have to manually do something like the following, for making it a row vector:

```julia
plot(a, lab = "Parameter = " .* string.(reshape(param, 1, :)))
```

t-bltg:
This is documented: https://docs.juliaplots.org/stable/input_data/#columns-are-series

Use a matrix for your arguments: `param = [10 20]`.

Issue #1367, opened 2018-01-20
Wrong label printed when adding a series


improbable-22:
This plots four lines, but the last two are incorrectly labelled "five","six": 
```
using Plots
plot(randn(10,2), lab=["one", "two"])
plot!(randn(10,2) .- 2, lab=["three", "four", "five", "six"]) 
```
Using ` lab=["three", "four"]`, as I did at first, gives an error. 

Using a single string for either of the commands (`lab="one & two"` or `lab="three & four"`) works as expected. 

I have only tried this on GR backend, as I seem to have broken my python installation and unicodeplots doesn't seem to support a vector of labels. REPL and IJulia behave the same. Julia 0.6.0 on Mac OS, and now Plots 0.15.0.

apalugniok:
```
using Plots
plot(randn(10,2), lab=["one" "two"])
plot!(randn(10,2) .- 2, lab=["three" "four"])
```
Will produce what you'd like. 

As for why the way presented does something weird. Plots interprets every column as belonging to a single series. So when you pass a 1 column vector that vector of labels gets copied as the label for each series. 
e.g.
```
using Plots
plot(randn(10,2), lab=["one", "two"])
p = plot!(randn(10,2) .- 2, lab=["three", "four", "five", "six"]) 
p[1][3][:label] # accessing the label for the third series

4-element Array{String,1}:
 "three"
 "four" 
 "five" 
 "six"  
```

GR attempts to access this vector to get a single label using the series index (which causes an error if it's not long enough), PyPlot will interpret the whole vector as the label. 
![example](https://user-images.githubusercontent.com/22132297/35183891-ec8d16ee-fde4-11e7-9b2f-0f686efa568d.png)
I think the PyPlot behavior is more correct given what Plots is (and I guess should be 😆 ) doing. So in my opinion we should make GR behave in line with PyPlot @mkborregaard, @daschw ?

improbable22:
Ah, thank you, I would never have guessed! 

I always forget that vectors are thought of as columns, not just one-index objects... but given that, PyPlot's behaviour makes sense, and agrees with how `plot(::Matrix)` works.

daschw:
PyPlot definitely has the "correct" behaviour, althoug I doubt that passing a Vector of Strings is really is intended in many cases. Maybe we should add a warning, asking the users if they wanted to pass a RowVector.

pkofod:
[...]

Issue #996, opened 2017-08-03
Array comprehensions in labels work in GR but not in PlotlyJS


dpsanders:
```
plot([x->sin(x+i) for i in 1:5], label=["Hello $i" for i in 1:5])
```

gives correct labels in GR but not in PlotlyJS.
In PlotlyJS it is necessary to make the labels into a matrix:

```
plot([x->sin(x+i) for i in 1:5], label=reshape(["Hello $i" for i in 1:5], 1, 5))
```

This seems to be an actual bug.
In any case, the distinction between when to use vectors versus matrices seems pretty confusing.

ChrisRackauckas:
This is an unnatural consequence of the "columns are series" setup of Plots.jl. Each column of the labels matches a series. However, I think `label` should be special-cased so that `Vector` becomes the proper `RowVector` since the `Vector` case just isn't useful IMO.

dpsanders:
Yes, agreed.

What I don't understand is why the behaviour is different for the two different backends.

mkborregaard:
I think the distinction between vectors and matrices is clear - matrix columns are series (even if there's only one row), vector points are data points :-)
@ChrisRackauckas I don't see anything wrong with automatically transposing `label` if the length corresponds to the number of series (other than it may cause confusion by overshadowing the columns-are-series fundamental Plots logic).
@dpsanders no that's confusing - in fact the bug is currently in GR though. Any suggestions for making this clearer to users?

PR #2183, opened 2019-09-11
RFC: Print vectors passed to label as vectors in GR

### LINKED ISSUE #2179: Multiple legend labels with `pyplot()` are misformatted. Works fine under `gr()`
`gr(); plot(rand(10,5),label=[1,2,3,4,5])` results in:

image

but with `pyplot(); plot(rand(10,5),label=[1,2,3,4,5])` results in this presumably incorrect behavior:

image

```
PyPlot v"2.8.2"
Plots v"0.26.2"
PyPlot.PyCall.pyimport("matplotlib").__version__ == "3.1.1"
Julia version v"1.4.0-DEV.106"
```

daschw:
Fixes #2179 and partially (for GR) #2182 
```julia
using Plots
plot(rand(10,5),label=[1,2,3,4,5])
```
label_gr

```julia
plot([1,2],[1.5,2.5], labels = [:o])
scatter!([1,2],[1,2], labels = [:s])
```
used to fail with gr and now gives
label_gr2

I'm actually not really sure what the use cases for passing vectors as labels could be. The legends in the plots above don't look that nice. But I think we should strive consistency among different backends and stick to http://docs.juliaplots.org/latest/input_data/#columns-are-series.

Any opinions?

Related issues:
#901
#996
#1367
#1550
#2179
#2182

### DIFF
--- a/src/backends/gr.jl
+++ b/src/backends/gr.jl
@@ -874,7 +874,6 @@ function gr_display(sp::Subplot{GRBackend}, w, h, viewport_canvas)
     # has to be done now due to a potential adjustment to the plotarea given an outer legend.
     legendn = 0
     legendw = 0
-    legendi = 0
     if sp[:legend] != :none
         GR.savestate()
         GR.selntran(0)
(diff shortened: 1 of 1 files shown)


JackDevine:
This seems logical, it is good to see that you found a case where the docs support this change on a philosophical level.

I agree that the legends you show do not look very nice, although the alternative is the gr behaviour shown in #2179 .

Just to play the devils advocate, you could justify the gr behaviour in #2179 by saying that an nx1 array or a 1xn array are just vectors, so you should treat them the same. That justification seems a little unjulian to me, although it does follow the Plots.jl philosophy of "do what I mean".

FWIW, my vote is that this PR implements the correct behaviour.

daschw:
There does not seem to be any objection. So I'll go ahead and merge. For now this at least adds consistency. If we want to automagically treat vectors of labels as row vectors we should do this in a uniform approach for all backends and not just for GR

PR #2186, opened 2019-09-12
Fix plotting ribbons as 2-tuples

asinghvi17:
…e{Any,Any}, plottatributes)

@ChrisRackauckas @daschw 

Fixes #2185

PR #2188, opened 2019-09-15
Simplify adding new test imgages via PlotReferenceImages.jl

daschw:


daschw:
With this it is now possible to update or add to the Plots reference images for testing following these steps:
- Add PlotReferenceImages.jl and dev it:
```julia
pkg> dev PlotReferenceImages

pkg> add PlotReferenceImages
```
- Fork https://github.com/JuliaPlots/PlotReferenceImages.jl and add your fork as remote in `~/.julia/dev/PlotReferenceImages`.
- Checkout a new branch in PlotReferenceImages.
- If you want to add a new test example **at the end of** the `_examples` array in [`~/.julia/dev/Plots/src/examples.jl`](https://github.com/JuliaPlots/Plots.jl/blob/master/src/examples.jl).
- Run the Plots test suite:
```julia
pkg> test Plots
```
- If test images differ or new ones are added a Gtk window opens showing the images and asking you whether to save the new image. Add the new images, if you consider them an improvement.
- `git add` the new images in `~/.julia/dev/PlotReferenceImages`.
- Submit a PR to PlotReferenceImages
- Submit a PR to Plots with the changes in examples.
Once the PlotReferenceImages PR is merged the Plots PR tests should pass.

The current open PRs requiring these actions are #2062 (@yha) and #2186 (@asinghvi17). 
Sorry that you had to wait so long for this.

asinghvi17:
@daschw, what is the effect of adding an image to the middle of the examples?  The reason was to show off ribboning, which wouldn't make sense to put right at the end.

daschw:
The files in PlotReferenceImages are named after there index in `Plots._examples`. So, if you add a new example all later examples will be compared to the wrong file.
Technically, updating all examples after the new one and submitting the change to PlotReferenceImages would also be fine and allow the tests to pass. I just thought putting them at the end was simpler and more transparent.

> The reason was to show off ribboning, which wouldn't make sense to put right at the end.

I don't think that there is any intended structure in the order of the examples.

Issue #2136, opened 2019-08-09
Noninteractive display in Juno under plotlyjs()


BoundaryValueProblems:
Thanks to the recent effort of @daschw , we can now enjoy the ```PlotlyJS``` if we use it directly without using ```Plots```. However, under ```Plots```, ```plotlyjs``` backend still shows a noninteractive figure on Juno's plot pane: the top part of the figure does not show the interaction menu. My current configuration is: Julia: v1.1.1, Atom: v1.39.1, ink: v0.10.12, julia-client: v0.9.4.
The versions of various appropriate julia packages I installed are:
```
  [c52e3926] Atom v0.9.1
  [ad839575] Blink v0.12.0 #master (https://github.com/JunoLab/Blink.jl.git)
  [e5e0dc1b] Juno v0.7.1
  [47be7bcc] ORCA v0.2.1
  [f0f68f2c] PlotlyJS v0.12.5 #master (https://github.com/sglyon/PlotlyJS.jl.git)
  [91a5bcdd] Plots v0.26.1 #master (https://github.com/JuliaPlots/Plots.jl.git)
  [0f1e0344] WebIO v0.8.9 #master (https://github.com/JuliaGizmos/WebIO.jl.git)
```
The situation is the same regardless of the platform (macOS, Windows 10, Ubuntu/Linux).
I truly wish that we could use ```plotlyjs``` backend of ```Plots``` under Juno very soon; I have been waiting for this since Julia v1.0.0, i.e., almost a year...

BoundaryValueProblems:
@daschw and @pfitzseb : the situation is still the same with all the latest versions of Atom; Juno; Plots.jl; PlotlyJS.jl, etc. Using ```plotlyjs()``` under Plots.jl, the interactive menu at the top does not appear in a plot pane of Atom/Juno. On the other hand, if we directly do ```using PlotlyJS``` followed by some plotting functions, then everything works. So, we should be close to solve this really long-standing problem, right? What is really the source of this problem at this point of time?

pfitzseb:
Turns out there was a missing `show` method, probably because I forgot to implement that after reworking something. The PR referenced above should fix that though:
![image](https://user-images.githubusercontent.com/6735977/64953998-a199ba80-d884-11e9-9f9d-a22bbf0289e6.png)

BoundaryValueProblems:
[...]

PR #2189, opened 2019-09-16
allow html output in Juno

pfitzseb:
e.g. for the PlotlyJS backend.

Fixes https://github.com/JuliaPlots/Plots.jl/issues/2136.

Not sure why the `collect` is necessary, but apparently `plt[:size]` sometimes is a generator.


daschw:
That's amazing, thanks @pfitzseb ! Finally `plotly` plots are shown in Juno's plot pane!

PR #2190, opened 2019-09-16
make Juno plots devicePixelRatio aware

pfitzseb:
Together with https://github.com/JunoLab/Atom.jl/pull/179 and https://github.com/JunoLab/atom-julia-client/pull/609 this should make sure plots in Juno are DPI-aware.

I didn't actually test this yet, but it should work :P


daschw:
Great!

Issue #1335, opened 2018-01-02
alternative extension for tikz plots


tpapp:
Saving `savefig("file.tikz")` using the `tex` format would be convenient since it is supported by the [tikzscale](https://www.ctan.org/pkg/tikzscale) LaTeX package, and `.tex` isn't.

My understanding is that adding `"tikz" => tex` to [`_savemap`](https://github.com/JuliaPlots/Plots.jl/blob/1949b6fe0f6afef3b1d7d0e0d39a78510a7966c1/src/output.jl#L69) would accomplish this.

Just wanted to check that this is OK before I submit a PR.

mkborregaard:
:+1: from me - @pkofod any opinions otherwise?

pkofod:
Was this ever followed up upon? This sounds reasonable.

mkborregaard:
Nope, it wasn't followed up on, but I also think it sounds reasonable. @tpapp still up for submitting a PR? :-)

tpapp:
@mkborregaard: sorry, I kind if moved to [PGFPlotsX.jl](https://github.com/KristofferC/PGFPlotsX.jl) for my plotting needs.

mkborregaard:
NO! :-O :cry:

PR #2191, opened 2019-09-17
add tikz ending

BeastyBlacksmith:
Enables saving the tex-output with ".tikz"-ending.

Fixes https://github.com/JuliaPlots/Plots.jl/issues/1335.


daschw:
Thanks!

Issue #2197, opened 2019-09-19
Way to pass keywords to the backend


BeastyBlacksmith:
It would be nice to have a way to pass backend-specific keywords.
One idea is, to have a `backend_options` keyword that takes a `Dict` or `NamedTuple` that gets passed as keywords to the backend.

mkborregaard:
Do you feel like experimenting with a PR for one of the backends? It's been suggested several times.

BeastyBlacksmith:
I did take a quick look on it, but the backend <-> frontend interface is not as clear-cut, as i had imagined and I get lost in the Plots pipeline quite quickly. So at least till the beginning of next year I won't have time.

mkborregaard:
Have a look how this keyword arg was added and passed through to PyPlot https://github.com/JuliaPlots/Plots.jl/pull/1009/files

BeastyBlacksmith:
Actually I see, that `extra_kwargs` does already exist, it is "just" not passed at the moment.

BeastyBlacksmith:
Maybe it is at the wrong spot though, since it belongs to the `Plot` object, but as far as I can see, the backends operate on the `Subplot` level?

mkborregaard:
I'm not sure what you mean by that?

BeastyBlacksmith:
Let put it as a question: How would a backend query the plotattributes?

mkborregaard:
The backend processes the Plot object and builds up a `series_list` containing the keywords passed to the backend. So I guess the extra_kw args should just be appended to the `series_list` without any transformations.

BeastyBlacksmith:
A bit of a memo to myself :
- extra_kwargs needs to be actually be populated
- this requires different handling of unsupported kwargs, like only checking the Plots specific kwargs and put the rest in extra_kwargs

isentropic:
perhaps:
`plot(x1, x2, backendkeyword="parameter=value")`
aliases:
`plot(x1, x2, bk="parameter=value")`  
`plot(x1, x2, backend_keywords="parameter=value, parameter2=value2")`

BeastyBlacksmith:
I think it would make most sense if you were checkout #2200 and see, if you can make it work for pyplot too

PR #2200, opened 2019-09-20
passing of extra keywords to pgfplotsx

BeastyBlacksmith:
This showcases how extra keyword passing could be realized, by populating the existing `extra_kwargs` and pass it in the pgfplots backend.

Partially fixes https://github.com/JuliaPlots/Plots.jl/issues/2197.


BeastyBlacksmith:
Currently the interface is take all arguments, warn if it is an unsupported Plots argument, pass it to the backend otherwise.
E.g.: `plot(x, y, arrow = true)` would warn and
`plot(x, y, texlabel="what a label")` would pass `texlabel` on.
The alternative would be the more explicit `plot(x,y, xlimits=(0,10), extra_kwargs=Dict(:do=>"that"))` or `plot(x,y, xlimits=(0,10), extra_kwargs=(do = "that"))` and keep warning about usupported keyword arguments. Preferences?

BeastyBlacksmith:
Also while this works for something like `plot(1:5, texlabel="test")` it does not work at the moment if you do `plot!(x->x^2, 1,5)` afterwards.
I think this is due to `extra_kwargs` being in `Plot.attr` rather than `Subplot.attr`.

BeastyBlacksmith:
@mkborregaard , @daschw: this works for me. Would be good to get some feedback regarding the interface (https://github.com/JuliaPlots/Plots.jl/pull/2200#issuecomment-535960763).
Have not checked yet how difficult it would be for the other backends.

daschw:
> Would be good to get some feedback regarding the interface (#2200 (comment)).

I would go with the current implementation. `kwarg = val` just feels more user-friendly than `extra_kwargs = (kwarg = val)`.

BeastyBlacksmith:
After my experience with https://github.com/JuliaPlots/Plots.jl/pull/2252 I can say, that this would need `extra_kwargs` dict on every layer: Plot, Subplot, Axis, Series(, Segments).

daschw:
> After my experience with #2252 I can say, that this would need extra_kwargs dict on every layer: Plot, Subplot, Axis, Series(, Segments).

Does this mean that this PR would have to be changed to have `extra_plot_kwargs`, `extra_subplot_kwargs`, `extra_axis_kwargs` and `extra_series_kwargs`?

I think that's a little much. IMO Plots does not have to provide all the functionality of each backend. The idea is rather to have a common interface for the plotting utilities most used and demanded by the community. For special, detailed tweaking of plots going beyond Plots attributes I'd point the users to use the backend directly.

daschw:
However, that's just my opinion and I can also see the point of having these `extra_kwargs`. If we really need them on different layers, I think my previous statement regarding `kwarg = val` vs `extra_kwargs = (kwarg = val)` is no longer valid. In that case the `kwarg = val` is no longer possible, right? So we'd have to go with the second option.

mkborregaard:
I agree with Daniel. 

But, what about allowing passing a `Dict` of `Dicts`, eg `Dict(:Subplot => Dict(:colorbar => :extend))`, it would be terribly clunky but allow people with highly particular needs a workaround to resolve stuff?

daschw:
That's a good idea.

BeastyBlacksmith:
Turned it into Dict of Dicts and collect unsupported kwargs into the series-dict, since that is probably the most useful.
Alternatively could only allow passing them explicitly.

BeastyBlacksmith:
@mkborregaard, @daschw I just change the interface a bit as you can see in the added tests and I think it is quite okay from the user perspective.
Its a bit annoying for the backends that extra keywords for the plot is stored in `extra_plot_kwargs` and the others in `extra_kwargs`.
So happy to hear what you think.

BeastyBlacksmith:
Demo:
```julia
    plot(plot(1:5, title="Line"), scatter(1:5, title="Scatter", extra_kwargs=Dict(:subplot=>Dict("axis line shift" => "10pt"))))
```
image

daschw:
Looks good! I don't have a better idea than `extra_plot_kwargs`. Does this really require a 2.0 release?

BeastyBlacksmith:
No, it was more a "should be done before 2.0"
Probably minor version bump?

PR #2217, opened 2019-10-11
fix html output of video animations

### LINKED ISSUE #2216: HTML output for video animation broken
Plots.jl producies invalid HTML for mp4 animations. As a result, videos dont get displayed properly in Jupyter.

Solution: Add close-quotation mark on this line https://github.com/JuliaPlots/Plots.jl/blob/v0.27.0/src/animation.jl#L107

daschw:
closes #2216.
Thanks @gwater


gwater:
thanks!

PR #2226, opened 2019-10-21
Image extent half-pixel margin in pyplot

nlw0:
Adds a half-pixel margin to the extent for image plots so that the pixel coordinates match the grid, i.e. the center of each square pixel lies over its integral index numbers, and each square covers half a pixel before and after that.


daschw:
Thanks!

Issue #1428, opened 2018-03-09
Unevenly spaced x and/or y displayed as even in heatmap  GR


yakir12:
Consider plotting a heatmap when the the axes are not evenly spaced:
```julia
using Plots
gr()
x = [1,2,3,4,5]
y = [1,3,5,7,15]
z = x .+ y'
heatmap(x, y, z)
```
wrongly results in:
![screenshot_2018-03-09_22 13 02](https://user-images.githubusercontent.com/3281957/37230264-165a0e14-23e7-11e8-8143-a677abc45569.png)
While the plotly backend correctly results in:
![screenshot_2018-03-09_22 11 39](https://user-images.githubusercontent.com/3281957/37230275-2963bfaa-23e7-11e8-8ce2-069fb88cd04f.png)
(these are screenshots)

mkborregaard:
But is the plotly one really "correct"? What are the borders you'd expect the turnover between categories to correspond to? It doesn't even look like 15 is in the middle on the top category for plotly. I'd say this is a bit hard to define.

mkborregaard:
`histogram2d` which uses some of this functionality assumes that the number is always at the center of the cell

mkborregaard:
But GR's behaviour is clearly wrong:
```
using PlotRecipes
marginalhist(randn(10000), randn(10000), bins = [-4,-2,-1,0,0.5,2,5])
```
<img width="590" alt="skaermbillede 2018-03-10 kl 12 39 54" src="https://user-images.githubusercontent.com/8429802/37241848-3e2bb61c-2460-11e8-9d4b-9b0c7299f178.png">
This looks very wrong, note also the coloring of the marginal histograms
This is completely wrong in Plotly too, it seems the axes become automatically linked:
<img width="620" alt="skaermbillede 2018-03-10 kl 12 41 56" src="https://user-images.githubusercontent.com/8429802/37241866-725cc2be-2460-11e8-85f6-5711e7a18a76.png">

Really terrible.

yakir12:
Yea, I see what you mean about Plotly. I guess I was just checking to see if the colored patches were unevenly sized (which they are). Tricky.

aramirezreyes:
Is there some work going on regarding this behavior?

mkborregaard:
If it's not fixed yet then most likely not. Are you trying to take a look?

lmh91:
At last JuliaCon I talked to @jheinen regarding uneven spaced heatmaps with GR. 
[...]

PR #2234, opened 2019-11-01
Nonuniform heatmaps are now possible with the GR backend.

lmh91:
Regarding the issue #1428.

### DIFF
--- a/src/backends/gr.jl
+++ b/src/backends/gr.jl
@@ -920,11 +920,26 @@ function gr_display(sp::Subplot{GRBackend}, w, h, viewport_canvas)
             outside_ticks = true
             for ax in (sp[:xaxis], sp[:yaxis])
                 v = series[ax[:letter]]
-                if length(v) > 1 && diff(collect(extrema(diff(v))))[1] > 1e-6*std(v)
-                    @warn("GR: heatmap only supported with equally spaced data.")
-                end
             end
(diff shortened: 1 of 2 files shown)


yha:
This breaks plotting heatmaps on non-linear scales. 
I think the support for directly specifying edges should be added by extending the `heatmap_edges` function (probably by a new function wrapping it) in `utils.jl`, so that it can be used by other backends as well. This can also avoid the code duplication in this PR, and deal with non-linear axes (as `heatmap_edges` already does).

lmh91:
Good points!
I am still learning how Plots.jl works...
I hope I adjusted it in a proper way.

src/backends/gr.jl
```diff
@@ -1317,23 +1314,32 @@ function gr_display(sp::Subplot{GRBackend}, w, h, viewport_canvas)
...
+                nx, ny = length(series[:x]), length(series[:y])
+                z_normalized = map(x -> GR.jlgr.normalize_color(x, zmin, zmax), z)
+                colors = Int32[round(Int32, 1000 + _i * 255) for _i in z_normalized]
+                xmin, xmax, ymin, ymax = xy_lims 
+                rmax = data_lims[4]
+                GR.setwindow(-rmax, rmax, -rmax, rmax)
+                if ymin > 0 
+                    @warn "'ymin[1] > 0' (rmin) is not yet supported."
+                end
+                @show series[:y][end]
```
daschw: There's a left-over @show.

oschulz:
Thanks for taking this on, Lukas!

Issue #1261, opened 2017-11-23
Reinterpret -Inf and Inf as NaN


mkborregaard:
Sometimes input arrays have -Inf and Inf in them, which makes Plots throw an error. Sometimes they are produced internally, if someone calls `yscale = :log` and there are zeros in the data. @dpsanders and @StefanKarpinski suggested on Slack that these should be ignored, which would happen automatically if they're converted into NaN. We need to find the right place in the pipeline to do it.

mkborregaard:
I've earlier argued that it is user error to pass an array with 0s in it and expect a log scale to work. What convinced me to try to deal with this is the difficulty, user side, in handling this gracefully as both x and y inputs will need to be reduced to non-0 slots.

ma-laforge:
That seems like a good idea - as long as there is an option for backends to handle these cases themselves (if they can do so elegantly).

Of course, this is more for the cases where it is not Plots.jl itself that throws the errors.

An other option would be implement something like what InspectDR does:

![demo9_export](https://user-images.githubusercontent.com/11205411/33272842-746b4fd8-d359-11e7-8194-db837543153b.png)

InspectDR supports a mode which basically scans through the data before plotting, strips out NaN data, and overlays a "NaN" dataset so the user can see which points are problematic.  This only works if only one of the two (x,y) coordinates is a NaN, though.

mkborregaard:
NaN's are generally handled by the backends - so the idea is to change Inf to NaN for better handling by the backends. Most backends just drop data points where one or both is NaN

ma-laforge:
Fair enough.  Though I believe InspectDR actually handles Inf correctly as well (drawing vertically up to the positive/negative edge of the plot).  That way, when the user zooms out, it always looks as if the line goes off to infinity.

If I forgot to implement it, there is no reason why I cannot make it happen - so it would be nice to have the option of passing off infinite data to the backend (as long as it does not cause Plots.jl itself to crash).

PR #2236, opened 2019-11-03
Convert infinite values to NaN

yha:
Ref #1261. The discussion there includes some other ideas for how to plot infinite values, so maybe should be left open.
An example which plots incorrectly on GR before this change:
```
plot([1,Inf,2,3])
```
Corresponding reference image: JuliaPlots/PlotReferenceImages.jl#49

### DIFF
--- a/src/series.jl
+++ b/src/series.jl
@@ -14,7 +14,10 @@ const SeriesData = Union{AVec{<:DataPoint}, Function, Surface, Volume}
 
 prepareSeriesData(x) = error("Cannot convert $(typeof(x)) to series data for plotting")
 prepareSeriesData(::Nothing) = nothing
-prepareSeriesData(s::SeriesData) = handlemissings(s)
+prepareSeriesData(s::SeriesData) = handleinfinites(handlemissings(s))
+
+handleinfinites(s) = s
+handleinfinites(s::AbstractArray{<:MaybeNumber}) = [isinf(x) ? NaN : x for x in s]
 
 handlemissings(v) = v
 handlemissings(v::AbstractArray{<:MaybeNumber}) = replace(v, missing => NaN)
--- a/src/series.jl
+++ b/src/series.jl
@@ -22,8 +22,6 @@ prepareSeriesData(s::Surface{<:AMat{<:MaybeNumber}}) = Surface(prepareSeriesData
 prepareSeriesData(s::Surface) = s  # non-numeric Surface, such as an image
 prepareSeriesData(v::Volume) = Volume(prepareSeriesData(v.v), v.x_extents, v.y_extents, v.z_extents)
 
-prepareSeriesArray(a::AbstractArray{<:MaybeNumber}) = [ismissing(x) || isinf(x) ? NaN : x for x in a]
-
 # default: assume x represents a single series
 convertToAnyVector(x, plotattributes) = Any[prepareSeriesData(x)]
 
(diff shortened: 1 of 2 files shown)


src/series.jl
```diff
@@ -14,7 +14,10 @@ const SeriesData = Union{AVec{<:DataPoint}, Function, Surface, Volume}
 
 prepareSeriesData(x) = error("Cannot convert $(typeof(x)) to series data for plotting")
 prepareSeriesData(::Nothing) = nothing
-prepareSeriesData(s::SeriesData) = handlemissings(s)
+prepareSeriesData(s::SeriesData) = handleinfinites(handlemissings(s))
+
+handleinfinites(s) = s
+handleinfinites(s::AbstractArray{<:MaybeNumber}) = [isinf(x) ? NaN : x for x in s]
```
mkborregaard: This copies the input data. We already do that in one place, in they `computexy` functions. Could we handle that there so we don't copy twice?
Alternatively, given that we do copy the input data, it should be safe to use `replace!` here.

src/series.jl
```diff
@@ -14,7 +14,10 @@ const SeriesData = Union{AVec{<:DataPoint}, Function, Surface, Volume}
...
 
 handlemissings(v) = v
 handlemissings(v::AbstractArray{<:MaybeNumber}) = replace(v, missing => NaN)
```
mkborregaard: and maybe here too?

yha:
We can't rely on `compute_xyz` making a copy, because it's only called after `convertToAnyVector`.
Also, we can't just `replace!`, since `NaN`s may not be compatible with the vector type, and we also want to remove `Missing` from the type.
The new version of this PR copies only once in `prepareSeriesData`, but there's still additional copying in `compute_xyz`. Also, I realized that functions returning infinities or `missing`s can't be handled in this design. I think both these problems can be fixed by handling `plot(::Function, x)` entirely in a recipe rather with `compute_xyz`, but that should probably be another PR.

mkborregaard:
But `ConvertToAnyVector` is always called, no?

mkborregaard:
> since NaNs may not be compatible with the vector type

Which Vector type has `Inf` but not `NaN`?

yha:
> But `ConvertToAnyVector` is always called, no?

Yes, but before `compute_xyz`, with the original input vectors. If it doesn't make a copy, it would replace infinities and missings in the original input.

> Which Vector type has `Inf` but not `NaN`?

Probably none. But you could get  a `v::Vector{Union{Int,Missing}}`, and then `replace!(v,missing=>NaN)` would error.

mkborregaard:
>  If it doesn't make a copy, it would replace infinities and missings in the original input

I don't think that would ever happen?

> But you could get a v::Vector{Union{Int,Missing}}, and then replace!(v,missing=>NaN) would error.

We have `Int`s this late in the pipeline?

yha:
Apparently we do have raw input vectors at this point in the pipeline, including `Int`s.
If I change the method in lines 17:19 from `replace!(..., map(float,a))` to `replace!(..., a)` I get
```
plot([1,missing])   # => InexactError: Int64(NaN)
x = [1.0,missing,Inf]
plot(x)
x      # => [1.0, NaN, NaN]
```

yha:
I also updated the example to show `missing`s as well, and test the case of `Vector{Union{Int,Missing}}`.

Incidentally, is there a reason for this check guarding the image comparison tests?
https://github.com/JuliaPlots/Plots.jl/blob/464c0271f026e53037c64a30c3dc3b40e0dd1ec3/test/runtests.jl#L52
I just commented it out to run them on Windows and it seems to work fine, except one failing test.

mkborregaard:
I guess that was back when ImageMagick was used everywhere and flaky

daschw:
We don't want to run the image comparison tests on Appveyor because we already test them on Travis, it takes very long on Appveyor and some plots failed due to different fonts between Windows/Linux or other similar issues.
But I guess we could also use some Appveyor environment variable for this check instead of `Sys.islinux()` - in fact, we should.

yha:
I think so, yes.

Issue #2239, opened 2019-11-05
Inline animations in Jupyter always displays the first animation


johanbluecreek:
I have an issue where inline animations in Jupyter does not update the plot. That is, if I take the example from the [Plots docs](https://docs.juliaplots.org/latest/), The Lorenz attractor, and I change code by removing the `every 10` argument, then I get the animation and it is drawn very slowly, not like in the docs. So I add `every 10` back, and I still get the same animation back, the slow one. I can verify that they are different, because the `tmp.gif` saved is different, it plots fast, as in the docs.

I can even take say
```julia
@gif for i=1:10
heatmap(rand(10,10))
end
```
in a separate Cell, and it will still show the Lorenz attractor.

Since the stable version of `Plots.jl` (v0.27.0) has an issue with displaying inline output of `mp4()` I have also reproduced this on the dev branch (464c0271f026e53037c64a30c3dc3b40e0dd1ec3). And the same problem is present for `mp4()` as with `@gif`.

So, summarised:
* Steps to reproduce:
  - Take the example from [Plots docs](https://docs.juliaplots.org/latest/)
  - Change the plotting in any manner or plot a new animation
  - Only the first animation will be displayed
* Julia version: 1.2.0 (stable official binary)
* `Plots.jl` versions: v0.27.0 and devel (464c0271f026e53037c64a30c3dc3b40e0dd1ec3)
* Back-end: GR
* Functions affected: `mp4()` and `@gif`

I could not find this issue already reported.

daschw:
Thanks for reporting! Reloading the browser tab shows the current image. However, I see that this is not a nice workflow. I have no idea how to fix this within Plots. Any suggestions are very welcome.

johanbluecreek:
Thanks for your reply.

> Reloading the browser tab shows the current image. 

Ah, that is a work-around at least. It does present another issue though: If I make two different animations, it is always the last one made before reloading. It seems to make it impossible to have more than one animation in the notebook at once (unless you have one `@gif` and one `mp4()`, that works.).

How does the embedding into the notebook work? Is it the `tmp.gif`/`tmp.mp4` files that gets embedded? Then could it be that it is not updated because it has the same name? Perhaps generating unique file-names would avoid this? I perhaps have the time to check in a few hours. If that works, I guess it creates an unwanted problem of increasingly more `tmp.*` files...

johanbluecreek:
[...]

PR #2240, opened 2019-11-06
animation.jl: using randstring() to generate unique tmp names (fix: 2239)

johanbluecreek:
Browser caching prevents animations from being updated if names are not
unique. This solves Issue #2239.

### DIFF
--- a/src/animation.jl
+++ b/src/animation.jl
@@ -21,9 +21,9 @@ function frame(anim::Animation, plt::P=current()) where P<:AbstractPlot
     push!(anim.frames, filename)
 end
 
-giffn() = (isijulia() ? "tmp.gif" : tempname()*".gif")
-movfn() = (isijulia() ? "tmp.mov" : tempname()*".mov")
-mp4fn() = (isijulia() ? "tmp.mp4" : tempname()*".mp4")
+giffn() = (isijulia() ? "tmp_"*randstring()*".gif" : tempname()*".gif")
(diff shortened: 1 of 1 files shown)


src/animation.jl
```diff
@@ -21,9 +21,9 @@ function frame(anim::Animation, plt::P=current()) where P<:AbstractPlot
     push!(anim.frames, filename)
 end
 
-giffn() = (isijulia() ? "tmp.gif" : tempname()*".gif")
-movfn() = (isijulia() ? "tmp.mov" : tempname()*".mov")
-mp4fn() = (isijulia() ? "tmp.mp4" : tempname()*".mp4")
+giffn() = (isijulia() ? "tmp_"*randstring()*".gif" : tempname()*".gif")
```
daschw: Have you tried 
```julia
giffn() = tempname() * ".gif"
```
instead as well? Is it necessary to have a different behavior for IJulia?

johanbluecreek: Yes, I did. That made nothing show up inline at all. I'm not sure why nothing showed up, but I suspected that there was some reason for that and that is the reason there was a `isijulia()` if there to begin with. I tested it both with `gif()` and `mp4()`, both showed nothing with `tempname()`

If that is the preferred solution I can try and figure out why, and see if I can work around it.

daschw: I was just curious. Whatever works is fine for me.

daschw:
Thanks a lot!

Issue #2238, opened 2019-11-04
plot no longer plots two graphic at the same time correctly


alea54:
It seems that with 0.27, Plot no longer plots two functions at the same time correctly.

try

plot([sin,cos],0,pi)

![Capture d’écran de 2019-11-04 16-34-55](https://user-images.githubusercontent.com/57280477/68133660-1e5e2200-ff21-11e9-9805-a4138ba5f97e.png)

The result was correct with 0.26

daschw:
Thanks for reporting!
@BeastyBlacksmith might this be related to https://github.com/JuliaPlots/Plots.jl/pull/2204/files and/or https://github.com/JuliaPlots/PlotUtils.jl/pull/68?

BeastyBlacksmith:
Most likely the issue should be [here](https://github.com/JuliaPlots/Plots.jl/blob/a76e756f4f0de5046964111d5cc9b2b0280ab017/src/series.jl#L520). I can take a look at the weekend.

PR #2242, opened 2019-11-08
fix multiplot of functions

BeastyBlacksmith:
fixes https://github.com/JuliaPlots/Plots.jl/issues/2238


daschw:
Awesome, thanks!

PR #2243, opened 2019-11-09
Alternative fix for #2239

antoine-levitt:
Reverts https://github.com/JuliaPlots/Plots.jl/pull/2240 to avoid multiple temp files. I have only tested it with gif files.

### DIFF
--- a/src/animation.jl
+++ b/src/animation.jl
@@ -104,15 +104,18 @@ end
 # write out html to view the gif
 function Base.show(io::IO, ::MIME"text/html", agif::AnimatedGif)
     ext = file_extension(agif.filename)
-    write(io, if ext == "gif"
-        "<img src=\"data:image/gif;base64," * base64encode(read(agif.filename)) * "\" />"
+    if ext == "gif"
+        html = "<img src=\"data:image/gif;base64," * base64encode(read(agif.filename)) * "\" />"
     elseif ext in ("mov", "mp4")
-          "<video controls><source src=\"data:video/$ext;base64," *
-          base64encode(read(agif.filename)) *
-          "\" type = \"video/$ext\"></video>"
+        mimetype = ext == "mov" ? "video/quicktime" : "video/mp4"
+        html = "<video controls><source src=\"data:$mimetype;base64," *
+               base64encode(read(agif.filename)) *
+               "\" type = \"video/$ext\"></video>"
     else
         error("Cannot show animation with extension $ext: $agif")
-    end)
+    end
+
+    write(io, html)
     return nothing
 end
 
--- a/src/animation.jl
+++ b/src/animation.jl
@@ -104,12 +104,12 @@ end
 # write out html to view the gif
 function Base.show(io::IO, ::MIME"text/html", agif::AnimatedGif)
     ext = file_extension(agif.filename)
-    link = relpath(agif.filename)
-    link *= "?"*randstring() #to foil browser cache, see https://github.com/JuliaPlots/Plots.jl/issues/2239
     write(io, if ext == "gif"
-        "<img src=\"$link\" />"
+        "<img src=\"data:image/gif;base64," * base64encode(read(agif.filename)) * "\" />"
     elseif ext in ("mov", "mp4")
-        "<video controls><source src=\"$link\" type=\"video/$ext\"></video>"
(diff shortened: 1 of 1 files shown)


antoine-levitt:
OK so I was happy with this solution, until I tried to export. Of course, there are explicit links to the gifs and so I can't just export to HTML. Can't Plots include the gif in the html, just like with regular static plots?

antoine-levitt:
So that works:
```diff
diff --git a/src/animation.jl b/src/animation.jl
index 659353e5..15a76a17 100644
--- a/src/animation.jl
+++ b/src/animation.jl
@@ -107,7 +107,7 @@ function Base.show(io::IO, ::MIME"text/html", agif::AnimatedGif)
     link = relpath(agif.filename)
     link *= "?"*randstring() #to foil browser cache, see https://github.com/JuliaPlots/Plots.jl/issues/2239
     write(io, if ext == "gif"
-        "<img src=\"$link\" />"
+        "<img src=\"data:image/gif;base64," * base64encode(read(agif.filename)) * "\" />"
     elseif ext in ("mov", "mp4")
         "<video controls><source src=\"$link\" type=\"video/$ext\"></video>"
     else
```
and using that I managed to export a notebook with animations to a HTML file. If people think this is a good idea I can update the PR.

antoine-levitt:
Do you mean the commits in the branch, or the diff above? I think the solution in the diff is much better, because it works when exporting notebooks. But it's a bigger change and has some drawbacks (the .ipynb inflate in size), so I don't know if it's acceptable.

pfitzseb:
Also note that data URIs [are probably limited in size](https://stackoverflow.com/questions/695151/data-protocol-url-size-limitations) (although I'm not sure if those answers are still relevant).

You could also try [this](https://jupyter-contrib-nbextensions.readthedocs.io/en/latest/nbextensions/export_embedded/readme.html) for exporting.

antoine-levitt:
I exported and read a 28M notebook with a few animated gifs using chrome, so I don't think those limitations are there anymore?

The extension is interesting, although it would certainly be great if that worked out of the box.

daschw:
I'm in favor of your diff suggestion @antoine-levitt . Exporting is definitely nice to have. I don't use jupyter notebooks at all, though, so I'm probably not the right person to judge if the file size is acceptable.

daschw:
@mkborregaard @asinghvi17 @yha any opinions?

mkborregaard:
I'm in favour

antoine-levitt:
Thanks, changed the PR. Can anybody test it with videos? For some reason I have not been able to make them work at all.

yha:
> @mkborregaard @asinghvi17 @yha any opinions?

I also rarely use IJulia, but I would think having large notebooks when animations are embedded is an improvement over incorrect display or not being able to export.

> Thanks, changed the PR. Can anybody test it with videos? For some reason I have not been able to make them work at all.

I tested it `mp4` (`a=@animate .. end; mp4(a)`) and it worked. With `mov` I get a widget saying "Invalid source", but I also get that on master (the temporary `.mov` file is created and seems fine).

asinghvi17:
Doesn't `mov` need `MIME"video/quicktime"`?

asinghvi17:
How about a lookup table:
```julia
function ext2mime(ext::String)
    ext == "gif" && return "image/gif"
    ext == "mp4" && return "video/mp4"
    ext == "mov" && return "video/quicktime"
end
```
and then we just interpolate `ext2mime(ext)`?

antoine-levitt:
Thanks! Fixed it.

antoine-levitt:
Even with that it's not enough to make chrome like it. Linking to the file directly works so that's not the problem. Firefox has no problem with it though.

Funnily enough, https://secure.phabricator.com/T13135 suggests faking the type to be video/mp4 even though it's a mov, and changing that does make it work with both firefox and chrome. I'd say Plots shouldn't have to work around bugs in browsers, but it's web stuff, so who knows what twisted rules apply there...

daschw:
>  I'd say Plots shouldn't have to work around bugs in browsers

Agreed. Should we merge this, then?

antoine-levitt:
Fine with me!

daschw:
Thanks a lot for this improvement @antoine-levitt !

antoine-levitt:
Thanks for the review and the merge!

PR #2264, opened 2019-11-19
Fix pgfplots legend fg/bg transparency.

korsbo:
Fix `fg_legend=:transparent` and `bg_legend=:transparent` for the PGFPlots backend.

```julia
using Plots
pgfplots()
plot(x->x, fg_legend=:transparent, bg_legend=:transparent)
```
Before PR:
before

After PR:
after

### DIFF
--- a/src/backends/pgfplots.jl
+++ b/src/backends/pgfplots.jl
@@ -476,16 +476,10 @@ function _update_plot_object(plt::Plot{PGFPlotsBackend})
         if haskey(_pgfplots_legend_pos, legpos)
             kw[:legendPos] = _pgfplots_legend_pos[legpos]
         end
-        cstr, a = pgf_color(plot_color(sp[:background_color_legend]))
-        a == 0 && (cstr = "none") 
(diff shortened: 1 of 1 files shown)


korsbo:
Hi,
That looks better, I'll change/test it tomorrow!

daschw:
Great!

korsbo:
`"fill opacity = $bg_alpha"` also affected the font opacity, so I added `"text opacity = ..."`. Text opacity reads the alpha from the `:legendfontcolor` attribute. This works but it feels somewhat inelegant since the actual colour of `:legendfontcolor` is still ignored. 

Alternatives are:
- leave it be (it does work after all).
- set `"text opacity = 1` statically (regressive but consistent).
- fix the legend font colour too (I'm not sure how, nor how often this would be used).

daschw:
I think ideally we'd fix the legend font color too. However, don't know how to do this right now either and I think this PR is already a big improvement. Maybe the legendfontcolor can be fixed in a separate PR.

korsbo:
No problem @daschw, it is me who should thank you and the other main contributors of this package which I use daily!

PR #2273, opened 2019-11-26
Fix PlotlyJS integration with WebIO

### LINKED ISSUE #2272: PlotlyJS Backend Broken with WebIO
The Plotly backend is integrated with WebIO. The default display does not use WebIO as it defines it's own `display` method for `text/html` which Jupyter/IJulia uses. However, if the plot is embedded further inside a "WebIO thing" (e.g. an `@manipulate` macro block from Interact.jl), it does invoke the WebIO render pathway.

https://github.com/JuliaPlots/Plots.jl/blob/d2ca5e3bf31f4a8f87ace6cfd11d8032681dbd7b/src/backends/plotlyjs.jl#L44-L47

The issue is that, after this, `plt.o` is `nothing` and so nothing is rendered.

travigd:
Fixes #2272.

### DIFF
--- a/src/backends/plotlyjs.jl
+++ b/src/backends/plotlyjs.jl
@@ -42,8 +42,8 @@ _display(plt::Plot{PlotlyJSBackend}) = display(plotlyjs_syncplot(plt))
 
 @require WebIO = "0f1e0344-ec1d-5b48-a673-e5cf874b6c29" begin
     function WebIO.render(plt::Plot{PlotlyJSBackend})
-        prepare_output(plt)
-        WebIO.render(plt.o)
+        plt_html = sprint(show, MIME("text/html"), x)
+        return WebIO.render(dom"div"(innerHTML=plt_html))
     end
 end
 


src/backends/plotlyjs.jl
```diff
@@ -42,8 +42,8 @@ _display(plt::Plot{PlotlyJSBackend}) = display(plotlyjs_syncplot(plt))
 
 @require WebIO = "0f1e0344-ec1d-5b48-a673-e5cf874b6c29" begin
     function WebIO.render(plt::Plot{PlotlyJSBackend})
-        prepare_output(plt)
-        WebIO.render(plt.o)
+        plt_html = sprint(show, MIME("text/html"), x)
```
daschw: I have not tested this, but isn't `x` undefined here?

travigd: This is what I get for trying to write PR's using GitHub's web interface. You're correct.

Lilyhanig:
Is this ever going to get integrated?

daschw:
> Is this ever going to get integrated?

If it ever gets updated.

BoundaryValueProblems:
I am wondering whether this would help the latest plotlyjs backend problem I reported #2462 .

PR #2289, opened 2019-11-29
Update dependencies

daschw:


KristofferC:
These bounds make sense. The only thing I am a bit worried about is that StatsBase and GR seems to be pushing out "breaking" releases very quickly. Perhaps that is just an information problem though in that they simply don't know that minor is considered breaking pre 1.0.

daschw:
OK, thanks for checking. There's nothing we can do about the "breaking" releases issue with GR and StatsBase here in Plots though, right?

PR #2299, opened 2019-12-03
Use GR.drawimage for uniform heatmaps

yha:
`drawimage` is faster and supports transparency, so use it if possible. See jheinen/GR.jl#275.
Also moves conversion of colors to GR's format to a function, which supports general color spaces, so e.g., HSV images can be plotted.

PR #2300, opened 2019-12-03
gr: optimize 'heatmap' logic

jheinen:


yha:
I think this is superseded by #2299.
Is there a reason to use `GR.cellarray` rather than `GR.drawimage` (which also supports alpha values)?

daschw:
This PR, in the end, only defined a function `is_equally_spaced` that is not used anywhere and probably equivalent to `is_uniformly_spaced` in #2299?

jheinen:
Sorry, my mistake.

daschw:
Don't worry. Should we revert this, or just remove `is_equally_spaced` in a new PR?

PR #2304, opened 2019-12-05
Generalized array fixes

michakraus:
Updates plots in many places to use the generalized arrays interface for array indices, thus automatically supporting anything that implements the AbstractArray interface. This involves e.g.:
- replace many uses of size with axes
- replace 1:length(v) with eachindex(v),
- replace zeros(Int, length(v)) with similar(Array{Int}, axes(v))

(see https://docs.julialang.org/en/latest/devdocs/offset-arrays/#Generalizing-existing-code-1)

This PR mostly brings #999 up-to-date.

### DIFF
--- a/src/args.jl
+++ b/src/args.jl
@@ -1098,7 +1098,7 @@ function extractGroupArgs(v::AVec, args...; legendEntry = string)
     if n > 100
         @warn("You created n=$n groups... Is that intended?")
     end
-    groupIds = Vector{Int}[filter(i -> v[i] == glab, 1:length(v)) for glab in groupLabels]
+    groupIds = Vector{Int}[filter(i -> v[i] == glab, eachindex(v)) for glab in groupLabels]
     GroupBy(map(legendEntry, groupLabels), groupIds)
 end
 
@@ -1106,7 +1106,7 @@ legendEntryFromTuple(ns::Tuple) = join(ns, ' ')
 
 # this is when given a tuple of vectors of values to group by
 function extractGroupArgs(vs::Tuple, args...)
-    isempty(vs) && return GroupBy([""], [1:size(args[1],1)])
+    isempty(vs) && return GroupBy([""], [axes(args[1],1)])
     v = map(tuple, vs...)
     extractGroupArgs(v, args...; legendEntry = legendEntryFromTuple)
 end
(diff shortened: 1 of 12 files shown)


daschw:
Awesome! Did you check if the tests pass locally and if it works correctly with e.g. OffsetArrays?

michakraus:
Yes, I tested locally, which was fine, and I did some simple tests with OffsetArrays and my own custom array types, which also worked. There are two issues, though:
- In recipes.jl, lines 412-438, there is some code related to heatmap which I could not fully understand and so did not update. This still needs to be fixed.
- I only tested with the PlotlyJS backend. I just realised that there still seem to be issues with GR and PyPlot. I'll look into that.

daschw:
Plotly(JS) seems to convert `x`, `y`, and `z` to "regular" arrays in `plotly_data` using `collect`, which does not happen in GR and PyPlot. Hence, the latter fail when trying to index with `rng` from `iter_segements`, which does not correspond to the offset indices.

michakraus:
I see. Great! I fixed the SegmentsIterator so that GR and PyPlot work, but now Plotly doesn't work anymore... for that very reason. Any suggestions on how to handle this?

daschw:
Is the `collect` in `plotly_data(v)` (https://github.com/JuliaPlots/Plots.jl/blob/master/src/backends/plotly.jl#L417) required? Maybe you could add a new method `plotly_data(a::AbstractArray) = a` without `collect`?

daschw:
I'm a little worried that changing the SegmentsIterator would require a vector input of e.g. `linecolor` or `linestyle` to be an `OffsetVector` too for `y` being an `OffsetVector`. Maybe this the correct interpretation though ...

michakraus:
Adding `plotly_data(a::AbstractArray) = a` without `collect` seems to do the trick.

I changed the SegmentsIterator to have a variable starting and ending index instead of 1 and n=length(y). That should (and seems to) guarantee compatibility with standard arrays.

michakraus:
PGFPlots shows a similar error as Plotly did. It doesn't like the generalized ranges. It seems there must be another array conversion, maybe in https://github.com/JuliaPlots/Plots.jl/blob/master/src/backends/pgfplots.jl#L183 ?
PGFPlotsX seems to work (all tests of test/test_pgfplotsx.jl passing and sample plots looking fine). I can't run InspectDR as there is some library issue I can't easily resolve (on macOS).

daschw:
It works with InspectDR as well :tada:

daschw:
Thanks a lot @michakraus !

PR #2310, opened 2019-12-06
fix ribbons and interaction between background and fill between

BeastyBlacksmith:
Relates to bug reports in https://discourse.julialang.org/t/confidence-interval-with-plots-and-pgfplots/31440/25


BeastyBlacksmith:
Hmm.. I changed the tests to actually check, if latex runs without erroring, but the DockerImage does not have latex installed.

KristofferC:
Maybe https://github.com/tpapp/texlive-julia-minimal-docker could be used somehow.

daschw:
Simon tried this in #2313. Not sure, what did not work and why the PR has been closed.

BeastyBlacksmith:
I always got permission denials whose source are a mystery to me. Also the current plot test suite has a bunch of other dependencies. So either we would need to extend the minimal docker image or we only run the pgfplotsx tests on the docker image ( haven't tried this ).

BeastyBlacksmith:
I just commented the offending tests for now and declare this as ready to merge.
Tests failing on 1.0 are unrelated to the changes (its the artifacts timestamp issue).

PR #2320, opened 2019-12-11
Support colons in layout keyword argument.

korsbo:
Allow more flexible layout specification for plot stacking: `plot(plt1, plt2, plt3; layout=(:,2))`. A colon in the layout tuple will be recalculated to the smallest value which will allow all plots to fit.

As per a suggestion by Spencer Russel in slack.

This does make the code a bit harder to follow so I'd be happy to make adjustments if anyone can see a neater solution.

### DIFF
--- a/src/layouts.jl
+++ b/src/layouts.jl
@@ -487,6 +487,7 @@ function layout_args(sztup::NTuple{2, Integer})
 end
 
 layout_args(n_override::Integer, n::Integer) = layout_args(n)
+layout_args(n, sztup::NTuple{2, Integer}) = layout_args(sztup)
 
 function layout_args(n, sztup::Tuple{Colon, Integer})
     nc = sztup[2]
--- a/src/layouts.jl
+++ b/src/layouts.jl
@@ -469,7 +469,7 @@ function layout_args(plotattributes::KW)
 end
 
 function layout_args(plotattributes::KW, n_override::Integer)
-    layout, n = layout_args(get(plotattributes, :layout, n_override))
+    layout, n = layout_args(n_override, get(plotattributes, :layout, n_override))
     if n != n_override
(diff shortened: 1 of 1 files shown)


korsbo:
I have spotted a mistake, hold up on merging.

korsbo:
It now works for the following inputs:
```julia
julia> plt = [plot(rand(2,4)) for i in 1:10]
10-element Array{Plots.Plot{Plots.GRBackend},1}:
 Plot{Plots.GRBackend() n=4}
 Plot{Plots.GRBackend() n=4}
 Plot{Plots.GRBackend() n=4}
 Plot{Plots.GRBackend() n=4}
 Plot{Plots.GRBackend() n=4}
 Plot{Plots.GRBackend() n=4}
 Plot{Plots.GRBackend() n=4}
 Plot{Plots.GRBackend() n=4}
 Plot{Plots.GRBackend() n=4}
 Plot{Plots.GRBackend() n=4}

julia> plot(plt...)

julia> plot(plt..., layout=(2,5))

julia> plot(plt..., layout=(4,:))

julia> plot(plt..., layout=(:,4))

```
Is there anything else that should be tested?

daschw:
I expected `plot(rand(10, 4), layout = (:, 1))` to work analogously `plot(rand(10, 4), layout = (4, 1)`, but I see that that would be ambiguous.

korsbo:
> I expected `plot(rand(10, 4), layout = (:, 1))` to work analogously `plot(rand(10, 4), layout = (4, 1)`, but I see that that would be ambiguous.

I would have preferred if this new notation was not limited to stacked plots, `plot(plt1, plt2,...)`, but then one would have to know how many subplots there should be when the layout is decided. I'm not very familiar with Plots.jl, but I could not instantly see a good way of getting this information.

daschw:
I don't think it's obvious to tell from `plot(rand(10, 4), layout = (:, 1))` if the user wants 4 subplots with 1 series or 1 subplot with 4 series.

PR #2326, opened 2019-12-13
Miscellaneous changes to PGFPlotsX backend

asinghvi17:
- Unified preamble pushes and added comments (so that the latex document is more readable)

TODO:
- [ ] Figure out a way to print non-standalone tex
- [ ] Improvements on the PGFPlotsX.jl end for more elegant documents

### DIFF
--- a/src/backends/pgfplotsx.jl
+++ b/src/backends/pgfplotsx.jl
@@ -6,10 +6,15 @@ Base.@kwdef mutable struct PGFPlotsXPlot
     function PGFPlotsXPlot(is_created, was_shown, the_plot)
         pgfx_plot = new(is_created, was_shown, the_plot)
         # tikz libraries
-        PGFPlotsX.push_preamble!(pgfx_plot.the_plot, "\\usetikzlibrary{arrows.meta}")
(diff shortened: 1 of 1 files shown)


BeastyBlacksmith:
What do mean by printing? To stdout? Saving as `.tex`-file is non-standalone by default.

src/backends/pgfplotsx.jl
```diff
@@ -6,10 +6,15 @@ Base.@kwdef mutable struct PGFPlotsXPlot
...
-        PGFPlotsX.push_preamble!(pgfx_plot.the_plot, "\\usetikzlibrary{backgrounds}")
         PGFPlotsX.push_preamble!(pgfx_plot.the_plot,
         """
+        % Load TikZ libraries
+        \\usetikzlibrary{arrows.meta, backgrounds}
+            
+        % Load PGFPlots libraries
+        \\usepgfplotslibrary{patchplots, fillbetween}
+            
+        % Set some useful keys
```
BeastyBlacksmith: Those keys are for defining the outer background color

PR #2347, opened 2019-12-28
Fix tests for PGXPlotsX backend

o01eg:
The backend uses `push!` and `append!` methods introduced in `PGFPlotsX` 1.2.0 version.

`Requires` incompatibilities will be fixed in https://github.com/KristofferC/PGFPlotsX.jl/pull/201


BeastyBlacksmith:
This does declare compability with `2.0` right? So better just use `"1.2.0"`.

o01eg:
@BeastyBlacksmith Fixed compatibility declaration.

BeastyBlacksmith:
Thanks :)

PR #2354, opened 2020-01-03
Fill between functions

BeastyBlacksmith:
```julia
julia> p = plot(1:5)

julia> plot!(p, 1:2:6, fill = 1)
```
So `fill = <series_index>` will fill between the current series and the series specified by index.

Happy to get some feedback, if it works for more complicated usecases also.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,7 +1,7 @@
 name = "Plots"
 uuid = "91a5bcdd-55d7-5caf-9e0b-520d859cae80"
 author = ["Tom Breloff (@tbreloff)"]
-version = "0.28.4"
+version = "0.28.5"
 
 [deps]
 Base64 = "2a0f44e3-6c83-55bd-87e4-b1978d98bd5f"
--- a/src/args.jl
+++ b/src/args.jl
@@ -786,7 +786,6 @@ function processFillArg(plotattributes::KW, arg)
         plotattributes[:fillrange] = arg
 
     elseif !handleColors!(plotattributes, arg, :fillcolor)
-
         plotattributes[:fillrange] = arg
     end
     # plotattributes[:fillrange] = fr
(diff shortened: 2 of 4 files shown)


mkborregaard:
This is breaking, as `fill = 1` will normally fill from the value `1.`- something that is used for quite a few recipes in the ecosystem.

BeastyBlacksmith:
That is good to know! Would passing a tuple of current series index and another series index be fine?

mkborregaard:
Yeah I don't think that would clash with anything

BeastyBlacksmith:
So I tried the tuple implementation and it kinda works, e.g.
```julia
julia> plot( 1:5, fill = 0, fillcolor = :match)
julia> plot!( 2:6, fill = ((2, 1), :red))
```
gives what is expected.
Ideally 
```julia
julia> plot!( 2:6, fill = (2, 1), fillcolor = :match)
```
should also work, but since `fill` is a magic keyword that is ambiguous AFAIU.
This could be alleviated if we'd export a wrapper type instead of using a tuple.
Also I don't see why the default `fillcolor` is transparent now.

daschw:
The issue with wrapper types is that they cannot be accessed in a Recipe that only depends on RecipesBase, but not on Plots.
Does 
```julia
plot!( 2:6, fillrange = (2, 1), fillcolor = :match)
```
work? Does pgfplotsx also support the standard fillrange as vector or number?

BeastyBlacksmith:
> Does
> ```julia
> plot!( 2:6, fillrange = (2, 1), fillcolor = :match)
> ```
> work?

No, you'd have to write
```julia
plot!( 2:6, fillrange = ((2, 1),), fillcolor = :match)
```

> Does pgfplotsx also support the standard fillrange as vector or number?

Good catch, it did and now it does again.

BeastyBlacksmith:
This looks fine to me now.

BeastyBlacksmith:
Actually this could also be implemented at the Plots level for all backends that support the Vector input to fillrange.
Somewhere along
```julia
plot( f1x, f1y, fillrange = [ f2(x) for x in f1x ] )
plot( f2x, f2y )
```
where `f2(x)` is either given by the user or an interpolation.

PR #2356, opened 2020-01-10
rework layouts

BeastyBlacksmith:
This drops the groupplots, since they block insets and didn't work with polar plots beforehand.
At the moment insets are working in exchange for gridlayouts.
Thus an alternative way for placing the axis has to be found and I have the feeling that the most flexible will be to use nodes to place the axes.


BeastyBlacksmith:
this is another episode of "you can't do better than the pgfplots backend". Please squash at merge.

Issue #1180, opened 2017-10-15
Let colorbar fonts be controlled by plot fonts


mkborregaard:
It would be nice if the colorbar font followed that of the plot in gr and plotlyjs - this functionality exists for pyplot().
Tagging @AnkurD who implemented this for pyplot in https://github.com/JuliaPlots/Plots.jl/pull/1108

Issue #463, opened 2016-08-29
Control colorbar size (and other attributes?)


ahwillia:
As far as I can tell, there isn't a good way to set the size/scale of the colorbar (see, e.g., https://github.com/stevengj/PyPlot.jl/issues/93)

The following would be nice:

```julia
X = randn(100,100)
heatmap(X; colorbar=:left)       # works with some backends already (not pyplot?)
heatmap(X; colorbar=0.5)        # half the default size
heatmap(X; colorbar=[-2,0,2])  # tick marks?
```

I did a bit of digging to see if I could figure out how to implement the scaling. In PyPlot, you can do the following:

<img width="443" alt="screen shot 2016-08-28 at 6 34 23 pm" src="https://cloud.githubusercontent.com/assets/636625/18038661/a25aa66c-6d4f-11e6-9aa3-aa4a0592b69b.png">

<img width="494" alt="screen shot 2016-08-28 at 6 34 07 pm" src="https://cloud.githubusercontent.com/assets/636625/18038662/a3afa7e2-6d4f-11e6-8048-503b3bfd2232.png">

So my original thought was to use the `fraction` keyword. But, it looks like `PyPlot.colorbar(...)` isn't called directly by Plots? It looks to me like a new axis with a hard-coded size is created here (am I parsing that correctly?):

https://github.com/tbreloff/Plots.jl/blob/master/src/backends/pyplot.jl#L816

Thanks again for your patient guidance through all of these details Tom.

mauro3:
It would also be great to have to usual font-selection keywords for colorbar ticklabels and colorbar_title.

mkborregaard:
See also #302

PR #2363, opened 2020-01-17
unite titlefonts

BeastyBlacksmith:
I am willing to take a stab at https://github.com/JuliaPlots/Plots.jl/issues/302. The question is: do we want to make this a breaking change?
E.g. remove `titlefontcolor`, `titlefontsize`, etc. and let those all be `PlotText` objects.
Implementing `:match` could be tricky.

Or non-breaking: Accept those arguments but remove the fields from the `Plot` object and construct the `PlotText` objects internally.
The issue here is that certain titles like `plot_title` or https://github.com/JuliaPlots/Plots.jl/issues/463 don't have the `*fontsize` keywords so they would need to get added for consistency.

Related: https://github.com/JuliaPlots/Plots.jl/issues/1180


daschw:
The first option would probably be nicer for internal development, BUT: If you only allow `PlotText` objects, it won't be possible to specify the title from within a recipe without depending on Plots anymore. So I'd rather go for the non-breaking change.

BeastyBlacksmith:
Good point

src/args.jl
```diff
@@ -288,13 +288,29 @@ const _series_defaults = KW(
...
-    :window_title                 => "Plots.jl",
+    :window_title                => "Plots.jl",
+    :window_titlefontsize        => 14,
+    :window_title_location       => :center,           # also :left or :right
+    :window_fontfamily           => :match,
+    :window_titlefontfamily      => :match,
+    :window_titlefonthalign      => :hcenter,
+    :window_titlefontvalign      => :vcenter,
+    :window_titlefontrotation    => 0.0,
+    :window_titlefontcolor       => :match,
```
daschw: Does the window title really have all these attributes? Isn't it just the text appearing in the gui window's title bar? In that case the font, etc. would depend on the OS, right? Or am I missing another use case for `window_title`?

BeastyBlacksmith: You are  right. Thanks for pointing out. I didn't really think here. I will delete these and improve the description of `window_title` instead.

BeastyBlacksmith:
I would say we don't need to actually create the `PlotText` objects, since we have the `*title` fields and the `*font` accessor-functions.
Only question that I have is how do we allow for string and `PlotText` input to the `*title` fields?

BeastyBlacksmith:
> The first option would probably be nicer for internal development, BUT: If you only allow `PlotText` objects, it won't be possible to specify the title from within a recipe without depending on Plots anymore. So I'd rather go for the non-breaking change.

This wouldn't be an issue, if `PlotText` would live in `RecipesBase`, right? Is there any downside to this? @daschw

daschw:
We want to keep RecipesBase as minimal as possible to reduce the barrier for other packages to take the depency. The same argument as for `PlotText` could be made for `Stroke`, `Brush`, `Font` (already required by `PlotText`) `Layout` and probably more. Also the idea of the recipe system is that other plotting libraries could hook into it as well. While currently only Plots uses it, the aim is that Makie can also consume RecipesBase recipes (I think).
So I'm not sure, if in practice there would be a big downside. But until now our strategy has been not to move too much from Plots to RecipesBase. What do you think @mkborregaard ?

mkborregaard:
Agreed, our policy has been not to move concrete types. 
BUT, we have a plottext constructor that takes tuples, and process them like a magic argument. I've used that e.g. here: https://github.com/richardreeve/Phylo.jl/blob/master/src/plot.jl#L103-L104
Could that approach not resolve it?

BeastyBlacksmith:
So the plan would be to have all text keywords like titles to use the `plottext` constructor to construct `PlotText` objects and then we can get rid of all the font related keywords, if I got that right.

BeastyBlacksmith:
added the magic arguments, so now you can construct them like the other fonts, too.
I think we can spare the `PlotText` refacturing for 2.0 and keep it as is?
Still has to be implemented from the backend side and maybe add aliases.
I can add an implementation for the pgfplotsx backend either on this PR or in a seperate PR.

BeastyBlacksmith:
I think this is fine now. @daschw, @mkborregaard  any objections?

PR #2371, opened 2020-01-27
Allow `label = nothing, :none, false`

### LINKED ISSUE #2314: label should accept false and nothing to disable
Currently to disable the label for a plot element we accept `label=""`

Would be nice if it also accepted `label=false` and possibly `label=nothing` with the same behaviour.

### LINKED ISSUE #2368: No label options should be expanded to include :none, false, and nothing
Currently, the supported way to have no labels is `label = ""`

The following options all get stringified:
`label = :none`
`label = nothing`
`label = false`

I would expect all of these to behave equivalently to `label=""`

daschw:
Close #2368
Close #2314


BeastyBlacksmith:
Ref: https://github.com/JuliaPlots/Plots.jl/issues/2314

PR #2372, opened 2020-01-28
New recipe: lens

### LINKED ISSUE #541: subplot views
copied from gitter:

> I was actually thinking through how you would automate a floating inset subplot that shows a zoomed in area. Not hard to do manually with Plots, but might be tricky to generalize
> 
> i had a quick thought about the floating subplot thing... it might be nice to implement a "subplot view"... basically a wrapper around another subplot, which inherits all the data and attributes from the wrapped subplot, but with optional overrides
> then you could create an inset floating subplot that is a zoomed-in view with one easy line of code
> i think that's pretty straightforward to implement... the only thing is that I might need to add an AbstractSubplot in a few places

BeastyBlacksmith:
The lens recipe should magnify a certain region and show it in an inset plot.
Current API is
```julia
pl = plot((1:5).^2)
plot!(pl, [1.2, 2.8], [3.4, 7.8], seriestype=:lens, inset=(1,bbox(0.4,0.,0.5,0.5)))
```
Where you pass the magnifying region as `x` and `y` values and the location of the inset as usually. 
Currently only relative bounding boxes are supported. Location of the linking line has to be improved.
Also drawing of the linking line should actually be outside of any plot, so it doesn't infer with axis limits of the parent plot, but I don't know if this is possible.
fixes #541

### DIFF
--- a/src/examples.jl
+++ b/src/examples.jl
@@ -311,6 +311,21 @@ const _examples = PlotExample[
         ],
     ),
     PlotExample(
+        "Lens",
+        "A lens lets you easyli magnify a region of a plot. x and y coordinates refer to the to be magnified region and the via the `inset` keyword the subplot index and the bounding box (in relative coordinates) of the inset plot with the magnified plot can be specified. Additional attributes count for the inset plot.",
+        [
+            quote
+                begin
+                    plot([(0, 0), (0, 0.9), (1, 0.9), (2, 1), (3, 0.9), (80, 0)])
+                    plot!([(0, 0), (0, 0.9), (2, 0.9), (3, 1), (4, 0.9), (80, 0)])
+                    plot!([(0, 0), (0, 0.9), (3, 0.9), (4, 1), (5, 0.9), (80, 0)])
+                    plot!([(0, 0), (0, 0.9), (4, 0.9), (5, 1), (6, 0.9), (80, 0)])
+                    lens!([1, 6], [0.9, 1.1], inset = (1, bbox(0.5, 0.0, 0.4, 0.4)))
+                end
+            end
+        ],
+    ),
+        * PlotExample(
         "Adding to subplots",
         """
         Note here the automatic grid layout, as well as the order in which new series are added
--- a/src/examples.jl
+++ b/src/examples.jl
@@ -315,7 +315,7 @@ const _examples = PlotExample[
         "A lens lets you easily magnify a region of a plot. x and y coordinates refer to the to be magnified region and the via the `inset` keyword the subplot index and the bounding box (in relative coordinates) of the inset plot with the magnified plot can be specified. Additional attributes count for the inset plot.",
         [quote
                 begin
-                    plot([(0, 0), (0, 0.9), (1, 0.9), (2, 1), (3, 0.9), (80, 0)])
+                    plot([(0, 0), (0, 0.9), (1, 0.9), (2, 1), (3, 0.9), (80, 0)], legend = :outertopright)
                     plot!([(0, 0), (0, 0.9), (2, 0.9), (3, 1), (4, 0.9), (80, 0)])
                     plot!([(0, 0), (0, 0.9), (3, 0.9), (4, 1), (5, 0.9), (80, 0)])
                     plot!([(0, 0), (0, 0.9), (4, 0.9), (5, 1), (6, 0.9), (80, 0)])
--- a/src/examples.jl
+++ b/src/examples.jl
@@ -311,23 +311,6 @@ const _examples = PlotExample[
         ],
     ),
     PlotExample(
-        "Lens",
-        "A lens lets you easily magnify a region of a plot. x and y coordinates refer to the to be magnified region and the via the `inset` keyword the subplot index and the bounding box (in relative coordinates) of the inset plot with the magnified plot can be specified. Additional attributes count for the inset plot.",
-        [quote
-                begin
-                    plot([(0, 0), (0, 0.9), (1, 0.9), (2, 1), (3, 0.9), (80, 0)], legend = :outertopright)
-                    plot!([(0, 0), (0, 0.9), (2, 0.9), (3, 1), (4, 0.9), (80, 0)])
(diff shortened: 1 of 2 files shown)


src/recipes.jl
```diff
@@ -801,6 +801,50 @@ end
 # note: don't add dependencies because this really isn't a drop-in replacement
 
 
+# ---------------------------------------------------------------------------
+# lens! - magnify a region of a plot
+@recipe function f(::Type{Val{:lens}}, plt::AbstractPlot)
```
mkborregaard: I'm surprised this works - it's the type recipe signature, which is usually just used for type conversions to vector. Should this not be a "user recipe" with a `@userplot` shorthand defined?

BeastyBlacksmith: I thought, this is the signature of a [plot recipe](https://docs.juliaplots.org/latest/recipes/#plot-recipes)

BeastyBlacksmith: also I would like to only define the `lens!` shorthand, since `lens` by itself doesn't make too much sense.

mkborregaard: Ah you're right. I don't think they are used anywhere in the ecosystem, but I may be wrong (it lists MarginalHist as an example, but looking down on that page clearly shows that marginalhist is a "user recipe"). Defining a `lens!` shorthand makes sense - but I feel that calling it with `seriestype` as done here feels unideomatic.

BeastyBlacksmith: What would be a more ideomatic alternative? I don't bother too much to change this.

mkborregaard: Just the lens shorthand? Sorry I have zero experience using plot recipes

BeastyBlacksmith: Okay, now I got it. You meant the way to call the recipe is unideomatic, I thought the implementation with using the `seriestype`. I will add the `lens!` shorthand.

BeastyBlacksmith:
I thought, adding a ghost subplot for the connecting line would help, but, since the bounding box of the parent box seems to get calculated afterwards it still affects his axis limits. Also some backends have problems if the inset plot is outside of the parent plot.

BeastyBlacksmith:
Easiest course of action would be to get rid of the connecting line altogether and just have the box + inset.

BeastyBlacksmith:
This is the best I can do without having access to the bounding box of the parent plot.
Please give it a try.

mkborregaard:
@BeastyBlacksmith are you waiting for me here?
BTW would this close https://github.com/JuliaPlots/Plots.jl/issues/541 ?

BeastyBlacksmith:
@mkborregaard Would be nice if you can confirm that I can't get the bounding box of the current plot at the recipe stage or if I am looking for the wrong thing and layout information is stored elsewhere.
Plus I am looking for a way to have the keywords passed be applied for only for the inset. I can make only parent or both but failed to do only inset.
Other than that #541 would be closed, yes.

mkborregaard:
Whether the bounding box is present depends on the recipe type. With this being a type recipe, I'm pretty sure it isn't  - if it were a series recipe it would be present as a key in a captured `plotattributes` object (as they are processed later in the pipeline where the bounding box is known), but then series recipes won't let you make another window.
With regards to the keywords - try `pop!`ing them from the `plotattributes` Dict and store them locally, them passing them in the inset.

BeastyBlacksmith:
Ok, that worked. Could use some testing before merge.
An example is
```julia
pl = plot((1:5).^2)
lens!(pl, [1.2, 2.8], [3.4, 7.8], inset=(1,bbox(0.5,0.2,0.5,0.5)))
```

mkborregaard:
You could add a testing image to `examples` maybe? That way it would become part of the testing suite - and be documented...

BeastyBlacksmith:
I added the example and made sure, I can easily revert the formatting changes, if they don't belong here.

src/examples.jl
```diff
@@ -311,6 +311,21 @@ const _examples = PlotExample[
         ],
     ),
     PlotExample(
+        "Lens",
+        "A lens lets you easyli magnify a region of a plot. x and y coordinates refer to the to be magnified region and the via the `inset` keyword the subplot index and the bounding box (in relative coordinates) of the inset plot with the magnified plot can be specified. Additional attributes count for the inset plot.",
```
mkborregaard: typo

src/examples.jl
```diff
@@ -2,503 +2,872 @@
...
+                    plot([(0, 0), (0, 0.9), (1, 0.9), (2, 1), (3, 0.9), (80, 0)])
+                    plot!([(0, 0), (0, 0.9), (2, 0.9), (3, 1), (4, 0.9), (80, 0)])
+                    plot!([(0, 0), (0, 0.9), (3, 0.9), (4, 1), (5, 0.9), (80, 0)])
+                    plot!([(0, 0), (0, 0.9), (4, 0.9), (5, 1), (6, 0.9), (80, 0)])
+                    lens!([1, 6], [0.9, 1.1], inset = (1, bbox(0.5, 0.0, 0.4, 0.4)))
+                end
+            end
+        ],
+    ),
+        * PlotExample(
```
mkborregaard: stray `*`

mkborregaard:
Nice - would you mind splitting the formatting changes to a separate PR?
Are you familiar with how to add a figure to PlotReferenceImages to make tests pass? (examples double as tests)

BeastyBlacksmith:
> Are you familiar with how to add a figure to PlotReferenceImages to make tests pass? (examples double as tests)
No, never did this.

daschw:
Firstly, I'd suggest putting new test_examples at the end of `Plots._examples`, because the test examples in PlotReferenceImages are named after their index in that Array. This way the proper images are compared to each other.

To update the PlotReferenceImages, you basically have to
1. `]dev PlotReferenceImages`, fork it, create a new branch, ...
2. Run `]test Plots`
3. If images differ a GUI opens show ing the old and new image and asking you if you want to save the changes.
4. After tests finished, add new images in your PlotReferenceImages branch, push to fork and PR

After the PlotReferenceImages PR is merged, the tests of this PR will pass

PR #2426, opened 2020-02-29
Fix shape markers for GR

daschw:
This allows to plot empty shapes as markers in GR.
```julia
using Plots
circle = Shape(Plots.partialcircle(0, 2π))
plot(1:10, marker = (circle, :transparent, 10))
```
marker

Furthermore this fixes reference image 21.
ref21


mkborregaard:
So awesome :heart:

PR #2427, opened 2020-03-01
use equal aspect ratio by default for images

### LINKED ISSUE #1898: use equal aspect ratio by default for images?
Normally the default aspect ratio behavior is fine, but specifically for `plot(image)` the image gets distorted and doesn't look good, and an equal ratio would make more sense. If there's other stuff in the plot, or an image is added to an existing plot, then it doesn't matter so much. But ideally I think we'd set `aspect_ratio=1` if the plot is created from just an image. Thoughts?

This was mentioned briefly here: https://github.com/JuliaPlots/Plots.jl/issues/1238#issuecomment-348755330

daschw:
close #1898 

Changes
```julia
using Plots, Images, TestImages
img = testimage("lighthouse")
p = plot(img[:, 250:450])
```
from
aspect_old

to
aspect_new


daschw:
Tested for GR, Plotly(JS) and PyPlot

mkborregaard:
nice :heart:

PR #2444, opened 2020-03-08
Allow image comparison tests on Windows.

yha:
This should allow the tests to run on Windows while still skipping them on AppVeyor.
I decreased the tolerance on Windows (because reverting commit ad4d7214e4116aa5cf0729fbbb8b07bb891e1510 outside Linux) otherwise a lot of stuff fails due to different fonts. There is still one such failure.


daschw:
Nice

PR #2458, opened 2020-03-14
implement the recipe_pipeline

BeastyBlacksmith:
This is how the new plot call chain would then look like


mkborregaard:
Travis says there's some dependency stuff

BeastyBlacksmith:
Might be the same issue we have on master?

mkborregaard:
probably

BeastyBlacksmith:
Kristoffer Carlsson (Slack)
Maybe a new version got released that is compatible with your compat but incompatible with one of the test dependencies?

mkborregaard:
I think it's VisualRegressionTests

daschw:
Looks good to me.

daschw:
How do we deal with `_typeAliases` here in this approach?

SebastianM-C:
I also tried to implement the ideas started here at https://github.com/JuliaPlots/Plots.jl/pull/2460
@daschw Regarding the `_typeAliases` I used the approach proposed by @mkborregaard with a kwarg that gets passed from the consumer package.

daschw:
I'll merge this into Michael's branch so we have a common foundation to start from tomorrow.

PR #2471, opened 2020-03-16
fix StaticArray plotting

daschw:
@BeastyBlacksmith That would also affect https://github.com/JuliaPlots/RecipePipeline.jl/pull/21

In general how do we deal with parallel development here? Will finally merging @mkborregaard recipeutils branch be a huge rebasing conflict mess?

### DIFF
--- a/src/series.jl
+++ b/src/series.jl
@@ -15,6 +15,7 @@ prepareSeriesData(x) = error("Cannot convert $(typeof(x)) to series data for plo
 prepareSeriesData(::Nothing) = nothing
 prepareSeriesData(t::Tuple{T, T}) where {T<:Number} = t
 prepareSeriesData(f::Function) = f
+prepareSeriesData(ar::AbstractRange{<:Number}) = ar
 function prepareSeriesData(a::AbstractArray{<:MaybeNumber})
     f = isimmutable(a) ? replace : replace!
     a = f(x -> ismissing(x) || isinf(x) ? NaN : x, map(float, a))
--- a/src/series.jl
+++ b/src/series.jl
@@ -15,9 +15,10 @@ prepareSeriesData(x) = error("Cannot convert $(typeof(x)) to series data for plo
 prepareSeriesData(::Nothing) = nothing
 prepareSeriesData(t::Tuple{T, T}) where {T<:Number} = t
 prepareSeriesData(f::Function) = f
-prepareSeriesData(a::AbstractArray{<:MaybeNumber}) = replace(
-                                    x -> ismissing(x) || isinf(x) ? NaN : x,
-                                    map(float,a))
+function prepareSeriesData(a::AbstractArray{<:MaybeNumber})
+    f = isimmutable(a) ? replace : replace!
+    a = f(x -> ismissing(x) || isinf(x) ? NaN : x, map(float, a))
+end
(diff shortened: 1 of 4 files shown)


src/series.jl
```diff
@@ -15,7 +15,7 @@ prepareSeriesData(x) = error("Cannot convert $(typeof(x)) to series data for plo
 prepareSeriesData(::Nothing) = nothing
 prepareSeriesData(t::Tuple{T, T}) where {T<:Number} = t
 prepareSeriesData(f::Function) = f
-prepareSeriesData(a::AbstractArray{<:MaybeNumber}) = replace!(
+prepareSeriesData(a::AbstractArray{<:MaybeNumber}) = replace(
```
mkborregaard: AFAICS this creates yet another copy of the input data, right? I think this was added because we knew that a copy is already done in `compute_xyz` so it's safe to keep modifying the array?

daschw: OK, thanks for the info! I'll see if I can fix this without adding another copy.

BeastyBlacksmith:
We should label PRs that affect the RecipePipeline refactor with a backport label, so when the recipeutils branch gets merged, we can quickly make sure that all changes got ported.

src/series.jl
BeastyBlacksmith: best thing would be to have a test for that

BeastyBlacksmith: the problem might be that copying a StaticVector returns a StaticVector.

mkborregaard: Yes I think so too

BeastyBlacksmith:
an alternative would be, to change `copy(x)` to `convert(Array, x)` in `compute_x` and  friends, depending on what expresses our intention better.

daschw:
I'll try.

mkborregaard:
The copy is there on purpose in Plots - I tried to deal with it here: https://github.com/JuliaPlots/Plots.jl/issues/1175
Plots does a lot of copying actually.

daschw:
also `convert(Array, x)` fails for surface.

daschw:
I just ran the tests without copy. The only issue is `push!`ing to plots in animations in example 2, as Michael found out in #1175. Maybe we should consider to disallow pushing series data to plots? We are redrawing the entire plot in each animation step anyway. Then we could :crossed_fingers: get rid of `copy` and move `compute_xyz` to RecipePipeline. ... just thinking out loud, I have not thought this through.

mkborregaard:
That's the only place the tests fails, but I think it could be a coverage thing. Essentially every time you add a new series in Plots you copy the plotattributes, and then that new series can modify it's local copy. If they are not copied, new series will change other series too. I think this is just the only example in the tests where we modify a series. But that's just my vague recollection of where I got stuck.

daschw:
Correction of my comment above: Animation example 31 fails too due to the same reason. I was thinking of digging into Plots push! code to make it work without copied input, i.e. checking for lengths of the series data and only pushing conditionally. Not sure if this can work and when I get to this though.

daschw:
Do you think we can merge this, so we have StaticArrays support and move the `copy` discussion in a new Issue/PR?

BeastyBlacksmith:
Looks good to me.

daschw:
Any objections @mkborregaard ?

mkborregaard:
none :-)

Issue #2482, opened 2020-03-17
Zero-spaced subplot stack with backend pgfplotsx().


theomb21:
Might it be possible to realize [multi-panel plots](https://github.com/JuliaPlots/Plots.jl/issues/376) within the `pgfplots()` backend?
It is [possible](https://discourse.julialang.org/t/subplots-without-gap-between-them/27491) with Package `PGFPlotsX`.

If it was possible, one could e.g. catch for the options of the `bottom_margin` (=0, maybe along with `link=:x`) and then add the group option? I'm afraid I wasn't able to figure it out how multiple subplots are assembled in the [backends source file](https://github.com/JuliaPlots/Plots.jl/blob/master/src/backends/pgfplotsx.jl). (happens [here](https://github.com/JuliaPlots/Plots.jl/blob/0d210843e7d853951a0abb57517af091bd7ac26e/src/backends/pgfplotsx.jl#L97) ?)

Last, thanks to all contributors and maintainers of the pgfplotsx() backend for the nice work. The PGFPlotsX package brings all the flexibility to custom design publication-ready graphics within Julia and incorporation into the versatile `Plots` saves code and gives access for the daily use by providing decent default-results.

mkborregaard:
"all contributors and maintainers of the pgfplotsx() backend" is @BeastyBlacksmith :-D :heart:

BeastyBlacksmith:
This is another example of "layouting is currently a mess". I take the opportunity and explain in a little more detail.

The layouting code is mostly [this section](https://github.com/JuliaPlots/Plots.jl/blob/0d210843e7d853951a0abb57517af091bd7ac26e/src/backends/pgfplotsx.jl#L145-153).
The problem is
- it takes width and height of the subplot bounding box, but should take the bounding box of the subplot plotarea
- it doesn't respect setting of margins (at the moment it is unclear to me whether this needs to be accounted for in the backend or if the frontent code already does this)
- it also does not check the `:link` keyword which would be responsible for what you ask for

Regarding the groupplots library showed in the discourse thread: I removed it on purpose, because it was incompatible with polar plots.

To answer your question: Yes, that is definitely doable, but I can't promise you any timeframe. If you wan't to take action you can always talk to me either here or on slack :).
I'm happy the backend is otherwise useful for you.

BeastyBlacksmith:
I made a start here (https://github.com/JuliaPlots/Plots.jl/pull/2486)

[...]

PR #2486, opened 2020-03-18
Layouting

BeastyBlacksmith:
Should fix layouting. Currently looks not too bad, but e.g.
```julia
plot(randn(100, 5), layout=@layout([a{0.1h}; b [c; d e]]), t=[:line :histogram :scatter :steppre :bar], leg=false, ticks=nothing, border=:none)
```
errors.

Fix: https://github.com/JuliaPlots/Plots.jl/issues/2482

### DIFF
--- a/src/backends/pgfplotsx.jl
+++ b/src/backends/pgfplotsx.jl
@@ -107,11 +107,11 @@ function (pgfx_plot::PGFPlotsXPlot)(plt::Plot{PGFPlotsXBackend})
             bb = bbox(sp)
             sp_width = width(bb)
             sp_height = height(bb)
-            dx, dy = bb.x0
+            dx, dy = bbox(sp).x0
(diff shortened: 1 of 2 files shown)


BeastyBlacksmith:
This looks actually better than the current setting

BeastyBlacksmith:
The above example still errors
```julia
plot(randn(100, 5), layout=@layout([a{0.2h}; b [c; d e]]), t=[:line :histogram :scatter :steppre :bar], leg=false, ticks=nothing, border=:none)
```
does not, but shows, that plots get really tiny.

BeastyBlacksmith:
I feel like this is how it is supposed to be, but the example still errors, since pgfplots is saying, that the plots are too small to have meaningful axis descriptions (which we don't have...). So all in all I am in favor of merging this.

PR #2492, opened 2020-03-19
Fix broken hdf5 "backend".

ma-laforge:
Add support for new Attr structure.


mkborregaard:
Thanks for being on top of this!
Could you maybe add a rough test for this backend to the tests so we don't break it unknowingly again?

ma-laforge:
That's a great idea. Thanks. I created a simple testset just now.

PR #2495, opened 2020-03-20
Add simple testset for HDF5 plots.

ma-laforge:


ma-laforge:
I will add more if I get a better idea as to what I should test.  But at least for now, if something breaks, an exception will be thrown (and we will notice data not making it through to the read operation).

daschw:
Thanks!

mkborregaard:
haha sniped

PR #2497, opened 2020-03-21
Allow for Plotly JSON saving

keorn:
In particular this is useful when using [Franklin](https://franklinjl.org/) to display PlotlyJS plots.


daschw:
Nice, thanks!

PR #2499, opened 2020-03-22
test on 1.4 and update version

daschw:


daschw:
Thanks! Should we stop supporting Julia 1.0?

BeastyBlacksmith:
I think that would be a bad move, since it is still LTS-version, I'll open an issue on Pkg. Let's see.

KristofferC:
This should be fixed in Pkg but if you just want to get going you can just do the `resolve` conditional on `VERSION`?

BeastyBlacksmith:
all green now

daschw:
Nice, still let's wait a little bit with this release. I want to think about how to coordinate this with RecipesBase.
We have here fixes for vector type recipes, that only work with the next Plots release. If RecipesBase users want to rely on these types of recipes they have to tell Pkg somehow that the next release of Plots is required (in general without depending on Plots).
I was thinking about releasing RecipesBase 1.0 and have the next Plots release require that version for this purpose. However, I'm not sure yet if that is the right approach, considering we just had a "breaking" RecipesBase release 0.8 and not all dependent Packages have updated to that yet. Kristoffer pointed out on Slack that frequent breaking releases can make a seemingly lightweight dependency quite annoying for other package developers.
However, if we decide that we have now a stable version of RecipesBase and announce maybe on Discourse, that all dependent packages can safely upgrade compat to 1.0, that could be fine, too (?).

We could of course also just release RecipesBase 0.8.1 and require that version here in Plots. Not sure ... It would be nice to have a RecipesBase 1.0. After all, the API is quite stable now.
Maybe I'm just over-complicating things ... Any input welcome!

mkborregaard:
My suggestion would be to do a noisy 1.0 release of RecipesBase and Plots - one of us (I wouldn't mind that to be me) could open issues on all the repos depending on RecipesBase stating that this version should NOT break any existing code and can be merged and released immediately, and that the publication of 1.0 heralds a new age of essentially no more semver-breaking releases (I think we can commit to that, at least for a long time). After 1.0 is out this should quiet down, and doing it in the aftermath of the 0.8 debacle might be an advantage rather than a disadvantage (as it signals we're solving the issue)

daschw:
This would also be my preferred option.
I'm still working on a local branch to fix aliases in recipes and I would like to also include that if I manage to finish it soon. I've come quite far, but now I'm struggling a little with inset subplots.

mkborregaard:
Sounds sweet - though I wonder, would allowing aliases in recipes make it significantly more difficult for other packages, such as Makie, to use recipes? The no-aliases does form a nice stringent syntax.

daschw:
I also had that concern. AFAIK, currently we support some aliases, like e.g. `st`, but others not, like `xlabel`. Maybe we should first have a discussion, whether to disallow or allow all? I'm not sure to what extent the currently working aliases are used in the ecosystem.

mkborregaard:
They are supposedly not allowed - I'm surprised about `st`. See bullet point #4 under this heading https://docs.juliaplots.org/latest/recipes/#Recipe-Syntax/Rules-1

daschw:
I think it was introduced in https://github.com/JuliaPlots/Plots.jl/commit/011f51fbd91a7dac440569426b24766f39aa813f

mkborregaard:
I see, I'm clearly not updated :joy:

daschw:
Neither are the docs ... Either we disallow it and include descriptive error messages like "Attribute not supported. Maybe you are trying to use an alias within a recipe ..." or we allow it, but strongly recommend in the docs to not use them to ensure future compatibility with other plotting pkgs like Makie.

daschw:
I'm not sure what I should keep working on right now.
- Allowing all aliases, would be less breaking, but I'm not even sure if I will manage to do this.
- Since you brought it up, I think disallowing all aliases in recipes would be cleaner solution. Not sure how much work this is in Plots now, but it's probably easier to maintain from there. Who knows what kind of weird corner cases for combination of user-input aliases and recipe aliases might pop up in the future.

In both cases we could add dedicated docs to RecipesBase for 1.0 (maybe roughly based on my vizcon talk) and highlight there the recommended/obligatory use of default attribute names.

mkborregaard:
I personally prefer the second option - but maybe that's just me being conservative :-)

PR #2504, opened 2020-03-25
Add LinearAlgebra to make example with `norm` work

nilshg:
As per Tamas' suggestion here: https://discourse.julialang.org/t/outdated-example-for-plots-surface-command/35219/6


daschw:
Thanks!

PR #2509, opened 2020-03-27
Revert "CompatHelper: bump compat for "StatsBase" to "0.33""

daschw:
Reverts JuliaPlots/Plots.jl#2506


daschw:
We can add that back later, when the tests for it pass.

mkborregaard:
Don't they pass now?

daschw:
Not sure, will check in https://github.com/JuliaPlots/Plots.jl/pull/2512

PR #2511, opened 2020-03-28
Don't convert to any vector.

daschw:
Apparently that is not needed.


mkborregaard:
That's a little bizarre :-)
Isn't this the function that always errors when a recipe isn't defined?

daschw:
Yes, it's also still there. It just does not convert to `Any` anymore, so I renamed it to `series_vector`, which better describes, what it actually does now.

mkborregaard:
haha great

Issue #2349, opened 2019-12-29
Annotations doesn't work with Date coordinate in GR backend


o01eg:
When I trying to place annotation to plot with Date coordinate I got error:

```julia
gr()
p = plot([Dates.Date(2019, 1, 1), Dates.Date(2019, 2, 1)], [3, 4])
annotate!(p, [([Dates.Date(2019, 1, 15)], 3.2, Plots.text("Test", :red, :center))])

Error showing value of type Plots.Plot{Plots.GRBackend}:
ERROR: MethodError: no method matching wctondc(::Date, ::Float64)
Closest candidates are:
  wctondc(::Real, ::Real) at /home/o01eg/.julia/packages/GR/oiZD3/src/GR.jl:2569
Stacktrace:
 [1] gr_display(::Plots.Subplot{Plots.GRBackend}, ::Measures.Length{:mm,Float64}, ::Measures.Length{:mm,Float64}, ::Array{Float64,1}) at /home/o01eg/.julia/packages/Plots/qZHsp/src/backends/gr.jl:1921
 [2] gr_display(::Plots.Plot{Plots.GRBackend}, ::String) at /home/o01eg/.julia/packages/Plots/qZHsp/src/backends/gr.jl:713
 [3] gr_display at /home/o01eg/.julia/packages/Plots/qZHsp/src/backends/gr.jl:671 [inlined]
 [4] _display(::Plots.Plot{Plots.GRBackend}) at /home/o01eg/.julia/packages/Plots/qZHsp/src/backends/gr.jl:1977
 [5] display(::Plots.PlotsDisplay, ::Plots.Plot{Plots.GRBackend}) at /home/o01eg/.julia/packages/Plots/qZHsp/src/output.jl:144
 [6] display(::Any) at ./multimedia.jl:323
 [7] #invokelatest#1 at ./essentials.jl:709 [inlined]
 [8] invokelatest at ./essentials.jl:708 [inlined]
 [9] print_response(::IO, ::Any, ::Bool, ::Bool, ::Any) at /tmp/portage/dev-lang/julia-1.3.0/work/julia-1.3.0/usr/share/julia/stdlib/v1.3/REPL/src/REPL.jl:156
 [10] print_response(::REPL.AbstractREPL, ::Any, ::Bool, ::Bool) at /tmp/portage/dev-lang/julia-1.3.0/work/julia-1.3.0/usr/share/julia/stdlib/v1.3/REPL/src/REPL.jl:141
[...]

PR #2522, opened 2020-03-31
Fix for #2349. Converts TimeTypes in annotations to numbers to allow their placement 

wfrgra:


daschw:
Thanks!
Just a note for future me: A cleaner solution would probably be to somehow integrate series_annotations and the type recipe  processing pipeline.

PR #2541, opened 2020-04-06
special case LaTeXStrings in sanitation

BeastyBlacksmith:
I haven't found a way to do this without adding the LaTeXStrings dep... Suggestions?


daschw:
If it's only needed in PGFPlotsX you could also wrap a 
```julia
@require LaTeXStrings = "b964fa9f-0449-5b57-a5c2-d3ea65f4040f" begin
    ...
end
```
around the method definitions.

PR #2544, opened 2020-04-06
use -O1 for code in Plots.jl

JeffBezanson:
On julia master this reduces the time to first plot for me by about 5 seconds. So far I have not detected any measurable slowdowns e.g. for plotting many points (the drawing itself inside gksterm is by far the limiting factor).

### DIFF
--- a/src/Plots.jl
+++ b/src/Plots.jl
@@ -1,5 +1,9 @@
 module Plots
 
+if isdefined(Base, :Experimental) && isdefined(Base.Experimental, Symbol("@optlevel"))
+    @eval Base.Experimental.@optlevel 1
+end
+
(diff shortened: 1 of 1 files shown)


daschw:
Nice, thanks a lot!

ChrisRackauckas:
Why not -O0? Does 1 offer some optimizations that helps Plots?

mkborregaard:
Another question: will this behaviour change if called by packagecompiler to form a system image?

JeffBezanson:
-O0 would be fine too. I made some changes to -O1 that make it very close to -O0 compiler performance while generating significantly better code, so I feel -O1 is overall a good default. It often pays for itself.

This option currently only affects the JIT; building a system image will only use whatever -O option is passed on the command line.

mkborregaard:
Perfect <3

Issue #2549, opened 2020-04-08
pgfplotsx legend in outer-position


matthieugomez:
Plotting a legend in outer position does not seem to work with `pgfplotsx`
```julia
using Plots
pgfplotsx()
plot(1:10, rand(10))
plot(1:10, rand(10), legend = :outertop)
#> ERROR: The latex command `lualatex jl_7bA9WF.tex` failed
```

BeastyBlacksmith:
Which Plots version is that? I don't get the tex-error, but also not the expected output.
As a workaround you can pass a tuple to `legend` like: `plot(1:10, rand(10), legend = (0.5,1.1))`

matthieugomez:
Thanks!

PR #2550, opened 2020-04-08
all the legend positions

BeastyBlacksmith:
Fix: https://github.com/JuliaPlots/Plots.jl/issues/2549


matthieugomez:
It seems like  `topleft` is wrong though, it's too much to the left.
`:topleft => ("at" => string((-1.02, 0.98)), "anchor" => "north west")`

BeastyBlacksmith:
:D, good catch! Quite a bit off

BeastyBlacksmith:
fixed by https://github.com/JuliaPlots/Plots.jl/commit/483e45db8a96d08b1d422b2e037f37bbb90ea739

PR #2553, opened 2020-04-09
minorgrid, minorticks and framestyle zerolines fix for pgfplotsx

daschw:
```julia
plot(rand(10), xminorgrid = true, yminorticks = true)
```
pgfplotsx-minorgrid

### DIFF
--- a/src/backends/pgfplotsx.jl
+++ b/src/backends/pgfplotsx.jl
@@ -1013,6 +1013,8 @@ function pgfx_axis!(opt::PGFPlotsX.Options, sp::Subplot, letter)
     # ticks on or off
     if axis[:ticks] in (nothing, false, :none) || framestyle == :none
         push!(opt, "$(letter)majorticks" => "false")
+    elseif framestyle in (:grid, :zerolines)
+        push!(opt, "$letter tick style" => PGFPlotsX.Options("draw" => "none"))
     end
 
     # grid on or off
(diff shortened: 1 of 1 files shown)


src/backends/pgfplotsx.jl
```diff
@@ -1093,6 +1096,37 @@ function pgfx_axis!(opt::PGFPlotsX.Options, sp::Subplot, letter)
...
+        # minor ticks
+        # NOTE: PGFPlots would provide "minor x ticks num", but this only places minor ticks
+        #       between major ticks and not outside first and last tick to the axis limits.
+        #       Hence, we hack around with extra ticks. Unfortunately this conflicts with
+        #       `:zerolines` framestyle hack. So minor ticks are not working with
+        #       `:zerolines`.
+        minor_ticks = get_minor_ticks(sp, axis, ticks)
+        if minor_ticks !== nothing
+            minor_ticks =
+                ispolar(sp) && letter == :x ? [rad2deg.(minor_ticks)[3:end]..., 360, 405] :
```
daschw, line 1109: I copied this from above. Why do we skip the second tick here as well?

BeastyBlacksmith: that code was written by @pkofod and I just translated it. The best way to find out is to test what happens if you include it ^^;

daschw: OK, no that does not work how I expected. I'll leave it as it is.

pkofod: :shrug: I don't remember much about this :)

PR #2557, opened 2020-04-10
Use RecipesPipeline

daschw:


daschw:
So probably we have to drop Julia 1.0 support with this. Do we test on 1.1 instead? Will this require a a Plots 1.1 release?

mkborregaard:
I think a minor version is called for, yes.

daschw:
:tada: fixed for julia 1.0 so we don't need Plots 1.1. The issue was an outdated precompile file. This whole factoring out the recipe pipeline reduced the generated precompile file significantly (~1/2). So we might want to check for TTFP regressions and maybe add precompile statements in RecipesPipeline.jl

daschw:
OK, actually TTFP improves with this compared to current master with julia 1.4 on my system from
```julia
julia> @time using Plots; display(plot(rand(10)))
 16.589053 seconds (25.16 M allocations: 1.259 GiB, 3.22% gc time)
Plot{Plots.GRBackend() n=1}
```
to
```julia
julia> @time using Plots; display(plot(rand(10)))
 13.957989 seconds (23.78 M allocations: 1.173 GiB, 3.91% gc time)
Plot{Plots.GRBackend() n=1}
```

Project.toml
```diff
@@ -1,7 +1,7 @@
 name = "Plots"
 uuid = "91a5bcdd-55d7-5caf-9e0b-520d859cae80"
 author = ["Tom Breloff (@tbreloff)"]
-version = "1.0.7"
+version = "1.0.8"
```
mkborregaard, line 4: still would have called a massive pipeline change a minor version

daschw: Oh sorry, I was too impatient with merging then. Should I do anything about this?

mkborregaard: I guess it's not that important :-)
More just us getting used to the post-1.0 world

Issue #2560, opened 2020-04-10
Number of minor ticks cannot be specified in PGFPlotsX backend


mbfstudios:
The implementation in #2553 does not allow the number of minor ticks to be specified. According to documentation the attribute `minorticks` can be an integer.

[This line](https://github.com/daschw/Plots.jl/blob/fa7f10a6e905b7acf7a05a9d412c7483fa7b1275/src/backends/pgfplotsx.jl#L1122):
`"major tick length" => axis[:minorticks] ? "0.1cm" : "0"`
raises a type error when `axis[:minorgrid]` is not boolean.

BeastyBlacksmith:
I suppose this is fixed

PR #2561, opened 2020-04-10
Fix error when setting minor ticks. See #2560

mbfstudios:
Fix type error when specifying number of minor ticks for PGFPlotsX backend. See #2560


daschw:
Thanks

PR #2564, opened 2020-04-12
Switch to ColorSchemes

daschw:
Part 3

https://github.com/JuliaPlots/PlotUtils.jl/pull/87
https://github.com/JuliaPlots/PlotThemes.jl/pull/53


mkborregaard:
Sorry could you quickly outline why we need this here? Couldn't PlotUtils just draw it's gradients instead of locally, and essentially keep the same interface? Why should Plots need changing an taking a direct dep on ColorSchemes?

daschw:
Yeah, actually we don't need the direct dependency. I just thought since we already have the indirect dependency via PlotUtils we could make ColorSchemes available to the user as well. We are actually keeping more or less the same interface. So we can drop the ColorSchemes dep.
I added a ColorPalette type in PlotUtils. This needs integration here.

mkborregaard:
I mean it's fine to take it just trying to follow your thought process here

mkborregaard:
I mean it's fine to take it just trying to follow your thought process here

daschw:
Anyway, I have to add some functionality I dropped for ColorGradients back in PlotUtils before we can go ahead with this.
I think I should remove the ColorSchemes dep again. I added it at some point during PR development without thinking to long about it, but it does not really provide more convenience for the users. Everything they could do with ColorSchemes they can do with ColorGradients and ColorPalettes from PlotUtils as well.

Issue #1527, opened 2018-05-15
How to make ylabel horizontal?


jongminyoon:
How to make ylabel horizontal? Does this functionality exist? 

yrotation only rotates the yticks, not the ylabel.

benfolsom:
Horizontal y-axis labels are very common in published work. It may be worth prioritizing this.

isentropic:
@daschw  I think the relevant attribute is `:guidefontrotation` accodring to my findings, only pgfplot takes use of it. This attribute is forgotten in other backends I think

daschw:
https://github.com/JuliaPlots/Plots.jl/blob/master/src/backends/pyplot.jl#L1111 I think here would be the place to start.

t-bltg:
Fixed by https://github.com/JuliaPlots/Plots.jl/pull/3529.

PR #2565, opened 2020-04-12
guidefontrotation fix for pyplot

isentropic:
Fixes guide rotation for pyplot (https://github.com/JuliaPlots/Plots.jl/issues/1527)
plot(rand(10), rand(10), yguide="yguide", xguide="xguide", yguidefontrotation=45, xguidefontrotation="-45")
Figure_1


isentropic:
I could not fix this for GR properly, GR is alien to me

daschw:
The tests fail because of fonts changes in GR that have not been updated in the reference images yet. Should be fixed with https://github.com/JuliaPlots/PlotReferenceImages.jl/pull/65.

Issue #2573, opened 2020-04-13
ylabel rotation with PyPlot


JackDunnNZ:
The PyPlot backend doesn't seem to be rotating the y label correctly:

```julia
using Plots
Plots.pyplot()
plot(1:10, 1:10, ylabel="abcdefg")
```

![image](https://user-images.githubusercontent.com/4717044/79126615-4d00e400-7d6e-11ea-8bc5-c66112c577e6.png)

This is with Plots v1.0.9, on v0.29.9 the label is rotated correctly

isentropic:
Do you mean the going over the y axis?

JackDunnNZ:
Sorry, here's how it is in v0.29.9:

![image](https://user-images.githubusercontent.com/4717044/79135534-783eff80-7d7d-11ea-8f3d-e806e4690bae.png)

It also has the correct rotation when using GR in v1.0.9:

![image](https://user-images.githubusercontent.com/4717044/79135924-1c28ab00-7d7e-11ea-9393-f0e6b259c9d6.png)

Although I'm not sure what is going on with the font compared to how it was in v0.29.9:

![image](https://user-images.githubusercontent.com/4717044/79135652-b2a89c80-7d7d-11ea-9f99-d22b90e04673.png)

isentropic:
Yeah, rotation only works with pyplot, and it has been only added in https://github.com/JuliaPlots/Plots.jl/releases/tag/v1.0.9

Unfortunately, I'm really unfamiliar with GR to able to fix it.

isentropic:
Oh, I understand what you mean. It is rotated without the specification of the attribute like: `"yguidefontrotation=45"`
@daschw Perhaps you could help. Maybe the default value for the font rotation is not processed in the pipeline so that it is interpreted as no rotation by default for pyplot.

isentropic:
https://github.com/JuliaPlots/Plots.jl/pull/2565 This needs to be taken off then

JackDunnNZ:
Thanks, that makes sense. Yes it looks like the default is now being interpreted as "no rotation" somehow

isentropic:
I can send a quickfix (dirtyfix) to this by,


by adding this: 
```
axis[:yguidefontrotation] = axis[:yguidefontrotation]  + 90
```
somewhere before
[...]

PR #2567, opened 2020-04-12
Isen dev

isentropic:
markerstrokealpha is no longer ignored: https://github.com/JuliaPlots/Plots.jl/issues/2447
Though, does not work in GR.
Omitting `markerstrokealpha` from keywords still works (naturally).

```
scatter(rand(3), rand(3), markerstrokealpha = .5, m=:star, 
markersize=100, markerstrokewidth=10, lw=10)
```

Figure_1

PR #2578, opened 2020-04-14
fixing default guidefontrotation direction for pyplot

isentropic:
I forgot that I was working on the same branch when at #2567, so that PR should be ommited.

```
plot(1:10, 1:10, ylabel="ylab", xlabel="xlab")

plot(1:10, 1:10, 1:10, ylabel="ylab", xlabel="xlab",zlabel="zlab")

plot(1:10, 1:10, 1:10, ylabel="ylab", xlabel="xlab",zlabel="zlab", yguidefontrotation=15)
```
Figure_1
Figure_2

Figure_3

### DIFF
--- a/src/backends/pyplot.jl
+++ b/src/backends/pyplot.jl
@@ -1113,7 +1113,13 @@ function _before_layout_calcs(plt::Plot{PyPlotBackend})
             end
             pyaxis."label"."set_fontsize"(py_thickness_scale(plt, axis[:guidefontsize]))
             pyaxis."label"."set_family"(axis[:guidefontfamily])
-            pyaxis."label"."set_rotation"(axis[:guidefontrotation])
+
+            if (letter == :y && !RecipesPipeline.is3d(sp))
+                pyaxis."label"."set_rotation"(axis[:guidefontrotation] + 90)
+            else
+                pyaxis."label"."set_rotation"(axis[:guidefontrotation])
+            end
+
             for lab in getproperty(ax, Symbol("get_", letter, "ticklabels"))()
                 lab."set_fontsize"(py_thickness_scale(plt, axis[:tickfontsize]))
                 lab."set_family"(axis[:tickfontfamily])


isentropic:
@daschw should fix https://github.com/JuliaPlots/Plots.jl/issues/2573 now

src/backends/pyplot.jl
```diff
@@ -1113,7 +1113,13 @@ function _before_layout_calcs(plt::Plot{PyPlotBackend})
             end
             pyaxis."label"."set_fontsize"(py_thickness_scale(plt, axis[:guidefontsize]))
             pyaxis."label"."set_family"(axis[:guidefontfamily])
-            pyaxis."label"."set_rotation"(axis[:guidefontrotation])
+
+            if (letter == :y && !RecipesPipeline.is3d(sp))
```
daschw, line 1117: How about adding
```julia
 || (letter == :z && RecipesPipeline.is3d(sp))
```
here? Or would add `-90` to the rotation in that case?

isentropic: Isn't having letter z equivalent of being 3d plot?

For 3d plots no rotations need to be specified

daschw: But the z label in the 3D Plots you posted should be rotated analogously to y labels in 2D Plots, no?

daschw: ... to save whitespace for longer labels

isentropic: In v0.29.9, before the guidefontrotation was ever introduced, this is behavior:
```
 p = plot(1:10, 1:10, 1:10, ylabel="ylab", xlabel="xlab",zlabel="zlab")
pyplot()
gr()
ploltly()
```
Figure_1

gr
newplot


That is in case of 3d plot, the guides are not rotated by default (in gr() and pyplot()). There is already some inconsistence among backends. So I kept what was default for pyplot back in 0.29.9 (to avoid breaking)

PR #2581, opened 2020-04-15
add legendtitle for pgfplotsx

daschw:
```julia
using Plots
pgfplotsx()
plot(rand(10), legendtitle = "Legend title")
```
pgfplotsxlegend

It does not consider legendtitlefontsize etc. yet, but it's a start.

### DIFF
--- a/src/backends/pgfplotsx.jl
+++ b/src/backends/pgfplotsx.jl
@@ -240,6 +240,16 @@ function (pgfx_plot::PGFPlotsXPlot)(plt::Plot{PGFPlotsXBackend})
                 PGFPlotsX.Axis
             end
             axis = axisf(axis_opt)
+            if sp[:legendtitle] !== nothing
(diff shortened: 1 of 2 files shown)


src/backends/pgfplotsx.jl
```diff
@@ -240,6 +240,16 @@ function (pgfx_plot::PGFPlotsXPlot)(plt::Plot{PGFPlotsXBackend})
                 PGFPlotsX.Axis
             end
             axis = axisf(axis_opt)
+            if sp[:legendtitle] !== nothing
+                push!(axis, PGFPlotsX.Options("\\addlegendimage{empty legend}" => nothing))
```
BeastyBlacksmith, line 244: actually it should be enough to push the string here, but if it works with the `Options` you can also leave it as it is

daschw: Oh, I was not aware of that, thanks for the info!

Issue #2505, opened 2020-03-26
Allow :auto for endpoints of limits


BeastyBlacksmith:
Would be nice to be able to specify `plot(1:5, ylims=(-5,:auto))`.

mkborregaard:
very! I've often thought that. I've also thought `:centered` or `:symmetric` to have the lims symmetric around 0. And I also think these should all apply to `clims` as well

isentropic:
PyPlot implements this as
```
plt.plot(rand(10))
plt.xlim(right=2)
```
Perhaps, allows users suppy lims as a named tuple?

akdor1154:
This would be useful even if just to let users enforce that their y-axes begin at zero. Unless this is currently possible another way? (framestyle=:origin enforces this on the x-axis too which isn't always appropriate)

daschw:
> Unless this is currently possible another way?

`ylims = (0, Inf)`

daschw:
... or maybe I don't understand what the behavior of `:auto` would be?

BeastyBlacksmith:
> > Unless this is currently possible another way?
> 
> `ylims = (0, Inf)`

That is actually what I was looking for. I'd argue that this is a bit hard to discover, because semantically `Inf` is kind of doing the opposite of what I would naively expect it to do.

mkborregaard:
Yes agreed

daschw:
> semantically Inf is kind of doing the opposite of what I would naively expect it to do

`ylims = (0, NaN)` works too. It would probably not be to hard to also accept `missing`, `nothing`, and/or `:auto`.

PR #2600, opened 2020-04-19
axis limits = :symmetric

BeastyBlacksmith:
I am not so sure about this one.
It is only a subset of @mkborregaard's ideal scenario (https://github.com/JuliaPlots/Plots.jl/issues/2505#issuecomment-604491221) and that we should also support `xlims = (:symmetric, 5)` or `xlims = (:symmetric, :round)`.
But that is a bit tricky.


mkborregaard:
great :+1:

Issue #2592, opened 2020-04-17
Intended behavior of `thickness_scaling` for `markersize`


isentropic:
Currently, this
```
pyplot()
scatter(rand(3), thickness_scaling=3)
```
![image](https://user-images.githubusercontent.com/24591123/79519952-56e95680-8090-11ea-9f46-e78a24c93bc8.png)

``` 
gr()
scatter(rand(3), thickness_scaling=3)
```
Produces different output in gr and pyplot

![image](https://user-images.githubusercontent.com/24591123/79519896-2d302f80-8090-11ea-89de-d2ab0f132ba6.png)
Also legend box is slightly differently shaped. GR: sharp rectangle, PyPlot: circle corners

Things that are different:

1. Tick marks (small indents at ticks)
    1. gr: have equal thickness with axes spine
    2. pyplot: seemingly unscaled

2. Markerstrokewidth
    1. gr: unscaled
    2. pyplot: upscaled thickness

3. Markerstrokewidth in the legend
    1. gr: unscaled
    2. pyplot: unscaled

4. Markersize both in the legend and ax
    1. gr: unscaled
    2. pyplot: unscaled

I want to fix these issues in PyPlot (I am inept at GR). Let me know what should be the desired outcome.

Additionally, increasing markersize does not increase the markersize in the legend, which I think needs to be implemented too, or given additional argument to accoun for this.

isentropic:
Perhaps, @daschw could clarify this?

daschw:
IMO tick marks and marker strokes should be scaled everywhere (legend and axis).
I think markersizes should be upscaled with thickness scaling, but I'm not sure about that.
@mkborregaard I remember you have already put a lot of thought into marker sizes. What do you say?

isentropic:
I think for now, I'll try get there tick marks and markerstroke fixed for pyplot

mkborregaard:
I think markersizes should stay constant, but markerstrokes should clearly scale. So I think the pyplot behaviour is more correct here

isentropic:
> 
> 
> I think markersizes should stay constant, but markerstrokes should clearly scale. So I think the pyplot behaviour is more correct here

I think we need a argument for `legendmarkersize` and `legendmarkerwidth`.

[...]

PR #2605, opened 2020-04-20
Tick marks now scale according to thickness_scaling

isentropic:
`scatter(rand(3), thickness_scaling=3)`

Before:

before

After;
after

Subtle, but becomes obvious for larger `thickness_scaling` values

Relevant to https://github.com/JuliaPlots/Plots.jl/issues/2592


daschw:
Awesome, thanks!

PR #2607, opened 2020-04-21
don't add .tex to .tikz

BeastyBlacksmith:
If we reach this point, we already checked if the extension is existent.
This used to append `.tex` if saving as `.tikz` file, which is unwanted.
ref: https://github.com/KristofferC/PGFPlotsX.jl/issues/222

### DIFF
--- a/src/output.jl
+++ b/src/output.jl
@@ -45,6 +45,7 @@ end
 eps(fn::AbstractString) = eps(current(), fn)
 
 function tex(plt::Plot, fn::AbstractString)
+  fn = addExtension(fn, "tex")
   io = open(fn, "w")
   show(io, MIME("application/x-tex"), plt)
   close(io)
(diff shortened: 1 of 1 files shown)


BeastyBlacksmith:
Or are those methods intended to be used independently from `savefig`?

daschw:
For `png` we expect `png("fn")` to do the same as `savefig("fn.png")`, but the others (svg, pdf, tex, ...) are not exported. So I guess not.

BeastyBlacksmith:
I think this is a better solution

eddieschoute:
Using `savefig(plot, "plot.tikz")` still gives me `plot.tex` instead. This cannot be included easily using `\includegraphics` with the `tikzscale` library (it doesn't recognize the extension as valid).

BeastyBlacksmith:
Which version of Plots are you on?

eddieschoute:
1.5.8

BeastyBlacksmith:
should be fixed with https://github.com/JuliaPlots/Plots.jl/pull/2901. Next time it is better to open an issue and link to the PR than commenting on the PR (that is easier to miss).

PR #2612, opened 2020-04-22
fix savefig extensions

daschw:


src/output.jl
```diff
@@ -88,10 +88,13 @@ const _savemap = Dict(
     "tex" => tex,
     "json" => json,
     "html" => html,
+    "tex" => tex,
```
BeastyBlacksmith, line 91: thats the same as line 88

BeastyBlacksmith:
Out of curiousity. What was the issue?

src/output.jl
daschw: oh, yeah

daschw:
apparantly all files are saved as `fn.ext.ext` instead of `fn.ext`

daschw:
... on julia 1.0

PR #2616, opened 2020-04-23
pyplot: legend marker scales independently 

isentropic:
Fix: https://github.com/JuliaPlots/Plots.jl/issues/2592
markers in the legend scale independently (cannot be modified by the user). pyplot
Here are examples:
```
# Default case
scatter(rand(3), thickness_scaling=1)
```
Figure_1
```
scatter(rand(3), thickness_scaling=4)
```
Figure_2

Since the user has explicit control of the marker parameters on the actual plot, changing them as `thickness_scaling` increases is not critical.

However, I still that the marker size on the actual plot should somehow scale too, but this needs discussion again.

### DIFF
--- a/src/backends/pyplot.jl
+++ b/src/backends/pyplot.jl
@@ -539,7 +539,7 @@ function py_add_series(plt::Plot{PyPlotBackend}, series::Series)
                     label = series[:label],
                     zorder = series[:series_plotindex] + 0.5,
                     marker = py_marker(_cycle(shapes,i)),
-                    s =  py_thickness_scale(plt, _cycle(series[:markersize],i) .^ 2),
+                    s =  py_thickness_scale(plt, _cycle(series[:markersize],i)).^ 2,
                     facecolors = py_color(get_markercolor(series, i), get_markercoloralpha(series, i)),
                     edgecolors = msc,
                     linewidths = lw,
@@ -570,7 +570,7 @@ function py_add_series(plt::Plot{PyPlotBackend}, series::Series)
                   push!(cur_y_list, _cycle(y,i))
 
                   push!(cur_color_list, _cycle(markercolor, i))
-                  push!(cur_scale_list, py_thickness_scale(plt, _cycle(series[:markersize],i) .^ 2))
+                  push!(cur_scale_list, py_thickness_scale(plt, _cycle(series[:markersize],i)).^ 2)
 
                   continue
                 end
@@ -580,7 +580,7 @@ function py_add_series(plt::Plot{PyPlotBackend}, series::Series)
                   zorder = series[:series_plotindex] + 0.5,
                   marker = prev_marker,
                   s = cur_scale_list,
-                  edgecolors = py_color(get_markerstrokecolor(series), get_markerstrokealpha(series)),  # Do we need include i?
+                  edgecolors = py_color(get_markerstrokecolor(series), get_markerstrokealpha(series)),
                   linewidths = py_thickness_scale(plt, series[:markerstrokewidth]),
                   facecolors = cur_color_list,
                   extrakw...
@@ -590,7 +590,7 @@ function py_add_series(plt::Plot{PyPlotBackend}, series::Series)
                 cur_y_list = [_cycle(y,i)]
 
                 cur_color_list = [_cycle(markercolor, i)]
-                cur_scale_list = [py_thickness_scale(plt, _cycle(series[:markersize],i) .^ 2)]
+                cur_scale_list = [py_thickness_scale(plt, _cycle(series[:markersize],i)) .^ 2]
 
                 prev_marker = cur_marker
             end
@@ -615,7 +615,7 @@ function py_add_series(plt::Plot{PyPlotBackend}, series::Series)
                 label = series[:label],
                 zorder = series[:series_plotindex] + 0.5,
                 marker = py_marker(series[:markershape]),
-                s = py_thickness_scale(plt, series[:markersize] .^ 2),
+                s = py_thickness_scale(plt, series[:markersize]) .^2,
                 edgecolors = py_color(get_markerstrokecolor(series), get_markerstrokealpha(series)),
                 linewidths = py_thickness_scale(plt, series[:markerstrokewidth]),
                 extrakw...
@@ -1320,10 +1320,10 @@ function py_add_legend(plt::Plot, sp::Subplot, ax)
                             linewidth = py_thickness_scale(plt, clamp(get_linewidth(series), 0, 5)),
                             linestyle = py_linestyle(:path, get_linestyle(series)),
                             marker = py_marker(_cycle(series[:markershape], 1)),
-                            markersize = py_thickness_scale(plt, 6), # 6 looks same the default
+                            markersize = py_thickness_scale(plt, 6), # 6 looks same the default, do not put ^2
                             markeredgecolor = py_color(single_color(get_markerstrokecolor(series)), get_markerstrokealpha(series)),
                             markerfacecolor = py_color(single_color(get_markercolor(series, clims)), get_markeralpha(series)),
-                            markeredgewidth = py_thickness_scale(plt, 1)
+                            markeredgewidth = py_thickness_scale(plt, 1.1)  # scales better
                         )
                     else
                         series[:serieshandle][1]
--- a/src/backends/pyplot.jl
+++ b/src/backends/pyplot.jl
@@ -1317,13 +1317,13 @@ function py_add_legend(plt::Plot, sp::Subplot, ax)
                     elseif series[:seriestype] in (:path, :straightline, :scatter)
                         PyPlot.plt."Line2D"((0,1),(0,0),
                             color = py_color(single_color(get_linecolor(series, clims)), get_linealpha(series)),
-                            linewidth = py_thickness_scale(plt, clamp(sp[:legendfontsize]/10, 0, 5)),
+                            linewidth = py_thickness_scale(plt, sp[:legendfontsize] / 8),
                             linestyle = py_linestyle(:path, get_linestyle(series)),
                             marker = py_marker(_cycle(series[:markershape], 1)),
-                            markersize = py_thickness_scale(plt, sp[:legendfontsize]),
+                            markersize = py_thickness_scale(plt, 0.8 * sp[:legendfontsize]),
                             markeredgecolor = py_color(single_color(get_markerstrokecolor(series)), get_markerstrokealpha(series)),
                             markerfacecolor = py_color(single_color(get_markercolor(series, clims)), get_markeralpha(series)),
-                            markeredgewidth = py_thickness_scale(plt, series[:markerstrokewidth] * sp[:legendfontsize] / series[:markersize])   # retain the markersize/markerstroke ratio from the markers on the plot
+                            markeredgewidth = py_thickness_scale(plt, 0.8 * series[:markerstrokewidth] * sp[:legendfontsize] / series[:markersize])   # retain the markersize/markerstroke ratio from the markers on the plot
                         )
                     else
                         series[:serieshandle][1]
--- a/src/backends/pyplot.jl
+++ b/src/backends/pyplot.jl
@@ -1317,7 +1317,7 @@ function py_add_legend(plt::Plot, sp::Subplot, ax)
                     elseif series[:seriestype] in (:path, :straightline, :scatter)
                         PyPlot.plt."Line2D"((0,1),(0,0),
                             color = py_color(single_color(get_linecolor(series, clims)), get_linealpha(series)),
-                            linewidth = py_thickness_scale(plt, clamp(get_linewidth(series), 0, 5)),
+                            linewidth = py_thickness_scale(plt, clamp(sp[:legendfontsize]/10, 0, 5)),
                             linestyle = py_linestyle(:path, get_linestyle(series)),
                             marker = py_marker(_cycle(series[:markershape], 1)),
                             markersize = py_thickness_scale(plt, sp[:legendfontsize]),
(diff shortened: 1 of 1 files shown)


isentropic:
Relevant issue https://github.com/JuliaPlots/Plots.jl/issues/2592

daschw:
I agree that the plot markersize should scale here too.

isentropic:
My issue with this is that, then there are 2 tweaking knobs on the markersize on the actual plot --- `thickness_scaling` and `markersize`. That is, `markersize` is not the only thing that determines the markersize on the plot anymore.
On the other side, I agree that `thickness_scaling` should set sensible defaults for all the parameters.

daschw:
> My issue with this is that, then there are 2 tweaking knobs on the markersize on the actual plot 

Well, that's the same for everything that thickness_scaling affects, like linewidth, fonts, etc.

isentropic:
Here are some updates
```
scatter(rand(3), thickness_scaling=5)
scatter(rand(3), thickness_scaling=5, markersize=24)
```
Figure_1
Figure_2


Interestingly enough, the code was trying to scale marker sizes already, but the $^2$ (squaring) was applied in the wrong place

jamblejoe:
Hi,
with the pyplot backend the new behaviour is as follows:
```
for markersize in [10,20,30]
for legendfontsize in [10, 20, 30]
for thickness_scaling in [1,2]
       p=plot([1,2,3], marker=:circle, markersize=markersize, linetype=:scatter, legend=true, 
                    legendfontsize=legendfontsize, thickness_scaling=thickness_scaling, 
                    title="ms:$markersize, lf:$legendfontsize, ths:$thickness_scaling")
end
end
end
```





















PS: I do not know how to shorten this comment. Please feel free to edit!

isentropic:
Should I merge this? @daschw

daschw:
To be honest, I'm not yet 100% convinced that the legend markersize should grow with the legend font size. I'm not saying that I I am convinced that it should not, I just think we should decide on a general consistent approach here first and then implement it for each backend.

- Legend linewidths match the plot linewidths up to a maximum value. They don't scale with legend fonts.
- With this PR (if I understand it correctly) marker sizes and strokewidths scale with the legend font, but do not match the plot markers.

I find this inconsistent. Maybe it's fine, but maybe we should aim for a more consistent general design. Should markers use the behavior of linewidths? Should the linewidth behavior be changed to the marker method in this PR?

BeastyBlacksmith:
I do think that markers in the legend should not exceed the fontsize.
So I would vote for matching the markersize of the plot and cap it at the fontsize.
And same for markerstrokewidth but cap at 0.25 markersize.

jamblejoe:
I think of the series plots and the legends as totally different plots with the only connection being the "same" markers/lines. Therefore, I do not immediately see any reason, why markers/lines should scale the same in both. The appearance of the markers/lines in the legend only has to fullfill one main purpose: to quickly match the corresponding data in the series plot. 

Maybe an example to clarify this. The following plot came up as a part of a bigger plot. There are thousands of points plotted with small circle markers. If the markersize would be the same in the legend, one could barely see it in the legend or it would hide features in the series plot.
grafik

I think the size of markers AND lines AND font in the legend should be of the same order by default.

isentropic:
> I do think that markers in the legend should not exceed the fontsize.
> So I would vote for matching the markersize of the plot and cap it at the fontsize.
> And same for markerstrokewidth but cap at 0.25 markersize.

I believe that upon scaling of the marker, we should retain strokewidth/markersize. So that strokewidth only scales to preserve this ratio. I totally agree with @jamblejoe about this.
Here's another issue I stumbled upon https://discourse.julialang.org/t/resize-legend-box-in-subplots-for-plots-jl/36141/7 in this case the marker is too large. Currently (in master) markersize in the legend is just hardcoded to take backend default value. I think it should be somewhat dynamic at least

jamblejoe:
> I believe that upon scaling of the marker, we should retain strokewidth/markersize. So that strokewidth only scales to preserve this ratio. 

> Currently (in master) markersize in the legend is just hardcoded to take backend default value. I think it should be somewhat dynamic at least

I agree with both! It should not be hardcoded to take the backend's default value! 

The question now is: on what should markersize in the legend depend? on the fontsize? or should the fontsize depend on the markersize? Do we want to be able to scale them individually and make them by default scale equally?

daschw:
@jamblejoe you make a good point with your example with the small markers above.
So I think I agree to have the legend markersizes and linewitdths scale with the font size. With this we can no longer distinguish two lines with the same color but different widths in the legend, but this approach is limited anyway.

Here are my opinions to your questions:
> on what should markersize in the legend depend? on the fontsize?

yes

> or should the fontsize depend on the markersize?

no, would be weird to have different font sizes if we have different markersizes.

> Do we want to be able to scale them individually and make them by default scale equally?

I'd say make them scale equally by default.

isentropic:
An additional argument supporting this would be that if the legendbox scales with the fontsize, then should all the elements inside the legendbox including the marker, even the line widths I think for lineplots. 

```
plot(rand(1000), rand(1000),lw=0.1, legendfont=30)
```
`lw` scales with legendfont
image

`lw` takes the value from the actual plot
image

floswald:
not sure how related that is, but I get an error on this branch:

```
px = @df x plot(:year, :value, group = :region,
	title = titles[i],
	titlefontsize=10,
	legend = :topleft,
	linewidth=2,marker = (:circle,3),
	legendfontsize = 5)

ERROR: MethodError: no method matching get_color_palette(::Symbol, ::RGBA{Float64}, ::Int64)
Closest candidates are:
  get_color_palette(::Any, ::Any) at /Users/florian.oswald/.julia/packages/PlotUtils/Rebxf/src/colorschemes.jl:327
  get_color_palette(::ColorGradient, ::Any) at /Users/florian.oswald/.julia/packages/PlotUtils/Rebxf/src/colorschemes.jl:328
Stacktrace:
 [1] _update_subplot_colors(::Plots.Subplot{Plots.GRBackend}) at /Users/florian.oswald/.julia/dev/Plots/src/args.jl:1404
 [2] _update_subplot_args(::Plots.Plot{Plots.GRBackend}, ::Plots.Subplot{Plots.GRBackend}, ::Dict{Symbol,Any}, ::Int64, ::Bool) at /Users/florian.oswald/.julia/dev/Plots/src/args.jl:1488
 [3] _subplot_setup(::Plots.Plot{Plots.GRBackend}, ::Dict{Symbol,Any}, ::Array{Dict{Symbol,Any},1}) at /Users/florian.oswald/.julia/dev/Plots/src/pipeline.jl:258
 [4] plot_setup! at /Users/florian.oswald/.julia/dev/Plots/src/pipeline.jl:157 [inlined]
 [5] recipe_pipeline!(::Plots.Plot{Plots.GRBackend}, ::Dict{Symbol,Any}, ::Tuple{Array{Int64,1},Array{Float64,1}}) at /Users/florian.oswald/.julia/packages/RecipesPipeline/dZsrD/src/RecipesPipeline.jl:84
 [6] _plot!(::Plots.Plot{Plots.GRBackend}, ::Dict{Symbol,Any}, ::Tuple{Array{Int64,1},Array{Float64,1}}) at /Users/florian.oswald/.julia/dev/Plots/src/plot.jl:167
 [7] plot(::Array{Int64,1}, ::Vararg{Any,N} where N; kw::Base.Iterators.Pairs{Symbol,Any,NTuple{7,Symbol},NamedTuple{(:group, :title, :titlefontsize, :legend, :linewidth, :marker, :legendfontsize),Tuple{Array{Int64,1},String,Int64,Symbol,Int64,Tuple{Symbol,Int64},Int64}}}) at /Users/florian.oswald/.julia/dev/Plots/src/plot.jl:57
 [8] add_label(::Array{String,1}, ::Function, ::Array{Int64,1}, ::Vararg{Any,N} where N; kwargs::Base.Iterators.Pairs{Symbol,Any,NTuple{7,Symbol},NamedTuple{(:group, :title, :titlefontsize, :legend, :linewidth, :marker, :legendfontsize),Tuple{Array{Int64,1},String,Int64,Symbol,Int64,Tuple{Symbol,Int64},Int64}}}) at /Users/florian.oswald/.julia/packages/StatsPlots/lRhuT/src/df.jl:153
 [9] (::Main.LandUse.var"#93#96"{Array{String,1},Int64})(::DataFrames.DataFrame) at ./array.jl:0
 [10] plot_ts_all(::Array{Main.LandUse.Country,1}) at /Users/florian.oswald/.julia/dev/LandUse/src/plotter.jl:49
```

it's puzzling becuase I don't specify any color at all.

daschw:
> not sure how related that is, but I get an error on this branch:

That's unrelated to the changes in this PR. A rebase on current master and `]up` should hopefully fix that.

BeastyBlacksmith:
These are good examples, I think it is a good idea.

src/backends/pyplot.jl
```diff
@@ -1317,13 +1317,13 @@ function py_add_legend(plt::Plot, sp::Subplot, ax)
...
-                            linewidth = py_thickness_scale(plt, clamp(get_linewidth(series), 0, 5)),
+                            linewidth = py_thickness_scale(plt, clamp(sp[:legendfontsize]/10, 0, 5)),
                             linestyle = py_linestyle(:path, get_linestyle(series)),
                             marker = py_marker(_cycle(series[:markershape], 1)),
-                            # markersize = py_thickness_scale(plt, series[:markersize]), # In case we decide that markersize needs to be scaled in the legend too
+                            markersize = py_thickness_scale(plt, sp[:legendfontsize]),
                             markeredgecolor = py_color(single_color(get_markerstrokecolor(series)), get_markerstrokealpha(series)),
                             markerfacecolor = py_color(single_color(get_markercolor(series, clims)), get_markeralpha(series)),
-                            markeredgewidth = py_thickness_scale(plt, series[:markerstrokewidth])
+                            markeredgewidth = py_thickness_scale(plt, series[:markerstrokewidth] * sp[:legendfontsize] / series[:markersize])   # retain the markersize/markerstroke ratio from the markers on the plot
```
daschw, line 1326: actually how about `0.8 * sp[:legendfontsize]`? I'm not sure how this looks with multiple series.

BeastyBlacksmith: I actually like the idea, but I also plot zero width markers from time to time and that would cause trouble here.

daschw: Oh, sorry, my comment was meant for line 1323.

src/backends/pyplot.jl
```diff
@@ -1317,13 +1317,13 @@ function py_add_legend(plt::Plot, sp::Subplot, ax)
                     elseif series[:seriestype] in (:path, :straightline, :scatter)
                         PyPlot.plt."Line2D"((0,1),(0,0),
                             color = py_color(single_color(get_linecolor(series, clims)), get_linealpha(series)),
-                            linewidth = py_thickness_scale(plt, clamp(get_linewidth(series), 0, 5)),
+                            linewidth = py_thickness_scale(plt, clamp(sp[:legendfontsize]/10, 0, 5)),
                             linestyle = py_linestyle(:path, get_linestyle(series)),
                             marker = py_marker(_cycle(series[:markershape], 1)),
-                            # markersize = py_thickness_scale(plt, series[:markersize]), # In case we decide that markersize needs to be scaled in the legend too
+                            markersize = py_thickness_scale(plt, sp[:legendfontsize]),
```
daschw, line 1323: `0.8 * sp[:legendfontsize]` here?

daschw:
How about
```julia
ms = first(series[:markersize])
msw = first(series[:markerstrokewidth])
legend_ms, legend_msw = if ms > 0
    0.8 * sp[:legendfontsize], 0.8 * sp[:legendfontsize] * msw / ms
else
    0, 0.8 * sp[:legendfontsize] * msw / 8
end
```
for the legend markers?
With this I get in GR:
```julia
scatter(rand(10, 4), markerstrokewidth = [0 1 2 3])
```
legendmarker

```julia
scatter(rand(10, 4), markerstrokewidth = [0 1 2 3], legendfontsize = 16)
```
legendmarkerbig

isentropic:
> How about
> ```julia
> ms = first(series[:markersize])
> msw = first(series[:markerstrokewidth])
> legend_ms, legend_msw = if ms > 0
>     0.8 * sp[:legendfontsize], 0.8 * sp[:legendfontsize] * msw / ms
> else
>     0, 0.8 * sp[:legendfontsize] * msw / 8
> end
> ```
> for the legend markers?
> With this I get in GR:
> ```julia
> scatter(rand(10, 4), markerstrokewidth = [0 1 2 3])
> ```
> legendmarker
> 
> ```julia
> scatter(rand(10, 4), markerstrokewidth = [0 1 2 3], legendfontsize = 16)
> ```
> legendmarkerbig
> 

I can't understand the if statement here, if markersize > 0? Isn't this always the case?

daschw:
You are probably right, we might just use the first version. I was just thinking of avoiding division by zero errors if a user passes markersize = 0, but I can't think of a use case for this.

src/backends/pyplot.jl
daschw:
```suggestion
                            markeredgewidth = py_thickness_scale(plt, 0.8 * series[:markerstrokewidth] * sp[:legendfontsize] / series[:markersize])   # retain the markersize/markerstroke ratio from the markers on the plot
```

src/backends/pyplot.jl
daschw:
```suggestion
                            markersize = py_thickness_scale(plt, 0.8 * sp[:legendfontsize]),
```

src/backends/pyplot.jl
```diff
@@ -1317,13 +1317,13 @@ function py_add_legend(plt::Plot, sp::Subplot, ax)
                     elseif series[:seriestype] in (:path, :straightline, :scatter)
                         PyPlot.plt."Line2D"((0,1),(0,0),
                             color = py_color(single_color(get_linecolor(series, clims)), get_linealpha(series)),
-                            linewidth = py_thickness_scale(plt, clamp(get_linewidth(series), 0, 5)),
+                            linewidth = py_thickness_scale(plt, clamp(sp[:legendfontsize]/10, 0, 5)),
```
daschw, line 1320:
```suggestion
                            linewidth = py_thickness_scale(plt, sp[:legendfontsize] / 8),
```

jamblejoe:
Is this issue only related to pyplot? Or to other backends as well?

isentropic:
Does it look good now? @daschw

PR #2657, opened 2020-05-01
allow passing a function to `clims`

mileslucas:
## Enhancement

This would allow users passing a function to clims like
```julia
heatmap(img, clims=myfunc)
```
The function should have a signature that takes in series values and returns a tuple. This is actually the default behavior, because the `:auto` clims use `Plots.ignorenan_extrema`. This will be useful for people who want to write special functions for plotting based on the limits instead of the normalization used. I have a particular use case based on a routine used literally _everywhere_ in astronomy and some related imaging fields. 

## Roadmap
I initially thought this would be a simple "if `get(kw, :clims) isa Function` then calculate the limits differently", but the three `get_clims` functions are slightly disconnected, since `:clims` isn't a `Series` attribute.

My proposed solution (so far) is to add this function
```julia
get_clims(::Series, func=ignorenan_extrema)
```
and then rework the other `get_clims` to use this method. Part of that will require writing a simple dispatch for the 3 types of expected values for `:clims`. These function definitions should avoid introducing any deprecations.


mileslucas:
well, the rewrite happened quicker than I thought it would take 😅

I've tested that `heatmap(img)`, `heatmap(img, clims=extrema)`, and `heatmap(img, clims=(-1, 1))` all work appropriately by hand.

I assume the first and third versions above are tested already, so where should I add tests for the second version?

src/utils.jl
```diff
@@ -410,44 +410,46 @@ xlims(sp_idx::Int = 1) = xlims(current(), sp_idx)
...
-        isfinite(clims[1]) && (zmin = clims[1])
-        isfinite(clims[2]) && (zmax = clims[2])
+        get_clims(series, op)
     end
     return zmin <= zmax ? (zmin, zmax) : (NaN, NaN)
 end
 
-function get_clims(series::Series)
+"""
+    get_clims(::Series, op=Plots.ignoranan_extrema)
```
BeastyBlacksmith, line 439: there is a typo here:  should be `ignorenan_extrema`

BeastyBlacksmith:
If I read this correctly, then currently the last series always defines the `clims`.
I would propose to change that, such that `cmin` is the minimal `op(z)` of all series and `cmax` accordingly.

mileslucas:
That’s how it works, currently. In the subplot version of get_clims it starts as -Inf to Inf and in the for loop it updates via _update_clims with the old and the new.

BeastyBlacksmith:
Though you should check, what changed with the contour example: https://docs.juliaplots.org/latest/generated/gr/#gr-ref22-1

daschw:
It's weird that the images only differ on julia 1.0, but not on julia 1.4 in the travis tests. I'll close and reopen to retrigger the tests.

mileslucas:
appveyor seems to be ~6 hours behind in the test queue.

daschw:
Don't worry about appveyor, it hasn't worked for us in a long time. Thanks a lot for this!

PR #2663, opened 2020-05-03
Added webm support for animations.

dietercastel:
I tested it with the [circle example](https://docs.juliaplots.org/latest/animations/) (s/gif/webm/) and that seemed to work just fine.

However when I tried testing it in a Jupyter notebook (for viewing) the webm function didn't seem to be exported properly (while it is properly exported and usable in the REPL). This blocks the testing on that front (for me at least), unless anyone has a solution for that?

### DIFF
--- a/src/Plots.jl
+++ b/src/Plots.jl
@@ -108,6 +108,7 @@ export
     gif,
     mov,
     mp4,
+    webm,
     animate,
     @animate,
     @gif,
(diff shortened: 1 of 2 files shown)


dietercastel:
With [a swift response on discourse](https://discourse.julialang.org/t/how-to-use-developed-package-in-ijulia/38676) I was able to test the Jupyter notebook interaction as well. Result below:
Screenshot 2020-05-03 at 19 22 22

daschw:
Nice, for me this works perfectly in jupyter. Thanks a lot!

dietercastel:
My pleasure! I've been enjoying Plots.jl quite a bit already so glad I could contribute. :-)

PR #2673, opened 2020-05-05
Removed arrow3d from gr.jl in backends/

komi1230:
Hello.

I'm making GR wrapper in Common Lisp (see [this](https://github.com/komi1230/kai)), and I've found [GR](https://github.com/sciapp/gr) doesn't provide a function to draw 3D arrows.

This `polyline` function is confusing and may make bugs.

### DIFF
--- a/src/backends/gr.jl
+++ b/src/backends/gr.jl
@@ -176,14 +176,6 @@ function gr_polyline3d(x, y, z, func = GR.polyline3d; arrowside = :none, arrowst
         # if we found a start and end, draw the line segment, otherwise we're done
         if istart > 0 && iend > 0
             func(x[istart:iend], y[istart:iend], z[istart:iend])
-            if arrowside in (:head,:both)
-                gr_set_arrowstyle(arrowstyle)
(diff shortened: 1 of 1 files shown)


daschw:
Why exactly do you want these lines removed from Plots?

komi1230:
GR provides a function to draw arrow only for 2D.
There is nothing for 3D.
See [this search](https://github.com/sciapp/gr/search?q=drawarrow&unscoped_q=drawarrow)

For example, this works fine:

```julia
GR.drawarrow(x[i], y[i], x[i+1], y[i+1])
```

But this doesn't work:

```julia
GR.drawarrow(x[i], y[i], z[i], x[i+1], y[i+1], z[i+1])
```

So, if you write `polyline3d(x, y, z, ; arrowside = :head)` in the code, this crashes.

Of course, this error doesn't happen if you don't provide a keyword for `arrowside`.
But I think it's better to remove needless lines for simplicity and avoidance of errors.

daschw:
OK, I see, thanks for the hint and the contribution!

PR #2679, opened 2020-05-06
pyplot hexbin

isentropic:
Fixing https://github.com/JuliaPlots/Plots.jl/issues/2491
Currently after my accidental direct commit https://github.com/JuliaPlots/Plots.jl/commit/28ebc0ccd893f39ab9518d181ef8c3852ab02454  this works:

```
 hexbin(rand(1000), rand(1000), thickness_scaling=2, bins=(5,8))
```
Note, color is wrong
image

However, specifying color directly helps
```
hexbin(rand(1000), rand(1000), thickness_scaling=2, bins=(5,8), c=:blues)
```
image

Additionally, `clims` are not calculated correctly too.
I cannot find `hexbin` recipe to try to understand how it gets processed in the `recipes.jl`

### DIFF
--- a/src/backends/pyplot.jl
+++ b/src/backends/pyplot.jl
@@ -612,7 +612,7 @@ function py_add_series(plt::Plot{PyPlotBackend}, series::Series)
             alpha = series[:fillalpha],
             cmap = py_fillcolormap(series),  # applies to the pcolorfast object
             zorder = series[:series_plotindex],
-            extrakw...  # for some reason vmin and vmax are NaN??? hexbin works without passing vmin and vmax
+            # extrakw...  # We are not supporting clims for hexbin as calculation of bins is not trivial
         )
         push!(handles, handle)
     end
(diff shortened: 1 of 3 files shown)


daschw:
I am not familiar with hexbin. What should the color be, a gradient?

daschw:
There is no recipe, pylot is the only backend that supports it. I think you could fix the colors by adding the method `RecipesPipeline.is_surface(::Type{Val{:hexbin}}) = true`.

isentropic:
> 
> 
> I am not familiar with hexbin. What should the color be, a gradient?

it is almost same as 2d histogram, but binned into hexagons

isentropic:
I think this is ready to be merged.
```
hexbin(rand(1000), rand(1000), thickness_scaling=2, bins=(15,6), linewidth=1)
```
image

This is still at the early stage, as there is no support for `clims` as binning is not trivial. Heck even, the number of bins is not intuitive to understand, but it now at least plots something for anyone who is interested in hexbins. It looks pretty but definitely not easy to implement from scratch, from recipes.

PR #2680, opened 2020-05-06
Allow labels and titles to be passed as vectors on all backends

### LINKED ISSUE #901: Multiple labels must be in a row vector
When plotting with labels put in a column vector or tuple, the labels are printed as "trace 0", "trace 1", and so forth. A MWE to reproduce this is:
```julia
using Plots
plotly()
xs = 1:10
f(x) = [0, 1x, 2x]
plot(xs, hcat(map(f, xs)...)', labels=["Total", "Channel 1", "Channel 2"])
```
For someone unfamiliar with Plots (me), this behaviour is quite unexpected—and as far is I can tell, undocumented. I think it would be nice if `plot` would accept labels in any sequence type, not just row vectors.

daschw:
fix #901

I know Plots has the "columns are series" philosophy but especially for titles and labels another interpretation than having each element of a vector map to a series/subplot does not really make sense and having to `permutedims` vectors of strings when passing multiple labels at once is a recurring gripe. So I think we can make an exception here.

See also https://github.com/JuliaPlots/Plots.jl/pull/2183 and https://github.com/JuliaPlots/Plots.jl/issues/901#issuecomment-616550691

### DIFF
--- a/src/args.jl
+++ b/src/args.jl
@@ -1103,6 +1103,10 @@ function RecipesPipeline.preprocess_attributes!(plotattributes::AKW)
         plotattributes[:colorbar] = convertLegendValue(plotattributes[:colorbar])
     end
 
+    if haskey(plotattributes, :label) && plotattributes[:label] isa AbstractVector
+        plotattributes[:label] = permutedims(plotattributes[:label])
+    end
+
     # framestyle
     if haskey(plotattributes, :framestyle) && haskey(_framestyleAliases, plotattributes[:framestyle])
         plotattributes[:framestyle] = _framestyleAliases[plotattributes[:framestyle]]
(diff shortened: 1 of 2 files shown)


BeastyBlacksmith:
Well, if we were to be consistent here `label = [l1, l2]` would label each segment of the series, e.g. each point in a scatter plot.
However, that is currently not the case and debatable how useful it is.

daschw:
I don't really think multiple legend entries for a single series makes sense. Isn't the whole idea of a series to group datapoints that belong together and link them visually (same linstyle/color/width/marker...) If you want two sets of datapoints with different appearance and different labels, why have them in a single series?

BeastyBlacksmith:
There are some edge cases, like you use `zcolor` and there is some special point you want to label, but I would say in that case an annotation is better suited than a legend entry.

PR #2681, opened 2020-05-06
remove StatsPlots recipe warning

### LINKED ISSUE #2678: Warnings for moved recipes check Main only
I'm developing a module Foo that is `using` both Plots and StatsPlots, and in it I have a function `my_boxplot` which calls `boxplot`. If I run

```
using Foo
my_boxplot()
```

then I get the following warning because I haven't also run `using StatsPlots`.

https://github.com/JuliaPlots/Plots.jl/blob/3d693d88e07b0d61c7f501cfe1b65baaabe57337/src/args.jl#L1113-L1115

Can this warning be turned off if StatsPlots is already being used by the module?

daschw:
fix #2678

### DIFF
--- a/src/args.jl
+++ b/src/args.jl
@@ -1108,12 +1108,6 @@ function RecipesPipeline.preprocess_attributes!(plotattributes::AKW)
         plotattributes[:framestyle] = _framestyleAliases[plotattributes[:framestyle]]
     end
 
-    # warnings for moved recipes
-    st = get(plotattributes, :seriestype, :path)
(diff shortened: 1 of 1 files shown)


mkborregaard:
Are you sure this is what we want? see my comment back on #2678

daschw:
Yeah, you are right. I'm not sure if we want this. Your suggestion using Recipes is obviously the preferred way to handle this and for this use case the warning can be really helpful for users to have that warning.

PR #2699, opened 2020-05-13
Wrap more GR backend code with `withenv`

asinghvi17:


BeastyBlacksmith:
Nice, do you have an idea, if this solves the issue of orphan subprocesses?

asinghvi17:
I'm not sure.  The diff is a little weird, and I'm not sure why CI is failing - will have to check it out locally next week or so.

However, given that setting `GKSwstype` seemed to work in [this discourse post](https://discourse.julialang.org/t/being-in-julia-and-killing-julia-in-package-manager-stops-not-all-subprocesses-resulting-in-problems-later), I think this might just work.

src/backends/gr.jl
```diff
@@ -1958,13 +1958,13 @@ function _display(plt::Plot{GRBackend})
...
-        rm(filepath)
+        withenv("GKSwstype" => "pdf", "GKS_FILEPATH" => filepath) do
+            gr_display(plt)
+            GR.emergencyclosegks()
+            content = string("\033]1337;File=inline=1;preserveAspectRatio=0:", base64encode(open(read, filepath)), "\a")
+            println(content)
+            rm(filepath)
+        end
     else
         ENV["GKS_DOUBLE_BUF"] = true
```
BeastyBlacksmith, line 1969: shouldn't we use `withenv` here also?

asinghvi17:
It looks like this messes with the display, so shouldn't be merged now.  Will let you know when it's done...

daschw:
ok, thanks

isentropic:
Does this mean that updating ENV would no longer be necessary?

daschw:
I think this can be closed. Feel free to reopen if you disagree.

PR #2700, opened 2020-05-13
pgfplotsx colorbar improvements

daschw:
Allow to disable colorbar or set positions left, right, top and bottom. Furthermore, inherint colorbar ticklabel font from plot ticklabel font.

Before:
pgfplotsx_colorbar_old

Now: (difference is in the colorbar tick label font size)
pgfplotsx_colorbar_new

pgfplotsx_colorbar_bottom

### DIFF
--- a/src/backends/pgfplotsx.jl
+++ b/src/backends/pgfplotsx.jl
@@ -206,41 +206,42 @@ function (pgfx_plot::PGFPlotsXPlot)(plt::Plot{PGFPlotsXBackend})
             # As it is likely that all series within the same axis use the same
             # colormap this should not cause any problem.
             for series in series_list(sp)
-                    if hascolorbar(series)
-                        cg = get_colorgradient(series)
-                        cm = pgfx_colormap(get_colorgradient(series))
-                        PGFPlotsX.push_preamble!(
-                            pgfx_plot.the_plot,
-                            """\\pgfplotsset{
-                            colormap={plots$(sp.attr[:subplot_index])}{$cm},
-                            }""",
-                        )
+                if hascolorbar(series)
+                    cg = get_colorgradient(series)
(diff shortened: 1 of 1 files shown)


src/backends/pgfplotsx.jl
```diff
@@ -206,41 +206,52 @@ function (pgfx_plot::PGFPlotsXPlot)(plt::Plot{PGFPlotsXBackend})
...
-                ),
-                "point meta max" => get_clims(sp)[2],
-                "point meta min" => get_clims(sp)[1],
-            )
+                    "xticklabel style" => pgfx_get_ticklabel_style(sp, sp[:xaxis]),
+                    "yticklabel style" => pgfx_get_ticklabel_style(sp, sp[:yaxis]),
+                )
+                if sp[:colorbar] === :top
+                    push!(colorbar_style,
+                        "at" => string((0.5, 1.05)),
```
BeastyBlacksmith, line 242: I think I used `1.02` everywhere else in the code, so maybe use it here also or does it look ugly?

daschw: From my quick comparisons 1.05 looks rather like the margin for `colorbar = :right`

BeastyBlacksmith: also fine

BeastyBlacksmith: for reference in the manual is stated that `colorbar right` defines `/pgfplots/colorbar shift/.style={xshift=0.3cm},`

BeastyBlacksmith: anyway, that can wait, until there are problems with the way it is in this PR

daschw: Yes I also do not yet consider subplot titles for `colorbar = :top`. But I agree, this can be fixed in a separate PR.

Issue #2092, opened 2019-07-05
Scatter3d depthshade (pyplot backend)


CVerbist1:
In a 3D scatter in base PyPlot you have an option "depthshade" that can be turned on/off. When using Plots, and pyplot as a backend, depthshade seems to be on by default but I couldn't find a way to turn it off. Is this possible to turn off or a possibility to add it? 
![deptshade](https://user-images.githubusercontent.com/22769287/60716395-912da100-9f1f-11e9-9338-7b021947e138.png)

SKYnSPACE:
Me neither. I couldn't find the way to turn that off. Just tricked the plot3d with a dummy value located behind of the scene having marker size 0.

isentropic:
@BeastyBlacksmith  perhaps I could a relevant attribute like :pydepthshade?

BeastyBlacksmith:
Cleanest way would be, if we had https://github.com/JuliaPlots/Plots.jl/issues/2197.

isentropic:
perhaps:
`plot(x1, x2, backendkeyword="parameter=value")`
aliases:
`plot(x1, x2, bk="parameter=value")`  
`plot(x1, x2, backend_keyword="parameter=value")`

BeastyBlacksmith:
Since that is merged, you can now could get a value in the backend code via `get(series[:extra_kwargs], :depthshade, true)`

isentropic:
@CVerbist1 @SKYnSPACE let me know your feedback about this. Refer to https://github.com/JuliaPlots/Plots.jl/pull/2703 and let me know if this is what you wanted

PR #2703, opened 2020-05-14
python depthshade

isentropic:
Fix https://github.com/JuliaPlots/Plots.jl/issues/2092
with help of extrakeywords by @BeastyBlacksmith

### DIFF
--- a/src/backends/pyplot.jl
+++ b/src/backends/pyplot.jl
@@ -590,10 +590,10 @@ function py_add_series(plt::Plot{PyPlotBackend}, series::Series)
         else
             # do a normal scatter plot
 
+            # Add depthshade option for 3d plots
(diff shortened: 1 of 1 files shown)


isentropic:
MWE
```
p = scatter(rand(10), rand(10), rand(10))  # false by default (to match behavior of GR)

scatter!(p, rand(10), rand(10), rand(10), depthshade=true)
```

image

BeastyBlacksmith:
Even better would be, if you find the correct place where you just write `merge!(extrakw, series[:extra_kwargs])` and it works automagically

isentropic:
It is difficult for all possible extra kws, as for example depthshade is a 3d scatter exclusive, so I think this is the right place for it.
However for some other global extrakws, it would be very good

PR #2711, opened 2020-05-19
deprecate pgfplots

BeastyBlacksmith:
We removed it from the docs, I closed all issues. Maybe its time.


mkborregaard:
Should we also move the backend file to a deprecated folder?

PR #2718, opened 2020-05-20
improve title- and legendlocation

BeastyBlacksmith:


src/backends/pgfplotsx.jl
```diff
@@ -703,6 +695,14 @@ pgfx_get_colorbar_pos(s) =
     get((left = " left", bottom = " horizontal", top = " horizontal"), s, "")
 pgfx_get_colorbar_pos(b::Bool) = ""
 
+pgfx_get_title_pos(s) =
+    get((
+        left = ("at" => "{(0,1)}", "anchor" => "south west"),
+        right = ("at" => "{(1,1)}", "anchor" => "south east"),
+        ), s, "{(0.5,1)}")
+pgfx_get_title_pos(t::Tuple) = ("at" => "{$(string(t))}", "anchor" => "south")
+pgfx_get_title_pos(nt::NamedTuple) = ("at" => "{$(string(nt.at))}", "anchor" => string(nt.anchor))
```
korsbo, line 704: This looks good, could the same be done for annotations and legends?

BeastyBlacksmith: I added it for legends, annotations is a different story...

Issue #2704, opened 2020-05-14
[BUG] Plotly plot is black in Juno


judober:
When plotting in Juno with Plots, the Plotly as well as the plotlyjs backend produce only black figures. Gr works though.
## Details
I can't point my finger to it, but I think it started after upgrading plots from 1.2.0 but I'm not sure since going back didn't help.

### Backends

This bug occurs on ( insert `x` below )

Backend      | yes | no  | untested
-------------|-----|-----|---------
gr (default) |     |   x  |
pyplot       |     |     | x
plotly       |   x  |     |
plotlyjs     |    x |     |
pgfplotsx    |     |     |x
inspectdr    |     |     |x

### Versions

Plots.jl version: 1.2.5
Backend  version: Plotly v0.3.0
Output of `versioninfo()`:
Julia Version 1.4.1
Commit 381693d3df* (2020-04-14 17:20 UTC)
Platform Info:
  OS: macOS (x86_64-apple-darwin18.7.0)
  CPU: Intel(R) Core(TM) i5-3470S CPU @ 2.90GHz
  WORD_SIZE: 64
  LIBM: libopenlibm
  LLVM: libLLVM-8.0.1 (ORCJIT, ivybridge)
Environment:
  JULIA_NUM_THREADS = 4

BoundaryValueProblems:
@daschw and @pfitzseb ; I also have the same problem all over again after I updated Atom editor v1.47.0; julia-client v0.12.5; ink v0.12.4 while the Plots.jl is now v1.3.0. Unlike before I reported https://github.com/JuliaPlots/Plots.jl/issues/2462, which was resolved mid April, this time, issuing `gui()` does not work for the plotlyjs backend. Interestingly, for the plotly backend, after some plotting commands, if I issue `gui()`, a web browser page is created and plots are displayed, i.e., the plotly backends works if one issues `gui()`. But of course, this is not a normal behavior anyways. I would greatly appreciate it if you could take a look at this issue. It's very bad if this happens each time when Atom is updated...

BeastyBlacksmith:
FWIW, I can reproduce

hhaensel:
It is an atom problem, also the atom-html-preview is affected :-(

hhaensel:
wait a moment before merging, I forgot to commit a correction ...

BeastyBlacksmith:
> wait a moment before merging, I forgot to commit a correction ...

[...]

PR #2719, opened 2020-05-20
Add support for mathjax (latex) and extra_plot_kwargs in plotly backend

hhaensel:
Currently, the plotly backend does not support latexstrings in the Juno IDE, whereas it works in IJulia, if the correct fonts are copied into the mathjax folder.

I propose to add an `:include_mathjax ` plot attribute that works similar to the python version:

- `:include_mathjax == ""` (default): no mathjax header)
- `:include_mathjax == "cdn"` include the standard online version of the header
- `:include_mathjax == "`<filename?config=xyz>"` include a user-defined file

Unfortunately, the x-axis fails to display properly with mathjax enabled. This can be worked around by removing :domain from the attributes and adding :automargin for :yaxis. That also helps in the case of large number at the y-axis.

I have therefore also implemented `extra_plot_kwargs` for the plotly backend.

Here are three example plots that demonstrate the usage:
```
using Plots, LaTeXStrings
plotly()

# if Conda and IJulia are installed, then the mathjax files are locally available
# one might think of adding an option "local" that includes the following line ...
import Conda.ROOTENV
function local_mathjax()
    joinpath(ROOTENV, "Lib", "site-packages", "notebook", "static", "components", "MathJax",
            "MathJax.js?config=TeX-AMS-MML_HTMLorMML-full")
end

p = Plots.plot(1:4, [[1,4,9,16]*10000, [0.5, 2, 4.5, 8]],
    labels = [L"\alpha_{1c} = 352 \pm 11 \text{ km s}^{-1}";
              L"\beta_{1c} = 25 \pm 11 \text{ km s}^{-1}"] |> permutedims,
    xlabel = L"\sqrt{(n_\text{c}(t|{T_\text{early}}))}",
    ylabel = L"d, r \text{ (solar radius)}",
    yformatter = :plain
)

p = Plots.plot(1:4, [[1,4,9,16]*10000, [0.5, 2, 4.5, 8]],
    include_mathjax = local_mathjax(),
    labels = [L"\alpha_{1c} = 352 \pm 11 \text{ km s}^{-1}";
              L"\beta_{1c} = 25 \pm 11 \text{ km s}^{-1}"] |> permutedims,
    xlabel = L"\sqrt{(n_\text{c}(t|{T_\text{early}}))}",
    ylabel = L"d, r \text{ (solar radius)}",
    yformatter = :plain
)

p = Plots.plot(1:4, [[1,4,9,16]*10000, [0.5, 2, 4.5, 8]],
    include_mathjax = local_mathjax(),
    labels = [L"\alpha_{1c} = 352 \pm 11 \text{ km s}^{-1}";
              L"\beta_{1c} = 25 \pm 11 \text{ km s}^{-1}"] |> permutedims,
    xlabel = L"\sqrt{(n_\text{c}(t|{T_\text{early}}))}",
    ylabel = L"d, r \text{ (solar radius)}",
    yformatter = :plain,
    extra_plot_kwargs = KW(:yaxis => KW(:automargin => true), :xaxis => KW(:domain => "auto"))
)
```
mathjax_no_header
mathjax_with_header
mathjax_with_header_and_automargin

### DIFF
--- a/src/backends.jl
+++ b/src/backends.jl
@@ -374,6 +374,7 @@ function _initialize_backend(pkg::PlotlyBackend)
     catch
         @info "For saving to png with the Plotly backend ORCA has to be installed."
     end
+    push!(_plot_defaults, :include_mathjax => "")
 end
 
 const _plotly_attr = merge_with_base_supported([
@@ -417,6 +418,7 @@ const _plotly_attr = merge_with_base_supported([
     :tick_direction,
     :camera,
     :contour_labels,
+    :include_mathjax,
   ])
 
 const _plotly_seriestype = [
--- a/src/backends.jl
+++ b/src/backends.jl
@@ -374,7 +374,6 @@ function _initialize_backend(pkg::PlotlyBackend)
     catch
         @info "For saving to png with the Plotly backend ORCA has to be installed."
     end
-    push!(_plot_defaults, :include_mathjax => "")
 end
 
 const _plotly_attr = merge_with_base_supported([
@@ -418,7 +417,6 @@ const _plotly_attr = merge_with_base_supported([
     :tick_direction,
     :camera,
     :contour_labels,
-    :include_mathjax,
   ])
 
 const _plotly_seriestype = [
--- a/src/backends/plotly.jl
+++ b/src/backends/plotly.jl
@@ -328,7 +328,7 @@ function plotly_layout(plt::Plot)
             plotattributes_out[:direction] = "counterclockwise"
         end
 
-        plotattributes_out
+        plotattributes_out = recursive_merge(plotattributes_out, sp.attr[:extra_kwargs])
     end
 
     # turn off hover if nothing's using it
--- a/src/backends/plotly.jl
+++ b/src/backends/plotly.jl
@@ -328,7 +328,7 @@ function plotly_layout(plt::Plot)
             plotattributes_out[:direction] = "counterclockwise"
         end
 
-        plotattributes_out = recursive_merge(plotattributes_out, sp.attr[:extra_kwargs])
+        plotattributes_out
     end
 
     # turn off hover if nothing's using it
--- a/src/backends/plotly.jl
+++ b/src/backends/plotly.jl
@@ -810,7 +810,7 @@ html_body(plt::Plot{PlotlyBackend}) = plotly_html_body(plt)
 const ijulia_initialized = Ref(false)
 
 function plotly_html_head(plt::Plot)
-    local_file = ("file://" * plotly_local_file_path)
+    local_file = ("file:///" * plotly_local_file_path)
     plotly =
         use_local_dependencies[] ? local_file : "https://cdn.plot.ly/plotly-latest.min.js"
 
--- a/src/backends/plotly.jl
+++ b/src/backends/plotly.jl
@@ -814,9 +814,9 @@ function plotly_html_head(plt::Plot)
     plotly =
         use_local_dependencies[] ? local_file : "https://cdn.plot.ly/plotly-latest.min.js"
 
-    include_mathjax = get(plt.attr, :include_mathjax, "")
+    include_mathjax = get(plt[:extra_plot_kwargs], :include_mathjax, "")
     mathjax_file = include_mathjax != "cdn" ? ("file://" * include_mathjax) : "https://cdnjs.cloudflare.com/ajax/libs/mathjax/2.7.7/MathJax.js?config=TeX-MML-AM_CHTML"
-    mathjax_head = include_mathjax == "" ? "" : "<script async src=\"$mathjax_file\"></script>\n\t\t"
+    mathjax_head = include_mathjax == "" ? "" : "<script src=\"$mathjax_file\"></script>\n\t\t"
 
     if isijulia() && !ijulia_initialized[]
         # using requirejs seems to be key to load a js depency in IJulia!
--- a/src/backends/plotly.jl
+++ b/src/backends/plotly.jl
@@ -813,6 +813,11 @@ function plotly_html_head(plt::Plot)
     local_file = ("file://" * plotly_local_file_path)
     plotly =
         use_local_dependencies[] ? local_file : "https://cdn.plot.ly/plotly-latest.min.js"
+
+    include_mathjax = get(plt.attr, :include_mathjax, "")
+    mathjax_file = include_mathjax != "cdn" ? ("file://" * include_mathjax) : "https://cdnjs.cloudflare.com/ajax/libs/mathjax/2.7.7/MathJax.js?config=TeX-MML-AM_CHTML"
+    mathjax_head = include_mathjax == "" ? "" : "<script async src=\"$mathjax_file\"></script>\n\t\t"
+
     if isijulia() && !ijulia_initialized[]
         # using requirejs seems to be key to load a js depency in IJulia!
         # https://requirejs.org/docs/start.html
@@ -826,7 +831,7 @@ function plotly_html_head(plt::Plot)
         """)
         ijulia_initialized[] = true
     end
-    return "<script src=$(repr(plotly))></script>"
+    return "$mathjax_head<script src=$(repr(plotly))></script>"
 end
 
 function plotly_html_body(plt, style = nothing)
--- a/src/backends/plotly.jl
+++ b/src/backends/plotly.jl
@@ -336,7 +336,7 @@ function plotly_layout(plt::Plot)
         plotattributes_out[:hovermode] = "none"
     end
 
-    plotattributes_out
+    plotattributes_out = recursive_merge(plotattributes_out, plt.attr[:extra_plot_kwargs])
 end
 
 function plotly_layout_json(plt::Plot)
--- a/src/utils.jl
+++ b/src/utils.jl
@@ -302,6 +302,11 @@ Base.convert(::Type{Vector{T}}, rng::AbstractRange{S}) where {T<:Real,S<:Real} =
 
 Base.merge(a::AbstractVector, b::AbstractVector) = sort(unique(vcat(a,b)))
 
+# recursively merge kw-dicts, e.g. for merging extra_kwargs / extra_plot_kwargs in plotly)
+recursive_merge(x::AbstractDict...) = merge(recursive_merge, x...)
+# if values are not AbstractDicts, take the last definition (as does merge)
+recursive_merge(x...) = x[end]
+
 nanpush!(a::AbstractVector, b) = (push!(a, NaN); push!(a, b))
 nanappend!(a::AbstractVector, b) = (push!(a, NaN); append!(a, b))
 


src/backends.jl
```diff
@@ -374,6 +374,7 @@ function _initialize_backend(pkg::PlotlyBackend)
     catch
         @info "For saving to png with the Plotly backend ORCA has to be installed."
     end
+    push!(_plot_defaults, :include_mathjax => "")
```
BeastyBlacksmith, line 377: This should be handled differently. This line needs to be removed.

src/backends.jl
```diff
@@ -417,6 +418,7 @@ const _plotly_attr = merge_with_base_supported([
     :tick_direction,
     :camera,
     :contour_labels,
+    :include_mathjax,
```
BeastyBlacksmith, line 421: This also

src/backends/plotly.jl
```diff
@@ -810,9 +810,14 @@ html_body(plt::Plot{PlotlyBackend}) = plotly_html_body(plt)
 const ijulia_initialized = Ref(false)
 
 function plotly_html_head(plt::Plot)
-    local_file = ("file://" * plotly_local_file_path)
+    local_file = ("file:///" * plotly_local_file_path)
     plotly =
         use_local_dependencies[] ? local_file : "https://cdn.plot.ly/plotly-latest.min.js"
+
+    include_mathjax = get(plt.attr, :include_mathjax, "")
```
BeastyBlacksmith, line 817: this should be `get(plt[:extra_plot_kwargs], :include_mathjax, "")`
Call site would be `plot(..., include_mathjax = "cdn", extra_kwargs = :plot)`.

src/utils.jl
```diff
@@ -302,6 +302,11 @@ Base.convert(::Type{Vector{T}}, rng::AbstractRange{S}) where {T<:Real,S<:Real} =
 
 Base.merge(a::AbstractVector, b::AbstractVector) = sort(unique(vcat(a,b)))
 
+# recursively merge kw-dicts, e.g. for merging extra_kwargs / extra_plot_kwargs in plotly)
+recursive_merge(x::AbstractDict...) = merge(recursive_merge, x...)
```
BeastyBlacksmith, line 306: Is this neccessary? Usually both dicts are only one level deep, or?

BeastyBlacksmith:
For the axis attributes: If I understood correctly we want `KW(:yaxis => KW(:automargin => true), :xaxis => KW(:domain => "auto")` always when `include_mathjax` is set, right?
Maybe even have it generally? ( I didn't use the plotly backend )
I would suggest to check for `include_mathjax` when the axis dictionaries are processed and add those then.

src/backends.jl
hhaensel: Done

src/backends.jl
hhaensel: Done

src/backends/plotly.jl
hhaensel: Done

src/utils.jl
hhaensel: I think we need it, because we want to merge the properties of respective axes. Just using `merge` on `plotattributes_out` would empty the rest of the arguments, which have been populated during `plotly_layout()`.

Maybe, I am wrong?

hhaensel:
> For the axis attributes: If I understood correctly we want `KW(:yaxis => KW(:automargin => true), :xaxis => KW(:domain => "auto")` always when `include_mathjax` is set, right?
> Maybe even have it generally? ( I didn't use the plotly backend )
> I would suggest to check for `include_mathjax` when the axis dictionaries are processed and add those then.

Not necessarily, it goes wrong, when the legend contains latex strings. And setting automargin to true puts the plot slightly in a different position, which might be annoying if you want to create animations ...

hhaensel:
... or fix the problem with wrong positioning of the grid...

hhaensel:
when I test
```p = plot(rand(10),
    xlabel = "\$\\alpha \\, F_\\mathrm{test}\$",
    ylabel = L"\Gamma^2", include_mathjax = local_mathjax(), 
    yaxis = KW(:automargin => true),
    extra_kwargs = :plot)
```
I receive a warning

```
┌ Warning: Skipped yaxis arg Dict{Symbol,Any}(:automargin => true)
└ @ Plots C:\Users\m136270\.julia\dev\Plots\src\axes.jl:74
```

BeastyBlacksmith:
That is, because `yaxis` is already a Plots plotattribute, so you can't use it for extra_kwargs.
You'd need to call `plot(..., extra_kwargs = Dict( :subplot => Dict( :yaxis => ... )) )`

src/utils.jl
BeastyBlacksmith: How does this work out with multiple subplots? Something like
```julia
Plots.plot(rand(4,4), [[1,4,9,16]*10000, [0.5, 2, 4.5, 8], rand(4), rand(4)],
    layout = 4,
    include_mathjax = local_mathjax(),
    extra_plot_kwargs = KW(:yaxis => KW(:automargin => true), :xaxis => KW(:domain => "auto"))
)
```

BeastyBlacksmith:
To explain a bit:
There are three entrypoints for `extra_kwargs` at the moment (toggled, by the `extra_kwargs` keyword)
 - `:plot` plot wide extra_kwargs stored in `plt[:extra_kwargs]`
 - `:subplot` extra_kwargs that count per subplot stored in `plt.subplots[index][:extra_kwargs]`
 - `:series` (default) extra_kwargs that count per series stored in `plt.series_list[index][:extra_kwargs]`
Keywords that are not already attributes by Plots get then collected in the corresponding container. However, if you want to pass mixed attributes ( or ones that are already Plots plotattributes) you can pass a dict to `extra_kwargs`, like `Dict( :plot => Dict(...), :subplot => Dict(...), :series => Dict(...))`, thats a bit unwieldly and should only be needed, if the user want to make some rarely used tweaks.

BeastyBlacksmith:
I would suggest an API in the spirit of 
```julia
plot( ..., automargin = (false, true), # automargin for x/yaxis
            , domain = (xvalue, yvalue),
            , extra_kwargs = :subplot
)
```
for setting these attributes

hhaensel:
Thanks for the explanation, that makes things clearer :-)
For normal plots without subplots one can also do 

```
p = Plots.plot(1:4, [[1,4,9,16]*10000, [0.5, 2, 4.5, 8]],
    labels = [L"\alpha_{1c} = 352 \pm 11 \text{ km s}^{-1}";
              L"\beta_{1c} = 25 \pm 11 \text{ km s}^{-1}"] |> permutedims,
    xlabel = L"\sqrt{(n_\text{c}(t|{T_\text{early}}))}",
    ylabel = L"d, r \text{ (solar radius)}",
    yformatter = :plain,
    extra_plot_kwargs = KW(
        :include_mathjax => local_mathjax(), 
        :yaxis => KW(:automargin => true),
        :xaxis => KW(:domain => "auto")
        ),
)
```
which would be more readable than
```
p = Plots.plot(1:4, [[1,4,9,16]*10000, [0.5, 2, 4.5, 8]],
    labels = [L"\alpha_{1c} = 352 \pm 11 \text{ km s}^{-1}";
              L"\beta_{1c} = 25 \pm 11 \text{ km s}^{-1}"] |> permutedims,
    xlabel = L"\sqrt{(n_\text{c}(t|{T_\text{early}}))}",
    ylabel = L"d, r \text{ (solar radius)}",
    yformatter = :plain,
    extra_kwargs = KW(:plot => KW(:include_mathjax => local_mathjax()),
                      :subplot => KW(:yaxis => KW(:automargin => true), :xaxis => KW(:domain => "auto"))
                      ),
)
```
Is this recommended usage?

BeastyBlacksmith:
That is fine, I am still not convinced, that `automargin` and `domain` should be plot attributes rather than subplot attributes. So maybe
```julia
p = Plots.plot(1:4, [[1,4,9,16]*10000, [0.5, 2, 4.5, 8]],
    labels = [L"\alpha_{1c} = 352 \pm 11 \text{ km s}^{-1}";
              L"\beta_{1c} = 25 \pm 11 \text{ km s}^{-1}"] |> permutedims,
    xlabel = L"\sqrt{(n_\text{c}(t|{T_\text{early}}))}",
    ylabel = L"d, r \text{ (solar radius)}",
    yformatter = :plain,
    extra_plot_kwargs = KW(
        :include_mathjax => local_mathjax(), 
        ),
   automargin = ..., 
   domain = ...,
   extra_kwargs = :subplot
)
```
could work

hhaensel:
I agree. Thanks for your support.

I currently don't have the time to spend on the API, should we merge nevertheless?

BeastyBlacksmith:
I can't test this right now, because of https://github.com/JuliaPlots/Plots.jl/issues/2704.. Can you show me what the output of
```julia
Plots.plot(rand(4,4), [[1,4,9,16]*10000, [0.5, 2, 4.5, 8], rand(4), rand(4)],
           layout = 4,
           include_mathjax = "cdn",
           xlabel = L"\sqrt{(n_\text{c}(t|{T_\text{early}}))}",
           ylabel = L"d, r \text{ (solar radius)}",
       )
```
looks like?

BeastyBlacksmith:
and the version with `automargin` and `domain`?

hhaensel:
you could test with atom 1.46 or disable the plot-pane

BeastyBlacksmith:
We can of course split the PR to merge the mathjax keyword for now and spare the rest for later

BeastyBlacksmith:
Okay, disabling plot pane worked.
```julia
p = Plots.plot(rand(4,4), [[1,4,9,16]*10000, [0.5, 2, 4.5, 8], rand(4), rand(4)],
           labels = [L"\alpha_{1c} = 352 \pm 11 \text{ km s}^{-1}";
                     L"\beta_{1c} = 25 \pm 11 \text{ km s}^{-1}"] |> permutedims,
           xlabel = L"\sqrt{(n_\text{c}(t|{T_\text{early}}))}",
           ylabel = L"d, r \text{ (solar radius)}",
           yformatter = :plain,
           layout = 4,
           extra_plot_kwargs = KW(
               :include_mathjax => "cdn", 
               :yaxis => KW(:automargin => true),
               :xaxis => KW(:domain => "auto")
               ),
       )
```
This is the reason I am hesitant to merge this as is. Setting `automargin` and `domain` should count for all subplots in the specified `plot` call

hhaensel:
With the latest commit it is a bit better, but still not perfect. automargin seems to do strange things, when there is more than one plot in a row. And also the vertical spacing is not correct :-(

```
p = Plots.plot(rand(4,4), [[1,4,9,16]*10000, [0.5, 2, 4.5, 8], rand(4), rand(4)],
    labels = [L"\alpha_{1c} = 352 \pm 11 \text{ km s}^{-1}";
             L"\beta_{1c} = 25 \pm 11 \text{ km s}^{-1}"] |> permutedims,
    xlabel = L"\sqrt{(n_\text{c}(t|{T_\text{early}}))}",
    ylabel = L"d, r \text{ (solar radius)}",
    yformatter = :plain,
    layout = 4,
    extra_plot_kwargs = KW(:include_mathjax => "cdn"),
    yaxis = KW(:automargin => true),
    xaxis = KW(:domain => "auto"),
    extra_kwargs = :subplot,
)
```
newplot

hhaensel:
But the result without extra_kwargs is also not convincing ...
newplot
Anyhow, it seems, as if plotly always uses automargin=true for latex plotting, which then corrupts the multiplot pages. I've seen this a while ago for 3D plots.

BeastyBlacksmith:
Yeah, the layouting is suboptimal either way. I think, we can merge plot-wide extra_kwargs into the output dictionary, but it doesn't make much sense for the subplots extra_kwargs, since that is not how it currently works in this backend. You can call `Plots.plotly_layout(plt)` and you will see some `yaxis1-4` keys instead of `yaxis`.
I am going to merge this now and further improvements can come later.
Thanks again, for doing this PR :)

BeastyBlacksmith:
Adding the documentation label as a reminder to add this to a LaTeX-Workflow section in the docs

hhaensel:
It's been a pleasure, thanks for merging :-)

Issue #2198, opened 2019-09-19
Increasing dpi moves colorbar out of position for PyPlot backend


jacobcook1995:
Hi, came across some slightly strange behaviour whilst making heatmap plots. Essentially when I increase the dpi using the GR backend e.g.

```julia
gr(dpi=200)
x = collect(1.0:10.0)
y = collect(1.0:10.0)
z = rand(10,10)
heatmap(x,y,z)
```

I get the expected result (a normal heatmap with double the pixels per inch)

![test2](https://user-images.githubusercontent.com/32497818/65255691-3e16c380-daf6-11e9-98f2-fb10b5d102ea.png)

However when I run identical code using the PyPlot backend,

```julia
pyplot(dpi=200)
x = collect(1.0:10.0)
y = collect(1.0:10.0)
z = rand(10,10)
heatmap(x,y,z)
```

The colorbar of the heatmap has now been shifted into the plotting area.

![test2](https://user-images.githubusercontent.com/32497818/65255845-82a25f00-daf6-11e9-9553-c92482be4e6b.png)

I'm guessing that this isn't intended behaviour but I have no idea what is causing this.

As I'm not trying to produce particularly high quality plots at the moment there's an obvious work round of just using the automatic dpi value, but I was wondering if there might be a more sophisticated work around?

daschw:
Thanks for reporting! That's not the intended behaviour.

jacobcook1995:
Managed to get the figure I wanted in the end by increasing the right margin of the plot, i.e. this

```julia
    pyplot(dpi=200)
    x = collect(1.0:10.0)
    y = collect(1.0:10.0)
    z = rand(10,10)
    heatmap(x,y,z,right_margin=20*mm)
```

produced this

![test](https://user-images.githubusercontent.com/32497818/65617553-dad4d780-dfb4-11e9-80a9-cbabb3ca71a0.png)

Issue #1755, opened 2018-09-19
Colorbar placement


briochemc:
MWE:
```julia
function peaks(x, y)
    z = 3 * (1 - x)^2 * exp(-x^2 - (y + 1)^2)
    z += -10 * (x / 5 - x^3 - y^5) * exp(-x^2 - y^2)
    z += -1/3 * exp(-(x+1)^2 - y^2)
    return z
end

x = range(-3, stop = 3, length = 100)
y = range(-2, stop = 2, length = 80)
levels = collect(-4:0.25:4)
vmin = minimum(levels)
vmax = maximum(levels)

using Plots
pyplot() # Set the backend to PyPlot
plt = contourf(x, y, peaks, levels = levels, clim = (vmin, vmax),
              colorbar_title = "\$\\phi\$",
              colorbar = :right,
              xlabel = "\$x\$",
              ylabel = "\$y\$")
savefig(plt, "test.png")
```
produces the following:

![test](https://user-images.githubusercontent.com/4486578/45737177-6e502b80-bc30-11e8-93c3-d982d6d0ebe9.png)

It seems the colorbar is centered with [the axes + the labels], instead of being aligned (and thus centered) with the axes only. Is it possible to align it with the axes of the plot only like in the example below?

![example](https://i.stack.imgur.com/1CRZW.png)

lnacquaroli:
This is related to `pyplot` as `gr` produces your desired result.

briochemc:
OK I will move the issue to [PyPlot.jl](https://github.com/JuliaPy/PyPlot.jl) 🙂

briochemc:
Actually, `using PyPlot` works fine, so this is related to `using Plots` (reopening).

lnacquaroli:
Yes, I was talking about the pyplot backend inside Plots, not the PyPlot itself. Sorry for the short misleading message before.

briochemc:
I mean I don't mind using `PyPlot` directly but I thought I should point out that colorbar issue here... Is there a place to raise issues for backends rather than packages? (Me still a Julia noob)

isentropic:
@briochemc would you please check out https://github.com/JuliaPlots/Plots.jl/pull/2722 and tell me your review

Issue #484, opened 2016-09-13
Keep colorbar next to contour plot when aspect ratio is changed in PyPlot


stillyslalom:
When the aspect ratio is changed so the plot area becomes laterally compressed, the colorbar stays on the far right side of the figure, leaving extra whitespace between the plot area and the colorbar. Setting the size manually (`size = (500,500)`) puts the colorbar in the correct position, but it is larger than it should be.

```
using Plots; pyplot()
data = rand(10,10)
p = contour(data, aspect_ratio = 1)
```

PR #2722, opened 2020-05-21
pyplot colorbar fixes

isentropic:
Fix: https://github.com/JuliaPlots/Plots.jl/issues/1755
Fix: https://github.com/JuliaPlots/Plots.jl/issues/2198
Fix: https://github.com/JuliaPlots/Plots.jl/issues/484
Some improvements to pyplot colorbars, because colorbars deserve some <3

There were multiple issues with them

```
plot(rand(10), line_z=1e10.*rand(10), title="asd", titlefont=50)
```
Before:
image

After:
image

However `dpi` option still very buggy
https://github.com/JuliaPlots/Plots.jl/issues/2198

I cannot figure out why this happens, perhaps @daschw could help out?

### DIFF
--- a/src/backends/pyplot.jl
+++ b/src/backends/pyplot.jl
@@ -1007,9 +1007,9 @@ function _before_layout_calcs(plt::Plot{PyPlotBackend})
             # note: the colorbar axis is positioned independently from the subplot axis
             fig = plt.o
 
-            #  There is no point specifying colorbar size, hence rand(4) as it will get
+            #  There is no point in specifying colorbar size, hence zeros(4). It will get
             #  updated at function _update_plot_object(plt::Plot{PyPlotBackend})
-            cbax = fig."add_axes"(rand(4), label = string(gensym()))
+            cbax = fig."add_axes"(zeros(4), label = string(gensym()))
             cb = fig."colorbar"(handle; cax = cbax, kw...)
             cb."set_label"(sp[:colorbar_title],size=py_thickness_scale(plt, sp[:yaxis][:guidefontsize]),family=sp[:yaxis][:guidefontfamily], color = py_color(sp[:yaxis][:guidefontcolor]))
 
--- a/src/backends/pyplot.jl
+++ b/src/backends/pyplot.jl
@@ -1006,9 +1006,17 @@ function _before_layout_calcs(plt::Plot{PyPlotBackend})
             # create and store the colorbar object (handle) and the axis that it is drawn on.
             # note: the colorbar axis is positioned independently from the subplot axis
             fig = plt.o
-            cbax = fig."add_axes"([0.8,0.1,0.03,0.8], label = string(gensym()))
+
+            #  There is no point specifying colorbar size, hence rand(4) as it will get
+            #  updated at function _update_plot_object(plt::Plot{PyPlotBackend})
+            cbax = fig."add_axes"(rand(4), label = string(gensym()))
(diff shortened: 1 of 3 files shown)


src/backends/pyplot.jl
```diff
@@ -1384,10 +1392,14 @@ function _update_plot_object(plt::Plot{PyPlotBackend})
...
         if haskey(sp.attr, :cbar_ax)
             cbw = sp.attr[:cbar_width]
-            # this is the bounding box of just the colors of the colorbar (not labels)
-            ex = sp[:zaxis][:extrema]
-            has_toplabel = !(1e-7 < max(abs(ex.emax), abs(ex.emin)) < 1e7)
-            cb_bbox = BoundingBox(right(sp.bbox)-cbw+1mm, top(sp.bbox) +  (has_toplabel ? 4mm : 2mm), _cbar_width-1mm, height(sp.bbox) - (has_toplabel ? 6mm : 4mm))
+
+            # Colorbar should be placed right to the axis with same vertical size
+            # we can extract bbox from axis object to make it match to it
+            cb_bbox = BoundingBox(right(py_bbox(ax)) + _cbar_width/2,  # Some paddding between colorbar and the ax
```
isentropic, line 1398: For some reason, `py_bbox(ax)` no longer refers to the actual bounding box of the axis with increasing dpi. I would assume that `py_bbox(ax)` always shows the bounding box of the axes (axis frames)

src/backends/pyplot.jl
```diff
@@ -1006,9 +1006,17 @@ function _before_layout_calcs(plt::Plot{PyPlotBackend})
             # create and store the colorbar object (handle) and the axis that it is drawn on.
             # note: the colorbar axis is positioned independently from the subplot axis
             fig = plt.o
-            cbax = fig."add_axes"([0.8,0.1,0.03,0.8], label = string(gensym()))
+
+            #  There is no point specifying colorbar size, hence rand(4) as it will get
+            #  updated at function _update_plot_object(plt::Plot{PyPlotBackend})
+            cbax = fig."add_axes"(rand(4), label = string(gensym()))
```
daschw, line 1012: Wouldn't `zeros` or `ones` be a better choice than `rand` then?

isentropic: Right, perhaps I should use zeros(4) as it is cognitively easier than rand(4)

isentropic:
@daschw I reworked some parts of the colorbars. Now a lot of things work better. Notably:

https://github.com/JuliaPlots/Plots.jl/issues/1755
https://github.com/JuliaPlots/Plots.jl/issues/2198
https://github.com/JuliaPlots/Plots.jl/issues/484

as well http://docs.juliaplots.org/latest/generated/pyplot/#pyplot-ref32-1

I found that this new behavior would serve as much better default
Answers found in https://stackoverflow.com/questions/18195758/set-matplotlib-colorbar-size-to-match-graph

src/utils.jl
```diff
@@ -983,7 +983,7 @@ px2inch(px::Real)       = float(px / PX_PER_INCH)
 inch2mm(inches::Real)   = float(inches * MM_PER_INCH)
 mm2inch(mm::Real)       = float(mm / MM_PER_INCH)
 px2mm(px::Real)         = float(px * MM_PER_PX)
```
isentropic, line 985: For some reason, I could not get these function to work. There was a type error. I ended up not using these anyways

PR #2732, opened 2020-05-26
Add fillalpha to :contourf seriestype for pyplot backend

Cornelius-G:
Adds the option to pass `fillalpha` keyword to `pyplot` backend when using `:contourf` seriestype.
Example:
```
plot(..., seriestype=:contourf, fillalpha=0.3, linewidth=2)
```

plot

(The surrounding line can be changed or removed using the `linewidth` keyword.)


BeastyBlacksmith:
Thanks!

PR #2734, opened 2020-05-26
Remove explicit Revise.track from atsign-require blocks

timholy:
Since Revise 2.4 and Requires 1.0, `@require` blocks have no longer needed an explicit call to `Revise.track`, as Revise and Requires talk to each other, and Revise discovers `include` statements and automatically adds the file to the track list. Indeed, as of Revise 2.7 (released this morning), doing so will trigger a depwarn. Until recently, under certain circumstances Plots could trigger a warning (https://github.com/timholy/Revise.jl/issues/431) that revealed a deeper race condition within Revise that required a redesign of certain pieces (https://github.com/timholy/Revise.jl/pull/491).

However, there's one exception to the depwarn policy: knowing that I would be submitting this PR, I didn't want the nice developers of Plots to be routinely annoyed by reports of deprecations from people using a new version of Revise but an older version of Plots. So I baked an exception into Revise's code: https://github.com/timholy/Revise.jl/blob/da6dc22590cd6ff6bcf5bbf6a5b3eb7a1dec2474/src/Revise.jl#L867. After all, I was the one who advised you to add a manual `Revise.track` in the first place, so it hardly seemed fair to annoy you.


timholy:
I should make one comment though: for this to work properly you need Requires 1.0. Neither Revise nor Requires depends on the other, so the only way to impose this requirement is in packages that use Requires. Shall I bump you to Requires 1? This issue only affects developers of Plots, so it's a bit of a "soft" dependency. Not quite sure what to do here.

BeastyBlacksmith:
Thanks a lot! Am I right in saying, that upon merging this we should drop Require v0.5 from compat?

BeastyBlacksmith:
Bumping is much appreciated.

timholy:
Done. I didn't want to add it along with the first since it's a `[compat]` requirement that only affects developers of Plots, not users of Plots. But as long as you're good with the stricter requirements, it will be much less baffling to developers who happen to be stuck on older versions.

timholy:
As usual, let me know if Revise doesn't work properly for you after this! It should be solid and even has a test named "LikePlots," but it's a complex piece of machinery and regressions are always possible. I want to hear about them.

daschw:
Will do, thanks!

PR #2739, opened 2020-05-27
fix type restriction in process_clims

mileslucas:
Noticed a bug in the code I added here (classic!)

Previously if you used mixed types in the tuple call it would fail eg.
`process_clims((1, 3.0))` would produce a `MethodError`

This loosens the type restriction (should have been this way from the beginning!).


daschw:
Nice, thanks!

PR #2751, opened 2020-06-03
plotly legend fixes

isentropic:
Fixes for plotly legends. Now works consistently with other backends:
```
plot(rand(10), legend=:left)

plot(rand(10), legend=:topright)

plot(rand(10), legend=(1.02, 1.0))
```

image

image
image


isentropic:
perhaps should fix https://github.com/JuliaPlots/Plots.jl/issues/574

Issue #2730, opened 2020-05-26
[BUG] spy only working for sparse matrices


jamblejoe:
## Details

Is the following intendend behaviour?
```
julia> spy(sparse([1 2; 3 4]))

julia> spy([1 2; 3 4])
ERROR: MethodError: no method matching findnz(::Array{Float64,2})
Closest candidates are:
  findnz(::SparseArrays.AbstractSparseMatrixCSC{Tv,Ti}) where {Tv, Ti} at D:\buildbot\worker\package_win64\build\usr\share\julia\stdlib\v1.4\SparseArrays\src\sparsematrix.jl:1423
  findnz(::SparseVector{Tv,Ti}) where {Tv, Ti} at D:\buildbot\worker\package_win64\build\usr\share\julia\stdlib\v1.4\SparseArrays\src\sparsevector.jl:737
```

### Backends

This bug occurs on ( insert `x` below )

Backend      | yes | no  | untested
-------------|-----|-----|---------
gr (default) |  x  |     |
pyplot       |  x  |     |
plotly       |     |     |
plotlyjs     |     |     |
pgfplotsx    |     |     |
inspectdr    |     |     |

### Versions

Plots.jl version: v1.2.2

BeastyBlacksmith:
The fix would be to add a method to `findnz` for dense matrices (without type piracy).

jamblejoe:
But isn't this function defined in `SparseArrays`? Should another method `findnz(::AbstractArray)` be defined in `Plots.jl`?

BeastyBlacksmith:
Thats right, that function needs two methods:
```julia
Plots.findnz(a::SparseArray) = SparseArrays.findnz(a)
Plots.findnz(a::AbstractArray) = ...
```

jamblejoe:
Sorry, to bother you. I quickly read the Julia style guide on type piracy. Does your code snippet already avoid type piracy? 

And how would you make `findnz` visible? By 
```
import SparseArrays.findnz
```
?

BeastyBlacksmith:
> Sorry, to bother you. I quickly read the Julia style guide on type piracy. Does your code snippet already avoid type piracy?
> 

Type piracy would be if I added a `SparseArrays.findnz(a::AbstractArray)`.
> And how would you make `findnz` visible? By
> 
> ```
> import SparseArrays.findnz
> ```
> 
> ?
[...]

PR #2755, opened 2020-06-03
Plots.findnz for sparse and non-sparse matrices

jamblejoe:
fixes https://github.com/JuliaPlots/Plots.jl/issues/2730

### DIFF
--- a/src/Plots.jl
+++ b/src/Plots.jl
@@ -10,7 +10,7 @@ using Reexport
 
 import GeometryTypes
 using Dates, Printf, Statistics, Base64, LinearAlgebra, Random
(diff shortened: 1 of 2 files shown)


BeastyBlacksmith:
That is working fine or do you still get errors locally?

jamblejoe:
No errors. The problem was that `SparseArray` does not exist. The correct abstract supertype is `AbstractSparseMatrix`. I changed that and now it works fine. `Plots.findnz` is restricted to 2-dimensional Arrays now.

BeastyBlacksmith:
Good work, thank you !

Issue #2207, opened 2019-10-03
legend = :outerright is missing from the documentation


feanor12:
The GR backend has implemented an :outer legend position, which is not reflected in the documentation.

tzxiong:
Is there a horizontal alignment of legend entries rather than the default vertical alignment? I want to put it horizontally in the ":outertop" position.

feanor12:
From a quick glimpse at the gr.jl backend file I would say there should be the options: :outertop, : outerbottom, :outerleft and :outerright.

Issue #2377, opened 2020-02-02
legend=:right does not show up in some plots in multiple plot layout


leandromartinez98:
```
x = rand(10); y = rand(10);
plot(layout=(2,1))
plot!(x,y,legend=:right,subplot=1)
plot!(x,y,legend=:right,subplot=2)
```

Results in:

![image](https://user-images.githubusercontent.com/31046348/73610075-7bb24f80-45b2-11ea-89de-3aabdf9fd3af.png)

That is, the legend does not appear in the first plot.

The same problem occurs with `:left`, but not with the other options, as `:bottomright`, `:bottomleft`, etc. For those, the legends appear as expected.

daschw:
Thanks for reporting! Also related: https://github.com/JuliaPlots/Plots.jl/issues/2369

Issue #2369, opened 2020-01-26
legend position with outer legends and vertical subplots


mkborregaard:
Consider this MWE
```julia
d16mer = Dict("Making BioSeq"=> 0.719,
    "Iterating kmers"=> 0.104,
    "Checking sym kmer"=> 0.063,
    "Reading line"=> 0.06,
    "Creating kmer iterator"=> 0.037
    )

dcpg = Dict("Making BioSeq"=> 0.695,
    "Reading line"=> 0.06,
    "GpC detection"=> 0.046,
    "Length of seq"=> 0.032,
    "Indexing seq"=> 0.038
    )

function bp(dict; kw...)
    dict["Misc"] = 1 .- sum(values(dict))
    vals = reverse(collect(values(dict)))'
    labs = permutedims(reverse(collect(keys(dict))))
    groupedbar(vals, label = labs, bar_position = :stack, orientation = :horizontal, legend = :outerright, yticks = []; kw...)
end

a = bp(d16mer)
b = bp(dcpg)
```
Then 
```julia
plot(a,b, size = (1200,300))
```
<img width="690" alt="Screenshot 2020-01-26 at 11 09 56" src="https://user-images.githubusercontent.com/8429802/73133678-7db56500-402c-11ea-9ee4-e7e8fc564234.png">

but
```julia
plot(a,b, layout = (2,1))
```
<img width="569" alt="Screenshot 2020-01-26 at 11 10 19" src="https://user-images.githubusercontent.com/8429802/73133682-8312af80-402c-11ea-95ca-89b62807e756.png">

mkborregaard:
Note also that the subplot sizes differ due to the legend text size, and that `link = :x` is overridden.

Issue #2202, opened 2019-09-25
Legend placement with gr appears buggy


dpo:
We noticed [here](https://github.com/JuliaSmoothOptimizers/JSOSolvers.jl/issues/2) that the legend for `:bottomright` is not at the bottom right. A short example of the problem can be recreated with
```julia
plot([scatter(rand(10,2), rand(10,2), legend=:bottomright) for i = 1:5]...,
	 layout=(1,5), size=(2000,400))
```
We initially opened this issue over at [GR.jl](https://github.com/jheinen/GR.jl/issues/243) but it turns out it might be an issue with Plots.jl instead.

Thanks.

PR #2770, opened 2020-06-09
Legends documentation and some todos

isentropic:
Closes: https://github.com/JuliaPlots/Plots.jl/issues/2207
I'll add a `good first issue` for plotly legends once this gets merged. I believe [this](https://github.com/JuliaPlots/Plots.jl/blob/c2df99cc6600f7fc0c6564d98fc3473c1c77519f/src/backends/plotly.jl#L340) should implemented properly as in [here](https://github.com/JuliaPlots/Plots.jl/blob/c2df99cc6600f7fc0c6564d98fc3473c1c77519f/src/backends/gr.jl#L567)


BeastyBlacksmith:
I don't see the benefit of adding todo comments to the code over opening an issue.
Also the GR legend implementation has quite a few bugs.
Non-exhaustive list:
#2377, #2369, 2330, #2202, #1776, #1589

isentropic:
Removed the `todo` comment

BeastyBlacksmith:
Thats fine, I just would't take it as a _proper_ reference implementation.

isentropic:
> Thats fine, I just would't take it as a _proper_ reference implementation.

I'll then open `good first issue` without referencing the gr implementation. Improving [this](https://github.com/JuliaPlots/Plots.jl/blob/c2df99cc6600f7fc0c6564d98fc3473c1c77519f/src/backends/plotly.jl#L340) be straightforward for a new contributor

isentropic:
I believe [this](https://github.com/JuliaPlots/Plots.jl/blob/c2df99cc6600f7fc0c6564d98fc3473c1c77519f/src/backends/pgfplotsx.jl#L666) is a better example. @BeastyBlacksmith 

I also have a question now, why does `pgfplotx` default to `outertopright` I thought about doing the same with `plotly` but then, as all other backends default to `topright` (within) the plot, I decided that it should default to `topright`

PS.
I'm impressed how well `pgfplotsx` handles colorbar positioning and margins, something I could not really achieve in https://github.com/JuliaPlots/Plots.jl/pull/2772

BeastyBlacksmith:
The actual default value for legend positioning is `:best`, which leaves room for interpretation.
You might know that `pyplot` for example places the legend where the most white space is when it would hide data.
So `:outertopright` is what seemed `best` to me, when implementing the backend, which is not always ideal as some examples in the documentation show.

isentropic:
Perhaps then, plotly should fallback to outerright as best too. I thought best really means topright unless the backend has a better idea

isentropic:
> The actual default value for legend positioning is `:best`, which leaves room for interpretation.
> You might know that `pyplot` for example places the legend where the most white space is when it would hide data.
> So `:outertopright` is what seemed `best` to me, when implementing the backend, which is not always ideal as some examples in the documentation show.

I believe that `pgfplotsx()` should default to `topright` to match GR. https://github.com/JuliaPlots/Plots.jl/blob/709a8a96093b512693b8703d195b5a5da0636a13/src/backends/gr.jl#L571

topright still takes around 10% of the plotarea so it should not be a problem. If a user desires they can manually position the legend
@BeastyBlacksmith

PR #2779, opened 2020-06-11
Plotly fixed version

fonsp:
Currently, Plots.jl loads `https://cdn.plot.ly/plotly-latest.min.js`, which is an unbounded dependency of this project - when plotly releases `2.0`, for example, Julia code using Plots.jl might no longer work - even if their dependency on `Plots.jl` has its version fixed.

To keep getting plotly's latest features, you could (test and) update the dependency version inside `Plots.jl` whenever you see fit.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,7 +4,7 @@
 .DS_Store
 examples/.ipynb_checkpoints/*
 examples/meetup/.ipynb_checkpoints/*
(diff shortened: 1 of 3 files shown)


isentropic:
This is made to have a dependency free backend.
Do you think that there's similar url with a fixed version?

PR #2786, opened 2020-06-15
better exponents

isentropic:
PyPlot improved exponents, additionally brought `:engineering` notation support.

`:engineering` is same as `:scientific` but it makes the powers of 10 being always divisible by 3

### DIFF
--- a/src/axes.jl
+++ b/src/axes.jl
@@ -181,16 +181,10 @@ function optimal_ticks_and_labels(sp::Subplot, axis::Axis, ticks = nothing)
 
     labels = if any(isfinite, unscaled_ticks)
         formatter = axis[:formatter]
-        if formatter == :auto
-            # the default behavior is to make strings of the scaled values and then apply the labelfunc
-            map(labelfunc(scale, backend()), Showoff.showoff(scaled_ticks, :auto))
-        elseif formatter == :plain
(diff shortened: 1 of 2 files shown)


src/axes.jl
```diff
@@ -181,16 +181,10 @@ function optimal_ticks_and_labels(sp::Subplot, axis::Axis, ticks = nothing)
...
-            # the default behavior is to make strings of the scaled values and then apply the labelfunc
-            map(labelfunc(scale, backend()), Showoff.showoff(scaled_ticks, :auto))
-        elseif formatter == :plain
-            # Leave the numbers in plain format
-            map(labelfunc(scale, backend()), Showoff.showoff(scaled_ticks, :plain))
-        elseif formatter == :scientific
-            Showoff.showoff(unscaled_ticks, :scientific)
-	elseif formatter == :latex
-	    map(x -> string("\$", replace(convert_sci_unicode(x), '×' => "\\times"), "\$"), Showoff.showoff(unscaled_ticks, :auto))
+        if formatter in (:auto, :plain, :scientific, :engineering)
```
isentropic, line 184: `:auto` falls back to `:scientific` for large numbers (internally in showoff), so there is no reason to treat these differently.

PR #2791, opened 2020-06-16
Fix #2771: Add aliases

### LINKED ISSUE #2771: [FR] topmargin should alias top_margin
As well as bottom, left, right. By convention every attribute should be accessible without spaces as per julian naming convention i believe

RashikaKarki:
Added Aliases according to the naming convention. I am not sure if this is the best way to deal with this issue. Need review.
Fix #2771

### DIFF
--- a/src/args.jl
+++ b/src/args.jl
@@ -502,6 +502,13 @@ end
 # -----------------------------------------------------------------------------
 
 
+# margin
+add_aliases(:left_margin , :leftmargin )
+add_aliases(:top_margin , :topmargin)
+add_aliases(:bottom_margin , :bottommargin)
+add_aliases(:right_margin ,:rightmargin)
+
+
 # colors
 add_aliases(:seriescolor, :c, :color, :colour)
 add_aliases(:linecolor, :lc, :lcolor, :lcolour, :linecolour)
--- a/src/args.jl
+++ b/src/args.jl
@@ -20,7 +20,14 @@ function add_non_underscore_aliases!(aliases::Dict{Symbol,Symbol})
     end
 end
 
-
+function add_non_underscore_aliases!(aliases::Dict{Symbol,Symbol}, args::Vector{Symbol}) 
+    for arg in args 
+        s = string(arg) 
+        if '_' in s 
+            aliases[Symbol(replace(s, "_" => ""))] = arg 
+        end 
+    end 
+end 
 # ------------------------------------------------------------
 
 const _allAxes = [:auto, :left, :right]
@@ -630,7 +637,7 @@ for arg in keys(_series_defaults)
     _keyAliases[makeplural(arg)] = arg
 end
 
-
+add_non_underscore_aliases!(_keyAliases)
 
 # -----------------------------------------------------------------------------
 


BeastyBlacksmith:
I think we can go even further and add a `add_non_underscore_aliases!(_keyAliases)` below [here](https://github.com/JuliaPlots/Plots.jl/blob/c2df99cc6600f7fc0c6564d98fc3473c1c77519f/src/args.jl#L625).

isentropic:
I further suggest that all the possible aliases need to be done with `add_non_underscore_aliases!(_keyAliases)`. That is, every manually typed alias that could have been generated with `add_non_underscore_aliases!(_keyAliases)` should be generated with this function.

This would be helpful long-term, as future contributors will understand that this is the way.

I wonder about the first time to plot issue though. Essentially, calculating most of the aliases would scratch on the time to first plot. Is there a way to have unimpacted?

RashikaKarki:
@BeastyBlacksmith 
Sure. I will add that.

RashikaKarki:
@isentropic Do you mean replacing aliases which as done using ```add_aliases``` with ```add_non_underscore_aliases!(_keyAliases)```?

BeastyBlacksmith:
@isentropic  I doubt this adds much time to `using`. So first step would be to profile that.

isentropic:
@RashikaKarki Yeah, for example this https://github.com/JuliaPlots/Plots.jl/blob/99c8bfe12d04fd2db77a247d29a723668e0606dc/src/args.jl#L618 needs to become 
```
add_aliases(:contour_labels, :clabels, :clabs)
```
with `:contourlabels` alias being added by the function (not manually as it is now)

BeastyBlacksmith:
> needs to become
> 
> ```
> add_aliases(:contour_labels, :clabels, :clabs)
> ```

Even better would be 
```
   add_aliases(:contour_labels, :c_labels, :c_labs)
```
and then we get the non underscore versions via the function. It would also be good to add some tests, that this works as expected.

@isentropic You never know beforehand. If you are curious, try it.

isentropic:
> > needs to become
> > 
> > ```
> > add_aliases(:contour_labels, :clabels, :clabs)
> > ```
> 
> Even better would be 
> ```
>    add_aliases(:contour_labels, :c_labels, :c_labs)
> ```
> and then we get the non underscore versions via the function. It would also be good to add some tests, that this works as expected.
> 
> @isentropic You never know beforehand. If you are curious, try it.

This is more correct and would lead to more consistency across-the-board

BeastyBlacksmith:
@RashikaKarki if you want, we can merge this with https://github.com/JuliaPlots/Plots.jl/pull/2791#issuecomment-644618688 to fix the issue and then the refactor could be a separate PR or we wait with the merge until the aliases are refactored. Your call, both is fine.

RashikaKarki:
> 
> 
> I think we can go even further and add a `add_non_underscore_aliases!(_keyAliases)` below [here](https://github.com/JuliaPlots/Plots.jl/blob/c2df99cc6600f7fc0c6564d98fc3473c1c77519f/src/args.jl#L625).

@BeastyBlacksmith I will add this line and I believe the exact  `add_non_underscore_aliases!(_keyAliases)` line is to be added right?

RashikaKarki:
> 
> 
> > needs to become
> > ```
> > add_aliases(:contour_labels, :clabels, :clabs)
> > ```
> 
> Even better would be
> 
> ```
>    add_aliases(:contour_labels, :c_labels, :c_labs)
> ```
> 
> and then we get the non underscore versions via the function. It would also be good to add some tests, that this works as expected.
> 
> @isentropic You never know beforehand. If you are curious, try it.

 `add_aliases(:contour_labels, :c_labels, :c_labs)` will add keys `:c_labs` and `:c_labels` to dict `_keyAliases` but since `add_non_underscore_aliases!()`  function will be taking keys only into account for removing the underscore I believe to add `:contour_labels` as alias we need to tweak  `add_non_underscore_aliases!()`  function a bit, right?

BeastyBlacksmith:
We could make use of a method that takes a vector of symbols and a dictionary, such that we can do
```julia
add_non_underscorealiases!(_keyAliases, keys(_series_defaults))
```

RashikaKarki:
@BeastyBlacksmith I have added the function

`add_aliases(:left_margin,:leftmargin)`
`add_aliases(:top_margin, :topmargin)`
`add_aliases(:bottom_margin, :bottommargin)`
`add_aliases(:right_margin, :rightmargin)`

This part of code can now be replaced by:

`const margin = Vector{Symbol}()`
`append!(margin, [:left_margin , :top_margin , :bottom_margin , :right_margin])`
`add_non_underscore_aliases!(_keyAliases, margin )`

BeastyBlacksmith:
What did you do, that the diff is so large?

src/args.jl
```diff
@@ -1,1649 +1,1665 @@
...
+end
+
+function add_non_underscore_aliases!(aliases::Dict{Symbol,Symbol}, args::Vector{Symbol})
+    for arg in args
+        s = string(arg)
+        if '_' in s
+            aliases[Symbol(replace(s, "_" => ""))] = arg
+        end
+    end
+end
```
RashikaKarki, line 28: This was the only code I added. Even I am quite surprised with these large differences.

RashikaKarki:
https://github.com/JuliaPlots/Plots.jl/blob/044836f686db282f6f86d32d0c125cb57591aa79/src/args.jl#L21-L28

This was the only function I added. I am not sure why it's showing such a large difference.

daschw:
> This was the only function I added. I am not sure why it's showing such a large difference.

Really weird, could it be a soft vs hard tab issue, maybe?

RashikaKarki:
Could this be because of the difference in code line position?

RashikaKarki:
I think I can try to revert back to the previous commit and again try to add this function.

RashikaKarki:
@BeastyBlacksmith I have reverted back to initial commit and added the function.

BeastyBlacksmith:
thanks. I'll merge this for now and if you want to take a stab at the refactor, just open a new PR.

isentropic:
Perhaps `thickness_scaling` deserves to be put in the aliases list, with aliases like `thickness` and `plot_scale`

Issue #2783, opened 2020-06-13
[BUG] Layouts meta-issue


isentropic:
I was trying to crack why Plots layouting does not always work perfectly and stumbled upon the 2016 year Toms issues and how he had in mind in the first place. I believe not many people to this day understand exactly how layouting works.

Perhaps, it would be a good idea to have comprehensive developer docs http://docs.juliaplots.org/latest/pipeline/ and explain concepts thoroughly with linking relevant code. But this is a digression, better documentation is always desired and all packages.

This is relevant to GR and PyPlot any other backend that would use `_update_min_padding!` callback. Upon conservation with @BeastyBlacksmith  I learned that pgfplotsx handles this a little differently

First of all, let us start with an example
```
p1 = plot(rand(10));
p2= plot(rand(10)); 
p3 = plot(rand(10)); 
p4 = plot(rand(10));

plot(p1,p2,p3,p4)  # looking well and good
```
![image](https://user-images.githubusercontent.com/24591123/84566524-c6d43e80-adac-11ea-968c-e749757cec00.png)

```
p1 = plot(rand(10), title="title", ylabel="label");
p2= plot(rand(10)); 
p3 = plot(rand(10)); 
p4 = plot(rand(10));

plot(p1,p2,p3,p4)  # looking good too!
```
![image](https://user-images.githubusercontent.com/24591123/84566600-87f2b880-adad-11ea-9c6a-ea5e4f6aca90.png)

This example is a typical MWE when layouting fails:
```
p1 = plot(rand(10));
p2= plot(rand(10)); 
p3 = plot(rand(10)); 
p4 = plot(rand(10), title="title", ylabel="label", xlabel="xlabel");

plot(p1,p2,p3,p4)  # p4 layouting fails miserably
```
Note how p4 is not aligned at all
![image](https://user-images.githubusercontent.com/24591123/84566586-61cd1880-adad-11ea-89dc-e41c7776fa38.png)

Issues reporting the same things: (perhaps should be deleted to keep things cleaner):
https://github.com/JuliaPlots/Plots.jl/issues/2237
https://github.com/JuliaPlots/Plots.jl/issues/1947

Why this happens:
[This code](https://github.com/JuliaPlots/Plots.jl/blob/ebc93b69353b06a0d3bdf42f09cdd82d1cf1701b/src/layouts.jl#L309) is responsible for updatting the padding of near-lying subplots, when a subplot contains sticking out elements (legends, colorbars, guides).

[...]

Issue #2695, opened 2020-05-12
[FR] by default align subplots at origin


floswald:
I think that this behaviour:

```
julia> plot(plot(1:10),plot(rand(10).*0.01),plot(1:10),plot(1:10),l=(2,2),link = :x)
```
![Screenshot 2020-05-11 17 08 05](https://user-images.githubusercontent.com/1021558/81693842-ccaecb00-9460-11ea-8de3-8429ccfd50c6.png)

is not desirable as a default setting. `left_margin` of the bottom right plot should be padded until the point `x=0` in the upper and lower right panels align. I hear from @BeastyBlacksmith on slack that this may be hard to achieve with plots.jl, but potentially worth posting here. He also points out that the `pgfplotsx` backend delivers the desired result. 

thanks.

Issue #2237, opened 2019-11-04
Subplot horizontal alignment off (y-axes not all aligned) in regular grid


OminousOmen:
I am trying to make a 2x2 grid of plots, all set to the same size.  While the first column of the grid behaves as expected, with the y-axes aligned, the second column has the issue that the bottom right plot is slightly too far over to the left, breaking the alignment of the y axes.  Below is my MWE code; I have tried and seen this on both PyPlot and GR backends.  Attached is `savefig()` output from my system.

Julia version is 1.2.0, current released packages for Plots.jl and backends used, on Linux/AMD64.

```Julia
using Plots
using LaTeXStrings

C_theta = [0.0, 1.13]
C_delta = [0.0, 0.84]
C_M = [5.0, 2.1]

B_theta = [2.42, 0.0]
B_delta = [-0.35, 0.0]
B_M = [2.1, 5.0]

xlimvals = (-0.075, 2.5)
ylimvals_delta = (-0.9, 0.025)
ylimvals_M = (1.8, 5.1)
p_MWE_1 = plot(C_theta, -C_delta, ylab=L"$\delta$", legend = :none, grid = :none, color = :black, xlims = xlimvals, ylims = ylimvals_delta)
p_MWE_2 = plot(C_theta, C_M, xlab=L"$\theta_\gamma - \theta_2$", ylab=L"$M$", legend = :none, grid = :none, color = :black, xlims = xlimvals, ylims = ylimvals_M)

p_MWE_3 = plot(B_theta, B_delta, legend = :none, grid = :none, color = :black, xlims = xlimvals, ylims = ylimvals_delta)
p_MWE_4 = plot(B_theta, B_M, xlab=L"$\theta_\gamma - \theta_2$", legend = :none, grid = :none, color = :black, xlims = xlimvals, ylims = ylimvals_M)


p_composite_MWE = plot(p_MWE_1, p_MWE_3, p_MWE_2, p_MWE_4, layout = (2,2), show = true)
```
[plot_alignment_issue_MWE.pdf](https://github.com/JuliaPlots/Plots.jl/files/3802656/plot_alignment_issue_MWE.pdf)

fredcallaway:
Bump. I am having the same issue. Note that matplotlib handles this correctly.

![julia](https://user-images.githubusercontent.com/7883156/70395601-3583aa00-19ce-11ea-8bd5-d323f3145d43.png)

[...]

Issue #1947, opened 2019-03-10
Heatmaps with linked axis not aligned


yha:
```
using Plots; pyplot()
let x = 0:5, z=0:2
    plot( heatmap(x,x,+), heatmap(x,x,+),
          heatmap(x,z,+), heatmap(x,z,+),
          link=:x, layout=(2,2), size=(400,200) )
end
```
![heatmap-align](https://user-images.githubusercontent.com/4170948/54088668-23a2a700-4369-11e9-8367-ed37a033c97a.png)

In both columns, the x axis is not correctly aligned. (Also, the default spacing is clearly not enough, but this is easy to work around with `...margin` attributes.)

PR #2799, opened 2020-06-18
layouts fixed

isentropic:
This PR has the yet highest `brain_power_used / lines_changed` ratio for me :smile: 
Fixes:
#2783
#2237
#1947
#2695

Perhaps there were other issues reporting some misalignment.

Plots before:
[layouts before improvement.pdf](https://github.com/JuliaPlots/Plots.jl/files/4796454/layouts.before.improvement.pdf)
Plots after:
[layouts improved.pdf](https://github.com/JuliaPlots/Plots.jl/files/4796455/layouts.improved.pdf)

### DIFF
--- a/src/backends/pyplot.jl
+++ b/src/backends/pyplot.jl
@@ -1263,7 +1263,7 @@ function _update_min_padding!(sp::Subplot{PyPlotBackend})
     # optionally add the width of colorbar labels and colorbar to rightpad
     if haskey(sp.attr, :cbar_ax)
         bb = py_bbox(sp.attr[:cbar_handle]."ax"."get_yticklabels"())
-        sp.attr[:cbar_width] = _cbar_width + width(bb) + (sp[:colorbar_title] == "" ? 0px : 30px)
+        sp.attr[:cbar_width] = width(bb) + (sp[:colorbar_title] == "" ? 0px : 30px)
         rightpad = rightpad + sp.attr[:cbar_width]
     end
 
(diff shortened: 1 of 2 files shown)


src/backends/pyplot.jl
```diff
@@ -1263,7 +1263,7 @@ function _update_min_padding!(sp::Subplot{PyPlotBackend})
     # optionally add the width of colorbar labels and colorbar to rightpad
     if haskey(sp.attr, :cbar_ax)
         bb = py_bbox(sp.attr[:cbar_handle]."ax"."get_yticklabels"())
-        sp.attr[:cbar_width] = _cbar_width + width(bb) + (sp[:colorbar_title] == "" ? 0px : 30px)
+        sp.attr[:cbar_width] = width(bb) + (sp[:colorbar_title] == "" ? 0px : 30px)
```
isentropic, line 1266: `_cbar_width` is already contained in the `bb`

src/layouts.jl
```diff
@@ -407,10 +405,10 @@ function update_child_bboxes!(layout::GridLayout, minimum_perimeter = [0mm,0mm,0
         # this is the minimum perimeter as decided by this child's parent, so that
         # all children on this border have the same value
         min_child_perimeter = [
-            c == 1  ? layout.minpad[1] : 0mm,
-            r == 1  ? layout.minpad[2] : 0mm,
-            c == nc ? layout.minpad[3] : 0mm,
-            r == nr ? layout.minpad[4] : 0mm
+            c == 1  ? layout.minpad[1] : pad_left[c],
```
isentropic, line 408: My best short explanation:
minimum padding should be propagated to all the children in the grid.

Additionally, perhaps I should add a clarifying comment that `minimum_perimeter` actually means `minimum_padding`. It took me a while to understand this. Should I?

isentropic:
This is due to colorbars not having a consistent width nor rules for scaling. Theres always more than actually required padding added for colorbars. I do not think it is related to grid layouting.

PR #2800, opened 2020-06-18
Recipe for GeometryBasics.Point

yha:
since `GeometryTypes` is now deprecated


yha:
Looks like tests with Julia 1.0 fail because `GeometryBasics` requires Julia 1.2.
Not sure what to do about that. Is it possible to depend on `GeometryBasics` conditionally based on Julia version? 
Otherwise, if `Plots` should continue supporting 1.0 and `GeometryBasics` doesn't, maybe the `Point` recipe needs to go inside `GeometryBasics` or to a different package?

mkborregaard:
I'm not against capping at 1.2. I don't think backward compatibility is awfully important at this stage - what do yuo say @daschw ?

daschw:
I agree, our tests on 1.0 don't pass anyway ...

yha:
In this case, I think this can be merged as-is.
I don't think it's necessary to explicitly require 1.2 in `Plots` when it's already required by a dependency. And the failures on nightly seem unrelated.

Issue #1499, opened 2018-04-27
Rounded corners with pyplot


stakaz:
Hello, when using pyplot I get rounded corners for markers and ugly rounding in errorbars, see the picture.

![screenshot_rounded_corners](https://user-images.githubusercontent.com/998651/39347858-096ac7fa-49f5-11e8-8876-3747e50ecd2c.png)

The code for this is:

```julia
using Plots
pyplot()

default(
markershape = :square,
markersize = 5,
markeralpha = 1.0,
markercolor = :match,
markerstrokewidth = 2,
markerstrokealpha = 1.0,
markerstrokecolor = :auto,
markerstrokestyle = :solid
)

x = [1,2,3,4,6]
y = [0.2,0.4,0.6,0.2,0.8]
yerr = [0.1,0.12,0.14,0.08,0.3]

scatter(x,y,err=yerr,color=:blue)
savefig("test.pdf")
```

Is this a "style" decision and has it something to do with pyplot itself? But with the same matplotlibrc file and using PyPlot I get sharp corners.

Thanks

isentropic:
This pr should fix your issue, additionally the reason why error caps look a little off-center is I believe https://github.com/matplotlib/matplotlib/issues/7233

PR #2802, opened 2020-06-19
line caps, legendbox

isentropic:
some minor fixes in pyplot

legend boxes have additional padding to match that of other backends
fixes https://github.com/JuliaPlots/Plots.jl/issues/1499 and makes the linecaps match the style of GR
I think the line joining style looks a lot better in pyplot

more info here https://matplotlib.org/3.1.0/gallery/lines_bars_and_markers/joinstyle.html

pyplot()
image
image

gr()
image
image

### DIFF
--- a/src/backends/pyplot.jl
+++ b/src/backends/pyplot.jl
@@ -437,18 +437,19 @@ function py_add_series(plt::Plot{PyPlotBackend}, series::Series)
             #     end
             #     push!(handles, handle)
             # else
-                for (i, rng) in enumerate(iter_segments(series))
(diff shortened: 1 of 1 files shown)


isentropic:
pyplot errobar caps look a little off-center, i believe the reason for which is  matplotlib/matplotlib#7233
one hack could be to comment out this line https://github.com/JuliaPlots/Plots.jl/blob/709a8a96093b512693b8703d195b5a5da0636a13/src/backends/pyplot.jl#L128

It will, however, make the error caps behave the same as in pgfplotsx i believe

PR #2807, opened 2020-06-22
Fix #2330, #2202: Position legends closer to the borders

### LINKED ISSUE #2202: Legend placement with gr appears buggy

### LINKED ISSUE #2330: Legend positioned far from borders overlaping data - GR
Legend position for :bottom[right-left] of :top[right-left] is far from the plot border so, in many cases, it goes so much to the center that overlaps the data points unnecessarily.
See:
https://discourse.julialang.org/t/how-to-change-legends-place-with-plots-gr/7294/18

It may be related with this older issue?
https://github.com/JuliaPlots/Plots.jl/issues/2202

Diaga:
Fix #2330 
Fix #2202
Images below for plot made using this snippet _(changing legend in each case)_: 
```
plot([scatter(rand(10,2), rand(10,2), legend=:topleft) for i = 1:5]...,
	 layout=(1,5), size=(2000, 400))
```

__Bottom Left:__
- Before:
bottomleft-before
- After:
bottomleft-after

__Bottom Right:__
- Before:
bottomright-before
- After:
bottomright-after

__Top Left:__
- Before:
topleft-before

- After:
topleft-after

__Top Right:__
- Before:
topright-before

- After:
topright-after

Needs review.

### DIFF
--- a/src/backends/gr.jl
+++ b/src/backends/gr.jl
@@ -565,11 +565,10 @@ gr_view_xcenter(viewport_plotarea) = 0.5 * (viewport_plotarea[1] + viewport_plot
 gr_view_ycenter(viewport_plotarea) = 0.5 * (viewport_plotarea[3] + viewport_plotarea[4])
 
 function gr_legend_pos(sp::Subplot, w, h, viewport_plotarea)
+    legend_leftw, legend_rightw, legend_textw, x_legend_offset, y_legend_offset = w
     s = sp[:legend]
     typeof(s) <: Symbol || return gr_legend_pos(s, w, h, viewport_plotarea)
     str = string(s)
-    x_legend_offset = (viewport_plotarea[2] - viewport_plotarea[1]) / 30
-    y_legend_offset = (viewport_plotarea[4] - viewport_plotarea[3]) / 15
     if str == "best"
         str = "topright"
     end
@@ -581,18 +580,18 @@ function gr_legend_pos(sp::Subplot, w, h, viewport_plotarea)
     if occursin("right", str)
         if occursin("outer", str)
             # As per https://github.com/jheinen/GR.jl/blob/master/src/jlgr.jl#L525
-            xpos = viewport_plotarea[2] + 0.11 + ymirror * gr_axis_width(sp, sp[:yaxis])
+            xpos = viewport_plotarea[2] + x_legend_offset + legend_leftw + ymirror * gr_axis_width(sp, sp[:yaxis])
         else
-            xpos = viewport_plotarea[2] - x_legend_offset - w
+            xpos = viewport_plotarea[2] - legend_rightw - legend_textw
         end
     elseif occursin("left", str)
         if occursin("outer", str)
-            xpos = viewport_plotarea[1] - 0.05 - w - !ymirror * gr_axis_width(sp, sp[:yaxis])
+            xpos = viewport_plotarea[1] - !ymirror * gr_axis_width(sp, sp[:yaxis]) - x_legend_offset * 2 - legend_rightw - legend_textw
         else
-            xpos = viewport_plotarea[1] + 0.08 + x_legend_offset
+            xpos = viewport_plotarea[1] + legend_leftw + x_legend_offset
         end
     else
-        xpos = (viewport_plotarea[2]-viewport_plotarea[1])/2 - w/2 + 0.04 + viewport_plotarea[1]
+        xpos = (viewport_plotarea[2]-viewport_plotarea[1])/2 + viewport_plotarea[1] + legend_leftw - legend_rightw - legend_textw - x_legend_offset * 2
     end
     if occursin("top", str)
         if s == :outertop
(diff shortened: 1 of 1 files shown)


BeastyBlacksmith:
I am not quite sure that this absolute offset is the correct approach. It does look better at the examples, but will this hold on all scales?
Ideally the position of the legend is at a fixed distance with respect to the width/ height of the plotarea.

Diaga:
I think the scaling issue arises because e.g, a 0.06 offset is a lot more significant when the size of window is large. Therefore, using a small offset as 0.01, the same issue can only arise when the window is 6 times larger.
I went with this approach because this is how it is being done right now. I am happy to change the offset to something that is calculated dynamically to plots width and height if you think it would be better.

An issue with using current constant offsets can be specifically with legend:=left _(since plot width would start before yaxis)_. I could not find how yaxis position is being calculated for each plot, and we would need to take that into account to make it completely dynamic.

BeastyBlacksmith:
the positions and extents of the subplots and their plotareas are calculated here ( https://github.com/JuliaPlots/Plots.jl/blob/b6234a9ea0b226355d83ca865f5914cef7e0a6ea/src/backends/gr.jl#L957 ). And the yaxis would then be from the lower left corner of the plotarea to the upper left corner.

BeastyBlacksmith:
After playing around a bit, I am pretty sure, that the legendwidth `w` is not calculated correctly.
A good example for testing is
```julia
plot([scatter(rand(10,2), rand(10,2), legend=pos) for pos = [:topleft, :bottomleft, :outertopright, :bottomright, :outertop, :topright]]...,
                layout=(2,3), size=(2400, 800))
```

Diaga:
Mentioned that in #2808.

Diaga:
After new changes:

BeastyBlacksmith:
All the `right` ones are a little bit too far

Diaga:
https://github.com/JuliaPlots/Plots.jl/blob/b6234a9ea0b226355d83ca865f5914cef7e0a6ea/src/backends/gr.jl#L1848-L1850

This adds 0.02 offset to the right of the legend box, increasing its width.
Adding/Subtracting this offset in ```x_pos``` will fix this. Should I proceed?

After:
2202-2330-v2

Diaga:
_Just throwing it out here:_
The ```w``` parameter in the ```gr_legend_pos``` serve little purpose because it corresponds to the legend **text** size and not the actual legend box size. Maybe we should update that?

BeastyBlacksmith:
It's not very clear to me where all those numbers come from, so if we can get rid of them or give them some meaningful comments that would be great.
Can you calculate the actual legendwidth and use that instead of the legend text width?

BeastyBlacksmith:
In a follow up PR we should also adjust the amount of whitespace within the legend box

Diaga:
Alright, I will do that.

Regarding whitespace, I understand what is causing this behavior, will open up an issue.

Diaga:
After:
2202-2330-v3

Diaga:
Thanks @daschw, I'll look into it and let you know.

Diaga:
The first image looks fine in the white background as below:

Colorful Background:
2202-2330-color
White Background:
2202-2330-white

The second image is messing up because the ```guide``` width/axis width is not catered in the ```viewplot_area``` here:

https://github.com/JuliaPlots/Plots.jl/blob/2f4eb7b5583e1524f51dd3ae4e49c5a011122267/src/backends/gr.jl#L1017-L1025

I think these should be fixed in follow up PRs because the causes of these issues are different. I'll be interested to take them up.

daschw:
Ok, great, thanks a lot for this.

BeastyBlacksmith:
Just as a clarification, to state what is desired behavior, I do think we can merge this for now, since it is already quite an improvement: 
> The first image looks fine in the white background as below:
> 
The upper border of inner `top`-legends should be just below the end of the axis-spine.
So this needs some improvement. Also compare with the examples of the other backends.
However, it's fine to adjust this in a different PR.

> The second image is messing up because the `guide` width/axis width is not catered in the `viewport_plotarea` here:

`viewport_plotarea` should never be changed, instead, if the outer legends are outside of the viewport the `viewport_subplot` should be extended.

Diaga:
However, changing ```viewport_subpolot``` may mean not adhering strictly to the ```size``` kwarg. Are we fine with that?

BeastyBlacksmith:
Thats a good point. So for now adjusting the plotarea is the way to go.
In the long term, we could rethink, if keeping the plotarea constant when `size` is not explicitly given is a better approach or not.

PR #2809, opened 2020-06-23
[WIP] Fix #1881, Ticks are now visible for filled contours without grid=:none parameter

### LINKED ISSUE #1881: Ticks of contour plots with PyPlot
I found a strange behavior on ticks of contour plot with PyPlots.

Default setting does not make ticks:
```
using Plots; pyplot()
contour(-2:0.1:2,-2:0.1:2,(x,y)->x^2+y^2, fill=true, levels=50)
```
contour1

If I put a grid option, I get ticks:
```
`using Plots; pyplot()
contour(-2:0.1:2,-2:0.1:2,(x,y)->x^2+y^2, fill=true, levels=50, grid=:none)
```
contour2

It seems better to use the second case as default.
Thank you for your development.

LakshyaKhatri:
Fixes #1881 

I discussed this issue with @isentropic and he suggested to add a `zorder` argument in the below call:
https://github.com/JuliaPlots/Plots.jl/blob/47590b2538b9d0387b14d46eb49e7257f904dde6/src/backends/pyplot.jl#L850
I tried adding `zorder` with a high value in other places too but it didn't work. 

Turned out the axes were drawn below the countors when `grid` was set to true:
https://github.com/JuliaPlots/Plots.jl/blob/47590b2538b9d0387b14d46eb49e7257f904dde6/src/backends/pyplot.jl#L1144-L1151
and hence the ticks were also hiding:
fig

Removing this line:
https://github.com/JuliaPlots/Plots.jl/blob/47590b2538b9d0387b14d46eb49e7257f904dde6/src/backends/pyplot.jl#L1151

or setting `ax."set_axisbelow"(true)` to `false` is producing the following results:
Figure_1

If this is the right solution, I will update the same for other backends too.

### DIFF
--- a/src/backends/pyplot.jl
+++ b/src/backends/pyplot.jl
@@ -1148,7 +1148,6 @@ function _before_layout_calcs(plt::Plot{PyPlotBackend})
                     linestyle = py_linestyle(:line, axis[:gridstyle]),
                     linewidth = py_thickness_scale(plt, axis[:gridlinewidth]),
                     alpha = axis[:gridalpha])
-                ax."set_axisbelow"(true)
             else
                 pyaxis."grid"(false)
             end


BeastyBlacksmith:
The question is, whether we want to show gridlines on top also or not?

daschw:
We don't do it for the other backends, right?

BeastyBlacksmith:
I think pgfplotsx does so, too. I planned to change that though... however I think it would be nice to be able to enable them, if they are wanted. So basically it would be nice, if we could have different defaults per seriestype.

LakshyaKhatri:
So, the defaults should not contain gridlines or the other way around?

BeastyBlacksmith:
I think for this PR gridlines should either not be displayed or drawn behind the contour.

LakshyaKhatri:
In the prior case these two example calls will result in the same plot:
```contour(-2:0.1:2,-2:0.1:2,(x,y)->x^2+y^2, fill=true, levels=50)```
and 
```contour(-2:0.1:2,-2:0.1:2,(x,y)->x^2+y^2, fill=true, levels=50, grid=:none)```



I will go with the suggestion you provide.

**Comment by isentropic** [comment]
I think gridlines should be behind the contour but ticks on top

**Comment by LakshyaKhatri** [comment]
Yes, I am working on the required changes

**Comment by LakshyaKhatri** [comment]
> I think for this PR gridlines should either not be displayed or drawn behind the contour.

I tried almost everything, but the latter case doesn't seem to work. The best solution I found till now is  [this thread](http://matplotlib.1069221.n5.nabble.com/Axis-ticks-zorder-td3957.html) from matplotlib forum. Any help or suggestions?

**Comment by LakshyaKhatri** [comment]
I discussed this issue with the maintainer of Matplotlib and he has suggested [these solutions](https://discourse.matplotlib.org/t/plot-contour-between-gridlines-and-ticks/21327/4?u=lakshyakhatri). Kindly look into the discussion and I think then we can decide what will be the best solution :smile: 

cc: @BeastyBlacksmith

**Comment by BeastyBlacksmith** [comment]
I would probably go with disabling the normal ticks and then draw ticks with the secondary axis feature.
But in the end it comes down to what works best for you.

**Comment by LakshyaKhatri** [comment]
Okay, the secondary ticks work:
Figure_2

but the issue is, it makes all the axis operations dependent on the secondary axis. It also increases the complexity by giving us different handles for each secondary axis (in this case two different handels for bottom and left axis).

**Comment by BeastyBlacksmith** [comment]
It does increase the complexity, but if we can keep that isolated to the part where axis decorations are built it's worth the flexibility we gain, IMO.
Would be good, if you push the changes, so we can have a look at the code.
Also interested about @isentropic's opinion.

**Comment by isentropic** [comment]
I think we should not overcomplicate things with secondary axis and stuff make that issue closed due to matplotlib limitations and thats it.

If we have secondary axis that might overcomplicate things a lot. Because: 

1 Plots calculates the ticks
2 series are drawn on axes, so if we have image with scatter on top, they should be drawn on different axes, which in turn would ruin the order. I guess drawing scatter on top of image wouldn't work (or the other way)
3 This would absolutely make no sense for future maintainers that will work with the codebase. I'm very sure that this would cause problem in understanding imshow and fixing future issues associated with it, and eventually someone will clear it up and make as it is now.

I understand that this issue had turned to be way more complex than it should have been, but I really think that fixing this is not worth our efforts.

You have done a great job @LakshyaKhatri  I can only say that it is sad that such a thoroughly researched issue happens not to have a simple solution.

**Comment by LakshyaKhatri** [comment]
calling `secondary_xaxis`  and `secondary_yaxis` on `Axes` object is giving me a key error in julia!

```julia
ax."secondary_xaxis"("bottom", zorder=2.0)
```

```julia
Error showing value of type Plots.Plot{Plots.PyPlotBackend}:
ERROR: KeyError: key "secondary_xaxis" not found
Stacktrace:
 [1] getproperty(::PyCall.PyObject, ::String) at /home/aries/.julia/packages/PyCall/zqDXB/src/PyCall.jl:301
 [2] _before_layout_calcs(::Plots.Plot{Plots.PyPlotBackend}) at /home/aries/.julia/dev/Plots/src/backends/pyplot.jl:1154
 [3] prepare_output(::Plots.Plot{Plots.PyPlotBackend}) at /home/aries/.julia/dev/Plots/src/plot.jl:177
 [4] display(::Plots.PlotsDisplay, ::Plots.Plot{Plots.PyPlotBackend}) at /home/aries/.julia/dev/Plots/src/output.jl:149
 [5] display(::Any) at ./multimedia.jl:323
 [6] #invokelatest#1 at ./essentials.jl:712 [inlined]
 [7] invokelatest at ./essentials.jl:711 [inlined]
 [8] print_response(::IO, ::Any, ::Bool, ::Bool, ::Any) at /buildworker/worker/package_linux64/build/usr/share/julia/stdlib/v1.4/REPL/src/REPL.jl:161
 [9] print_response(::REPL.AbstractREPL, ::Any, ::Bool, ::Bool) at /buildworker/worker/package_linux64/build/usr/share/julia/stdlib/v1.4/REPL/src/REPL.jl:146
 [10] (::REPL.var"#do_respond#38"{Bool,REPL.var"#48#57"{REPL.LineEditREPL,REPL.REPLHistoryProvider},REPL.LineEditREPL,REPL.LineEdit.Prompt})(::Any, ::Any, ::Any) at /buildworker/worker/package_linux64/build/usr/share/julia/stdlib/v1.4/REPL/src/REPL.jl:729
 [11] #invokelatest#1 at ./essentials.jl:712 [inlined]
 [12] invokelatest at ./essentials.jl:711 [inlined]
 [13] run_interface(::REPL.Terminals.TextTerminal, ::REPL.LineEdit.ModalInterface, ::REPL.LineEdit.MIState) at /buildworker/worker/package_linux64/build/usr/share/julia/stdlib/v1.4/REPL/src/LineEdit.jl:2354
 [14] run_frontend(::REPL.LineEditREPL, ::REPL.REPLBackendRef) at /buildworker/worker/package_linux64/build/usr/share/julia/stdlib/v1.4/REPL/src/REPL.jl:1055
 [15] run_repl(::REPL.AbstractREPL, ::Any) at /buildworker/worker/package_linux64/build/usr/share/julia/stdlib/v1.4/REPL/src/REPL.jl:206
 [16] (::Base.var"#764#766"{Bool,Bool,Bool,Bool})(::Module) at ./client.jl:383
 [17] #invokelatest#1 at ./essentials.jl:712 [inlined]
 [18] invokelatest at ./essentials.jl:711 [inlined]
 [19] run_main_repl(::Bool, ::Bool, ::Bool, ::Bool, ::Bool) at ./client.jl:367
 [20] exec_options(::Base.JLOptions) at ./client.jl:305
 [21] _start() at ./client.jl:484

```

LakshyaKhatri:
Thanks @isentropic for the appreciation. I'll keep contributing to the Julia community. Feel free to close this PR :smile:

BeastyBlacksmith:
Hmm... disabling the gridlines is still an option

LakshyaKhatri:
Yes that will be an easy hack.

PR #2821, opened 2020-06-27
Update julia version

daschw:
Require julia version 1.3 (for lower versions the tests fail to precompile GeometryBasics) and (hopefully) fix tests.


daschw:
for reference, the failure on julia 1.2 for GeometryBasics: https://github.com/daschw/Plots.jl/runs/813719952#step:8:218

PR #2832, opened 2020-06-30
Use SnoopCompileBot

daschw:
This should automatically update the precompile files using SnoopCompileBot and a corresponding Github Action as described in https://timholy.github.io/SnoopCompile.jl/stable/bot/
This removes the current precompile file and hopefully PRs with new precompile files are opened after this gets merged.
@timholy or @aminya it would be great if you could take a look at this and check if I'm doing something wrong here.


daschw:
The SnoopCompile runs can be checked at https://github.com/daschw/Plots.jl/actions/runs/153066245 and here you can take a look at the PR that would be created: https://github.com/daschw/Plots.jl/pull/4

daschw:
I guess we should decide if we want to have different precompile files per OS (does this make sense?). Otherwise we don't have to run the action on all OSs. And, since nightly is already 1.6 I should add a run for some Julia 1.5 version

timholy:
LGTM but when it comes to the bot, @aminya is the real expert.

timholy:
Also worth saying, I think you guys will see some gains wrt to load times within another month or so on nightly...there are still several crucial PRs to merge, the most important being https://github.com/JuliaLang/julia/pull/35877 (which is currently on hold for good reasons). Outside of Julia & Pkg, only https://github.com/JuliaWeb/HTTP.jl/pull/549 is waiting to be merged (as long as you're using at least FixedPointNumbers 0.8.1, ColorTypes 0.10.4). For me the combination of all of those drops TTFP by almost an additional factor of 2 compared to 1.5rc1.

daschw:
Sounds great and huge thanks to everyone who's involved in this!

daschw:
Looking at https://github.com/daschw/Plots.jl/pull/4/files now it like that the precompile files are identical for each OS, so probably we do not require multi-os.

deps/SnoopCompile/snoop_bench.jl
```diff
@@ -0,0 +1,12 @@
+using SnoopCompile
+
+snoop_bench(
+    BotConfig(
+        "Plots",
+        os = ["linux", "windows", "macos"],
+        else_os = "linux",
+        version = ["1.3", "1.4", "nightly"],
+        else_version = "1.4",
```
aminya, line 9: I think it is better to use the `yml_path` option to prevent conflicts.
```suggestion
		yml_path= "SnoopCompile.yml"
        else_os = "linux",
        else_version = "1.4",
```

deps/SnoopCompile/snoop_bot.jl
```diff
@@ -0,0 +1,12 @@
+using SnoopCompile
+
+snoop_bot(
+    BotConfig(
+        "Plots",
+        os = ["linux", "windows", "macos"],
+        else_os = "linux",
+        version = ["1.3", "1.4", "nightly"],
+        else_version = "1.4",
```
aminya, line 9: I think it is better to use the `yml_path` option to prevent conflicts.
```suggestion
        yml_path= "SnoopCompile.yml",
        else_os = "linux",
        else_version = "1.4",
```

.github/workflows/SnoopCompile.yml
```diff
@@ -0,0 +1,97 @@
...
+      GKSwstype: "100"
+      PLOTS_TEST: "true"
+    runs-on: ${{ matrix.os }}
+    strategy:
+      fail-fast: false
+      matrix:
+        version:   # NOTE: the versions below should match those in your botconfig
+          - '1.3'
+          - '1.4'
+          # - '1.5'
```
aminya, line 26: Enabling this since nightly is 1.6 now? 🤔 
```suggestion
           - '1.5'
```

daschw: I tried but it did not work: https://github.com/daschw/Plots.jl/actions/runs/153176644

daschw: All 1.5 tests are failing there.

aminya: Maybe try `-'1.5.0-rc1'`.

@SaschaMann shouldn't 1.5 install `'1.5.0-rc1'` by default? julia-setup fails to do this.

daschw: ok, let's try

daschw:
@aminya I guess, once we're sure this works I should uncomment
```
    #  - 'master'  # NOTE: to run the bot only on pushes to master
```
in SnoopCompile.yml?

.github/workflows/SnoopCompile.yml
```diff
@@ -0,0 +1,97 @@
...
+      GKSwstype: "100"
+      PLOTS_TEST: "true"
+    runs-on: ${{ matrix.os }}
+    strategy:
+      fail-fast: false
+      matrix:
+        version:   # NOTE: the versions below should match those in your botconfig
+          - '1.3'
+          - '1.4'
+           - '1.5.0-rc1'
```
daschw, line 26:
```suggestion
          - '1.5.0-rc1'
```

.github/workflows/SnoopCompile.yml
SaschaMann: > @SaschaMann shouldn't 1.5 install '1.5.0-rc1' by default? julia-setup fails to do this.

`1.5` doesn't include prereleases, you need to use `1.5.0-0` to include all prereleases.

(I'm in the process of adding 1.5.0-rc1 right now, right now it points at 1.5.0-beta1)

aminya:
> @aminya I guess, once we're sure this works I should uncomment
> 
> ```
>     #  - 'master'  # NOTE: to run the bot only on pushes to master
> ```
> 
> in SnoopCompile.yml?

If you want to get precompile updates in pushes to other branches (under JuliaPlots/Plots) don't uncomment this. But if you want precompile updates only on pushes to master, then do that.

I don't think this be an issue but may happen in rare cases: Once this PR is merged and bot PR is merged as well, it also may worth mentioning in the contribution section of readme or docs to set `should_precompile=false` in case some error happens because of precompilation. This is very rare, but it is good to know.

daschw:
> it also may worth mentioning in the contribution section of readme or docs to set `should_precompile=false`

Where exactly do we set that?

aminya:
> > it also may worth mentioning in the contribution section of readme or docs to set `should_precompile=false`
> 
> Where exactly do we set that?

https://github.com/daschw/Plots.jl/pull/4/files#diff-2270ce88ed8290c3b808f5ea76728946R1
Again, this is very rare, and for Plots, this will not be a thing as it has used hardcoded precompile signatures for a long time.

.github/workflows/SnoopCompile.yml
daschw: So, `1.5.0-0` would be recommended?

SaschaMann: Yes.

(1.5.0-rc1 should be available now)

daschw: Thanks!

daschw: With `1.5.0-0` I get
```

Run julia-actions/setup-julia@latest
0s
    PLOTS_TEST: true
Run julia-actions/setup-julia@latest
  with:
    version: 1.5.0-0
    arch: x64
    show-versioninfo: false
  env:
    GKS_ENCODING: utf8
    GKSwstype: 100
    PLOTS_TEST: true
##[error]Unexpected HTTP response: 404
```
https://github.com/daschw/Plots.jl/runs/824362165

daschw:
OK, this seems to work now. Thanks a lot @aminya @timholy and @SaschaMann for your support!

.github/workflows/SnoopCompile.yml
SaschaMann: My bad, it should've been `^1.5.0-0`. Sorry! 🤦

daschw: That does not work either: https://github.com/daschw/Plots.jl/actions/runs/155304982

daschw: But that's an issue with SnoopCompile.jl cc: @aminya

PR #2837, opened 2020-07-02
Add label for `label=0` kwarg

### LINKED ISSUE #2834: [BUG] `label=0` produces no legend, whereas `label=1` does
## Details
It seems that `0` is used as a sentinel value somewhere. 

`Plots.plot(rand(100); label=1)` plots a legend with label 1
`Plots.plot(rand(100); label=0`) plots no legend.
`Plots.plot(rand(100); label="0"`) plots a legend with label 0.

I think it is good style to ensure the label is a string, but I'd say this is a bug since some values are converted to strings, and others are not.

### Backends

gr (default) |  x   |     |
pyplot       |     |     |x
plotly       |  x   |     |
plotlyjs     |     |     |x
pgfplotsx    |     |     |x
inspectdr    |     |     |x

### Versions

Plots.jl version: Plots v1.5.0
Backend  version (`]st -m`): GR v0.50.1
Output of `versioninfo()`:
```
julia> versioninfo()
Julia Version 1.4.1
Commit 381693d3df* (2020-04-14 17:20 UTC)
Platform Info:
  OS: macOS (x86_64-apple-darwin18.7.0)
  CPU: Intel(R) Core(TM) i7-4960HQ CPU @ 2.60GHz
  WORD_SIZE: 64
  LIBM: libopenlibm
  LLVM: libLLVM-8.0.1 (ORCJIT, haswell)
Environment:
  JULIA_EDITOR = atom  -a
  JULIA_NUM_THREADS = 4
  JULIA_BASE_PATH = /Applications/Julia-1.4.app/Contents/Resources/julia/share/julia/base
  JULIA_STDLIB_PATH = /Applications/Julia-1.4.app/Contents/Resources/julia/share/julia/stdlib
```

LakshyaKhatri:
Fixes #2834

### DIFF
--- a/src/args.jl
+++ b/src/args.jl
@@ -1639,7 +1639,7 @@ function _update_series_attributes!(plotattributes::AKW, plt::Plot, sp::Subplot)
     # set label
     label = plotattributes[:label]
     label = (label == "AUTO" ? "y$globalIndex" : label)
-    label = label == 0 ? "0" : label
+    label = label === 0 ? "0" : label
     label = label in (:none, nothing, false) ? "" : label
     plotattributes[:label] = label
 
--- a/src/args.jl
+++ b/src/args.jl
@@ -1651,7 +1651,7 @@ function _update_series_attributes!(plotattributes::AKW, plt::Plot, sp::Subplot)
     end
 
     # set label
-    plotattributes[:label] = label_to_string(plotattributes[:label], globalIndex)
+    plotattributes[:label] = label_to_string.(plotattributes[:label], globalIndex)
 
     _replace_linewidth(plotattributes)
    plotattributes
--- a/src/args.jl
+++ b/src/args.jl
@@ -1642,9 +1642,7 @@ function _update_series_attributes!(plotattributes::AKW, plt::Plot, sp::Subplot)
     end
 
     # set label
-    label = plotattributes[:label]
-    label = label_to_string(label, globalIndex)
-    plotattributes[:label] = label
+    plotattributes[:label] = label_to_string(plotattributes[:label], globalIndex)
 
     _replace_linewidth(plotattributes)
    plotattributes
--- a/src/args.jl
+++ b/src/args.jl
@@ -20,14 +20,14 @@ function add_non_underscore_aliases!(aliases::Dict{Symbol,Symbol})
     end
 end
 
-function add_non_underscore_aliases!(aliases::Dict{Symbol,Symbol}, args::Vector{Symbol}) 
-    for arg in args 
-        s = string(arg) 
-        if '_' in s 
-            aliases[Symbol(replace(s, "_" => ""))] = arg 
-        end 
-    end 
-end 
+function add_non_underscore_aliases!(aliases::Dict{Symbol,Symbol}, args::Vector{Symbol})
+    for arg in args
+        s = string(arg)
+        if '_' in s
+            aliases[Symbol(replace(s, "_" => ""))] = arg
+        end
+    end
+end
 # ------------------------------------------------------------
 
 const _allAxes = [:auto, :left, :right]
@@ -1639,6 +1639,7 @@ function _update_series_attributes!(plotattributes::AKW, plt::Plot, sp::Subplot)
     # set label
     label = plotattributes[:label]
     label = (label == "AUTO" ? "y$globalIndex" : label)
+    label = label == 0 ? "0" : label
     label = label in (:none, nothing, false) ? "" : label
     plotattributes[:label] = label
 
--- a/src/args.jl
+++ b/src/args.jl
@@ -1572,8 +1572,17 @@ end
 label_to_string(label::Bool, series_plotindex) = label ? label_to_string(:auto, series_plotindex) : ""
 label_to_string(label::Nothing, series_plotindex) = ""
 label_to_string(label::Missing, series_plotindex) = ""
-label_to_string(label::Symbol, series_plotindex) = label==:auto ?  "y$series_plotindex" : string((label==:none ? "" : label))
+function label_to_string(label::Symbol, series_plotindex)
+    if label==:auto
+        return string("y", series_plotindex)
+    elseif label==:none
+        return ""
+    else
+        throw(ArgumentError("unsupported symbol $(label) passed to `label`"))
+    end
+end
 label_to_string(label, series_plotindex) = string(label)  # Fallback to string promotion
+
 function _update_series_attributes!(plotattributes::AKW, plt::Plot, sp::Subplot)
     pkg = plt.backend
     globalIndex = plotattributes[:series_plotindex]
--- a/src/args.jl
+++ b/src/args.jl
@@ -235,7 +235,7 @@ const _bar_width = 0.8
 # -----------------------------------------------------------------------------
 
 const _series_defaults = KW(
-    :label             => "AUTO",
+    :label             => :auto,
     :colorbar_entry    => true,
     :seriescolor       => :auto,
     :seriesalpha       => nothing,
@@ -1569,6 +1569,11 @@ function _slice_series_args!(plotattributes::AKW, plt::Plot, sp::Subplot, comman
     return plotattributes
 end
 
+label_to_string(label::Bool, series_plotindex::Int) = label ? label_to_string(:auto, series_plotindex) : ""
+label_to_string(label::Nothing, series_plotindex::Int) = ""
+label_to_string(label::Missing, series_plotindex::Int) = ""
+label_to_string(label::Symbol, series_plotindex::Int) = label==:auto ?  "y$series_plotindex" : string((label==:none ? "" : label))
+label_to_string(label, series_plotindex::Int) = string(label)  # Fallback to string promotion
 function _update_series_attributes!(plotattributes::AKW, plt::Plot, sp::Subplot)
     pkg = plt.backend
     globalIndex = plotattributes[:series_plotindex]
@@ -1638,9 +1643,7 @@ function _update_series_attributes!(plotattributes::AKW, plt::Plot, sp::Subplot)
 
     # set label
     label = plotattributes[:label]
-    label = (label == "AUTO" ? "y$globalIndex" : label)
-    label = label === 0 ? "0" : label
-    label = label in (:none, nothing, false) ? "" : label
+    label = label_to_string(label, globalIndex)
     plotattributes[:label] = label
 
     _replace_linewidth(plotattributes)
--- a/src/args.jl
+++ b/src/args.jl
@@ -1569,11 +1569,11 @@ function _slice_series_args!(plotattributes::AKW, plt::Plot, sp::Subplot, comman
     return plotattributes
 end
 
-label_to_string(label::Bool, series_plotindex::Int) = label ? label_to_string(:auto, series_plotindex) : ""
-label_to_string(label::Nothing, series_plotindex::Int) = ""
-label_to_string(label::Missing, series_plotindex::Int) = ""
-label_to_string(label::Symbol, series_plotindex::Int) = label==:auto ?  "y$series_plotindex" : string((label==:none ? "" : label))
-label_to_string(label, series_plotindex::Int) = string(label)  # Fallback to string promotion
+label_to_string(label::Bool, series_plotindex) = label ? label_to_string(:auto, series_plotindex) : ""
+label_to_string(label::Nothing, series_plotindex) = ""
+label_to_string(label::Missing, series_plotindex) = ""
+label_to_string(label::Symbol, series_plotindex) = label==:auto ?  "y$series_plotindex" : string((label==:none ? "" : label))
+label_to_string(label, series_plotindex) = string(label)  # Fallback to string promotion
 function _update_series_attributes!(plotattributes::AKW, plt::Plot, sp::Subplot)
     pkg = plt.backend
     globalIndex = plotattributes[:series_plotindex]


daschw:
Thanks, consider
```julia
plot(rand(10), label = false)
```
with your changes though. It shows a plot with the label "0".
Maybe instead removing `false` from the line
```julia
label = label in (:none, nothing, false) ? "" : label
```
might be a more consistent approach. However this might break existing user code in an unexpected way. I'm not sure. What's your opinion on this?

Another option is to check for the type:
```julia
julia> 0 == false
true

julia> 0::Bool
ERROR: TypeError: non-boolean (Int64) used in boolean context
Stacktrace:
 [1] top-level scope at none:0

julia> false::Int
ERROR: TypeError: in typeassert, expected Int64, got Bool
Stacktrace:
 [1] top-level scope at none:0
```

daschw:
Also
```julia
julia> 0 === false
false
```
could be helpful.

LakshyaKhatri:
In removing `false` from
```julia
label = label in (:none, nothing, false) ? "" : label
```
a plot with `label=false` produces plot with the label "`False`".
`label === 0` works perfectly fine.

daschw:
> a plot with label=false produces plot with the label "False"

I know. The question is, if maybe we want this to be consistent with `label=true`? I'm not sure about this. Your solution is definitely less breaking, but somehow also less consistent. Although this inconsistency (`label=false` vs `label=true`) is really just a corner case.

LakshyaKhatri:
I got your point after the `label=true` example, you are telling the right thing. :thinking:

daschw:
@BeastyBlacksmith @isentropic what are your opinions on this?

isentropic:
I think label should checked for being bool or none, nothing else it should be parsed to a string first, [this](https://github.com/JuliaPlots/Plots.jl/issues/2834#issuecomment-652157850) might help on how to resolve if its a bool or not. It should be done early in the pipeline I think

isentropic:
```
[i in (:none, nothing, false) for i in [0, 1, true, false, :none]
```
returns
```
5-element Array{Bool,1}:
 1
 0
 0
 1
 1
```
This is somehow unintuitive, 0 is true, 1 is false. This is very confusing I suggest julian style like in [this comment](https://github.com/JuliaPlots/Plots.jl/issues/2834#issuecomment-652157850)

BeastyBlacksmith:
I think `label = true` should show the default label and not be the same as `label = "true"`.
Other than that I second the implementation in this PR

isentropic:
Nailing it cleanly with multiple dispatch seems to be a good solution to me. We can reuse these functions for other arguments like `title`

daschw:
That's a good idea.

LakshyaKhatri:
> I think `label = true` should show the default label and not be the same as `label = "true"`.
> Other than that I second the implementation in this PR
@BeastyBlacksmith Your doubt is corect, actually it produces a label with a string "True"

BeastyBlacksmith:
> Your doubt is corect, actually `label = true` produces a label with a string "True"

Probably only on the pyplot-backend, since julia `true` is python `True`.

LakshyaKhatri:
> Probably only on the pyplot-backend, since julia `true` is python `True`.

No, on gr it produces label "true" and on plotly it produces a label "trace 0".

BeastyBlacksmith:
That is a good reason to get rid of it ^^

isentropic:
Maybe I'm pushing an overkill here, but again, I do not think that doing any comparisons is a good idea.
Because:
`0.0 == false` gives `true`. I'm not sure what else can do this. Fairly sure if given hex, bit, longint 0, it would break. I think go for multiple dispatch would be better

isentropic:
I'm sorry for the confusion here, maybe I'm not understanding this issue at all :confused:

daschw:
something like
```julia
if label === true
    # default label
elseif label === false || label in (:none, nothing)
    # no label
else
    # convert to string
end
```
should do it, no?

LakshyaKhatri:
This seems a good idea!

BeastyBlacksmith:
Maybe at `missing` too in the no label case

isentropic:
```
label_to_string(label::Bool) = label ? label : ""
label_to_string(label::Nothing) = ""
label_to_string(label::Missing) = ""
label_to_string(label::PlotText # perhaps another case for future) = 
```

daschw:
> `label_to_string(label::Bool) = label ? label : ""`

we do not want this, as discussed above. this would show "true" or "True" as label. Otherwise, of course multiple dispatch is also an option, but not necessary IMO.

isentropic:
> > `label_to_string(label::Bool) = label ? label : ""`
> 
> we do not want this, as discussed above. this would show "true" or "True" as label. Otherwise, of course multiple dispatch is also an option, but not necessary IMO.
My mistake, it should have been,
```
label_to_string(label::Bool) = label ? string(label) : ""
```
but yeah, there are 1000 ways to skin a cat. Let's do it the way it is, but make sure that the label is promoted to `String` what do you think?

isentropic:
In case the pattern
```
if label === true
    # default label
elseif label === false || label in (:none, nothing)
    # no label
else
    # convert to string
end
```
repeats too many times for `title, guide, window_title` and such, it should be delegated to some function is `utils.jl` for sure. But let this be a long term maintainability goal

LakshyaKhatri:
I didn't understand the `label == "AUTO"` case!

isentropic:
> I didn't understand the `label == "AUTO"` case!

it is meant for auto `y1, y2, y3` assignment like in:
image

LakshyaKhatri:
Yes, I got that. I was telling that I might need to add:
```julia
if label === true || label == "AUTO"
    # default label
...
```
to handle the `AUTO` case

isentropic:
In addition, why is  `"AUTO"` not `:auto` even? Because there is currently no way to make it a literal `"AUTO"` string. @daschw has more experience for about this for sure.
But yeah, you are right "AUTO" case needs to be handled too.

LakshyaKhatri:
I don't know I should add on top of all this or not, but:
```julia
plot(rand(10); label=missing)
```
is giving:
```julia
ERROR: TypeError: non-boolean (Missing) used in boolean context
Stacktrace:
 [1] _update_series_attributes!(::RecipesPipeline.DefaultsDict, ::Plots.Plot{Plots.PyPlotBackend}, ::Plots.Subplot{Plots.PyPlotBackend}) at /home/aries/.julia/dev/Plots/src/args.jl:1641
 [2] add_series!(::Plots.Plot{Plots.PyPlotBackend}, ::RecipesPipeline.DefaultsDict) at /home/aries/.julia/dev/Plots/src/pipeline.jl:302
 [3] _process_seriesrecipe(::Plots.Plot{Plots.PyPlotBackend}, ::RecipesPipeline.DefaultsDict) at /home/aries/.julia/packages/RecipesPipeline/wolJ9/src/series_recipe.jl:44
 [4] _process_seriesrecipes!(::Plots.Plot{Plots.PyPlotBackend}, ::Array{Dict{Symbol,Any},1}) at /home/aries/.julia/packages/RecipesPipeline/wolJ9/src/series_recipe.jl:25
 [5] recipe_pipeline!(::Plots.Plot{Plots.PyPlotBackend}, ::Dict{Symbol,Any}, ::Tuple{Array{Float64,1}}) at /home/aries/.julia/packages/RecipesPipeline/wolJ9/src/RecipesPipeline.jl:95
 [6] _plot!(::Plots.Plot{Plots.PyPlotBackend}, ::Dict{Symbol,Any}, ::Tuple{Array{Float64,1}}) at /home/aries/.julia/dev/Plots/src/plot.jl:167
 [7] plot(::Array{Float64,1}; kw::Base.Iterators.Pairs{Symbol,Missing,Tuple{Symbol},NamedTuple{(:label,),Tuple{Missing}}}) at /home/aries/.julia/dev/Plots/src/plot.jl:57
 [8] top-level scope at REPL[77]:1
 [9] eval(::Module, ::Any) at ./boot.jl:331
 [10] eval_user_input(::Any, ::REPL.REPLBackend) at /buildworker/worker/package_linux64/build/usr/share/julia/stdlib/v1.4/REPL/src/REPL.jl:86
 [11] run_backend(::REPL.REPLBackend) at /home/aries/.julia/packages/Revise/BqeJF/src/Revise.jl:1184
 [12] top-level scope at REPL[3]:0
```
In our if-else case adding:
```julia
label in (:none, nothing, missing)
```
will throw the same error. We have to compare `missing` explicitly before comparing the label with `false` :(

BeastyBlacksmith:
best to use `ismissing` to check for missingness

isentropic:
Yeah, this is happening because `in` is syntactic sugar for comparison `==`, and there is no way around it.

I think dispatch would work best here because again, a user could throw at you completely unexpected type of label, but Plots should be malleable and provide  

And let us make "AUTO" into :auto too while we are here right? (@BeastyBlacksmith @daschw )

```
label_to_string(label::Bool) = label ? string(label) : ""
label_to_string(label::Nothing) = ""
label_to_string(label::Missing) = ""
label_to_string(label::Symbol) = label==:auto ?  "y$globalIndex" : string(label)   # Or make this String not Symbol if we decide to keep "AUTO"
# label_to_string(label::PlotText # perhaps another case for future) = # something
label_to_string(label) = string(label)  # Fallback to string promotion

```
This way it is more robust I think

LakshyaKhatri:
can't we use a Dict to return the corresponding value? (sorry in advance XD)

isentropic:
> can we use a `Dict` to return the corresponding value? (sorry in advance XD)

This is your PR after all, I think in theory you can make a mapping. If you think this would be functional and wouldn't break that's a solution too.

BeastyBlacksmith:
```julia
label_to_string(label::Bool) = label ? string(label) : ""
```
That should be
```julia
label_to_string(label::Bool) = label ? label_to_string(:auto) : ""
```
then.

isentropic:
Do not worry too much about all of this discussion, @LakshyaKhatri . This is not a major change, do what you think would pass all the tests and not break anything. If you satisfy all the potential use cases it would be good. Your goal after all is just get this part to work as expected.

daschw:
>  That should be
`label_to_string(label::Bool) = label ? label_to_string(:auto) : ""`
then.

`label_to_string` should also take the series index as input parameter then.

isentropic:
> > That should be
> > `label_to_string(label::Bool) = label ? label_to_string(:auto) : ""`
> > then.
> 
> `label_to_string` should also take the series index as input parameter then.

Do you think vectorizing with `.`(dot) syntax would just work?
```julia
label_to_string(label::Array) =[label_to_string(i) for in label]
```
?
```

**Comment by LakshyaKhatri** [comment]
I will go with the multiple dispatch method if you allow.

**Comment by LakshyaKhatri** [comment]
I have a few doubts about the current commit:
 - Have I defined the methods in the right place?
 - Is `Int` is correct type annotation for series index?
 - should I expand:
```julia 
label_to_string(label::Symbol, series_plotindex::Int) = label==:auto ?  "y$series_plotindex" : string((label==:none ? "" : label))
```
 - into more readable manner (or use another approach)?

src/args.jl
```diff
@@ -1569,6 +1569,11 @@ function _slice_series_args!(plotattributes::AKW, plt::Plot, sp::Subplot, comman
     return plotattributes
 end
 
+label_to_string(label::Bool, series_plotindex::Int) = label ? label_to_string(:auto, series_plotindex) : ""
+label_to_string(label::Nothing, series_plotindex::Int) = ""
+label_to_string(label::Missing, series_plotindex::Int) = ""
+label_to_string(label::Symbol, series_plotindex::Int) = label==:auto ?  "y$series_plotindex" : string((label==:none ? "" : label))
+label_to_string(label, series_plotindex::Int) = string(label)  # Fallback to string promotion
```
daschw, line 1576: You don't need a type annotation for series_plotindex in all these cases

src/args.jl
```diff
@@ -1638,8 +1643,7 @@ function _update_series_attributes!(plotattributes::AKW, plt::Plot, sp::Subplot)
 
     # set label
     label = plotattributes[:label]
-    label = (label == "AUTO" ? "y$globalIndex" : label)
-    label = label in (:none, nothing, false) ? "" : label
+    label = label_to_string(label, globalIndex)
     plotattributes[:label] = label
```
daschw, line 1647: Instead of the three lines above, you could just do `plotattributes[:label] = label_to_string(plotattributes[:label], globalIndex)`. Both is fine for me though.

LakshyaKhatri: I was doing the same thing, but then thought it will complicate the code :sweat_smile:

daschw:
The tests fail because of line 80 in test/test_pgfplotsx.jl. Apparently now 5 instead of 6 label entries are returned. Not exactly, sure why this changed here and what's the correct value, 5 or 6.

BeastyBlacksmith:
> `label_to_string(label::Symbol, series_plotindex::Int) = label==:auto ?  "y$series_plotindex" : string((label==:none ? "" : label))`

IMO we shouldn't convert symbols to strings here, because we use symbols for special actions and adding any other action later on would be a breaking change then.
So I would write this as
```
function label_to_string(label::Symbol, series_plotindex)
    if label==:auto
        return string("y", series_plotindex)
    elseif label==:none
        return ""
    else
        throw(ArgumentError("unsupported symbol $(label) passed to `label`")
    end
end
```

src/args.jl
LakshyaKhatri: Done

src/args.jl
```diff
@@ -1637,10 +1651,7 @@ function _update_series_attributes!(plotattributes::AKW, plt::Plot, sp::Subplot)
     end
 
     # set label
-    label = plotattributes[:label]
-    label = (label == "AUTO" ? "y$globalIndex" : label)
-    label = label in (:none, nothing, false) ? "" : label
-    plotattributes[:label] = label
+    plotattributes[:label] = label_to_string(plotattributes[:label], globalIndex)
```
BeastyBlacksmith, line 1654:
```suggestion
    plotattributes[:label] = label_to_string.(plotattributes[:label], globalIndex)
```
This should fix the tests.

LakshyaKhatri: will it be okay to ask how? (because it was working even without broadcasting)

LakshyaKhatri:
will it be okay to ask how? (because it was working even without broadcasting) :thinking:

BeastyBlacksmith:
It did fall back to
```julia
julia> string([1,2,3])
"[1, 2, 3]"
```
before

BeastyBlacksmith:
Well done

daschw:
Thanks @LakshyaKhatri !

PR #2844, opened 2020-07-03
fix errorbar color

daschw:
```julia
using Plots

x = [0.5:10 1:10]
y = rand(10, 2) .+ [0 1]
err = [(rand(10), rand(10)) (rand(10), rand(10))]

plot(scatter(x, y, yerror = err), scatter(x, y, yerror = err, msc = :auto))
```

Before
errorbar_old

Now
errorbar_new

### DIFF
--- a/src/examples.jl
+++ b/src/examples.jl
@@ -973,8 +973,8 @@ const _examples = PlotExample[
                     surf = Measurement.((1:10) .* (1:10)', rand(10,10))
 
                     plot(
-                        scatter(x, [x y], msw = 0),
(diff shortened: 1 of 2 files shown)


daschw:
```julia
plot(plot(x, y, yerror = err), plot(x, y, yerror = err, msc = :auto))
```

Before
errorbar_line_old

Now
errorbar_line_new

daschw:
I think it would make more sense if the left plots where `markerstrokecolor = :auto` and the right ones are `markerstrokecolor = :match`. In our current implementation it's the other way around.

PR #2854, opened 2020-07-09
Horizontal legends

BeastyBlacksmith:
While preparing an implementation for horizontal legends in the pgfplotsx backend I got an idea.

Because I wanted to avoid creating another keyword, I hijacked the `legend` keyword, so now I can create legend columns like
```
plot(rand(4,4), legend = :horizontal)
```
h

```
plot(rand(4,4), legend = 2)
```
h2

However, now I can't change the position of the legend at the same time.
So my idea was to always convert the input to named tuples, something like
|`legend` in | `legend` out |
|---|---|
| `:top` | `(pos = :top, columns = 1)`
| `:horizontal` | `(pos = :best, columns = -1)` |
| `(:top, 2)` | `(pos = :top, colums = 2)` |

And passing the NamedTuple would be also allowed. Do you think this is a good idea?

### DIFF
--- a/src/args.jl
+++ b/src/args.jl
@@ -1208,12 +1208,16 @@ function convertLegendValue(val::Symbol)
         :none
     elseif val in (:right, :left, :top, :bottom, :inside, :best, :legend, :topright, :topleft, :bottomleft, :bottomright, :outertopright, :outertopleft, :outertop, :outerright, :outerleft, :outerbottomright, :outerbottomleft, :outerbottom, :inline)
         val
+    elseif val == :horizontal
+        -1
     else
         error("Invalid symbol for legend: $val")
     end
 end
+convertLegendValue(val::Real) = val
 convertLegendValue(val::Bool) = val ? :best : :none
 convertLegendValue(val::Nothing) = :none
+convertLegendValue(v::Union{Tuple, NamedTuple}) = convertLegendValue.(v)
 convertLegendValue(v::Tuple{S,T}) where {S<:Real, T<:Real} = v
 convertLegendValue(v::AbstractArray) = map(convertLegendValue, v)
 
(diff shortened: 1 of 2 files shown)


daschw:
I like it.

isentropic:
perhaps we could have something line `legend_layout` that is by default `:vertical`
or it could be a tuple for specific dimensions or

isentropic:
good idea I think

isentropic:
To make it even more simpler, it could just be called, `legend_columns` referring to reqiured columns number, or possible symbol `:horizontal` to indicate that it should 1 row only.

aramirezreyes:
I just came here to say I am very interested in this feature.

BeastyBlacksmith:
I dicided to not go the named tuple road and instead make handling of keywords more ergonomic. Unfortunately at the moment this would be breaking unless we can make recipes consume aliased keywords as well.

BeastyBlacksmith:
@daschw I have a hard time coming with critical tests, but this one works, but probably I got wrong what you meant with color matching.
```
julia> struct A end
julia> @recipe function f(a::A)
         legendfontcolor --> :magenta
         1:5
       end

julia> plot(A())
```

BeastyBlacksmith:
At this point I'd really like to make this change, which is why I am considering this ready for review. I am trying to summarize this PR:

## Frontend reorganization
 - key component of this PR is the macro `@add_attributes` (which I added earlier)
   - this takes a `struct` definition and recurses into its fields to create keywords by chaining the field names with the structs' name with underscore
   - it also creates the pluralized and non-underscore aliases for these keywords
 - as the first example this is done for the new `Legend` struct, but in the long run I imagine all of Plots' keywords corresponding to components being created this way
 - as a consequence we have consistent naming for these keywords, though we need to change them throughout the codebase which is potentially breaking, but is intended to be kept non-breaking via aliasing
 - thats why this removes the warning about aliases in recipes (good examples for where this still breaks sought)
 
## New functionality
 - additionally this implements placing legend entries in columns implemented in the `pgfplotsx` backend (including horizontal legends)

Issue #2120, opened 2019-08-03
How to change the formula font to "CMU-Serif"?


Photonico:
In Plots, I can change the font to "CMU-Serif".
But when I use latex, the font cannot be changed.
How can I change it?

![Problem](https://raw.githubusercontent.com/ConAntares/Problems/master/2019/20190803a.png)

Another question is how to change the color and line weight of axes, border, and the border of label?

lnacquaroli:
Since you are using `LaTeXStrings`, have you tried using `L"ABC \alpha"` instead of `raw`. I believe the docs answer the other questions.

lnacquaroli:
I just noticed you did in one ...

ConAntares:
> 
> 
> I just noticed you did in one ...

I am writing my article. This problem is bothering me.
Do you plan to solve it? or when will you solve it?

JackDevine:
I am not completely sure that I understand your problem.

Have you tried using PGFPlots? That backend can produce LaTeX code that you can `\input` into your document and the plot will inherit the font from the rest of your article.

ConAntares:
> 
> 
> I am not completely sure that I understand your problem.
> 
> Have you tried using PGFPlots? That backend can produce LaTeX code that you can `\input` into your document and the plot will inherit the font from the rest of your article.

My problem is that the font of formula is uncontrolled, I want to change it to CMU Serif. Your method is too inconvenient.

ConAntares:
> 
> 
> I just noticed you did in one ...

I have solved with:
`PyPlot.rc("text", usetex = "true")
PyPlot.rc("font", family = "CMU Serif")`

But, I cannot find which attribution control the border of the label.
![Problem](https://raw.githubusercontent.com/ConAntares/Problems/master/2019/20190804d.png)

JackDevine:
Looks good!

`fg_legend=:transparent` should remove the box around the legend.

ConAntares:
> 
> 
> Looks good!
> 
> `fg_legend=:transparent` should remove the box around the legend.

Thanks a lot for your help,
I also have two questions,
First:  In the legend, the line's endpoint is not smooth. 
![1](https://raw.githubusercontent.com/ConAntares/Problems/master/2019/20190804e.png)
Second: The width of Top/Right borders and Bottom/Left borders are different.
![2](https://raw.githubusercontent.com/ConAntares/Problems/master/2019/20190804f.png)

Could you tell me the solution or fix these bugs?

[...]

Issue #2042, opened 2019-05-30
Plotting with Serif Font, scientific notation


isentropic:
Following [897](https://github.com/JuliaPlots/Plots.jl/issues/897), I attempted to make the plot look different,
```
using Plots
pyplot()
upscale = 1 #8x upscaling in resolution
fntsm = Plots.font("Latin Modern Math", 10.0*upscale)
fntlg = Plots.font("Latin Modern Math", 14.0*upscale)
default(titlefont=fntlg, guidefont=fntlg, tickfont=fntsm, legendfont=fntsm)
default(size=(800*upscale,600*upscale)) #Plot canvas size
default(framestyle=:box)
default(dpi=300) #Only for PyPlot - presently broken
```

The problem is that backends have trouble rendering powers of 10:
![image](https://user-images.githubusercontent.com/24591123/58609794-e0aafc80-82e3-11e9-820a-0f7904f9e10a.png)

PyPlot handles this as follows (example):
![image](https://user-images.githubusercontent.com/24591123/58609813-020be880-82e4-11e9-8623-ad0f9d5f8c7e.png)

Are there any additional parameters that require tuning?

isentropic:
Additionally,
`pyplotjs()` handles this as follows:
![image](https://user-images.githubusercontent.com/24591123/58609979-b6a60a00-82e4-11e9-80ee-75f96f1d901c.png)

`gr()` does not change the fonts at all, and makes the picture huge (this was mentioned in previous issues)

ma-laforge:
Hi @isentropic,

I just want to start off by explaining there was a ***slight*** bug with the code snippet you used above.  You should replace the `fntsm` & `fntlg` expressions with the following:

```
fntsm = Plots.font("sans-serif", pointsize=round(10.0*upscale))
fntlg = Plots.font("sans-serif", pointsize=round(14.0*upscale))
```
(see comments by nathancunn in #897 for explanations)

Also, I assume you generated the above plots with code similar to the following:
```
x = 0:1e-6:6e-6
y1 = 0 .* x
y2 = y1 .+ 1

plot(x, [y1, y2])
```
If not, please provide such code for clarity.

ma-laforge:
As for the trouble rendering powers of 10:

### Background
[...]

PR #2858, opened 2020-07-13
pyplot serif fonts fixed

isentropic:
perhaps someone remembers old discussion here https://github.com/JuliaPlots/Plots.jl/issues/2042

I took a crack on solving it based on https://matplotlib.org/tutorials/text/mathtext.html 

The issue is that maplotlib does not really allow specify `mathtext.fontspec` per element, so I made every tick be surrounded with `\mathregular` which solves this issue.
Now pyplot can handle `serif` font somewhat intuitively.


isentropic:
```

julia> plot(10 .^(1:9), fontfamily="serif")

julia> plot(10 .^(1:9), fontfamily="sans")

```
image
image

isentropic:
Perhaps fixes https://github.com/JuliaPlots/Plots.jl/issues/2120 too

PR #2873, opened 2020-07-23
Update SnoopCompile.yml

BeastyBlacksmith:
I don't think we need to run on all branches and nightly.
Also we don't have benchmarks, right?


daschw:
We have https://github.com/JuliaPlots/Plots.jl/blob/master/deps/SnoopCompile/snoop_bench.jl but we can run the bot only on pushes to master.

isentropic:
rc2 just came out too.

Issue #2741, opened 2020-05-29
[FR] Some Ideas for PGFPlotsX


benneti:
Hi together,

thank you for creating this library it really makes plotting a breeze!
As I cannot comment in the PGFPlotsX meta issue (and I think this discussion may fit there I will simply comment here, all of this is with Plots v1.3.5).
All of this can be discussed with the minimal working example:

``` julia
using LinearAlgebra
using Statistics
using Plots
pgfplotsx()
default(tickfontsize=11, legendfontsize=11, guidefontsize=11, framestyle=:box)
default(gridlinewidth=1)
default(linewidth=2)
# additional packages for labels
push!(PGFPlotsX.CUSTOM_PREAMBLE, raw"\usepackage{amsmath}")
push!(PGFPlotsX.CUSTOM_PREAMBLE, raw"\usepackage{physics}")
push!(PGFPlotsX.CUSTOM_PREAMBLE, raw"\usepackage{nicefrac}")
using LaTeXStrings
φ(ω, γ) = (tmp = atan(- ω*γ/(1-ω^2)); tmp > 0 ? tmp : tmp+π)
function latex(n::Rational; mulnum="", mulden="")
    num = mulnum != "" && n.num == 1 ? mulnum : "$(n.num)$(mulnum)"
    if n.den == 1 && mulden == ""
        return num
    end
    "\\frac{$(num)}{$(n.den)$(mulden)}"
end

zticks =  (π/4*(0:4), [LaTeXString(latex(n//4;mulnum="\\pi")) for n in 0:4])
pl = surface(range(0,5, length=20), range(0,3, length=20), φ, label="", color=:viridis)
title!(L"Solution for ($s\gg1$) $\ddot{x}(s) + \gamma \dot{x}(s) + x(s) = f \sin(\omega s)$")
xaxis!(L"Driving Frequency $\omega$")
yaxis!(L"Damping $\gamma$")
plot!(zlabel=L"Phaseshift $\varphi$", zticks=zticks)
# HACK for more features:
# this is necessary to access the plot directly
savefig("tmp.tex") # need to save once to create the_plot
pl.o.the_plot.elements[1].elements[1].options["colorbar style"] = PGFPlotsX.Options(
    "title" => pl.subplots[1][:colorbar_title],
    "xticklabel style" => Plots.pgfx_get_ticklabel_style(pl.subplots[1], pl.subplots[1][:xaxis]),
[...]

PR #2880, opened 2020-07-28
Update series args using plot!

### LINKED ISSUE #2851: [FR] Way to modify series labels after plot creation.
Currently labels have to be added at the same time the series data is plotted. Other attributes like titles and axis labels can be added and modified after the fact. I would like to be able to add and modify the legend after plot creation also. 

Ideally, this would be done in the same way as other attributes like `plot!(labels=stringarrayoflabels)`, but a new function like `labels!(stringarrayoflabels)` could work too.

This would be most helpful to me when a plot is created within a function, but I want to be able to pretty it up outside of the function to avoid passing a lot of variables back and forth.

Diaga:
Fixes #2851 as well.

```julia
begin
    default(titlefont = (20, "times"), legendfontsize = 18, guidefont = (18, :darkgreen), tickfont = (12, :orange), guide = "x", framestyle = :zerolines, yminorgrid = true)
    plot([sin, cos], -2π, 2π, label = ["sin(θ)" "cos(θ)"], title = "Trigonometric Functions", xlabel = "θ", linewidth = 2, legend = :outerbottom)
    plot!(label=["cos(θ)" "sin(θ)"], arrow=(:closed, 2.0))
end
```

Before:
2851-before

After: (Notice changed legend labels and arrows)
2851-after

### DIFF
--- a/.github/workflows/SnoopCompile.yml
+++ b/.github/workflows/SnoopCompile.yml
@@ -3,7 +3,7 @@ name: SnoopCompile
 on:
   push:
     branches:
-    #  - 'master'  # NOTE: to run the bot only on pushes to master
+      - 'master'  # NOTE: to run the bot only on pushes to master
 
 defaults:
   run:
@@ -24,7 +24,6 @@ jobs:
           - '1.3'
           - '1.4'
           - '1.5.0-rc1'
-          - 'nightly'
         os:        # NOTE: should match the os setting of your botconfig
           - ubuntu-latest
           - windows-latest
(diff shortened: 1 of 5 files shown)


Diaga:
May need documentation somewhere, if it should be merged.

BeastyBlacksmith:
Would be nice to add some tests.
This doesn't address the case mentioned in https://github.com/JuliaPlots/Plots.jl/issues/2741, does it?

Diaga:
Was tricky setting up LaTeX. No, it does not address #2741. 
Previously, series args could not be modified using ```plot!``` after initial plot creation. This PR allows updating of series.

BeastyBlacksmith:
Just linking my PR for a few cases that one has to consider: https://github.com/MLH-Fellowship/Plots.jl/pull/26

PR #2882, opened 2020-07-29
Add series type pixel

### LINKED ISSUE #2859: [FR] Series type pixel
Create a new series type that draws pixels. It is similar to `st=:scatter` with smallest size and no border if backend cannot draw pixels natively. PyPlot and GR are capable of this where the benefit is vastly faster plotting for huge data where line plots take too long

LakshyaKhatri:
Fixes #2859

### DIFF
--- a/src/recipes.jl
+++ b/src/recipes.jl
@@ -178,7 +178,7 @@ end
     x := x
     y := y
     markersize := 1
-    markerstrokewidth := 0
+    markerstrokewidth --> 0
     seriestype := :scatter
     ()
 end
--- a/src/recipes.jl
+++ b/src/recipes.jl
@@ -174,6 +174,14 @@ end
 end
 @deps scatterpath path scatter
 
+@recipe function f(::Type{Val{:pixel}}, x, y, z)
+    x := x
+    y := y
+    markersize := 1
+    markerstrokewidth := 0
+    seriestype := :scatter
+    ()
+end
 
 # ---------------------------------------------------------------------------
 # steps
--- a/src/recipes.jl
+++ b/src/recipes.jl
@@ -174,12 +174,15 @@ end
 end
 @deps scatterpath path scatter
 
+# ---------------------------------------------------------------------------
+# pixel
+
 @recipe function f(::Type{Val{:pixel}}, x, y, z)
     x := x
     y := y
+    seriestype := :pixel
     markersize := 1
     markerstrokewidth --> 0
-    seriestype := :scatter
     ()
 end
 
--- a/src/recipes.jl
+++ b/src/recipes.jl
@@ -185,6 +185,7 @@ end
     markerstrokewidth --> 0
     ()
 end
+@deps pixel scatter
 
 # ---------------------------------------------------------------------------
 # steps
--- a/src/recipes.jl
+++ b/src/recipes.jl
@@ -180,9 +180,9 @@ end
 @recipe function f(::Type{Val{:pixel}}, x, y, z)
     x := x
     y := y
-    seriestype := :pixel
-    markersize := 1
-    markerstrokewidth --> 0
+    seriestype := :scatter
+    markersize := 0.1
+    markerstrokewidth := 0
     ()
 end
 @deps pixel scatter
--- a/src/recipes.jl
+++ b/src/recipes.jl
@@ -181,7 +181,7 @@ end
     x := x
     y := y
     seriestype := :scatter
-    markersize := 0.1
+    markersize := 0.5
     markerstrokewidth := 0
     ()
 end
--- a/src/backends.jl
+++ b/src/backends.jl
@@ -355,6 +355,7 @@ const _gr_seriestype = [
     :wireframe,
     :volume,
     :shape,
+    :pixel,
 ]
 const _gr_style = [:auto, :solid, :dash, :dot, :dashdot, :dashdotdot]
 const _gr_marker = _allMarkers
@@ -566,6 +567,7 @@ const _pyplot_seriestype = [
     :scatter3d,
     :surface,
     :wireframe,
+    :pixel,
 ]
 const _pyplot_style = [:auto, :solid, :dash, :dot, :dashdot]
 const _pyplot_marker = vcat(_allMarkers, :pixel)
--- a/src/backends/gr.jl
+++ b/src/backends/gr.jl
@@ -1683,6 +1683,8 @@ function gr_display(sp::Subplot{GRBackend}, w, h, viewport_canvas)
 
         elseif st == :pixel
             GR.setmarkertype(GR.MARKERTYPE_DOT)
+            lc = get_linecolor(series, clims, 0)
+            gr_set_markercolor(lc)
             GR.polymarker(x, y)
 
         elseif st == :contour
--- a/src/backends/gr.jl
+++ b/src/backends/gr.jl
@@ -568,7 +568,7 @@ gr_view_ycenter(viewport_plotarea) = 0.5 * (viewport_plotarea[3] + viewport_plot
 
 function gr_legend_pos(sp::Subplot, w, h, viewport_plotarea)
     legend_leftw, legend_rightw, legend_textw, x_legend_offset = w
-    legend_dy, legendh, y_legend_offset = h 
+    legend_dy, legendh, y_legend_offset = h
     s = sp[:legend]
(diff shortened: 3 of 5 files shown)


src/recipes.jl
```diff
@@ -174,6 +174,18 @@ end
 end
 @deps scatterpath path scatter
 
+# ---------------------------------------------------------------------------
+# pixel
+
+@recipe function f(::Type{Val{:pixel}}, x, y, z)
```
BeastyBlacksmith, line 180: I don't think this recipe is needed.
Anyways, you can't return the same seriestype in a recipe with which you entered.
I'd recommend to have a look at daniels blog post: https://daschw.github.io/recipes/#how_do_recipes_actually_work

LakshyaKhatri: Thanks @BeastyBlacksmith :innocent: , I created a recipe as per the discussion with @isentropic.
I will be happy to hear new ideas. :smile:

BeastyBlacksmith: you can just delete it, should work without it

BeastyBlacksmith: `:pixel` is a "Atomic" seriestype so to speak

isentropic: do not we want to fallback to small markersizes if a backend can't support it?

isentropic: I doubt that all backends can support this natively

LakshyaKhatri: I will check it for each backend and then I will make the changes accordingly.

BeastyBlacksmith: in that case the output seriestype should be `:scatter`

isentropic: right, it should fallback to scatter

src/recipes.jl
```diff
@@ -174,6 +174,18 @@ end
...
+    x := x
+    y := y
+    seriestype := :pixel
```
BeastyBlacksmith, line 183:
```suggestion
    seriestype := :scatter
```

isentropic: Perhaps I do not understand the recipe system, but in a sense, pixel could be supported natively by both GR and Pyplot. However, others might not support it, in that case it should fallback to :scatter. Are we on the same page?

LakshyaKhatri: Hello @isentropic, sorry I am not able to work today. I have high fever. As soon as I start feeling better I will quickly read the recipe resource shared by @BeastyBlacksmith, then I might be able to understand things better.

LakshyaKhatri: @BeastyBlacksmith, since pyplot and gr supports the pixel plots natively, we can update the natively supported series to handle `:pixel`. For the other plots we can add pixel series to them and make them fallback to a scatter plot.

isentropic: Of course, feel free to take your time

BeastyBlacksmith: Backends declare in `backends.jl` which seriestypes they support natively and for those this recipe will not be called.

LakshyaKhatri: Hello @isentropic, I am finally recovered from the fever and continuing my work on this PR. I will push some changes soon.

LakshyaKhatri: So I read the resource shared by @BeastyBlacksmith. He was right about the `seriestype := :pixel` thing, that it should be changed to `seriestype := :scatter`. I guess, this is due to the recursive calling of different recipes.

LakshyaKhatri:
@isentropic, can you suggest me the native way for plotting pixel in gr? Right now I am using the default scatter plot with significantly small marker size. 

Also, @BeastyBlacksmith was right about the recipe. We don't need a recipe for this purpose we just have to update the natively supported series type and handle them in the backend files.

isentropic:
The recipe is needed as a fallback for backends that do not support pixel native lly. As for gr @jheinen should know about it more. I only saw him oncee mentioning about it. 

@jheinen Basically pixel, is a series type with smallest possible drawable unit. Naturally it doesn't have stroke only the color. In pyplot you can see it as `plot(Rand(10), ",")`

LakshyaKhatri:
> The recipe is needed as a fallback for backends that do not support pixel native lly.
@isentropic Right, I figured this out just after posting that comment :sweat_smile:

LakshyaKhatri:
We might need tests too for this feature, right?

BeastyBlacksmith:
I think adding an example would be a good idea

jheinen:
GR supports the drawing of "pixels" using `GR.polymarker()` with the marker style `MARKERTYPE_DOT` (`GR.setmarkertype(GR.MARKERTYPE_DOT)`). For most output types (PostScript, PDF, SVG, ...) however, there are no "pixels". In such cases GR "emulates" plotting of dots using "small" circles or stroke segments depending on the output resolution.

LakshyaKhatri:
Hello @isentropic, I tried the method suggested above, but plot results are not that good. As @jheinen said, in GR the plot is not actually a pixel, hence it blurs out on zooming or maximizing the window.

LakshyaKhatri:
Here are the results for `plot([sin, cos], st=:pixel)` GR:
Screenshot from 2020-08-04 18-22-17

_I am working on the colors and labels rn_

LakshyaKhatri:
cc: @isentropic

isentropic:
Let's see how labels and colors work out. Legend markers do not need to be point size too. Perhaps smaller than a typical marker, however not so small that color is not discernible.

LakshyaKhatri:
I have never worked on GR before (and GR is strange). I have made the colors to work but, I'm not sure if it is the standard way. Please guide me if I make a mistake at any point.

LakshyaKhatri:
In GR we are only plotting markers and hence (I guess) legend keys are not showing up for plots. Could we do something about that?

cc: @jheinen

LakshyaKhatri:
@BeastyBlacksmith could you tell me what type of example should I add?

BeastyBlacksmith:
Add one here:
https://github.com/JuliaPlots/Plots.jl/blob/806d4671ce7a0fe0f2e56f264c895c816c86f308/src/examples.jl#L11

Make sure you add to the end of the array. After that is done, it is necessary to update PlotReferenceImages like @Diaga did here: https://github.com/JuliaPlots/PlotReferenceImages.jl/pull/81

LakshyaKhatri:
Yes, I figured that out somehow. I am just confused that the example needs to be complex or what series should I plot?

BeastyBlacksmith:
What it shows is up to you, it should be reproducible and well suited to catch unwanted changes. If you can cover other functionality of Plots that isn't tested in any other example thats also good.
Bonus points if it looks pretty ( and the code is not too long ).

LakshyaKhatri:
> In GR we are only plotting markers and hence (I guess) legend keys are not showing up for plots. Could we do something about that?

I have tried almost everything, could I ask for some pointers?

cc: @isentropic

isentropic:
This is tricky, I'm not entirely sure about this. I'm a bit uncertain if this deserves a new recipe or it is just a new markershape.
It is not a markershape because it cannot have a stroke. 
It is a markershape because pyplot treats it as such.

I'm trying to connect all of this to Plots API that assumes that everymarker can have a stroke and a fill. But [matplotlib](https://matplotlib.org/3.3.0/api/_as_gen/matplotlib.markers.MarkerStyle.html#matplotlib.markers.MarkerStyle) clearly distinguishes fillable and plain markers

LakshyaKhatri:
What should we do then? Should we add a new markertype?

daschw:
We already have markers that have a stroke, but no fill (`:hline`, `:vline`, ...). So I don't really see a reason not to have a marker with just a fill and without stroke.

daschw:
Actually, markers `:+`, `:hline`, `:vline` change their color with `markercolor` and not with `markerstrokecolor`, so I guess we already have markers that only have a fill, but no stroke.

LakshyaKhatri:
@isentropic, should we add a new markertype instead? I might need to look for some documentation before.

isentropic:
@LakshyaKhatri then I guess this should be a new marker type then

isentropic:
Perhaps called as
`plot(Rand(10), m=:pixel)`

daschw:
I would go for a new markertype. I don't really see how `:pixel` is a different seriestype from `:scatter`. For GR this should (hopefully) just be a matter of adding `pixel = X,` to https://github.com/JuliaPlots/Plots.jl/blob/master/src/backends/gr.jl#L14-L39, where X is the value corresponding to  `GR.MARKERTYPE_DOT`.

PR #2893, opened 2020-08-06
text.usetex option hack pyplot

### LINKED ISSUE #2889: [BUG] Using pyplot backend and matplotlibrc with `text.usetex : True`, #2858 breaks tick labels
## Details

My matplotlibrc contains the line `text.usetex : True`. I don't totally understand what #2858 did, but tick labels are broken for me now. Tick labels are correct when just using the PyPlot package.

The first problem is just that it's not using Computer Modern (or Latin Modern) on the ticks, which makes me think it's not even necessarily calling LaTeX. Output of `plt1 = plot(1:5, title = "Test plot 1"); png(plt1, "test1")`:
test1

A further problem, it's just not parsing things correctly somehow as can be seen in the output of `plt2 = plot(1:5, title = "Test plot 2", xticks = ([2,4], ["two", "four"])); png(plt2, "test2")`:
test2

(my apologies to @isentropic, I wasn't around when you asked for feedback on this PR)

### Backends

PyPlot

### Versions

Plots.jl version: 1.5.6 and 1.5.7
Backend  version (`]st -m`): PyPlot v2.9.0
Output of `versioninfo()`:
```
Julia Version 1.5.0-rc1.0
Commit 24f033c951 (2020-06-26 20:13 UTC)
Platform Info:
  CPU: AMD Ryzen 5 3600 6-Core Processor
  WORD_SIZE: 64
  LIBM: libopenlibm
  LLVM: libLLVM-9.0.1 (ORCJIT, znver1)
```

### LINKED ISSUE #2890: [BUG] xticks property does not work with pyplot backend anymore
## Note:
My issue might be related or even a duplicate of #2889 

## Details
With one of the latest updates using the pyplot backend Julia, gives the
> ERROR: LoadError: MethodError: no method matching getindex(::Int64, ::UnitRange{Int64}) 

when plotting and saving a plot that contains the xticks property:

```
using Plots

x = 0:3600:5*86400
y = x

pyplot() # commenting this out makes the code work
plt = plot( x, y, xticks=( 0:86400:5*86400, 0:5))

savefig( plt, "test.png")
```

### Versions: Julia 1.5.0 and all latest packages

isentropic:
@benide 
Resolves #2890 
Resolves #2889

### DIFF
--- a/src/backends/pyplot.jl
+++ b/src/backends/pyplot.jl
@@ -861,7 +861,11 @@ function py_set_ticks(ax, ticks, letter, env)
         axis."set_ticks"(ticks)
     elseif ttype == :ticks_and_labels
         axis."set_ticks"(ticks[1])
-        tick_labels = [py_surround_latextext(ticklabel, env) for ticklabel in ticks[2]]
+        if pyrcparams["text.usetex"]
+            tick_labels = ticks[2]
+        else
+            tick_labels = [py_surround_latextext(ticklabel, env) for ticklabel in ticks[2]]
+        end
         axis."set_ticklabels"(tick_labels)
     else
         error("Invalid input for $(letter)ticks: $ticks")
@@ -1124,7 +1128,7 @@ function _before_layout_calcs(plt::Plot{PyPlotBackend})
             # Set ticks
             fontProperties = PyPlot.PyCall.PyDict(Dict("family" => axis[:tickfontfamily],
                                                        "size" => py_thickness_scale(plt, axis[:tickfontsize]),
-                                                       "useTex" => false,
+                                                       # "useTex" => false,
                                                        "rotation" => axis[:tickfontrotation]))
             getproperty(ax, Symbol("set_",letter,"ticklabels"))(getproperty(ax, Symbol("get_",letter,"ticks"))(), fontProperties)
 
@@ -1134,8 +1138,11 @@ function _before_layout_calcs(plt::Plot{PyPlotBackend})
 
             # Tick marks
             intensity = 0.5  # This value corresponds to scaling of other grid elements
-            pyaxis."set_tick_params"(direction = axis[:tick_direction] == :out ? "out" : "in", width=py_thickness_scale(plt, intensity),
-                                     length= 5 * py_thickness_scale(plt, intensity))
+            pyaxis."set_tick_params"(
+                direction = axis[:tick_direction] == :out ? "out" : "in",
+                width=py_thickness_scale(plt, intensity),
+                length= 5 * py_thickness_scale(plt, intensity)
+                                     )
 
             getproperty(ax, Symbol("set_", letter, "label"))(axis[:guide])
             if get(axis.plotattributes, :flip, false)
--- a/src/backends/pyplot.jl
+++ b/src/backends/pyplot.jl
@@ -842,7 +842,15 @@ function py_set_lims(ax, sp::Subplot, axis::Axis)
     getproperty(ax, Symbol("set_", letter, "lim"))(lfrom, lto)
 end
 
-py_surround_latextext(latexstring, env) = PyPlot.LaTeXStrings.latexstring(env, "{",latexstring[2:end-1],"}")
+function py_surround_latextext(latexstring, env)
+    if latexstring[1] == '$' && latexstring[end] == '$'
+        unenclosed = latexstring[2:end-1]
+    else
+        unenclosed = latexstring
+    end
+    PyPlot.LaTeXStrings.latexstring(env, "{", unenclosed, "}")
+end
+
 
 function py_set_ticks(ax, ticks, letter, env)
     ticks == :auto && return
@@ -861,11 +869,13 @@ function py_set_ticks(ax, ticks, letter, env)
         axis."set_ticks"(ticks)
     elseif ttype == :ticks_and_labels
         axis."set_ticks"(ticks[1])
+
         if pyrcparams["text.usetex"]
             tick_labels = ticks[2]
         else
             tick_labels = [py_surround_latextext(ticklabel, env) for ticklabel in ticks[2]]
         end
+
         axis."set_ticklabels"(tick_labels)
     else
         error("Invalid input for $(letter)ticks: $ticks")


benide:
@isentropic Works exactly as expected for both issues using LaTeX for rendering. When *not* using LaTeX for the rendering, #2890 has the same result as before, i.e., the interface has different requirements for tick labels with the PyPlot backend as compared with the other backends. This doesn't affect me, because I use the LaTeX renderer if I'm using PyPlot, but just so you're aware.

I think this PR is strictly an improvement regardless of my above caveat, but I wonder if it makes sense to open a low priority issue for this inconsistency issue?

BeastyBlacksmith:
I still haven't understood. Can you give a short example of the difference of the backends?

wg030:
> I still haven't understood. Can you give a short example of the difference of the backends?

@BeastyBlacksmith I think what @benide means is that, as shown in my issue #2890, right now
```
pyplot()
plt = plot( x, y, xticks=( 0:86400:5*86400, 0:5))
#plt = plot(x, y, xticks=( 0:86400:5*86400, ["\$0\$", "\$1\$", "\$2\$", "\$3\$", "\$4\$"])) # This is the only way to make it work!
```
creates an error, but
```
#pyplot()
plt = plot( x, y, xticks=( 0:86400:5*86400, 0:5))
```
works like a charm. This is inconsistent and I strongly agree that no matter how this issues are solved, the choice of the backend should not require the user to change the syntax.

BeastyBlacksmith:
I agree that
```julia
pyplot()
plt = plot( x, y, xticks=( 0:86400:5*86400, 0:5))
```
should not error.

isentropic:
I think [this function](https://github.com/JuliaPlots/Plots.jl/blob/8dcff1f73d24679b0011bf29da7b123862d782c6/src/backends/pyplot.jl#L845) needs to be smarter about checking for labels to contain `$$` signs around it. Thats it. I will see how can I make that check, because simple `==` comparison did not work for me

wg030:
> I think [this function](https://github.com/JuliaPlots/Plots.jl/blob/8dcff1f73d24679b0011bf29da7b123862d782c6/src/backends/pyplot.jl#L845) needs to be smarter about checking for labels to contain `$$` signs around it.

@isentropic One could try to modify this function, but due to the its name and the name of its first argument I would prefer to look at where it was called. So I would tend to modify or rather replace [this line](https://github.com/JuliaPlots/Plots.jl/blob/8dcff1f73d24679b0011bf29da7b123862d782c6/src/backends/pyplot.jl#L864) by some new code instead.

Here is a suggestion that I have written quickly and not tested yet, but maybe it's a start
```julia
tick_labels = Array{LaTeXString}(undef, length( ticks[2]))
for (i, tlabel) in enumerate(ticks[2])
  if tlabel isa String && length( tlabel) >= 2 && tlabel[1] == '$' && tlabel[end] == '$'
    tick_labels[i] = py_surround_latextext(tlabel, env)
  else
    tick_labels[i] = py_surround_latextext( string('$', tlabel, '$'), env)
  end
end
```

isentropic:
This works now:
```
x=1:10
y=1:10

plt = plot( x, y, xticks=( 1:5, ["\$0\$", "\$1\$", "\$2\$", "\$3\$", "\$4\$"]))
plt = plot( x, y, xticks=( 1:5, 1:5))

```
Please provide some feedback @benide @wg030 
I decided to modify [this](https://github.com/JuliaPlots/Plots.jl/blob/8dcff1f73d24679b0011bf29da7b123862d782c6/src/backends/pyplot.jl#L845) as else there are too many nested if else statements that make it even more confusing.

isentropic:
Please drop [some docs](https://github.com/JuliaPlots/PlotDocs.jl/pulls)  as was [promised](https://github.com/JuliaPlots/Plots.jl/issues/2889#issuecomment-669726447). I really like your workflow I think more users can benefit from it.

wg030:
> Please provide some feedback @benide @wg030

I have tested the latest commit 9edc6f5 of this pull request and I can confirm that it successfully solves my issue #2890.

However I started testing a little more and I figured out that if someone wanted to use "$" in his labels, this causes trouble. For example if someone wants his labels to be like "0 $" , "1 $" and so on because his x represetns a price in dollar, I personally would expect the following code to work correctly, but it does not altough it does not create an error either. The output is just pretty ugly on my machine containing some "\mathregular".

```julia
pyplot( )
plt = plot( x, y, xticks=( 1:5, ["0 \$", "1 \$", "2 \$", "3 \$", "4 \$", "5 \$"]))
```

isentropic:
> ```julia
> pyplot( )
> plt = plot( x, y, xticks=( 1:5, ["1 \$", "2 \$", "3 \$", "4 \$", "5 \$"]))
> ```
This happens because the dollar sign is being preemtively escaped with the dollar sign `$\mathregular{1$}$` does not obey latex logic.

So I suggest:
```
plt = plot( x, y, xticks=( 1:5, [raw"1\$", "2", "3", "4", "5"]))
# or
plt = plot( x, y, xticks=( 1:5, ["1\\\$", "2", "3", "4", "5"]))
```

wg030:
> So I suggest:
> 
> ```julia
> plt = plot( x, y, xticks=( 1:5, [raw"1\$", "2", "3", "4", "5"]))
> # or
> plt = plot( x, y, xticks=( 1:5, ["1\\\$", "2", "3", "4", "5"]))
> ```

Both suggestions work for me. However if we wanted to have a space between "1" and "$", I am not sure how to make the first suggestion work, but for the second it would be `plt = plot( x, y, xticks=( 1:5, ["1\\ \\\$", "2", "3", "4", "5"]))`.

Moreover I think these suggestions can just be considered a temporary solution because 
```julia
#pyplot( )
plt = plot( x, y, xticks=( 1:5, ["1 \$", "2 \$", "3 \$", "4 \$", "5 \$"]))
```
work as expected and as I said "I strongly agree that no matter how this issues are solved, the choice of the backend should not require the user to change the syntax.".
Hence dealing with such cases should be the job of the package and not the user in my opinion.

isentropic:
The issue is that backends interpret these strings differently, surely plotly or Unicode plots interpret and render things differently. So there is no way around this

wg030:
> So there is no way around this

I would disagree here.
As long as each backend can render each case in one way or the other, we can create an interface for Plots that is independent of the backend altough it might be complicated.
What needs to be done is that the universal input of the Plots interface are converted for each bakcend such that they fit to interpretation of the backend. This could and, as I said, should be done within the package.

isentropic:
While I agree that nothing is impossible, catering to each backend and how it interprets strings is too difficult from maintainability stand point. If a backend string rendering updates so will have to all plots code. Sometimes one backend can do more sometimes a backend can do less. For example, Unicode plots would not really render latex stuff. Or GR might update latex procesing or whatnot.

In short: 
1. Too difficult to maintain
2. Could cause unexpected behavior per backend
3. Better focus the effort to get this right on other more important things
4. Simpler code (as it is now) is better and would allow advanced users like yourself, hack your way into understanding

PR #2898, opened 2020-08-06
Tick formating fixes

### LINKED ISSUE #2810: [BUG] GR negative exponents look ugly
### Backends

gr (default) |  x  |     |
pyplot       |     | x   |
plotly       |     |   x |
plotlyjs     |     |   x |
pgfplotsx    |     |  x  |
inspectdr    |     |     | x

Both
```
plot(1e-12 .* rand(100), ticks=:native)
plot(1e-12 .* rand(100))
```
produce (similar)

image
image

Note how the `-` negative sign is rendered uglily, whereas pyplot and other backends look normal.
I believe it is internal to GR, perhaps @jheinen  could help. The relevent code I belive is [here](https://github.com/JuliaPlots/Plots.jl/blob/07146e84e5e507f24cfc7d43375425494cb8bfa4/src/backends/gr.jl#L749)

yha:
Fixes scientific tick formatting in GRand plotly. 
Fix  #2810
Also fixes `formatter=:engineering` in GR.


yha:
Not sure how to update the precompiles

src/utils.jl
```diff
@@ -1050,8 +1050,8 @@ guidefont(ax::Axis) = font(
 )
 
 # ---------------------------------------------------------------
-# converts unicode scientific notation unsupported by pgfplots and gr
-# into a format that works
+# converts unicode scientific notation, as returned by Showoff,
+# to a tex-like format (supported by gr, pyplot, and gpfplot).
```
BeastyBlacksmith, line 1054:
```suggestion
# to a tex-like format (supported by gr, pyplot, and pgfplots).
```

BeastyBlacksmith:
Can you post some before/ after pictures while I enable the precompile bot for all branches again?

yha:
demos for
```
plot( plot([-1e10,1e10], yformatter=:engineering),
      plot([-1e-10,1e-10]),
      plot([0.001,1000], yscale=:log10) )
```
### GR
before:
gr-before
after:
gr-after
### Plotly
before:
plotly-before
after:
plotly-after
### PyPlot
before/after (no change):
pyplot-before

yha:
Not sure why the change in ticks messes up the layout in plotly. Seems this extra space happens even with native ticks: `plot([-1e6,1e6], ticks=:native)` get a much bigger left margin than `plot([-1,1], ticks=:native)`

yha:
What needs to be done to trigger SnoopCompile on this branch?

isentropic:
Would you please explain why did Plotly got affected in the first place? I thought the changes are meant to be GR exclusive?
Other backends seemed to handle `-` sign just fine

src/backends/pyplot.jl
```diff
@@ -176,15 +176,7 @@ function add_pyfixedformatter(cbar, vals::AVec)
...
 
 function labelfunc(scale::Symbol, backend::PyPlotBackend)
-    if scale == :log10
-        x -> PyPlot.LaTeXStrings.latexstring("10^{$x}")
-    elseif scale == :log2
-        x -> PyPlot.LaTeXStrings.latexstring("2^{$x}")
-    elseif scale == :ln
-        x -> PyPlot.LaTeXStrings.latexstring("e^{$x}")
-    else
-        x -> PyPlot.LaTeXStrings.latexstring(convert_sci_unicode(x))
```
isentropic, line 186: I wonder if this else statement is still triggered?

isentropic: It is pretty important as it fixed `Serif` fonts for pyplot

src/backends/pyplot.jl
```diff
@@ -176,15 +176,7 @@ function add_pyfixedformatter(cbar, vals::AVec)
...
-    end
+    PyPlot.LaTeXStrings.latexstring ∘ labelfunc_tex(scale)
```
isentropic, line 179: I do not really understand this `\dot` operator, bu I assume it runs latexstring once again to cover the previous else statement

isentropic: I just wanna make sure that https://github.com/JuliaPlots/Plots.jl/pull/2858 does not break because of this

src/axes.jl
```diff
@@ -123,6 +123,14 @@ const _label_func = Dict{Symbol,Function}(
 )
 labelfunc(scale::Symbol, backend::AbstractBackend) = get(_label_func, scale, string)
 
+const _label_func_tex = Dict{Symbol,Function}(
+    :log10 => x -> "10^{$x}",
+    :log2 => x -> "2^{$x}",
+    :ln => x -> "e^{$x}",
+)
+labelfunc_tex(scale::Symbol) = get(_label_func_tex, scale, convert_sci_unicode)
```
isentropic, line 131: Good change I think

BeastyBlacksmith:
> What needs to be done to trigger SnoopCompile on this branch?

You'd need to rebase on master

src/backends/pyplot.jl
yha: That's a `\circ` (function composition), not a `\dot`. It should behave identically to the original function. I just factored out the part that's shared between backends to `labelfunc_tex`.
I confirmed now that plots with "serif" on pyplot are also unchanged by this PR.

src/backends/pyplot.jl
yha: It's triggered as the default value in `labelfunc_tex`

yha:
Rebased

yha:
Seems like the SnoopCompile script needs updating? https://github.com/yha/Plots.jl/runs/975918743

BeastyBlacksmith:
Hmm.. I see this error on master too, @aminya did anything change for the snoopcompile bot?

aminya:
I think so. SnoopCompileBot is in a separate repository now because of the issues of the monorepo. I am trying to make it work again on my account.

I think @timholy has registered SnoopCompile which does not include SnoopCompileBot anymore.

aminya:
> > It correctly updates the precompiles. Merge it to this branch: [yha#7](https://github.com/yha/Plots.jl/pull/7)
> 
> What did you change to make the precompiles work?

Nothing much 🤷‍♂️ I just removed the old precompile files and ran the CI again.

yha:
This PR now requires updating most of the reference images, since the change of dashes to minus signs causes failures in the image comparison tests. I only have easy access to Windows machines, but it's probably better to generate the images on linux, since that's where it's run in CI.
I tested locally without the dash-to-minus change and tests pass.

daschw:
I can update the reference images.

daschw:
Looks good, still requires resolving of some conflicts.

PR #2902, opened 2020-08-07
Enable snoopcompile bot on all branches again

BeastyBlacksmith:
Because of https://github.com/JuliaPlots/Plots.jl/pull/2898


BeastyBlacksmith:
@aminya why is the benchmark part failing?

aminya:
Just rerun the job. It seems a Github issue.

aminya:
Xvfb (which is used for running the graphical tests in Linux) fails to start on Ubuntu. I am not sure how this is related to Julia in any ways.

PR #2904, opened 2020-08-08
Fix error when setting animation loop count

ztangent:
When creating an animation with a custom loop count, via, e.g., `gif(anim; loop=-1)`, so `mov(anim; loop=-1)`, FFMPEG throws the following error:

```
[image2 demuxer @ 000001fdf77ec280] Unable to parse option value "-1" as boolean
[image2 demuxer @ 000001fdf77ec280] Error setting option loop to value -1. 
```

This PR fixes the error by moving the `-loop` flag to after the `-i` input flag in the call to `ffmpeg_exe`. I believe the error is caused because the `-loop` flag occuring before `-i` controls how the input is interpreted, whereas the `-loop` flag occuring after `-i` controls how many loops are in the output.


BeastyBlacksmith:
Thanks!

Issue #392, opened 2016-07-18
Feature Request: trisurf_plot aka mesh3d


krcools:
I really like the idea of this package and I am more and more relying upon it. These is one type of plots however that I encounter very frequently in my work and that does not seem to be supported by plots.jl: an interface for the plotting of triangular meshes in 3d. At least the PyPlot and PlotsJS backends provide support through their `trisurf_plot` and `mesh3d` API's, respectively. This leaves me with a mix of Plots calls and direct backend invocations.

Is this somehow already supported or are there any plans to support this API in the future?

tbreloff:
For pyplot and gr you can pass 3 vectors with seriestype surface and it
will do a trisurf.

krcools:
Thanks for the quick reply. If IIUC these trisurfs are always structured in the sense that they derive from a rectangular grid, correct? I am looking for functionality that allows me to create unstructured triangular meshes. Typical APIs expect x,y,z values, but in addition also i,j,k integer indices defining the grid. I know that PyPlot, PlotlyJS, and Matlab can do this.

![Matlab example](https://cloud.githubusercontent.com/assets/10086147/16913689/324d42ca-4ce4-11e6-81db-540d0707d8cb.png)

tbreloff:
The whole point of a trisurf is that it's not necessarily a grid.

krcools:
Thanks again. Possibly I am missing something obvious here, but I can't see how a three argument invocation of `surface` can produce a general trisurf. Issues #24 and #387 also seem to imply `Plots.jl` is not exposing this functionality for now.

tbreloff:
While I agree there's no way to specify arbitrary 3D meshes at this point, the "trisurf" functionality works fine:

```
using Plots; gr(); surface(rand(100),rand(100),rand(100))
```

![tmp](https://cloud.githubusercontent.com/assets/933338/16916681/1603e35c-4ccc-11e6-94c9-2881dc4e8d8b.png)

As you can see from the example, there's no need for the data to be on a grid, or even ordered in any way.  The "trisurf" creates a  surface, not a mesh, and this might be your point of confusion?

krcools:
Yes, I imagine our communication issues can be traced back to us using slightly different definitions. I'd like to partially shift blame for that to `matplotlib` though: the `plot_trisurf` API can accommodate  a `triangles` kwarg that takes an `Nx3` index buffer allowing for the description of any 3D unstructured mesh:

[...]

PR #2909, opened 2020-08-13
Add :mesh3d series type for Plotly

dwd31415:
This pull-request adds the possibility to plot arbitrary 3d meshes using the mesh3d plot type in Plotly. It can be seen as partial fix/implementation for #392. For backends other than Plotly there is a fallback recipe provided that however does not support specifying the triangles of the mesh. 

Example: 
```Julia
using Plots; plotly()

x=[0, 1, 2, 0]
y=[0, 0, 1, 2]
z=[2, 4, 2, 3]

i=[0, 0, 0, 1]
j=[1, 2, 3, 2]
k=[2, 3, 1, 3]

plot(x,y,z,seriestype=:mesh3d;i,j,k)
```
Output:
image

### DIFF
--- a/src/args.jl
+++ b/src/args.jl
@@ -38,7 +38,7 @@ const _axesAliases = Dict{Symbol,Symbol}(
 )
 
 const _3dTypes = [
-    :path3d, :scatter3d, :surface, :wireframe, :contour3d, :volume
+    :path3d, :scatter3d, :surface, :wireframe, :contour3d, :volume, :mesh3d
 ]
 const _allTypes = vcat([
     :none, :line, :path, :steppre, :steppost, :sticks, :scatter,
--- a/src/args.jl
+++ b/src/args.jl
@@ -288,7 +288,7 @@ const _series_defaults = KW(
                                      #     one logical series to be broken up (path and markers, for example)
     :hover              => nothing,  # text to display when hovering over the data points
     :stride             => (1,1),    # array stride for wireframe/surface, the first element is the row stride and the second is the column stride.
-    :connections	=> nothing,  # tuple of arrays to specifiy connectivity of a 3d mesh
+    :connections	  => nothing,  # tuple of arrays to specifiy connectivity of a 3d mesh
     :extra_kwargs       => Dict()
 )
 
--- a/src/args.jl
+++ b/src/args.jl
@@ -288,6 +288,7 @@ const _series_defaults = KW(
                                      #     one logical series to be broken up (path and markers, for example)
     :hover              => nothing,  # text to display when hovering over the data points
     :stride             => (1,1),    # array stride for wireframe/surface, the first element is the row stride and the second is the column stride.
+    :connections	=> nothing,  # tuple of arrays to specifiy connectivity of a 3d mesh
     :extra_kwargs       => Dict()
 )
 
--- a/src/backends/plotly.jl
+++ b/src/backends/plotly.jl
@@ -574,11 +574,19 @@ function plotly_series(plt::Plot, series::Series)
     elseif st == :mesh3d
 	plotattributes_out[:type] = "mesh3d"
         plotattributes_out[:x], plotattributes_out[:y], plotattributes_out[:z] = x, y, z
-        plotattributes_out[:colorscale] = plotly_colorscale(series[:fillcolor], series[:fillalpha])
+        
+	if :i in keys(series[:extra_kwargs]) && :j in keys(series[:extra_kwargs]) && :k in keys(series[:extra_kwargs])
+		plotattributes_out[:i] = series[:extra_kwargs][:i]
+		plotattributes_out[:j] = series[:extra_kwargs][:j]
+		plotattributes_out[:k] = series[:extra_kwargs][:k]
+	end
+
+	plotattributes_out[:colorscale] = plotly_colorscale(series[:fillcolor], series[:fillalpha])
         plotattributes_out[:opacity] = series[:fillalpha]
         if series[:fill_z] !== nothing
             plotattributes_out[:surfacecolor] = plotly_surface_data(series, series[:fill_z])
         end
+        plotattributes_out[:showscale] = false 
     else
         @warn("Plotly: seriestype $st isn't supported.")
         return KW()
--- a/src/backends/plotly.jl
+++ b/src/backends/plotly.jl
@@ -552,7 +552,7 @@ function plotly_series(plt::Plot, series::Series)
         plotattributes_out[:showscale] = hascolorbar(sp) && hascolorbar(series)
 
     elseif st in (:surface, :wireframe)
-	plotattributes_out[:type] = "surface"
+	  plotattributes_out[:type] = "surface"
         plotattributes_out[:x], plotattributes_out[:y], plotattributes_out[:z] = x, y, z
         if st == :wireframe
             plotattributes_out[:hidesurface] = true
--- a/src/backends/plotly.jl
+++ b/src/backends/plotly.jl
@@ -574,11 +574,19 @@ function plotly_series(plt::Plot, series::Series)
     elseif st == :mesh3d
 	plotattributes_out[:type] = "mesh3d"
         plotattributes_out[:x], plotattributes_out[:y], plotattributes_out[:z] = x, y, z
-        
-	if :i in keys(series[:extra_kwargs]) && :j in keys(series[:extra_kwargs]) && :k in keys(series[:extra_kwargs])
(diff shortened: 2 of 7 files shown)


src/pipeline.jl
```diff
@@ -331,9 +331,10 @@ function _override_seriestype_check(plotattributes::AKW, st::Symbol)
     if !RecipesPipeline.is3d(st) && !(st in (:contour, :contour3d))
         z = plotattributes[:z]
         if !isa(z, Nothing) &&
-           (size(plotattributes[:x]) == size(plotattributes[:y]) == size(z))
+           (size(plotattributes[:x]) == size(plotattributes[:y]) == size(z)) &&
+	   st !== :mesh3d
```
BeastyBlacksmith, line 335: This shouldn't be necessary since it should already been handled by `!RecipesPipeline.is3d(st)`

src/recipes.jl
```diff
@@ -1534,3 +1546,5 @@ julia> areaplot(1:3, [1 2 3; 7 8 9; 4 5 6], seriescolor = [:red :green :blue], f
         end
     end
 end
+
+is_3d(::Type{Val{:mesh3d}}) = true
```
BeastyBlacksmith, line 1550: This should extend `RecipesPipeline.is3d`

dwd31415: Should I open a separate pull request for RecipesPipeline for this then?

BeastyBlacksmith: yes, that would be good

dwd31415:
Hi Simon, 
thanks for the feedback.

Regarding the first point that sounds like a much cleaner interface. Am I right that adding `connection` as a series keyword would be achieved by adding it to `_series_defaults` in args.jl? It seems to work but I just wanted to check it is not some kind of hacky way.

BeastyBlacksmith:
> Am I right that adding `connection` as a series keyword would be achieved by adding it to `_series_defaults` in args.jl? It seems to work but I just wanted to check it is not some kind of hacky way.

You are right. That is the way to go.

src/backends/plotly.jl
```diff
@@ -574,11 +574,19 @@ function plotly_series(plt::Plot, series::Series)
...
-        
-	if :i in keys(series[:extra_kwargs]) && :j in keys(series[:extra_kwargs]) && :k in keys(series[:extra_kwargs])
-		plotattributes_out[:i] = series[:extra_kwargs][:i]
-		plotattributes_out[:j] = series[:extra_kwargs][:j]
-		plotattributes_out[:k] = series[:extra_kwargs][:k]
+       
+	if series[:connections] != nothing
+		if typeof(series[:connections]) <: Tuple{Array,Array,Array} 
+			i,j,k = series[:connections]
+			if length(i) == length(j) == length(k)
```
BeastyBlacksmith:
```suggestion
			if l!(ength(i) == length(j) == length(k))
```

dwd31415: Ah yes this was a typo, but it should already be fixed in the latest commit.

src/examples.jl
```diff
@@ -995,16 +995,40 @@ const _examples = PlotExample[
             scatter!(Point2.(eachcol(rand(d,1000))), alpha=0.25)
         end]
     ),
+    PlotExample(
+	"Mesh3d",
+	"Allows to plot arbitrary 3d meshes. If only x,y,z are given the mesh is generated automatically. You can also specify the connections using the connections keyword. This is only supported on the Plotly backend. The connections are specified using a tuple of arrays where each array contains the index of the first, second and third vertex of a triangle.",
```
BeastyBlacksmith, line 1000: Please make this a multiline string using triple quotes.

> The connections are specified using a tuple of arrays where each array contains the index of the first, second and third vertex of a triangle

That sounds to me like each array has three elements. Also better to speak about vectors here to not confuse it with matrices. Something like "each vector contains the 0-based indices of one point of a triangle, such that elements at the same position of these vectors form a triangle"

src/recipes.jl
```diff
@@ -913,6 +913,19 @@ end
 @deps pie shape
 
 
+# ---------------------------------------------------------------------------
+# mesh 3d replacement for non-plotly backends
+
+@recipe function f(::Type{Val{:mesh3d}}, x, y, z; i=nothing,j=nothing,k=nothing)
```
BeastyBlacksmith, line 919: this should also query `plotattributes[:connections]` instead of the keywords

src/examples.jl
dwd31415: Ok, I have updated it

src/recipes.jl
dwd31415: Oh yes, I forget that. I have changed it now.

src/args.jl
```diff
@@ -288,6 +288,7 @@ const _series_defaults = KW(
                                      #     one logical series to be broken up (path and markers, for example)
     :hover              => nothing,  # text to display when hovering over the data points
     :stride             => (1,1),    # array stride for wireframe/surface, the first element is the row stride and the second is the column stride.
+    :connections	=> nothing,  # tuple of arrays to specifiy connectivity of a 3d mesh
```
BeastyBlacksmith, line 291:
```suggestion
    :connections	 => nothing,  # tuple of arrays to specifiy connectivity of a 3d mesh
```

src/backends/plotly.jl
```diff
@@ -553,7 +552,7 @@ function plotly_series(plt::Plot, series::Series)
         plotattributes_out[:showscale] = hascolorbar(sp) && hascolorbar(series)
 
     elseif st in (:surface, :wireframe)
-        plotattributes_out[:type] = "surface"
+	plotattributes_out[:type] = "surface"
```
BeastyBlacksmith, line 555:
```suggestion
	  plotattributes_out[:type] = "surface"
```

src/examples.jl
```diff
@@ -1015,6 +1044,7 @@ _backend_skips = Dict(
         31, # animation
         32, # spy
         38, # histogram2d
+	47, # mesh3d
```
BeastyBlacksmith, line 1047:
```suggestion
	 47, # mesh3d
```

src/pipeline.jl
```diff
@@ -331,9 +331,9 @@ function _override_seriestype_check(plotattributes::AKW, st::Symbol)
     if !RecipesPipeline.is3d(st) && !(st in (:contour, :contour3d))
         z = plotattributes[:z]
         if !isa(z, Nothing) &&
-           (size(plotattributes[:x]) == size(plotattributes[:y]) == size(z))
+           (size(plotattributes[:x]) == size(plotattributes[:y]) == size(z))
```
BeastyBlacksmith, line 334:
```suggestion
           (size(plotattributes[:x]) == size(plotattributes[:y]) == size(z))
```

src/pipeline.jl
```diff
@@ -331,9 +331,9 @@ function _override_seriestype_check(plotattributes::AKW, st::Symbol)
     if !RecipesPipeline.is3d(st) && !(st in (:contour, :contour3d))
         z = plotattributes[:z]
         if !isa(z, Nothing) &&
-           (size(plotattributes[:x]) == size(plotattributes[:y]) == size(z))
+           (size(plotattributes[:x]) == size(plotattributes[:y]) == size(z)) 
             st = (st == :scatter ? :scatter3d : :path3d)
-            plotattributes[:seriestype] = st
+	    plotattributes[:seriestype] = st
```
BeastyBlacksmith, line 336:
```suggestion
	      plotattributes[:seriestype] = st
```

src/shorthands.jl
```diff
@@ -318,6 +318,27 @@ julia> scatter3d([0,1,2,3],[0,1,4,9],[0,1,8,27])
 """
 @shorthands scatter3d
 
+"""
+    mesh3d(x,y,z)
+    mesh3d(x,y,z;i,j,k)
```
BeastyBlacksmith, line 323:
```suggestion
```

dwd31415:
Okay great, then everything should be fine now.

BeastyBlacksmith:
Thanks!

dlfivefifty:
Does this support setting the colours of the triangle? Trying to plot a function a sphere: have the sphere working but now need the colours

dwd31415:
No I don't think this is supported. As a hacky workaround you could plot the triangles individually since you can specify the color of the whole mesh like this `mesh3d(x,y,z;color=:purple,connections=(i,j,k))`

dlfivefifty:
That's too bad... I'm guessing the Plotly backend supports it?

PR #2913, opened 2020-08-17
Fixing xfvb issues

aminya:
Trying to fix the xfvb issues in the CI

Based on the documentation: https://manpages.ubuntu.com/manpages/trusty/man1/xvfb-run.1.html
Some times the xfvb server might be busy, so this command finds a free server.

Running SnoopCompile action, the Ubuntu jobs now run without any issues.
https://github.com/aminya/Plots.jl/runs/993198724?check_suite_focus=true


aminya:
@daschw This should fix the random Xfvb failures.

daschw:
Great thanks!

PR #2930, opened 2020-08-25
Test SnoopCompileBot

aminya:
This shows that SnoopCompileBot works. You can merge it to update the precompiles if you want.

https://github.com/aminya/Plots.jl/actions/runs/223955247


BeastyBlacksmith:
I merged another PR already, so I think this one can be closed.

PR #2940, opened 2020-08-31
Fix marker shapes with segments

daschw:
Consider the following code:
```julia
using Plots

yv = ones(9)
ys = [1; 1; NaN; ones(6)]

plot(
   5 .- [yv 2ys 3yv 4ys],
   seriestype = [:path :path :scatter :scatter],
   markershape = [:utriangle, :rect],
   markersize = 8,
   color = [:red, :black],
)
```

### Before

This is what we get on master with different backends. No backend behaves as expected for the top series.

#### GR
marker_segmens_old_gr

#### PyPlot
marker_segmens_old_pyplot

#### Plotly
marker_segmens_old_plotlyjs

#### PGFPlotsX
marker_segmens_old_pgfplotsx

### After

This PR fixes this for GR, PyPlot and Plotly. I did not manage to fix it for PGFPlotsX yet.

#### GR
marker_segmens_new_gr

#### PyPlot
marker_segmens_new_pyplot

#### Plotly
marker_segmens_new_plotlyjs

I added above example as a new test example. https://github.com/JuliaPlots/PlotReferenceImages.jl/pull/83is required for the tests to pass.


daschw:
THis also fixes our broken doc image in http://docs.juliaplots.org/latest/input_data/#columns-are-series

PR #2962, opened 2020-09-09
fix error bars in log scale

### LINKED ISSUE #2961: [BUG] Disappearing errorbar under log scale
## Details
```
scatter([1:3], [0.1, 0.15, 0.08], yerror=[0.3, 0.1, 0.02], yscale=:log10, ylims=(10^-3,Inf))
```
image

### Backends

gr (default) |  x   |     |
pyplot       |     |     | x
plotly       |     |     | x
plotlyjs     |     |     | x
pgfplotsx    |     |     | x
inspectdr    |     |     | x

### Versions

Plots.jl version: v1.6.3
Backend  version (`]st -m`): GR v0.52.0
Output of `versioninfo()`:
```
Julia Version 1.5.0
Commit 96786e22cc (2020-08-01 23:44 UTC)
Platform Info:
```

Moelf:
fix #2961
before:
image
after:
image

### DIFF
--- a/src/recipes.jl
+++ b/src/recipes.jl
@@ -1090,6 +1090,12 @@ function error_coords(errorbar, errordata, otherdata...)
     return (ed, od...)
 end
 
+# clamp non-NaN values in an array to Base.eps(Float64) for log-scale plots
+function clamp_to_eps!(ary)
+    replace!(x -> x <= 0.0 ? Base.eps(Float64) : x, ary)
+    nothing
+end
+
 # we will create a series of path segments, where each point represents one
 # side of an errorbar
 
@@ -1102,6 +1108,9 @@ end
     else
         plotattributes[:x], plotattributes[:y], plotattributes[:z] =
             error_coords(xerr, x, y, z)
+    end
+    if :xscale ∈ keys(plotattributes) && plotattributes[:xscale] == :log10
+        clamp_to_eps!(plotattributes[:x])
     end
     ()
 end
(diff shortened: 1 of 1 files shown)


src/recipes.jl
```diff
@@ -1117,6 +1123,12 @@ end
         plotattributes[:y], plotattributes[:x], plotattributes[:z] =
             error_coords(yerr, y, x, z)
     end
+    for i in [:x, :y, :z]
+        _scale= Symbol(i, :scale)
+        if _scale ∈ keys(plotattributes) && plotattributes[_scale] == :log10
```
BeastyBlacksmith, line 1128: This applies also for other log scales and likewise we'd need to consider the case of x-errors with log-scale

Moelf: `i` will iterate over `:x, :y, :z` and clamp each of their error bar in the same fashion, so this should be fine?

mkborregaard:
I'm trying to understand why you have sub-zero error bars in log space in the first place?

Moelf:
that's a valid question. because the statistical error for counting experiment is `log(N)`, if `N<1`, the symmetric error is greater than `N` itself. While it's un-physical, currently we can't have asymmetric error bar. And it's often not easy to derive what's "correct" value for the lower half, so this is still a good visual cue to say: don't read too much into this value since it's error bar is large.

src/recipes.jl
BeastyBlacksmith: I am just saying that the `:xerror` recipe probably also needs this fix

daschw (approved):

PR #2968, opened 2020-09-13
bump compat for "GeometryBasics" to "0.3.1"

jaakkor2:
CompatHelper PR to add "0.3" failed https://github.com/JuliaPlots/Plots.jl/pull/2935, but should be fixed in https://github.com/JuliaGeometry/GeometryBasics.jl/pull/83, after which GeometryBasics 0.3.1 was released.


jaakkor2:
Error on nightly is due to Showoff.jl using Grisu that is not due to this PR
```
PGFPlotsX: Error During Test at D:\a\Plots.jl\Plots.jl\test\test_pgfplotsx.jl:14
  Got exception outside of a @test
  UndefVarError: Grisu not defined
```
See https://github.com/JuliaGraphics/Showoff.jl/issues/24

BeastyBlacksmith:
Thanks!

PR #2988, opened 2020-09-18
fine tune distance between axis and ticks text

Moelf:
since the same magic numbers appear everywhere I believe this is the correct "fix". Assuming others felt the same about ticks being too close.

```
a = rand(2,200).-0.5;
scatter(a[1,:].*3, a[2,:].*3, minorticks=2, framestyle=:origin)
```

Before:
image

After:
image


mkborregaard (approved):
Given that we use magic numbers anyway, might as well use some better ones.

Issue #1965, opened 2019-03-26
Plotting arrows with pgfplots() does not work


natschil:
The following works fine with the GR backend, but not with pgfplots:
```
using Plots
pgfplots()
Plots.plot([(0.0,0.0),(1.0,1.0)],line=arrow())
```

There is a warning message: `Warning: Keyword argument arrow not supported with Plots.PGFPlotsBackend().  Choose from:....`
followed by a long list of keyword arguments.

BeastyBlacksmith:
Still holds for the pgfplotsx backend, though `quiver` should work.

AshtonSBradley:
quiver has stopped working:
```julia
using Plots
gr()
# Arrow position.
x = collect(0:10)
y = fill(0., 10)
# Arrow direction.
u = fill(1., 10)
v = fill(1., 10)
# Arrow color.
c = collect(1:10)
quiver(x, y, quiver=(u, v))
```
runs, but `gr() -> pgfplotsx()` causes
```julia
DimensionMismatch("vectors must have same lengths")

Stacktrace:
 [1] hcat(::Array{Float64,1}, ::Array{Float64,1}, ::Array{Float64,1}, ::Vararg{Array{Float64,1},N} where N) at ./array.jl:1586
 [2] PGFPlotsX.TableData(::Array{Pair{Symbol,Array{Float64,1}},1}; scanlines::Int64, rowsep::Bool) at /Users/abradley/.julia/packages/PGFPlotsX/C6ahR/src/axiselements.jl:406
 [3] TableData at /Users/abradley/.julia/packages/PGFPlotsX/C6ahR/src/axiselements.jl:406 [inlined]
 [4] #Table#44 at /Users/abradley/.julia/packages/PGFPlotsX/C6ahR/src/axiselements.jl:474 [inlined]
 [5] Table at /Users/abradley/.julia/packages/PGFPlotsX/C6ahR/src/axiselements.jl:474 [inlined]
 [6] #Table#45 at /Users/abradley/.julia/packages/PGFPlotsX/C6ahR/src/axiselements.jl:477 [inlined]
 [7] PGFPlotsX.Table(::Array{Pair{Symbol,Array{Float64,1}},1}) at /Users/abradley/.julia/packages/PGFPlotsX/C6ahR/src/axiselements.jl:477
[...]

Issue #2290, opened 2019-11-29
PGFPlotsX meta-issue


BeastyBlacksmith:
I start to collect what is missing:

- [ ] allow files as input data
- [ ] fix histogram2d
- [ ] implement fillranges with fillbetween
- [ ] interface for filling between functions
- [ ] interface for patch plots
- [ ] fix layouts
- [x] can't plot two series with fillrange (https://github.com/JuliaPlots/Plots.jl/issues/2290#issuecomment-560427043)
- [x] `heatmap(1:2,1:2,[1 2; 3 4])` only shows a black square

daschw:
Can't plot two series with fillrange:
```julia
julia> plot(rand(10, 2), fillrange = 0)
! Illegal parameter number in definition of \pgfplots@curplotlist.
<to be read again>
1
l.118 \end{groupplot}

Error showing value of type Plots.Plot{Plots.PGFPlotsXBackend}:
ERROR: The latex command `lualatex jl_AE10.tmp.tex` failed
Stacktrace:
 [1] error(::String) at .\error.jl:33
 [2] #savepdf#71(::PGFPlotsX.LaTeXEngine, ::Array{String,1}, ::Int64, ::String, ::typeof(PGFPlotsX.savepdf), ::String, ::PGFPlotsX.TikzDocument) at C:\Users\Daniel\.julia\packages\PGFPlotsX\K4i95\src\tikzdocument.jl:196
 [3] #savepdf at .\none:0 [inlined]
 [4] #savepng#73(::PGFPlotsX.LaTeXEngine, ::Array{String,1}, ::Int64, ::typeof(PGFPlotsX.savepng), ::String, ::PGFPlotsX.TikzDocument) at C:\Users\Daniel\.julia\packages\PGFPlotsX\K4i95\src\tikzdocument.jl:322
 [5] #savepng at .\none:0 [inlined]
 [6] #save#65(::Bool, ::PGFPlotsX.LaTeXEngine, ::Array{String,1}, ::Int64, ::Bool, ::typeof(PGFPlotsX.save), ::String, ::PGFPlotsX.TikzDocument) at C:\Users\Daniel\.julia\packages\PGFPlotsX\K4i95\src\tikzdocument.jl:87
[...]

PR #2989, opened 2020-09-18
add arrow support

BeastyBlacksmith:
fix https://github.com/JuliaPlots/Plots.jl/issues/1965


KristofferC:
Should https://github.com/JuliaPlots/Plots.jl/issues/2290 be updated?

BeastyBlacksmith:
You are right, thank you !

PR #3002, opened 2020-09-23
Added support for 3d-plots of seriestype sticks

bernd1995:
I just found out that `seriestype=:sticks` is not supported in 3d-plots.
So I added a workaround for 3d plots and followed therefore the template of bezier-curves.

### DIFF
--- a/src/recipes.jl
+++ b/src/recipes.jl
@@ -262,13 +262,22 @@ end
         end
     end
     newx, newy = zeros(3n), zeros(3n)
+    newz = z!== nothing ? zeros(3n) : nothing
     for i = 1:n
         rng = (3i - 2):(3i)
         newx[rng] = [x[i], x[i], NaN]
-        newy[rng] = [_cycle(fr, i), y[i], NaN]
+        if z !== nothing
+            newy[rng] = [y[i], y[i], NaN]
+            newz[rng] = [_cycle(fr, i), z[i], NaN]
+        else
+            newy[rng] = [_cycle(fr, i), y[i], NaN]
+        end
     end
     x := newx
     y := newy
+    if z !== nothing
+        z := newz
+    end
     fillrange := nothing
     seriestype := :path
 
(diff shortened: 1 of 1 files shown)


BeastyBlacksmith:
what does it look like?

bernd1995:
Oh yeah sorry, I was a bit in a hurry when I opened the request.

So now with the added support 
```
using Plots
gr()
plot(rand(10),rand(10),rand(10),seriestype=:sticks,markershape=:circle)

``` 
produces the following figure.

bernd1995-sticks3d

The original version in [JuliaPlots:masters](https://github.com/JuliaPlots/Plots.jl) however creates the following figure

bernd1995-sticks3d_old

So it basically puts a 2D-stick plot on top of a 3D-scatter plot.

mkborregaard (approved):
nice

daschw:
Thanks a lot!

bernd1995:
Right now I realized one ambiguity I introduced with this edit.
If someone wants to label the dots (and uses the sticks just as helplines) there is a problem with the labeling.
```julia
gr();
x,y,z,c = rand(1,10),rand(1,10),rand(1,10),rand(1,10);
scatter(x,y,z,marker_z=c,marker=(:circle,5),line=(:black),st=:sticks,label = ["1" "2" "3" "4" "5" "6" "7" "8" "9" "10"])
```
stick_labels
Only the sticks receive labels while there is no possibility to label to dots (at least at first sight).

Any ideas/comments how to fix this?
There is probably something odd with the order of processing the separate recipes.

BeastyBlacksmith:
You could draw it separately:
```julia
x,y,z,c = rand(1,10),rand(1,10),rand(1,10),rand(1,10);
plot(x,y,z,line=(:black),st=:sticks)
scatter!(x,y,z,marker_z=c,marker=(:circle,5),st=:sticks,label = ["1" "2" "3" "4" "5" "6" "7" "8" "9" "10"])
```

bernd1995:
Oh yeah that's right of course.

Still I think the expected behavior should be:
  - sticks + marker: labels for markers
  - only sticks: labels for lines

Should we open an issue to discuss or should I just make another pull request because I already figured out how to do this?

BeastyBlacksmith:
Just open the PR

Issue #2708, opened 2020-05-16
[BUG] Heatmap with NaN fails with PGFPlotsX backend


nmheim:
## Details
```julia
using Plots
pgfplotsx()
x = rand(10,10)
x[1] = NaN
p = heatmap(x)
display(p)
```
produces
```
! Package pgfplots Error: the arguments of [mesh/rows=10,mesh/cols=10] assume 1
00 points, but I actually got N = 99 points! The data matrix appears to be inco
mplete or overcomplete!? [Use mesh/check=false to disable this message].

See the pgfplots package documentation for explanation.
Type  H <return>  for immediate help.
 ...                                              
                                                  
l.397 \end{axis}
              
ERROR: LoadError: The latex command `lualatex jl_qRmSBJ.tex` failed
Stacktrace:
 [1] error(::String) at ./error.jl:33
 [2] savepdf(::String, ::PGFPlotsX.TikzDocument; latex_engine::PGFPlotsX.LaTeXEngine, buildflags::Array{String,1}, run_count::Int64, tmp::String) at /home/niklas/.julia/packages/PGFPlotsX/C6ahR/src/tikzdocument.jl:197
 [3] save(::String, ::PGFPlotsX.TikzDocument; include_preamble::Bool, latex_engine::PGFPlotsX.LaTeXEngine, buildflags::Array{String,1}, dpi::Int64, showing_ide::Bool) at /home/niklas/.julia/packages/PGFPlotsX/C6ahR/src/tikzdocument.jl:86
 [4] save(::String, ::PGFPlotsX.TikzDocument) at /home/niklas/.julia/packages/PGFPlotsX/C6ahR/src/tikzdocument.jl:73
 [5] display(::PGFPlotsX.PGFPlotsXDisplay, ::PGFPlotsX.TikzDocument) at /home/niklas/.julia/packages/PGFPlotsX/C6ahR/src/tikzdocument.jl:351
 [6] _display(::Plots.Plot{Plots.PGFPlotsXBackend}) at /home/niklas/.julia/packages/Plots/5srrj/src/backends/pgfplotsx.jl:1274
 [7] display(::Plots.PlotsDisplay, ::Plots.Plot{Plots.PGFPlotsXBackend}) at /home/niklas/.julia/packages/Plots/5srrj/src/output.jl:150
 [8] display(::Any) at ./multimedia.jl:323
 [9] top-level scope at /home/niklas/test.jl:6
[...]

PR #3007, opened 2020-09-25
fix NaNs in heatmaps

BeastyBlacksmith:
This fixes the occurance of NaNs in heatmaps and therefore via the recipe also 2D Histograms.
It doesn't fix images though because the dimension of `x` and `y` are swapped there...

Fix https://github.com/JuliaPlots/Plots.jl/issues/2708


BeastyBlacksmith:
I think, this could in principle work with images now, but the example from the documentation crashes with
```
! Emergency stop.
\pgfplots@colormap@LUA@new ...\string \n ");
end
}
                                                  \endgroup \fi 
l.150033 }
        
End of file on the terminal!
```

PR #3015, opened 2020-09-28
fix horizontal range of hspan()

alusiani:
hspan!() has an odd behaviour in Plots v1.6.6, instead of plotting from -Inf to +Inf over x it plots from 1 to 2. Minimum example

```julia
julia> plot(0:9)
julia> hspan!([2,3])
```
The horizontal rectangle is drawn between 1 and 2 instead of between the minimum and maximum x range of the plot.

This small patch will draw the rectangle with the correct horizontal span, similarly to what vspan!() does.

Cheers,
\--
Alberto

### DIFF
--- a/src/recipes.jl
+++ b/src/recipes.jl
@@ -133,7 +133,7 @@ end
 
 @recipe function f(::Type{Val{:hspan}}, x, y, z)
     n = div(length(y), 2)
-    newx = repeat([1, 2, 2, 1, NaN], outer = n)
+    newx = repeat([-Inf, Inf, Inf, -Inf, NaN], outer = n)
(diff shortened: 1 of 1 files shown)


BeastyBlacksmith:
does that work on logarithmic x-axis?

alusiani:
> does that work on logarithmic x-axis?

Yes, works for me:

```julia
julia> plot(1:10, 1:10, xaxis=:log, yaxis=:log)
julia> hspan!([3,4])
```

and

```julia
julia> plot(1:10, 1:10, xaxis=:log)
julia> hspan!([3,4])
```

produce correct results for me.

Cheers,
\--
Alberto

PR #3016, opened 2020-09-30
Fixed plotly_colorbar_hack() to avoid additional marker

bernd1995:
After recently adding "support for 3d-plots of seriestype sticks" (https://github.com/JuliaPlots/Plots.jl/pull/3002) I recognized in my scatter-plots additional points which should not be there when using the `plotly()` backend.

The following code produced the following figure:
```julia
plotly();
x,y,z,c = rand(10),rand(10),rand(10),rand(10);
scatter(x,y,z,marker_z=c,marker=(:circle),st=:sticks);
```
markersticks_old

It seems that the main problem is that `plotly` uses a hack for creating the colorbar by adding one additional marker at the start of a new series. Therefore these two pinkish dots are added, one for the fill segments of the sticks and one for the general scatter plot. However, this behavior is unexpected because in `plotly_colorbar_hack()` the opacity and the size of this marker is set to 0.

In the end it turned out that a value of 0 (and also 0.0) is not working as markersize and/or markeropacity. Thus as an intermediate fix I set these values to a very small number and now it produces the following plot.
markerssticks_fixed

As a final side remark this also happens without `st=:sticks`:
```julia
scatter(x,y,z,marker_z=c,marker=(:circle),st=:sticks);
```
old: 
marker_old

fixed:
marker_fixed

### DIFF
--- a/src/backends/plotly.jl
+++ b/src/backends/plotly.jl
@@ -584,9 +584,9 @@ function plotly_series(plt::Plot, series::Series)
     elseif st == :mesh3d
 	plotattributes_out[:type] = "mesh3d"
         plotattributes_out[:x], plotattributes_out[:y], plotattributes_out[:z] = x, y, z
-       
+
 	if series[:connections] != nothing
-		if typeof(series[:connections]) <: Tuple{Array,Array,Array} 
+		if typeof(series[:connections]) <: Tuple{Array,Array,Array}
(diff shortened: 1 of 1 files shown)


BeastyBlacksmith (approved):
Looks great, thank you!

daschw:
Awesome, thanks!

PR #3017, opened 2020-09-30
Changed labeling-behaviour of :sticks

bernd1995:
As mentioned in https://github.com/JuliaPlots/Plots.jl/pull/3002#issuecomment-701293633 the labeling behavior of sticks is (at least to me ) misleading.

So I changed the behavior to the following:

### Sticks alone; column 1x5 array
```julia
x,y,z = rand(1,5),rand(1,5),rand(1,5);
sticks(x,y,z,label = ["1" "2" "3" "4" "5"])
```
Expected: 
  - Labels for lines
  - Different colors (columns are separate series)
stickslabel

### Sticks+Marker; column 1x5 array
```julia
x,y,z = rand(1,5),rand(1,5),rand(1,5);
sticks(x,y,z,marker=(:circle),label = ["a" "b" "c" "d" "e"])
```
Expected: 
  - Labels for markers
  - Line and markers share a color (works also with color set by `marker_z`)
sticksmarkerlabel

### Adjustments

It was necessary to set the markers (if they are added) as the primary series to achieve that. I am not sure if this introduces some other ambiguity elsewhere? However it didn't seem to make any problems by now.

### DIFF
--- a/src/recipes.jl
+++ b/src/recipes.jl
@@ -280,7 +280,6 @@ end
     end
     fillrange := nothing
     seriestype := :path
-    println(plotattributes[:linecolor])
     if plotattributes[:linecolor] == :auto && plotattributes[:marker_z] !== nothing && plotattributes[:line_z] == nothing
         line_z := plotattributes[:marker_z]
     end
--- a/src/recipes.jl
+++ b/src/recipes.jl
@@ -280,7 +280,8 @@ end
     end
     fillrange := nothing
     seriestype := :path
-    if plotattributes[:marker_z] !== nothing && plotattributes[:line_z] == nothing
+    println(plotattributes[:linecolor])
+    if plotattributes[:linecolor] == :auto && plotattributes[:marker_z] !== nothing && plotattributes[:line_z] == nothing
         line_z := plotattributes[:marker_z]
     end
 
--- a/src/recipes.jl
+++ b/src/recipes.jl
@@ -286,8 +286,6 @@ end
 
     # create a primary series for the markers
     if plotattributes[:markershape] != :none
-        tmplabel = deepcopy(plotattributes[:label])
-        label := ""
         primary := false
         @series begin
             seriestype := :scatter
@@ -296,7 +294,6 @@ end
             if z !== nothing
                 z := z
             end
-            label := tmplabel
             primary := true
             ()
         end
--- a/src/recipes.jl
+++ b/src/recipes.jl
@@ -280,11 +280,15 @@ end
     end
     fillrange := nothing
     seriestype := :path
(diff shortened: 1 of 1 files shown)


BeastyBlacksmith:
So how would I get monochrome sticks with colorful heads now?

bernd1995:
The linecolor attribute (and other line attributes as well) is untouched in the edits so by setting e.g. `line=(:black,2)` in the function call one gets thicker black lines with colorful heads.

BeastyBlacksmith:
What happens if I pass in row vectors as data?

bernd1995:
> The linecolor attribute (and other line attributes as well) is untouched in the edits so by setting e.g. `line=(:black,2)` in the function call one gets thicker black lines with colorful heads.

Okay I just checked it once again and found that it didn't work as expected. 
However, now I've fixed it and now `line=(:black,2)` works as expected.


> What happens if I pass in row vectors as data?

I assume you mean something like `x=rand(5)`?
Then the data is interpreted as one series. here some examples:
```julia
x,y,z,c = rand(5),rand(5),rand(5),rand(5);
plot1 = sticks(x,y,z,label = "1")
plot2 = sticks(x,y,z,marker_z = c,label = "1")  # line_z = marker_z unless explicit statement of line_z or linecolor
plot3 = sticks(x,y,z,marker_z = c, marker = (:circle), label ="a" )
```
###Plot 1
plot1
###Plot 2
Color in legend is always taken from the first line.
plot2
###Plot 3
Color in legend is always taken from the first marker.
plot3

src/recipes.jl
```diff
@@ -280,18 +280,24 @@ end
...
     fillrange := nothing
     seriestype := :path
+    if plotattributes[:linecolor] == :auto && plotattributes[:marker_z] !== nothing && plotattributes[:line_z] == nothing
+        line_z := plotattributes[:marker_z]
+    end
 
-    # create a secondary series for the markers
+    # create a primary series for the markers
     if plotattributes[:markershape] != :none
+        tmplabel = deepcopy(plotattributes[:label])
```
BeastyBlacksmith, line 289: is that neccessary or would a `copy` do?

bernd1995: `copy` won't do it since there is no copy method `copy(::String)`.
But a usual assignment `tmplabel = plotattributes[:label]` should be sufficient though. 
Just used `deepcopy` to be 100% sure.

BeastyBlacksmith: `deepcopy` is quite slow though. I think it should work without any copying of the labels, when you just switch the primary-ness, doesn't it?

bernd1995: Yes you are right.
So changing the primary-ness is sufficient.

PR #3020, opened 2020-10-02
latexify unicode

### LINKED ISSUE #3005: Example for trigonometric functions for PGFPlotsX uses unicode in labels which doesn't seem to work
http://docs.juliaplots.org/latest/generated/pgfplotsx/#pgfplotsx-ref42

Screenshot 2020-09-25 at 08 24 28

maybe the label should be changed to `raw"$\sin(\theta)$"`?

BeastyBlacksmith:
Fix #3005 
This adds a a new dependency, so in principle it would be better to take the other route.


daschw:
The other route being to modify the preamble? Would this be done in Plots or in PGFPlotsX? Would you have an idea what to set in the preamble?

BeastyBlacksmith:
This seems to be a font issue, I can do the following
```julia
push!(Plots.PGFPlotsX.DEFAULT_PREAMBLE, "\\usepackage{fontspec}")
push!(Plots.PGFPlotsX.DEFAULT_PREAMBLE, "\\setmainfont{CMU Serif")
```
and it works. The issue is, that we can't know whether the user has the CMU fonts installed

daschw:
Hm, hard to say, did you check the effect on Plots load time of adding Latexify?

daschw:
Is there a way to check if CMU is available?

BeastyBlacksmith:
I don't know from julia, but it is possible from latex (but said to be slow)

BeastyBlacksmith:
The package is quite lightweight, I would say:
```julia
 @time using Latexify
  0.159964 seconds (175.22 k allocations: 10.154 MiB)
```

daschw (approved):
I'm in favor of this.

Issue #1783, opened 2018-10-08
Is the build script still needed?


KristofferC:
The build script at https://github.com/JuliaPlots/Plots.jl/blob/master/deps/build.jl has effectively not been touched for 2 years. Is it still really needed?

SimonDanisch:
Could be removed as part of https://github.com/JuliaPlots/Plots.jl/pull/1777 I guess...

mkborregaard:
Maybe it could be run as part of initialising the `plotly` backend? Removing it would void the  backend, as it doesn't use the Plotly.jl package at all. And some users swear by it.

SimonDanisch:
>Removing it would void the backend

That's why I'd make it part of #1777, since it wouldn't void the backend with my changes - could also do the download thing in the __init__ if required by the ENV.

PR #3023, opened 2020-10-02
replace build.jl by Artifacts.toml

BeastyBlacksmith:
I'm still a bit shaky if that really works out. So if someone could test this branch, who uses this feature, I would be happy.

Fix: https://github.com/JuliaPlots/Plots.jl/issues/1783


daschw:
What exactly should be tested? If plotly works when building with this branch?

daschw (approved):
Works for me, also with `ENV["PLOTS_HOST_DEPENDENCY_LOCAL"] = "true"`

BeastyBlacksmith:
I don't really know when this environment variable is actually used... I guess it might be an IJulia thing. Building shouldn't be necessary anymore.

daschw:
Yes, works for me also in IJulia with `ENV["PLOTS_HOST_DEPENDENCY_LOCAL"] = true`.

BeastyBlacksmith:
Great, thanks for checking!

KristofferC:
nice!

fredrikekre:
I am pretty sure this does not work; artifact downloads has to be .tar.gz?

src/init.jl
```diff
@@ -74,16 +76,22 @@ function __init__()
         end
     end
 
-    if haskey(ENV, "PLOTS_HOST_DEPENDENCY_LOCAL")
-        use_local_plotlyjs[] = ENV["PLOTS_HOST_DEPENDENCY_LOCAL"] == "true"
-        use_local_dependencies[] = isfile(plotly_local_file_path) && use_local_plotlyjs[]
-        if use_local_plotlyjs[] && !isfile(plotly_local_file_path)
-            @warn("PLOTS_HOST_DEPENDENCY_LOCAL is set to true, but no local plotly file found. run Pkg.build(\"Plots\") and make sure PLOTS_HOST_DEPENDENCY_LOCAL is set to true")
+    if get(ENV, "PLOTS_HOST_DEPENDENCY_LOCAL", "false") == true
```
fredrikekre, line 79: This path is never taken, since what you get back is a the string `"true"` if you have set it to that, so that is never `== true`

BeastyBlacksmith:
> I am pretty sure this does not work; artifact downloads has to be .tar.gz?

From https://github.com/JuliaLang/Pkg.jl/issues/1921 and the iris exampe I got the impression it does work this way, if I don't use the artifact string macro

src/init.jl
BeastyBlacksmith: I changed that in https://github.com/JuliaPlots/Plots.jl/pull/3067

PR #3025, opened 2020-10-02
warn against `lims` being AbstractRange

### LINKED ISSUE #2914: [FR] lims with range silently ignored
Hello everybody,

I could not find a matching existing issue, there is some overlap with the feature request in [2687.](https://github.com/JuliaPlots/Plots.jl/issues/2687)

I would like to suggest that this
```
plot(0:10, sin.(0:10), ylims=(-1:1.5))
```
should at least give a warning. At the moment ylims is silently ignored. Using a tuple of course fixes the problem. Alternatively, allowing a range would also be fine from my perspective. Perhaps this even extends to other keyword arguments which might be ignored silently ?

Coming from gnuplot muscle memory kicked in and I needed to look a the code for some time to see my mistake.

Thank you very much for considering this !

Best Regards

Moelf:
fix #2914

### DIFF
--- a/src/args.jl
+++ b/src/args.jl
@@ -1489,10 +1489,6 @@ function _update_axis(axis::Axis, plotattributes_in::AKW, letter::Symbol, subplo
         # then get those args that were passed with a leading letter: `xlabel = "X"`
         lk = Symbol(letter, k)
         if haskey(plotattributes_in, lk)
-            # warn against using range in x,y,z lims
-            if k==:lims && plotattributes_in[lk] isa AbstractRange
-                @warn("$lk should be a Tuple")
-            end
             kw[k] = slice_arg(plotattributes_in[lk], subplot_index)
         end
     end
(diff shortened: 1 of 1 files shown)


daschw:
Thanks a lot, but this would be breaking as it changes
```julia
using Plots
plot(rand(10), xaxis = 3:3:9)
```
from
axisold
to
axisnew

Moelf:
@daschw I knew I missed something since that `_update_axis` is very general and no place to use multi-dispatch mechanics.
What do you think about this manual dispatched warning?

src/args.jl
```diff
@@ -1491,6 +1489,10 @@ function _update_axis(axis::Axis, plotattributes_in::AKW, letter::Symbol, subplo
         # then get those args that were passed with a leading letter: `xlabel = "X"`
         lk = Symbol(letter, k)
         if haskey(plotattributes_in, lk)
+            # warn against using range in x,y,z lims
+            if k==:lims && plotattributes_in[lk] isa AbstractRange
```
daschw, line 1493: Does this also warn for `plot(0:1, 0:1, lims = -2:2)`?

Moelf: it does, which is weird because I thought in that case `lk` won't be in the keys

daschw: Ok, wouldn't have expected that, but great!

daschw:
Yes, I think that's the way to go, thanks for the quick reply!

Moelf:
@daschw but in the case of `lims`, it warns 3 times (x,y,z) separately, which is kinda annoying

daschw:
Yes, we should find a workaround for that before merging, maybe something along the lines of https://github.com/JuliaPlots/Plots.jl/blob/master/src/args.jl#L1154-L1181?

Moelf:
```
julia> plot(0:10, sin.(0:10), ylims=-1:2)
┌ Warning: lims should be a Tuple, not UnitRange{Int64}.
└ @ Plots ~/.julia/dev/Plots/src/args.jl:1545

julia> plot(0:10, sin.(0:10), lims=-1:2)
┌ Warning: lims should be a Tuple, not UnitRange{Int64}.
└ @ Plots ~/.julia/dev/Plots/src/args.jl:1545
```

daschw (approved):
Great, thanks!

PR #3043, opened 2020-10-06
Update arg_desc.jl

logankilpatrick:
Add additional context for mirror attribute per: https://discourse.julialang.org/t/putting-xaxis-ticks-and-labels-on-top-not-bottom/47854/2


daschw:
Thanks, I'm fine with this change, but, in general, the `attribute` - `xattribute`, `yattribute`, `zattribute` pattern applies to all axis arguments. Maybe we should rather highlight this for all axis attributes in http://docs.juliaplots.org/latest/generated/attributes_axis/ than explaining for individual attributes. What do you think?

logankilpatrick:
I agree with that. Best to highlight the overall trend (if it apples to all attributes) and maybe include a few examples.

PR #3049, opened 2020-10-09
more friendly error when x,y shape mis-match

### LINKED ISSUE #3048: [FR] Better error message when plotting vector "v" with array "A" and length(v) is not equal to size(A,1)
From a student who was frustrated plotting something like

plot( rand(5), rand(1,5) ) 

and got 
```
BoundsError: attempt to access 1-element Array{Float64,1} at index [1:5]
```

it would be better if the error can read something like

number of elements in x (5)  is not equal to number of rows in the Y array (1)

or something even better.

Moelf:
fix #3048
```
julia> plot( rand(5), rand(1,5) )
┌ Warning: length(x) == 5, but only sees 1 in the other inputs.
└ @ Plots ~/.julia/dev/Plots/src/output.jl:160

julia> plotly()
[ Info: For saving to png with the Plotly backend PlotlyBase has to be installed.
Plots.PlotlyBackend()

julia> plot( rand(5,3), rand(4,5) )
┌ Warning: length(x) == 5, but only sees 4 in the other inputs.
└ @ Plots ~/.julia/dev/Plots/src/output.jl:160
```

### DIFF
--- a/src/output.jl
+++ b/src/output.jl
@@ -147,7 +147,21 @@ end
 
 function Base.display(::PlotsDisplay, plt::Plot)
     prepare_output(plt)
-    _display(plt)
+
+    try
+        _display(plt)
+    catch err
+        # need to be careful not to supress other kinds of error
+        # could use a more general error 'teleporter'
+        # 1. handle when length(x) != size(y,1)
+        if err isa BoundsError && length(err.i)==1
+            ex_nrows = length(err.i[1])
+            actual_nrows = size(err.a)[1]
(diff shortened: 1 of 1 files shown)


daschw:
I would maybe rather add a size check maybe somewhere around here https://github.com/JuliaPlots/RecipesPipeline.jl/blob/master/src/series.jl#L91 (not sure where exactly) than introducing a `try ... catch`.

Moelf:
continue in https://github.com/JuliaPlots/RecipesPipeline.jl/pull/65

PR #3050, opened 2020-10-09
WIP add GR hexbin

### LINKED ISSUE #3031: [FR] Add support for hexbin for GR
It seems `GR.jl` does support `hexbin` plots but this is not exposed via `Plots.jl`.

Moelf:
fix #3031

image

clearly it doesn't work properly at the moment. I tried to do something similar to what we do for `heatmap` but realized we actually use `GR.drawimage` for heatmaps.

The issue is `GR.hexbin` doesn't take any `color, min, max` info. But I thought we were able to "correct" it after we draw the main `hexbin`.

Can you mentor me on this a bit? @BeastyBlacksmith

### DIFF
--- a/src/backends.jl
+++ b/src/backends.jl
@@ -355,6 +355,7 @@ const _gr_seriestype = [
     :wireframe,
     :volume,
     :shape,
+    :hexbin,
 ]
 const _gr_style = [:auto, :solid, :dash, :dot, :dashdot, :dashdotdot]
 const _gr_marker = _allMarkers
--- a/src/backends/gr.jl
+++ b/src/backends/gr.jl
@@ -1545,6 +1545,8 @@ function gr_add_series(sp, series)
         else
             gr_draw_image(series, x, y, z, clims)
         end
+    elseif st == :hexbin
+        gr_draw_hexbin(series, x, y)
     end
 
     # this is all we need to add the series_annotations text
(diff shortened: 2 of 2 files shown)


BeastyBlacksmith:
Unfortunately, I am not very familiar with GR... maybe @daschw or @jheinen could help you out here

daschw:
I'll have a look soon.

sethaxen:
@Moelf are you sure this does not work properly? From what I can tell, `GR.hexbin` only shows a bin if its count is nonzero. So with `GR.hexbin(randn(10), randn(10))`, we should expect to see at most 10 bins, which is what we see here.

I fiddled with this locally, and I couldn't find a way to change the color of `hexbin`, even while calling `GR.hexbin` directly. Actually, I couldn't get any style attributes to change `hexbin`. e.g. setting `theme(:dark)`, `heatmap` has a dark plot pane, but `hexbin` has a white plot pane.

Moelf:
the plot works, as shown in the screenshot, but as I described, the problem is that `GR.hexbin` doesn't take the same customization arguments as others. I don't think we can make a hexbin that plot something completely out of style

sethaxen:
> as I described, the problem is that `GR.hexbin` doesn't take the same customization arguments as others

That's only part of the problem though. e.g. `GR.drawimage` does not take arguments to style the background and axes; that is handled elsewhere. But `GR.hexbin` appears to completely reset all existing styling to the plot, so if the background has already been set to be dark, `hexbin` resets it to be white.

@jheinen is this a GR bug?

jheinen:
You can not mix GR convenience functions (defined in `jlgr.jl`) with low-level GR functions. The GR backend in `Plots` currently only uses low-level functions - some functionality has even been re-written in Julia. I am aware that GR is not cleanly constructed at this point.

In any case, for what is desired here, further primitives must first be built into GR. So I will first have to provide low-level functions (for hexbin-style plots or images), that can then be used in the backend code.

PR #3067, opened 2020-10-15
Replace artifact by scratchspace

BeastyBlacksmith:
In https://github.com/JuliaPlots/Plots.jl/pull/3023#discussion_r505772221 it was pointed out that this path is never taken, this PR fixes that.

Question is open if that works out or if more needs to be done/ revert the artifact stuff.

Should close https://github.com/JuliaPlots/Plots.jl/issues/3071


BeastyBlacksmith:
@fredrikekre I replaced the artifact by the scratchspace. Could you have a look on this, if this is how it is supposed to be used?

fredrikekre:
Looks ok. For packaging purposes it would of course be nice to repack it as an artifact since you would not have to load Plots etc in order to instantiate it.

BeastyBlacksmith:
I can't quite follow. Where would that artifact be hosted?

fredrikekre:
Anywhere public.

PR #3069, opened 2020-10-15
add add_attributes macro

BeastyBlacksmith:
This macro does the following
```julia
 julia> @macroexpand 
           @add_attributes series struct Legend
                title::String = "default title"
           end
quote
    struct Legend
        title::String
    end
    begin
        _series_defaults[:legend_title] = "default title"
    end
end
```
I think it would be nice to add the non-underscore aliases also

### DIFF
--- a/src/Plots.jl
+++ b/src/Plots.jl
@@ -189,9 +189,9 @@ import RecipesPipeline: SliceIt,
 
 include("types.jl")
 include("utils.jl")
-include("components.jl")
 include("axes.jl")
 include("args.jl")
+include("components.jl")
 include("themes.jl")
 include("plot.jl")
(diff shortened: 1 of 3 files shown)


BeastyBlacksmith:
I added non_underscore and plural aliases and (hopefully) improved the alias generation in general. Might qualify for a minor release

daschw:
I'm not sure, I follow completely. Where would this new macro be used? In Plots or by Plots users?

BeastyBlacksmith:
That is intended for internal usage to ease the organization of attributes (I plan to use it for the Legend overhaul)

daschw:
OK, thanks for the explanation!

PR #3074, opened 2020-10-16
fix guide position in 3d plots in GR

daschw:
Changes
ax3dold
to
ax3dnew


daschw:
also fixed z axis guide rotation now
ax3dnewrot

ptoche:
Isn't the z label supposed to be rotated in the following? Thanks.

    using Plots
    gr()  # just in case
    t = range(0, 8π, length = 100)
    x = t .* map(cos, t)
    y = (0.1t) .* map(sin, t)
    z = 1:100
    plot(x, y, z, xlabel = "x", ylabel = "y", zlabel = "z")


a

PR #3075, opened 2020-10-17
use CompileBot 1

aminya:
This changes CompileBot from the master branch to the registered version.

Successful generated PR: https://github.com/aminya/Plots.jl/pull/26


daschw:
Thanks!

PR #3101, opened 2020-10-27
improve histogram docs

Moelf:


src/examples.jl
```diff
@@ -813,7 +813,9 @@ const _examples = PlotExample[
     ),
     PlotExample( # 38
         "Histogram2D (complex values)",
-        "",
+	"""
+	Other possible `normed` mode include `:probability`, where bin area (volumn) is not divided.
```
BeastyBlacksmith, line 817: It is a bit unclear to me, what you are trying to say here

Moelf: the `_make_hist` takes a `normed` argument, the example only shows `normed=true`, but under the hood, it calls the `normalize!` from `StatsBase`, with normalization mode default to `:pdf`, but there's actually another normalization mode, which is called `:probability`

BeastyBlacksmith: I think these modes are quite good covered in https://docs.juliaplots.org/latest/generated/attributes_series/.
This example aims to show how complex data can be visualized as a 2D Histogram.
I agree it would be good to also have a graphic that compares the different modes (like in spirit with https://github.com/JuliaPlots/PlotDocs.jl/issues/217), but here is not the right place.
Adding a different example would also be fine.

Moelf: ah, I see, makes sense

Issue #3063, opened 2020-10-14
[BUG] certain ticks omitted if x axis is a categorial string


oheil:
scatter (and plots) omit certain ticks if the x axis consists of categorial string values.

MWE:
```
using Plots
using DataFrames
data=DataFrame(x=string.(1:23),y=rand(Float64,23))
scatter(data[:x],data[:y],xaxis="x",yaxis="y")
```

![omitted_ticks](https://user-images.githubusercontent.com/5408557/95982798-de860700-0e20-11eb-8d34-70f66dead56b.jpg)

### Backends

This bug occurs on ( insert `x` below )

```
Backend      | yes | no  | untested
-------------|-----|-----|---------
gr (default) |  x  |     |
pyplot       |  x  |     |
plotly       |  x  |     |
plotlyjs     |  x  |     |
pgfplotsx    |     |     | x
inspectdr    |     |     | x
```

### Versions

Plots.jl version:
`Plots v0.28.4`
Backend  version (`]st -m`):
```
PyPlot v2.9.
PlotlyJS v0.14.0
```

Output of `versioninfo()`:
```
Julia Version 1.5.0
Commit 96786e22cc (2020-08-01 23:44 UTC)
Platform Info:
  OS: Windows (x86_64-w64-mingw32)
  CPU: Intel(R) Core(TM) i7-4770 CPU @ 3.40GHz
  WORD_SIZE: 64
  LIBM: libopenlibm
  LLVM: libLLVM-9.0.1 (ORCJIT, haswell)
Environment:
  JULIA_PKGDIR = c:\Program Files\Julia-0.6.2\packages\
  JULIA_SHELL = C:\Windows\System32\WindowsPowerShell\v1.0\powershell.exe
```

daschw:
You can add `xticks = :all` to your `scatter` command.

oheil:
Still it is surprising, that some ticks are omitted. That there exists a workaround doesn't mean it isn't a bug. Of course this is up to you to decide.

daschw:
I agree that defaulting to `ticks = :all` in these cases would be a better default.

daschw:
However, defaults are always tricky. Consider
```julia
x = [string(letter, num) for letter in 'a':'z' for num in 0:9]
```
and compare
```julia
plot(x, rand(length(x)))
```
[...]

PR #3105, opened 2020-10-27
more regular automatic categorical ticks

daschw:
Changes
```julia
using Plots
x = [string(letter, num) for letter in 'a':'z' for num in 0:9]
plot([plot(x[1:n], 1:n) for n in (15, 18, 29, 52)]...)
```
from
ticks_before
to
ticks_regular

@oheil would you consider this a fix for #3063?

### DIFF
--- a/src/axes.jl
+++ b/src/axes.jl
@@ -227,8 +227,9 @@ function get_ticks(sp::Subplot, axis::Axis; update = true)
                 if !isempty(dvals)
                     # discrete ticks...
                     n = length(dvals)
-                    rng = if ticks == :auto
-                        Int[round(Int,i) for i in range(1, stop=n, length=min(n,15))]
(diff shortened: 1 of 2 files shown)


oheil:
It looks much better and, more important, it doesn't look like a bug anymore. 
I am afraid that this is just a temporary solution as the number of ticks shown should be a function of the total width of the plot (for x axis), font size and longest string in the labels and if labels are plotted horizontal or vertical or at all... ok, too complex for now, but on the long run maybe worth to think of.
Yes, I like it now, and for the super fancy solution: it is always possible to take complete control over the ticks (I guess it is, isn't it?).

BeastyBlacksmith (approved):
I think the regularity is definitely an improvement. Could you add a unit test for this?

PR #3113, opened 2020-10-31
Switch to plotly-latest

daschw:


BeastyBlacksmith:
why plotly latest? That was changed deliberately recently : https://github.com/JuliaPlots/Plots.jl/pull/2779

daschw:
Thanks for the hint! To be honest I just wanted to check if legendtitle worked with a newer version of plotly - it does not - and opened the PR to keep the changes. But I guess we should not do this then. However, we could update to a more recent version.

BeastyBlacksmith:
> However, we could update to a more recent version.

Agreed. I don't have any objections to setting it to the current version

Issue #319, opened 2016-06-10
Recipe ideas


tbreloff:

ChrisRackauckas:
What about column scatter plots? They are pretty widely used in Biology since it's like a boxplot in that shows an error bar, but is also plotting the actual data values. At the bottom of this is a good example (made in R):

![13227_2016_40_fig12_html](https://cloud.githubusercontent.com/assets/1814174/15992153/e77428de-307a-11e6-8ed3-177f693bb5b6.jpg)

diegozea:
In that omage, also the upper bars with the significance level are useful and generally hard to make, but should be easy to make a recipe for that significance annotations with Plots.

ChrisRackauckas:
If adding significance annotations is possible that was be fantastic, but I thought it would be hard (it would be easy to tell it that a certain plot is *** significant, but then for it to automatically place the annotations? That seems difficult).

tbreloff:
No. Annotations should be fairly straightforward (depending on what it is
of course). You'd just create a new series with 'primary = false'.

tbreloff:
For anyone watching the repo, I just wanted to give a quick demo of features I just added: `iter_segments` and the `curves` series recipe.

![tmp](https://cloud.githubusercontent.com/assets/933338/16088975/491adae6-32f7-11e6-8c6a-8aa263340d6c.png)

If you can't figure out by the screenshot, I added a nice convenience iterator `iter_segments`, which takes 1 or more vectors (or anything that can be "cycled", which includes scalar values), and gives an iterator that, at each iteration, returns an integer range which is the next consecutive set of indices where the inputs are all non-NaN.  I'll be replacing things like `gr_polyline` with this sort of iteration logic, though it has other uses:

The `:curves` series recipe will split the x/y/z/fillrange/line_z coordinates into segments, and rebuild each segment using the given coordinates as "control points" in a bezier curve.  So you can create nice curves simply by specifying the control points separated by NaN's.  The optional `npoints` keyword arg specifies how many line segments to use to draw each curve segment.

mkborregaard:
Rather than choropleths, or as a way of doing that, it could be nice with a recipe to plot shapefiles, i.e. the objects returned by Shapefile.jl. Allowing colour would make an easy way to do choropleths, in addition to the other uses of shapefiles.

tbreloff:
This should be pretty straightforward. Can we add a recipe to shapefile.jl?

mkborregaard:
[...]

PR #3134, opened 2020-11-10
Proposes a 3D quiver

kellertuer:
This PR introduces a quiver functionality (at least partially, see below) for 3D quivers, cf. #319.

This extension can be seen by looking at
```julia
using Plots
x = [1.0, 2.0, 3.0]
y = [0.0, 1.0, 0.0]
z = [1.0, 0.0, 0.0]
u = [1.0, 1.0, 1.0]
v = [0.5, 0.5, 0.5]
w = [0.0, 0.0, 1.0]
s = quiver(x,y,z, quiver=(u,v,w))
```

which was previously a little strange (took `x` and `z` but `u` and `w`, first and last quiver component)
pre

and now it correctly looks like
post

Works/tested with
* GR
* PyPlot

## Questions
* What would be the best way to hack an arrow in 3D to make it work e.g. with `Plotly`?
* Which tests are required here?


BeastyBlacksmith:
That is a really good addition. I don't know about plotly, but as a test it would be sufficient to add an example at the end of the examples array here https://github.com/JuliaPlots/Plots.jl/blob/master/src/examples.jl, add the number to the excludes of the backends where it doesn't work and then do a PR against PlotReferenceImages to add the reference image for the GR backend.

kellertuer:
Thanks, I have to check how one can add things to that list, but then I'll do that for sure.

kellertuer:
Uh, nice! I was just starting to test and got lost in GR arrow head sizes and (maybe) missing scaling of quivers, when I wanted to add a 2D example (found a bug in my code, too). Will update/merge that here, then.

daschw:
I'm sorry about that commit mess above. I added a new test example. I hope that's fine.
```julia
using Plots

ϕs = range(-π, π, length=50)
θs = range(0, π, length=25)
θqs = range(1, π-1, length=25)

x = vec([sin(θ) * cos(ϕ) for (ϕ, θ) in Iterators.product(ϕs, θs)])
y = vec([sin(θ) * sin(ϕ) for (ϕ, θ) in Iterators.product(ϕs, θs)])
z = vec([cos(θ) for (ϕ, θ) in Iterators.product(ϕs, θs)])

u = 0.1 * vec([sin(θ) * cos(ϕ) for (ϕ, θ) in Iterators.product(ϕs, θqs)])
v = 0.1 * vec([sin(θ) * sin(ϕ) for (ϕ, θ) in Iterators.product(ϕs, θqs)])
w = 0.1 * vec([cos(θ) for (ϕ, θ) in Iterators.product(ϕs, θqs)])

quiver(x,y,z, quiver=(u,v,w))
```
quiver3

kellertuer:
oh, what a neat example!

kellertuer:
I was thinking about adding a 2D quiver example, too, but got lost since this
```
using Plots
xr = range(-π, π, length=41)
yr = range(-π, π, length=41)

X = vec([ x for x ∈ xr, y ∈ yr])
Y = vec([ y for x ∈ xr, y ∈ yr])
U = vec([ 0.12*sin(x)*cos(y) for x ∈ xr, y ∈ yr])
V = vec([ 0.12*cos(x)*sin(y) for x ∈ xr, y ∈ yr])

s = quiver(X, Y, quiver=(U,V))
```
has huge arrowheads in `GR`.
And maybe for another PR one could also introduce an auto-down/up-scaling option of the vectors (I think both peplos and Matlab do this)

daschw:
I think we can also investigate the arrowhead size issue and the 2D example in another issue/PR.

daschw:
I like the auto-scaling idea.

kellertuer:
Ok, sure.
For auto-scaling I have an idea, I will think about it and propose something maybe next week – something along the lines of „scale down to the longest vector being as long as the smallest distance of domain points.“

daschw:
Thanks for your contribution @kellertuer !

PR #3154, opened 2020-11-15
Default fontsize for pgfplotsx backend

willtebbutt:
@BeastyBlacksmith , this is as per our discussion on [this discourse thread](https://discourse.julialang.org/t/plots-jl-default-label-size/49454).

Is there any unit testing that you would like me to provide? Not sure how best to go about it.

edit: additionally, assuming this is an acceptable change, what's the Plots' policy regarding releases? Should I bump the minor / patch version, or do you generally wait to group new bits of code together?


BeastyBlacksmith:
I think it is fine as it is. You can check the existing pgfplotsx test for testing, if you want to add a test. Otherwise i will merge tomorrow and do a point release.

Issue #3157, opened 2020-11-20
Deprecation warnings for HDF5 backend


BeastyBlacksmith:
Running the tests print a bunch of deprecation warnings for the HDF5 backend. These should be easy to fix.

wg030:
All these warnings pissed me off for so many days that I have now created a pull request :D
See #3175. Would be cool if you merged it!

PR #3175, opened 2020-12-01
Handling Deprecation warnings for HDF5 backend

wg030:
I have updated hdf5.jl so that all deprecations warnings should be gone now!

Fix: #3157

### DIFF
--- a/src/backends/hdf5.jl
+++ b/src/backends/hdf5.jl
@@ -46,7 +46,7 @@ import Dates
 
 #Plots.jl imports HDF5 to main:
 import ..HDF5
-import ..HDF5: HDF5Group, HDF5Dataset
+import ..HDF5: Group, Dataset
 
 import ..Colors, ..Colorant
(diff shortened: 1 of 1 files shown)


BeastyBlacksmith (approved):
Great! Thank you

ranocha:
Thanks! It would be nice to have a new release of Plots.jl, since these warning still show up in the latest release (from November). We use Plots.jl and would like to run some tests with `@test_nowarn` which currently fails (and the only warnings we see are related to Plots.jl and HDF5).

wg030:
@ranocha I can't help you with a new release, but have you tried `add Plots#master`, works like a charm for me!

ranocha:
Thanks for that hint, @wg030. I have used `add Plots#master` locally, but we would like to be able to run these tests using a released version of Plots.jl such that everything works reliably also on standard installations (CI, users of our packages etc.).

PR #3178, opened 2020-12-04
Set extra mime only when HTML format

jks-liu:
Extra mine is not used by other format.

For example

```julia
using Plots
plotlyjs()
plot(rand(1000000), x->x, fmt=:png)
```

Above codes make jupyter file be about 60 MiB.
After changing, jupyter file is only 40 KiB.


daschw:
Thanks!

PR #3180, opened 2020-12-06
Fix #1531, Let xlims be set to Dates or DateTimes

### LINKED ISSUE #1531: xlim does not work with DateTimes
Similar to #1026,

```julia
> plot([DateTime(2016), DateTime(2017)], [3,2], xlim=(DateTime(2015), DateTime(2018)))

MethodError: no method matching optimize_datetime_ticks(::DateTime, ::DateTime; k_min=2, k_max=4)
```

and, like #1026, using `Dates.value` solves the issue.

ThatcherC:
Resolves #1531, allowing xlims to be set to Date or DateTime values. Also adds two unit tests that show xlims working with both Date and DateTime.

### DIFF
--- a/src/axes.jl
+++ b/src/axes.jl
@@ -94,6 +94,8 @@ function attr!(axis::Axis, args...; kw...)
                 for vi in v
                     discrete_value!(axis, vi)
                 end
+            #could perhaps use TimeType here, as Date and DateTime are both subtypes of TimeType
+            # or could perhaps check if dateformatter or datetimeformatter is in use
             elseif k == :lims && isa(v, Tuple{Date,Date})
                 plotattributes[k] = (v[1].instant.periods.value, v[2].instant.periods.value)
             elseif k == :lims && isa(v, Tuple{DateTime,DateTime})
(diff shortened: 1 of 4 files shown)


daschw:
Thanks! Sorry I merged your other PR before this now we have a conflict here. Could you rebase this PR on current master? I think ideally, we should somehow call type recipes on axes attributes like ylims (or ticks - where we special case for Dates types). However, until we come up with a clear general interface for that, I think special casing for Dates types here for lims ony is an improvement.

ThatcherC:
No problem! I'll rebase soon and push it back up. I agree there's probably a better way to solve the issue - hopefully this is helpful for now. One problem I noticed with this fix is that you can't use DateTimes for `xlims` if the series x data is Date objects, since Dates get converted to integer days while DateTimes get converted to integer milliseconds. Maybe an eventual solution could convert Dates and DateTimes both to the same numerical representation (like milliseconds)?

Also - should I worry about the Julia nightly CI checks? f6627f3 passed the Julia night/macos_latest check, but 9ba633b did not, even though there were no changes in the 9ba633b commit.

daschw:
Thanks! I don't think the nightly errors are related to the changes here. I like the idea of converting Dates and DateTimes to the same value internally.

nathanrboyer:
Can you also add support for the `Dates.Time` type?

PR #3189, opened 2020-12-10
various pyplot fixes

### LINKED ISSUE #3155: [BUG] Colorbar flows partially outside the figure when specifying xticks in pyplot backend
## Details

MWE:
```
using Plots; pyplot()
contour([3 2 3;1 2 3;5 6 7],colorbar=true)
contour([3 2 3;1 2 3;5 6 7],xticks = [1,5,2.0,2.5],colorbar=true)
```
The first figure is the correct one:

Screen Shot 2020-11-16 at 12 26 38 AM

The second figure is missing the right part of the colorbar. Upon further investigation, whenever I specify the largest xtick to also be the largest value of the x-axis, I got the correct colorbar, but if the xticks are all within the interval of the x-axis, specifying xticks can produce the following figure:

Screen Shot 2020-11-16 at 12 26 25 AM

### Backends

gr (default) |     |     | x
pyplot       |  x   |     |
plotly       |     |     |x
plotlyjs     |     |     |x
pgfplotsx    |     |     |x
inspectdr    |     |     |x

### Versions

Plots.jl version: [91a5bcdd] Plots v1.6.7
Backend  version (`]st -m`): [d330b81b] PyPlot v2.9.0
Output of `versioninfo()`:
Julia Version 1.5.2
Commit 539f3ce943 (2020-09-23 23:17 UTC)
Platform Info:
  CPU: Intel(R) Xeon(R) Platinum 8268 CPU @ 2.90GHz
  WORD_SIZE: 64
  LIBM: libopenlibm
  LLVM: libLLVM-9.0.1 (ORCJIT, cascadelake)

isentropic:
Quick summary of the changes
```
p = plot(rand(50), linez=rand(50),  thickness_scaling=2, cbar=:best, tick_direction=:in, framestyle=:box, cbartitle="two\nlines")
```
Before
image

After
image

Fix #3155

### DIFF
--- a/src/backends/pyplot.jl
+++ b/src/backends/pyplot.jl
@@ -920,8 +920,6 @@ function _before_layout_calcs(plt::Plot{PyPlotBackend})
             end
             kw[:spacing] = "proportional"
 
-            fig = plt.o
-
             if RecipesPipeline.is3d(sp) || ispolar(sp)
                 cbax = fig."add_axes"([0.9, 0.1, 0.03, 0.8])
                 cb = fig."colorbar"(handle; cax=cbax, kw...)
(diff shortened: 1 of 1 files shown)


isentropic:
should fix https://github.com/JuliaPlots/Plots.jl/issues/3155

src/backends/pyplot.jl
```diff
@@ -1010,15 +1036,11 @@ function _before_layout_calcs(plt::Plot{PyPlotBackend})
             PyPlot.PyCall.hasproperty(ax, axissym) || continue
             axis = sp[axissym]
             pyaxis = getproperty(ax, axissym)
-            if axis[:mirror] && letter != :z
```
isentropic, line 1013: Mirror will not work for 3d as matplotlib does not support it. Mirror right now does not work properly at all, so it is best to fix this later in a separate PR

src/backends/pyplot.jl
```diff
@@ -1202,13 +1225,26 @@ function _update_min_padding!(sp::Subplot{PyPlotBackend})
         end
     end
 
-    # optionally add the width of colorbar labels and colorbar to rightpad
-    if haskey(sp.attr, :cbar_ax)
-        bb = py_bbox(sp.attr[:cbar_handle]."ax"."get_yticklabels"())
-        sp.attr[:cbar_width] = width(bb) + (sp[:colorbar_title] == "" ? 0px : 30px)
-        rightpad = rightpad + sp.attr[:cbar_width]
+    if haskey(sp.attr, :cbar_ax) # Treat colorbar the same way
```
isentropic, line 1228: I think we might as well treat colorbar the same (as in matplotlib it is too an ax)

isentropic:
Irrelevant to this, but I want to revert the changes introduced when 
```
function py_surround_latextext(latexstring, env)
    if !isempty(latexstring) && latexstring[1] == '$' && latexstring[end] == '$'
    if !isempty(latexstring) && latexstring[1] == '$' && latexstring[end] == '$'
        unenclosed = latexstring[2:end-1]
        unenclosed = latexstring[2:end-1]
    else
    else
        unenclosed = latexstring
        unenclosed = latexstring
    end
    end
    PyPlot.LaTeXStrings.latexstring(env, "{", unenclosed, "}")
    PyPlot.LaTeXStrings.latexstring(env, "{", unenclosed, "}")
end
end
```
Hacky surrounding with dollar signs and also adding `\mathenvnormal{}` hurts more than helps I think. It was only to achieve serif fonts just work, but I think we should just document that users need to change math font in `Plots.pyrcparams` and if they want set `usetex=true` and what not. 
Also I'm planning to add more colorbar features soon as this gets resolved. I think we can solve the colorbar issue for 2d plots quite nicely with pyplot

isentropic:
@daschw do you think this is ok to merge?

daschw (approved):
Sorry, it took me so long! This looks great!

PR #3220, opened 2021-01-11
Correct version string of plotly cache file name

jks-liu:
Maybe we need a const string variable for this.


daschw:
Thanks! 
Yes, probably it would be good to define something like `plotly_filename = "plotly-1.57.1.min.js"` and use it for both download and local file.

jks-liu:
#3229 contains this PR, close it.

PR #3229, opened 2021-01-18
Proper Plotly JavaScript setup in IJulia

### LINKED ISSUE #174: plotly backend in notebook
Now that we are using requirejs  to load plotly (a good thing) we need to make sure that all calls to `Plotly` (javascript) functions happen after requirejs has loaded plotly.

We can do this with the following js:

```javascript
require(['plotly'], function(Plotly) {
    // Plotly code here
});
```

In PlotlyJS.jl I have changed all the display commands to wrap calls to `Plotly.newPlot` in this require statement. You can see the change [here](https://github.com/spencerlyon2/PlotlyJS.jl/commit/4c030351c2f02b13480e9b6562cbd8e5ca2c0770)

The reason I did this was because I noticed that in nbviewer if I had very simple plots (easy to generate/parse quickly) at the top of a notebook the javascript console complained about `window.Plotly` not being defined and the plot wasn't shown. However, plots later in the notebook showed up without a problem. 

with plotlyjs backend you get this for free, but I believe you might want to make these adjustments for the plotly backend

### LINKED ISSUE #2441: Plotly reference not found if Browser Cache on
I've found the strange behaviour that when using the Plotly interface in a Jupyter session, the plots will only work the first time. Afterwards the browser console shows 
```
Plotly: reference not found
```

After some debugging I found that disabling the browser cache strangely solves the problem. This behaviour was found on Firefox and Edge.

System:
I'm running current Jupyter notebook from Ubuntu LTS within the WSL on Windows 10. I've installed Plots, PlotlyJS and ORCA with all it's dependencies. Starting plots with plotly only shows the above mentioned error in the browser.

jks-liu:
Improvements:
- Interactive plots show correctly when page re-open or re-load
- Plotly JavaScript is loaded only once when first time use
- May fix #2441
- May fix #174

This PR contains items of #3220, so I'll close it.

### DIFF
--- a/src/backends/plotly.jl
+++ b/src/backends/plotly.jl
@@ -887,7 +887,7 @@ const ijulia_initialized = Ref(false)
 
 function plotly_html_head(plt::Plot)
     plotly =
-        use_local_dependencies[] ? ("file:///" * plotly_local_file_path[]) : "https://cdn.plot.ly/plotly-1.57.1.min.js"
+        use_local_dependencies[] ? ("file:///" * plotly_local_file_path[]) : "https://cdn.plot.ly/$(_plotly_min_js_filename)"
 
     include_mathjax = get(plt[:extra_plot_kwargs], :include_mathjax, "")
     mathjax_file = include_mathjax != "cdn" ? ("file://" * include_mathjax) : "https://cdnjs.cloudflare.com/ajax/libs/mathjax/2.7.7/MathJax.js?config=TeX-MML-AM_CHTML"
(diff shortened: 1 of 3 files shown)


src/backends/plotly.jl
```diff
@@ -883,43 +883,54 @@ plotly_series_json(plt::Plot) = JSON.json(plotly_series(plt), 4)
 html_head(plt::Plot{PlotlyBackend}) = plotly_html_head(plt)
 html_body(plt::Plot{PlotlyBackend}) = plotly_html_body(plt)
 
-const ijulia_initialized = Ref(false)
-
 function plotly_html_head(plt::Plot)
     plotly =
-        use_local_dependencies[] ? ("file:///" * plotly_local_file_path[]) : "https://cdn.plot.ly/plotly-1.57.1.min.js"
+        use_local_dependencies[] ? ("file:///" * plotly_local_file_path[]) : "https://cdn.plot.ly/$(_plotly_min_js_filename)"
```
isentropic, line 888: Why do we need remove the head?

brenhinkeller:
It looks like Plots CI has been failing ever since this PR, though it's not clear that this was caused by anything in this PR?

daschw:
The test failures are caused by breaking changes in the HDF5 backend. I'm working on it ...

Issue #3238, opened 2021-01-24
Time to first plot approximately doubled in Plots v1.10.0


KristofferC:
Using Plots 1.10.1 I get:

```jl
julia> begin
              @time using Plots
              @time p = plot(1:0.1:10, sin.(1:0.1:10))
              @time display(p)
          end
  2.670063 seconds (5.54 M allocations: 413.097 MiB, 2.73% gc time, 1.47% compilation time)
 13.580643 seconds (47.47 M allocations: 2.721 GiB, 9.40% gc time, 1.79% compilation time)
  4.033910 seconds (6.88 M allocations: 397.034 MiB, 2.04% gc time, 45.89% compilation time)
```

Using 1.6.6 I get:

```jl
julia> begin
              @time using Plots
              @time p = plot(1:0.1:10, sin.(1:0.1:10))
              @time display(p)
          end
  3.114219 seconds (6.67 M allocations: 501.160 MiB, 6.29% gc time, 1.19% compilation time)
  2.229045 seconds (4.22 M allocations: 258.580 MiB, 5.47% gc time, 99.78% compilation time)
  3.958535 seconds (6.84 M allocations: 396.392 MiB, 2.03% gc time, 48.27% compilation time)
```

Specifially, the `plot` call seem to have gotten 6x slower.

~~I haven't looked into it more than these two versions.~~

Seems to be from v1.9.1 to v1.10.0.

PR #3239, opened 2021-01-24
Add basic CI Benchmarking via BenchmarkCI

ianshmean:
A suggestion, based on a slack discussion about trying to help avoid performance regressions like https://github.com/JuliaPlots/Plots.jl/issues/3238

This setup will post a comment with the result after every push to a PR, which might be a bit noisy. 
That can be changed to a less visible print in the action as per https://github.com/tkf/BenchmarkCI.jl#printing-benchmark-result-optional

cc. @brenhinkeller

### DIFF
--- a/.github/workflows/benchmark.yml
+++ b/.github/workflows/benchmark.yml
@@ -6,6 +6,8 @@ on:
 jobs:
   Benchmark:
     if: "!contains(github.event.head_commit.message, '[skip ci]')"
+    env:
+      GKS_ENCODING: "utf8"
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v2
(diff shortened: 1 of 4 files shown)


ianshmean:
It seems this won't be able to complete until after its merged, as it's trying to compare to master, where the benchmarking files are missing.

After this is merged, there is a question about the nature of report generated, given we're not interested in the median/mean time of the tests, but the `max` for first time to plot evaluation.

Here's an example report comment https://github.com/tkf/BenchmarkCI.jl/pull/17#issuecomment-577510333

Perhaps this can be merged, then we refine in another PR?

BeastyBlacksmith (approved):

ianshmean:
It seems the order that the benchmarks is run isn't alphabetical or repeatable. I'm reworking them so that order doesn't matter

daschw:
Nice, thanks a lot!

PR #3240, opened 2021-01-24
Add basic CI Benchmarking via BenchmarkCI - Pt. 2

ianshmean:
Continuation of #3239 given merge was needed to make the comparison to master work

### DIFF
--- a/benchmark/benchmarks.jl
+++ b/benchmark/benchmarks.jl
@@ -1,10 +1,10 @@
 using BenchmarkTools
+using Plots
 
 const SUITE = BenchmarkGroup()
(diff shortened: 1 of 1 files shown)


ianshmean:
@daschw this one's ready to go now. The benchmarks pass in this PR, but there were some residual errors on master.

So after this is merged, future PRs will work. I plan to open a test PR after this with an intentional slowdown, to see how the report looks

ianshmean:
I did some testing on my fork, and got a report generated for a PR that added a `sleep(1)` to `__init__()`
https://github.com/ianshmean/Plots.jl/pull/5#issuecomment-766449608

PR #3242, opened 2021-01-25
reenable precompiles

KristofferC:
Since https://github.com/JuliaPlots/Plots.jl/pull/3169, precompile statements have been disabled for Plots.jl since the `isfile` check was buggy (it thought that the relative path was the directory of the file but it is actually the package root, it works differently than `include` in that regard). The `isfile` check therefore failed and no precompile statements were executed. This fixes it and also simplifies the script.

Fixes https://github.com/JuliaPlots/Plots.jl/issues/3238


src/precompile_includer.jl
```diff
@@ -1,44 +1,3 @@
-should_precompile = true
+include("../deps/SnoopCompile/precompile/linux/1.5/precompile_Plots.jl")
 
-
-# Don't edit the following! Instead change the script for `snoop_bot`.
```
KristofferC, line 4: I see this but it seems this file is edited before anyway. I don't really see the point of it either.

daschw: The point was to load different precompile statements for different OSs and julia versions. However, we got spammed with so many automatic snoopcompilebot PRs that we decided to only us julia stable on linux for generating precompile statements. So now it's pointless.

daschw:
Thanks so much!

KristofferC:
The benchmarks seem to go a bit crazy and spam e.g.

```
 GKS: GKS not in proper state. GKS must be either in the state WSAC or SGOP in routine POLYLINE
GKS: GKS not in proper state. GKS must be either in the state WSAC or SGOP in routine FILLAREA
GKS: GKS not in proper state. GKS must be either in the state WSAC or SGOP in routine FILLAREA
GKS: GKS not in proper state. GKS must be either in the state WSAC or SGOP in routine POLYLINE
GKS: GKS not in proper state. GKS must be either in the state WSAC or SGOP in routine POLYLINE
``` 

cc @ianshmean

ianshmean:
@tkf any idea why `BenchmarkCI.postjudge()` would succeed with `[ Info: Comment posted.` but no comment here?

daschw:
@jheinen Is there anything we can do about these `GKS: GKS not in proper state. ...` messages in our tests and benchmarks?

src/precompile_includer.jl
daschw: @aminya will this change be overwritten again by CompileBot?

ianshmean:
@KristofferC would you mind adding this before the `postjudge()` step in `.github/workflows/benchmark.yml` so even if the comment fails again, we can see the report in the CI
```
      - name: Print judgement
        run: julia -e 'using BenchmarkCI; BenchmarkCI.displayjudgement()'
```

src/precompile_includer.jl
KristofferC: There have been other manual changes to this file and it didn't get changed, which is why I assumed it was fine.

daschw: OK, great.

jheinen:
@dashw: You can set `ENV["GKSwstype"] = "100"` to use off-screen rendering and test backends like PNG, PDF or SVG using `savefig`.

ianshmean:
The benchmark comment silently failed to appear again but the printed report did catch the speed up:
```
Results
  =========

  A ratio greater than 1.0 denotes a possible regression (marked with :x:),
  while a ratio less than 1.0 denotes a possible improvement (marked with
  :whitecheckmark:). Only significant results - results that indicate possible
  regressions or improvements - are shown below (thus, an empty table means
  that all benchmark results remained invariant between builds).

                     ID                 time ratio               memory ratio
  ––––––––––––––––––––– –––––––––––––––––––––––––– ––––––––––––––––––––––––––
            ["display"] 0.00 (5%) :whitecheckmark: 0.00 (1%) :whitecheckmark:
               ["load"]              1.07 (5%) :x:                  1.00 (1%)
  ["load_plot_display"] 0.65 (5%) :whitecheckmark: 0.00 (1%) :whitecheckmark:
               ["plot"]                  1.03 (5%)              1.02 (1%) :x:
```

`display` and `plot` won't be informative until we can either reduce it to 1 benchmark run, or just pay attention to the `max` time for each.

`load_plot_display` is the classic  TTFP test `using Plots; display(plot(1:0.1:10, sin.(1:0.1:10)))` in a new session.

More detail here https://github.com/JuliaPlots/Plots.jl/pull/3242/checks?check_run_id=1762843608#step:8:7

daschw:
Great, thanks again @KristofferC , @ianshmean and @jheinen !

aminya:
So, `isfile` is buggy? Which Julia versions are affected?

KristofferC:
`isfile` is not buggy, the usage of it was just wrong.

aminya:
Root path is calculated using this:
https://github.com/aminya/CompileBot.jl/blob/cdeed7452abc9a8242ddf89e0af94ab27f35e9f7/src/precompile_include.jl#L32
```julia
    precompiles_rootpath = GoodPath(relpath(precompiles_rootpath_in, dirname(package_path)))
    
```

Then this is used example like this:
https://github.com/aminya/CompileBot.jl/blob/cdeed7452abc9a8242ddf89e0af94ab27f35e9f7/src/precompile_include.jl#L119
```julia
				precompile_file_name = "$precompiles_rootpath/$eachos/precompile_$package_name.jl"
```
```julia
                @static if (isfile("$precompile_file_name"))
                    include("$precompile_file_name")
                    _precompile_()
                end
```

So, how can I convert this to a format that `isfile` understands?

KristofferC:
The `isfile` check there should likely be `isfile(joinpath(@__DIR__, precompile_file_name))` so you get it relative to the current file (if that is what you want). Just as a note, `isfile("$precompile_file_name")` is just a worse way of writing `isfile(precompile_file_name)`.

aminya:
> The `isfile` check there should likely be `isfile(joinpath(@__DIR__, precompile_file_name))` so you get it relative to the current file (if that is what you want). 

Let me try that.

> As a note, `isfile("$precompile_file_name")` is just a worse way of writing `isfile(precompile_file_name)`.

This code is inside a string. 

https://github.com/aminya/CompileBot.jl/blob/cdeed7452abc9a8242ddf89e0af94ab27f35e9f7/src/precompile_include.jl#L120
```julia
            multistr = multistr * """
                @static if (isfile("$precompile_file_name"))
                    include("$precompile_file_name")
                    _precompile_()
                end
            """
```

PR #3246, opened 2021-01-26
CompileBot use default precompile path

aminya:
If you want to use one path for every version or operating system, you should delete these specific configs


daschw:
Now this gives
```julia
ERROR: LoadError: LoadError: could not open file /home/runner/work/Plots.jl/Plots.jl/deps/SnoopCompile/precompile/linux/1.5/precompile_Plots.jl
```

aminya:
It was because the file had been modified manually. After merging this, a new PR will be generated that adds the precomopiles.

daschw:
Great, thanks, so is this ready to go, or should we wait for the tests (or will they fail anyway)?

aminya:
This is ready to go. The PR generated by the previous run:

https://github.com/aminya/Plots.jl/pull/30/

aminya:
But maybe let's wait until it generates the files again.

aminya:
For strange reasons, CompileBot 1.0.5 is installed instead of 1.0.6 in my online run 😳. Hopefully, it works in your run.

PR #3255, opened 2021-01-27
Pairplot support 

TheCedarPrince:
Based off a Slack conversation where I saw that Plots.jl doesn't have native support for pair plotting, I am opening this PR to support the missing functionality!

Currently just a draft. @daschw - I will ping you more when this is ready for review/merging. Right now, is there anything in particular you want me to prioritize with this functionality? Thanks!

### DIFF
--- a/pairplot.jl
+++ b/pairplot.jl
@@ -1,0 +1,82 @@
+using Plots
+
+df = rand(5, 5) * 10 |> Matrix
+
+function pairplot(df)
+    rows, cols = size(df)
+    y = ones(3)
+    title = Plots.scatter(
+        y,
+        marker = 0,
+        markeralpha = 0,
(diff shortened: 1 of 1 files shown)


daschw:
Hi, sorry, apparently I was not fast enough on Slack :wink:
I would put the code in `src/recipes.jl` and implement it as a user recipe. This way users of `pairplot` can still modify Plots attributes and are not limited to the default values for e.g. `ytickfont` you set in the function. For inspiration you could have a look at the `corrplot` recipe in StatsPlots https://github.com/JuliaPlots/StatsPlots.jl/blob/master/src/corrplot.jl.

TheCedarPrince:
No worries! Probably good that it is here so we have a record of it @daschw. I will be working on this over the next couple days/weeks.

PR #3262, opened 2021-01-30
Add support for custom tick labels

ma-laforge:
Also:
 - Correctly register changes to "plot.layout"
 - Fix deprecated use of series[:marker[stroke][color/alpha]].

Requires InspectDR v0.3.10 (Added facilities for custom ticks).


daschw:
Nice, thanks!

PR #3267, opened 2021-01-31
Fix typo preventing minor ticks from working in 3D plots

sefffal:
Currently, minor ticks do not seem to work in plots with 3D axes. I noticed this bug while trying to use `Plots.showtheme(:ggplot2)` which was failing due to an undefined variable error.

This PR simply adds a missing underscore to fix the issue.


daschw:
Thanks!

PR #3277, opened 2021-02-07
don't stringify argument to `warn_on_recipe_aliases!` early

### LINKED ISSUE #3276: [PERFOMANCE] Slowdown caused by new type-computation in 1.6
Ref: https://github.com/JuliaLang/julia/issues/39509

The new short-form type computation in 1.6 slows down the computation of the user recipe signature here:

https://github.com/JuliaPlots/RecipesPipeline.jl/blob/c251015a5420818229cdfd284c0802a7db19085c/src/utils.jl#L232-L234

As a fix, it looks as if this string computation is only made to potentially display a warning in `warn_on_recipe_aliases!`:
https://github.com/JuliaPlots/Plots.jl/blob/a955ded5c87fff9265c416da8c3b614f0c624878/src/pipeline.jl#L5-L20

If instead the string is computed only if the warning is determined to appear, this slowdown should go away. This would probably also lead to a small speedup on 1.5 and earlier since the `userrecipe_signature_string` also shows up on the 1.5 profile in the issue above.

As far as I can tell, this would require changing 4 calls to `warn_on_recipe_aliases!` in RecipesPipeline by passing the argument directly, instead of stringifying first:
https://github.com/JuliaPlots/RecipesPipeline.jl/search?q=warn_on_recipe_aliases%21

jebej:
needs matching RecipesPipeline changes
fixes #3276

### DIFF
--- a/src/pipeline.jl
+++ b/src/pipeline.jl
@@ -5,13 +5,24 @@
 function RecipesPipeline.warn_on_recipe_aliases!(
     plt::Plot,
     plotattributes::AKW,
-    recipe_type,
-    signature_string
+    recipe_type::Symbol,
+    args
 )
     for k in keys(plotattributes)
         if !is_default_attribute(k)
             dk = get(_keyAliases, k, k)
             if k !== dk
+                if recipe_type == :user
+                    signature_string = RecipesPipeline.userrecipe_signature_string(args)
+                elseif recipe_type == :type
+                    signature_string = RecipesPipeline.typerecipe_signature_string(args)
+                elseif recipe_type == :plot
+                    signature_string = RecipesPipeline.plotrecipe_signature_string(args)
+                elseif recipe_type == :series
+                    signature_string = RecipesPipeline.seriesrecipe_signature_string(args)
+                else
+                    throw(ArgumentError("Invalid recipe type `$recipe_type`"))
+                end
                 @warn "Attribute alias `$k` detected in the $recipe_type recipe defined for the signature $signature_string. To ensure expected behavior it is recommended to use the default attribute `$dk`."
             end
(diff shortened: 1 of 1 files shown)


jebej:
This change should also lead to a small improvement on 1.5, though I haven't benchmarked.

On 1.6, we have before this PR:

```julia
julia> @benchmark begin Plots.plot([1:100 1:100]); Plots.plot!([1:100 1:100]); Plots.plot!([1:100 1:100]) end
BenchmarkTools.Trial:
  memory estimate:  936.56 KiB
  allocs estimate:  13454
  --------------
  minimum time:     5.430 ms (0.00% GC)
  median time:      5.576 ms (0.00% GC)
  mean time:        5.742 ms (1.05% GC)
  maximum time:     9.023 ms (32.11% GC)
  --------------
  samples:          871
  evals/sample:     1
```

after PR

```julia
julia> @benchmark begin Plots.plot([1:100 1:100]); Plots.plot!([1:100 1:100]); Plots.plot!([1:100 1:100]) end
BenchmarkTools.Trial:
  memory estimate:  631.64 KiB
  allocs estimate:  11087
  --------------
  minimum time:     1.228 ms (0.00% GC)
  median time:      1.247 ms (0.00% GC)
  mean time:        1.326 ms (2.84% GC)
  maximum time:     8.066 ms (64.06% GC)
  --------------
  samples:          3768
  evals/sample:     1
```

daschw:
Awesome, thanks!

jebej:
This will probably require some kind of bounding in Projects.toml so that users aren't left in a state with incompatible RecipesPipeline and Plots versions

daschw:
thanks for the hint!

Issue #2820, opened 2020-06-27
[FR] Is this warning really appropriate?


juliohm:
Whenever I call a plot recipe twice that uses a colorbar by default, I get this warning:

![image](https://user-images.githubusercontent.com/3345261/85921325-f164d380-b851-11ea-899f-d58a08e9e868.png)

I wonder if this is something that could be fixed? To me it seems like we should be able to plot multiple series on a scene and get the colorbar updated? Is the warning just reflecting a missing implementation?

mkborregaard:
It's nice enough with the information, but it could be an `@info` and less defensively phrased (keeping just the first sentence)

juliohm:
Is it something that is not amenable to fix? The way that series get plotted one after another inhibits the update of the colorbar in the current design, is that it?

daschw:
Maybe we could check if the two series share the same color gradient and omit the warning/info in that case.

juliohm:
That seems reasonable and very useful to have.

juliohm:
Another warning that could go away:

![image](https://user-images.githubusercontent.com/3345261/85929199-41ab5800-b889-11ea-97f1-45bf83381b72.png)

isentropic:
Perhaps these warnings should be handled by Plots, not backends? If we were to cross check the series range and colorbar, it can only be checked in Plots

juliohm:
I think a quick removal of the warnings would already be an improvement. They are not helping much.

isentropic:
We cannot just remove the warnings, if this gets unnoticed it might produce wrong plots I think. For now, it could be changed to @info but that is it

juliohm:
These two warnings show up in this tutorial: https://www.youtube.com/watch?v=Ee8K2Dxp-2k

The second warning could be removed without problems in my opinion. People will notice the issue with polar grid labels anyways.

yha:
> Maybe we could check if the two series share the same color gradient and omit the warning/info in that case.

The check does exist in the code, but stopped working because comparison of color gradients was broken in the transition to `ColorSchemes`. Should be fixed now on `PlotUtils` master (JuliaPlots/PlotUtils.jl#99).
I still think this should be a warning and not `@info`, because it means the colorbar is probably wrong. Once comparison of gradients works correctly, I can't think of a scenario where this warning is emitted and the plot is actually correct.

juliohm:
Closing as no longer valid.

PR #3299, opened 2021-02-13
Implement non-uniform polar heatmaps with the GR backend

### LINKED ISSUE #3149: [BUG] Nonuniform polar heatmaps in GR are displayed as if the axes were uniform
## Details

I have been trying to plot polar heatmaps with non-uniform axes with Plots.jl.
In my specific case, I have axes with differently high tick densities, so here a generic example:
```julia
using Plots
θ = π .+ 2*asin.(vcat(-(1:-0.001:0).^4,(0.001:0.001:1).^4))
r = sqrt.(0:1:100)
```
where the `θ` axis has a high tick density around `π`:
TickDensity

When plotting a heatmap, for example
```julia
heatmap(θ,r,r.*cos.(2θ'), projection = :polar)
```

- In `pyplot`, it gives:
NUpolar_PyPlot
which (except for the hole in the middle) is what I would expect.

- In `gr`, it gives:
NUpolar_GR
which does not agree with the expected outcome. I assume that the heatmap is displayed as if the axes were uniform, effectively stretching the points in regions with high density of axis ticks and compressing it in regions with low axis ticks.

This seems to be related with #1428.

### Backends

gr (default) | x  |     |
pyplot       |     | x  |
plotly       |     |     | x
plotlyjs     |     |     | x
pgfplotsx    |     |     | x
inspectdr    |     |     | x

### Versions

Plots.jl version: 
`Plots v1.9.0`

Backend  version: 
`GR v0.53.0`, `PyPlot v2.9.0`

Output of `versioninfo()`:
```
Julia Version 1.5.2
Commit 539f3ce943 (2020-09-23 23:17 UTC)
Platform Info:
  CPU: Intel(R) Core(TM) i7-7700 CPU @ 3.60GHz
  WORD_SIZE: 64
  LIBM: libopenlibm
  LLVM: libLLVM-9.0.1 (ORCJIT, skylake)
Environment:
  JULIA_DEPOT_PATH = /user/.julia:
  JULIA_NUM_THREADS = 4
  JULIA_PKGDIR = /user/.julia
  JULIA_PKG_DEVDIR = /user/.julia/dev
  JULIA_CXX_RTTI = 1
```

### LINKED ISSUE #3156: Problem with heatmap contour
Right now only the maximum value of y (r) is taken into account.

I am getting the above error when trying to use the heatmap polar. 

```julia
heatmap(delta, theta, reshape(tensile_strength_mpa_normal_fault_values, length(theta), length(delta)), proj =:polar, color=:coolwarm)
```

fhagemann:
Thanks to @danielkaiser (sciapp/gr@fcb2949ebde37db0d0833f16f8670cdfa9a85123) and @jheinen (jheinen/GR.jl@9b1dd41252bdf188539bebc5262765b024d73bd4) for implementing non-uniform polar cell arrays in the GR framework!

I tried to implement it in Plots.jl to fix #3149, but as a newbie would be very thankful for some feedback/reviews from the professionals here!

## Short summary of the changes in the code

- **Calling `GR.nonuniformpolarheatmap` for non-uniform polar heatmaps**
Polar heatmaps with GR will always be processed as `GR.nonuniformpolarheatmap`, just like non-uniform normal heatmaps are always processed as `GR.nonuniformheatmap`. The syntax between the two is very similar, just that GR.jl expects the angle array in units of degree rather than radians (hence the `rad2deg`).

- **Update warning to avoid displaying negative radii**
 Polar heatmaps will not be displayed if the first radius is smaller than 0. If the first value of the radius array is negative, one warning is implemented in Plots.jl (`@warn "'y[1] < 0' (rmin) is not yet supported."`) and another warning is given by GR itself (`Invalid radii specified`).

- **Update `heatmap_edges` to avoid negative radii**
In order to allow both, midpoints or edges, to be passed to the polar heatmap, the function `heatmap_edges` was expanded by an additional argument: `ispolar`. If the midpoints are passed and the first entry of the radius array is 0, `heatmap_edges` would return values smaller than 0, resulting in the heatmap not being displayed. `ispolar` will ensure that the edges are only expanded to smaller values than the minimum radius if the resulting value is still positive. This is probably not the most elegant way, so I would be very happy for some feedback!

### Initial issue (#3149)

To get back to the example of the initial issue
```julia
using Plots
θ = π .+ 2*asin.(vcat(-(1:-0.001:0).^4,(0.001:0.001:1).^4))
r = sqrt.(0:1:100)
heatmap(θ,r,r.*cos.(2θ'), projection = :polar)
```

used to be this before:

gr_old_nonuniform_polar_heatmap

and would now be this:

gr_new_nonuniform_polar_heatmap

### Possibility to plot non-2π polar heatmaps

A very nice side effect of this update is that it also allows plotting polar heatmaps that do not range over the whole 2π range.
These were previously stretched to the full 2π range, causing misleading plots.

Minimum example (similar to the one before but `θ` only ranging from `0` to `π` instead of `2π`) :
```julia
using Plots
θ = 0:π/100:π # notice that this is not full 2π
r = sqrt.(0:1:100)
heatmap(θ,r,r.*cos.(2θ'), projection = :polar)
```
used to give this:

gr_old_half_nonuniform_polar_heatmap

and now gives this:

gr_new_half_nonuniform_polar_heatmap

## Remarks

```julia
xmin, xmax, ymin, ymax = gr_xy_axislims(series[:subplot])
GR.setwindow(-ymax, ymax, -ymax, ymax)
```
I am not quite sure what these lines exactly do, as this PR also seems to add some white space around the polar heatmap which did not seem to be the case before:

Feel free to edit and push to this repository or comment!

Edit (daschw): fix #3156

### DIFF
--- a/src/backends/gr.jl
+++ b/src/backends/gr.jl
@@ -1578,10 +1578,8 @@ function gr_add_series(sp, series)
         GR.gr3.clear()
         dmin, dmax = GR.gr3.volume(y.v, 0)
     elseif st === :heatmap
-        if !ispolar(series)
-            # `z` is already transposed, so we need to reverse before passing its size.
-            x, y = heatmap_edges(x, xscale, y, yscale, reverse(size(z)))
-        end
+        # `z` is already transposed, so we need to reverse before passing its size.
+        x, y = heatmap_edges(x, xscale, y, yscale, reverse(size(z)))
         gr_draw_heatmap(series, x, y, z, clims)
     elseif st === :image
         gr_draw_image(series, x, y, z, clims)
@@ -1751,42 +1749,33 @@ end
 
 function gr_draw_heatmap(series, x, y, z, clims)
     fillgrad = _as_gradient(series[:fillcolor])
-    if !ispolar(series)
-        GR.setspace(clims..., 0, 90)
-        w, h = length(x) - 1, length(y) - 1
-        if is_uniformly_spaced(x) && is_uniformly_spaced(y)
-            # For uniformly spaced data use GR.drawimage, which can be
-            # much faster than GR.nonuniformcellarray, especially for
-            # pdf output, and also supports alpha values.
-            # Note that drawimage draws uniformly spaced data correctly
-            # even on log scales, where it is visually non-uniform.
-            colors = plot_color.(get(fillgrad, z, clims), series[:fillalpha])
-            rgba = gr_color.(colors)
-            GR.drawimage(first(x), last(x), last(y), first(y), w, h, rgba)
-        else
-            if something(series[:fillalpha], 1) < 1
-                @warn "GR: transparency not supported in non-uniform heatmaps. Alpha values ignored."
-            end
-            z_normalized = get_z_normalized.(z, clims...)
-            rgba = Int32[round(Int32, 1000 + _i * 255) for _i in z_normalized]
-            GR.nonuniformcellarray(x, y, w, h, rgba)
-        end
+    GR.setspace(clims..., 0, 90)
+    w, h = length(x) - 1, length(y) - 1
+    if !ispolar(series) && is_uniformly_spaced(x) && is_uniformly_spaced(y)
+        # For uniformly spaced data use GR.drawimage, which can be
+        # much faster than GR.nonuniformcellarray, especially for
+        # pdf output, and also supports alpha values.
+        # Note that drawimage draws uniformly spaced data correctly
+        # even on log scales, where it is visually non-uniform.
+        colors = plot_color.(get(fillgrad, z, clims), series[:fillalpha])
+        rgba = gr_color.(colors)
+        GR.drawimage(first(x), last(x), last(y), first(y), w, h, rgba)
     else
-        phimin, phimax = 0.0, 360.0 # nonuniform polar array is not yet supported in GR.jl
-        nx, ny = length(series[:x]), length(series[:y])
-        xmin, xmax, ymin, ymax = gr_xy_axislims(series[:subplot])
-        GR.setwindow(-ymax, ymax, -ymax, ymax)
-        if ymin > 0
-            @warn "'ymin[1] > 0' (rmin) is not yet supported."
-        end
-        if series[:y][end] != ny
-            @warn "Right now only the maximum value of y (r) is taken into account."
+        if something(series[:fillalpha], 1) < 1
+            @warn "GR: transparency not supported in non-uniform heatmaps. Alpha values ignored."
         end
         z_normalized = get_z_normalized.(z, clims...)
         rgba = Int32[round(Int32, 1000 + _i * 255) for _i in z_normalized]
-        GR.polarcellarray(0, 0, phimin, phimax, 0, ymax, nx, ny, rgba)
-        # Right now only the maximum value of y (r) is taken into account.
-        # This is certainly not perfect but nonuniform polar array is not yet supported in GR.jl
+        if !ispolar(series)
+            GR.nonuniformcellarray(x, y, w, h, rgba)
+        else
+            if y[1] < 0
+                @warn "'y[1] < 0' (rmin) is not yet supported."
+            end
+            xmin, xmax, ymin, ymax = gr_xy_axislims(series[:subplot])
+            GR.setwindow(-ymax, ymax, -ymax, ymax)
+            GR.nonuniformpolarcellarray(rad2deg.(x), y, w, h, rgba)
+        end
     end
 end
 
--- a/src/backends/gr.jl
+++ b/src/backends/gr.jl
@@ -1579,7 +1579,7 @@ function gr_add_series(sp, series)
         dmin, dmax = GR.gr3.volume(y.v, 0)
     elseif st === :heatmap
         # `z` is already transposed, so we need to reverse before passing its size.
-        x, y = heatmap_edges(x, xscale, y, yscale, reverse(size(z)))
+        x, y = heatmap_edges(x, xscale, y, yscale, reverse(size(z)), ispolar(series))
         gr_draw_heatmap(series, x, y, z, clims)
     elseif st === :image
         gr_draw_image(series, x, y, z, clims)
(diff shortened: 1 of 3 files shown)


fhagemann:
It seems like the tests are failing because GR polar heatmaps are compared to some kind of reference image (Julia 1 tests) or something unrelated (Julia nightly).

fhagemann:
This PR omits one warning addressed in #2820.

jheinen:
Thanks for the MR. I will check this tomorrow.

jheinen:
The problem should be fixed in GR#master.

```
using GR
θ = π .+ 2*asin.(vcat(-(1:-0.001:0).^4,(0.001:0.001:1).^4))
r = sqrt.(0:1:100)
nonuniformpolarheatmap(θ,r,r.*cos.(2θ'))
```
Screenshot 2021-02-16 at 15 35 19

... and for `θ = 0:π/1000:π` we get:

Screenshot 2021-02-16 at 15 38 44

fhagemann:
> The problem should be fixed in GR#master.
> 
> ```
> using GR
> θ = π .+ 2*asin.(vcat(-(1:-0.001:0).^4,(0.001:0.001:1).^4))
> r = sqrt.(0:1:100)
> nonuniformpolarheatmap(θ,r,r.*cos.(2θ'))
> ```

Thanks for your response!

Which exact problem are you referring to? 

Upgrading to GR#master and running your code example results in a warning (```Dimensions of color index array are invalid.```) and an empty plot:

Screenshot from 2021-02-16 16-23-10

jheinen:
Did you update the GR run-time? `ENV["GRDIR"]=""; ] build GR`

fhagemann:
No, I didn't. Getting the new GR runtime resolved this issue

```julia
julia > ENV["GRDIR"] = ""
julia > Pkg.build("GR")
```

Screenshot from 2021-02-16 16-38-44

Sorry for the false alarm!

oschulz:
Nice! @jheinen, do you plan new GR release soon? We have several people who'd love to use this, but not all in our group, so having a release will make it easier. :-)

oschulz:
@daschw , could you take a look a this PR? We kinda need this functionality to be able to use GR instead of PyPlot for our detector simulations (the devices are cylindrical, hence, polar plots ...).

daschw (approved):
Thanks, looks great! I will update the reference images, re-run the tests and merge afterwards.

fhagemann:
Any comments on the white space that are seen around the polar heatmap after my changes?

daschw:
Not yet, I'll have a look.

fhagemann:
Thanks!

I just checked that `xmin, xmax, ymin, ymax = gr_xy_axislims(series[:subplot])` returns `ymax = 11.077569095913589` which is higher than the maximum value of the `y` array (`10` in this case). 

One could set the window to `GR.setwindow(-maximum(y), maximum(y), -maximum(y), maximum(y))``instead, but that might be quite hard-coded..

fhagemann:
I think I found out where the axis limits come from:
`gr_xy_axislims`  in [src/backends/gr.jl#L310](https://github.com/JuliaPlots/Plots.jl/blob/d1969525719a30bc6b27ab8b0114ffd86fc9e09a/src/backends/gr.jl#L310) --> `gr_y_axislims` [src/backends/gr.jl#L308](https://github.com/JuliaPlots/Plots.jl/blob/d1969525719a30bc6b27ab8b0114ffd86fc9e09a/src/backends/gr.jl#L308) --> `axis_limits` [src/axes.jl#L469](https://github.com/JuliaPlots/Plots.jl/blob/f9ff2c8962d3650772f276d423b3981f918f2a11/src/axes.jl#L469)

The axis extrema ([src/axes.jl#L471](https://github.com/JuliaPlots/Plots.jl/blob/f9ff2c8962d3650772f276d423b3981f918f2a11/src/axes.jl#L471)) are `Plots.Extrema(-0.5, 10.0250628144669)`.
For polar plots ([src/axes.jl#L490](https://github.com/JuliaPlots/Plots.jl/blob/f9ff2c8962d3650772f276d423b3981f918f2a11/src/axes.jl#L490)) and `:auto` scaling ([src/axes.jl#L493](https://github.com/JuliaPlots/Plots.jl/blob/f9ff2c8962d3650772f276d423b3981f918f2a11/src/axes.jl#L493)), the ticks are widened by 10% of the overall axis range, which results in `ymax = 10.0250628144669 + 0.1 * 10.5250628144669 = 11.077569095913589` ([src/axes.jl#L495](https://github.com/JuliaPlots/Plots.jl/blob/f9ff2c8962d3650772f276d423b3981f918f2a11/src/axes.jl#L495)).

If these axis limits are intended, I would just leave it as is. 
Otherwise, I will change it to `GR.setwindow(-maximum(y), maximum(y), -maximum(y), maximum(y))`.

daschw:
I think for polar heatmaps your suggestion makes sense. I don't see a point in widened axes here.

fhagemann:
It looks like the axes are widened also in all the other backends, such that I would for now not change anything in the GR-specific code to leave it consistent. 

Speaking of which: using the `ispolar` argument added in `heatmap_edges` could also be used in the other backends, e.g. PyPlot ([src/backends/pyplot.jl#L651](https://github.com/JuliaPlots/Plots.jl/blob/master/src/backends/pyplot.jl#L651)), to ensure that the edges are non-negative. It seems like this would avoid the white circle in the centre of the PyPlot heatmap which I addressed in my initial issue #3149.


If the line 651 is `x, y = heatmap_edges(x, sp[:xaxis][:scale], y, sp[:yaxis][:scale], size(z))`, my example gives

old_pyplot_heatmap



`x, y = heatmap_edges(x, sp[:xaxis][:scale], y, sp[:yaxis][:scale], size(z), ispolar(sp))` would result in

new_pyplot_heatmap

daschw:
OK, then let's leave that for now. Actually, I'm not sure why the axes are widened for polar `heatmap`s in the first place, because `heatmap` is not in the series types that we automatically widen for. So I guess we should rather resolve this in a backend-independent way in a separate PR.

fhagemann:
Sounds like a good plan!
I guess we can then also close #3156 when merging.

oschulz:
Thanks a lot @jheinen for this feature in GR, @fhagemann for adding it to plots, and @daschw for the quick merge and new Plots release!

Issue #3297, opened 2021-02-12
[BUG] ymirror=true does not work with the pyplot backend


churli:
## Details
Using the pyplot backend I cannot manage to have the y-axis on the right.

MWE:
```julia
using Plots; pyplot()
plot(rand(100), ymirror=true)
```
This just outputs a plot with the y-axis on the left, as if `ymirror=false`.  
The problem also affects plots with 2 different y-axes via `twinx()`: in this case both axes end up superimposed on the left and I could not find a way to get a nice plot with 2 different y-axes.

### Backends

This bug occurs on ( insert `x` below )

Backend      | yes | no  | untested
-------------|-----|-----|---------
gr (default) |     |  x   |
pyplot       |  x   |     |
plotly       |     |     |  x
plotlyjs     |     |     |  x
pgfplotsx    |     |     |  x
inspectdr    |     |     |  x

### Versions

Plots.jl version: `v1.10.4` (latest)
Backend  version (`]st -m`): `v2.9.0`(latest)
Output of `versioninfo()`:
> Julia Version 1.5.3
> Commit 788b2c77c1 (2020-11-09 13:37 UTC)
> Platform Info:
>  OS: macOS (x86_64-apple-darwin18.7.0)
>  CPU: Intel(R) Core(TM) i7-8557U CPU @ 1.70GHz
>  WORD_SIZE: 64
>  LIBM: libopenlibm
>  LLVM: libLLVM-9.0.1 (ORCJIT, skylake)

isentropic:
I will fix this soon, by mistake it was taken out recently

churli:
Thanks for your quick reply @isentropic !
Could you please point me to the right file/function so that I can work around it while waiting for the fix?

Thanks!

isentropic:
The changes introduced here
https://github.com/JuliaPlots/Plots.jl/commit/62af64f08abd89ef90fdb5921839245617419a41

The plan was to separate xymirror somewhere else, as it only works for 2d plots in pyplot

isentropic:
@chfin please check out the PR. Before it gets merged, would you please perform some graphical tests and attach your findings in the PR. 
Test some of the commands and attach screenshots of behavior before and after the PR, please?

Attributes in mind:
```
(x)ymirror
framestyle=(:box, :axes, :semi) # I guess of other framestyles mirroring is ill defined, so don't bother checking those
twinx()
```

isentropic:
[...]

Issue #2956, opened 2020-09-05
[BUG] twinx() broken for pgfplots and plotly backends


Bebotron:
### Backends

This bug occurs on ( insert `x` below )

Backend      | yes | no  | untested
-------------|-----|-----|---------
gr (default) |     |  x   |
pyplot       |     |  x   |
plotly       |  x   |     |
plotlyjs     |  x   |     |
pgfplotsx    |  x  |     |
inspectdr    |     |     | x

## Details
I found this issue with `pgfplots()`, but I tried to reproduce it with the other backends and had another hiccup, though `pgfplots()` was the most egregious. Note I wasn't even able to make an initial plot with `insepctdr()`, which I don't really use so I didn't bother trying to figure out what's wrong there. The simple test code is:
```Julia
using Plots
x = [1,2,3]
y = [4,5,6]
plot(x,y1,ylim=(0,6))
plot!(twinx(),x,y2,ylim=(0,6))
```
this output the following for each backend. The behavior I expected from `pgfplots()` (and the other backends) to begin with is the one that `gr()` gives: yaxis ticks and labels on the right side. `plotly()` and `plotlyjs()` don't quite do this, while `pgfplots()` does something else entirely.

`gr()`

![gr](https://user-images.githubusercontent.com/24438203/92292724-e8405480-eedb-11ea-816f-6f0757814be8.png)

`pyplot()`

![pyplot](https://user-images.githubusercontent.com/24438203/92292710-d9f23880-eedb-11ea-8283-3420e9d9175c.png)

`plotly()`

![plotly](https://user-images.githubusercontent.com/24438203/92292898-e88d1f80-eedc-11ea-9427-5523ffe6238d.png)

`plotlyjs()`

![plotlyjs](https://user-images.githubusercontent.com/24438203/92292742-f9896100-eedb-11ea-9dca-e8de20e31ce6.png)

`pgfplots()`

![pgfplots](https://user-images.githubusercontent.com/24438203/92292975-415cb800-eedd-11ea-84cc-cc06284a2902.png)

With `inspectdr()`, as soon as I tried to plot anything, it would print some errors and kick me out of the REPL. That should be a separate issue, but for now I'm more concerned with `pgfplots()`.

[...]

Issue #2205, opened 2019-10-01
twinx @ pyplot: Right margin too small


fbruetting:
`Julia v1.1.1`
`Plots v0.26.3`

When I add `twinx` on `pyplot`, and give the second y-axis a name, I see this:
![Bildschirmfoto von 2019-10-01 19-30-55](https://user-images.githubusercontent.com/10520767/65985583-67ced380-e482-11e9-9be3-38d040a4e42f.png)

I have to give the right margin an extra 10mm for the graph to look usable. Even better would be an extra 2mm for all sides, so that there is a little space between the contents and the border. Additionally, the spacing between the second y axis and it’s label could use a bit more space.
```
left_margin		=  2mm,
right_margin	= 12mm,
top_margin		=  2mm,
bottom_margin	=  2mm,
```
![Bildschirmfoto von 2019-10-01 19-30-35](https://user-images.githubusercontent.com/10520767/65985702-9cdb2600-e482-11e9-9131-d5a4dfdf9ddb.png)

ssfrr:
This doesn't seem to be specific to pyplots - I see the same thing using GR

t-bltg:
Closed by https://github.com/JuliaPlots/Plots.jl/pull/4488.

PR #3300, opened 2021-02-15
twin, mirror fixes

isentropic:
Pyplot twin and mirror attribute fixes
Related issues: https://github.com/JuliaPlots/Plots.jl/issues/2205, pyplot plots not consistent with [GR](https://github.com/JuliaPlots/Plots.jl/issues/2956),  https://github.com/JuliaPlots/Plots.jl/issues/3297


isentropic:
Worth mentioning that pyplot does not support 3d plot mirroring or stuff, so this is 2d plot exclusive. matplotlib is primarily 2d plotting suite, so I doubt that it ever will

daschw (approved):
Nice!

Issue #3000, opened 2020-09-21
[BUG] Plots.jl is causing Documenter.jl build to hang


juliohm:
In the last 3 weeks some update in some indirect dependency of GeoStats.jl bumped the version of Plots.jl in the documentation build to 1.x and out of the sudden my documentation cannot be built.

```julia
] dev GeoStats
; cd .julia/dev/GeoStats/docs
] activate .
] instantiate
include("make.jl")
```

The issue seems to be related to SVG output, but I am not sure. Can you please help address this blocking issue? I was able to build the docs by removing all markdown pages that contained plot commands.The build was working just fine 4 weeks ago.

BeastyBlacksmith:
As a quick remedy you can pin the Plots version to the one that worked

juliohm:
I will end up doing this. Meanwhile, it seems that after hours waiting the documentation is being built. The update to Plots.jl apparently made the documentation build extremely slow, but I am waiting a few more hours here before I can conclude that.

juliohm:
@BeastyBlacksmith could you please help with this issue? I am still in a setup where the documentation with Plots.jl examples takes hours to finish. Pinning to an old version of Plots.jl is a suboptimal solution.

juliohm:
Anything that changed recently regarding the option `marker_z` in the GR backend that may have affected performance considerably with large point clouds? I think the plots I am having issue with are the ones where I try to do a colored scatter, but I am not sure. Still investigating this.

juliohm:
Ok, here is a MWE with the exact issue with `marker_z` incredibly slow: https://gist.github.com/juliohm/de0ceeb47b9076b626d9a0a793c658d6

daschw:
I'm afraid I introduced this performance regression in https://github.com/JuliaPlots/Plots.jl/pull/2940. The good news is that https://github.com/JuliaPlots/Plots.jl/pull/3320 provides a fix.

juliohm:
Looking forward to seeing the PR merged.

cc: @jheinen

jheinen:
I'm not sure whether this will fix the performance regression.  The `get_*` function in `gr_draw_marker(...)` cause too much load within the GR run-time.
```
    gr_set_bordercolor(get_markerstrokecolor(series, i));
    gr_set_markercolor(get_markercolor(series, clims, i));
    gr_set_transparency(get_markeralpha(series, i))
```
The problem is, that the colors are **not** indices into a pre-allocated color map.

[...]

Issue #2959, opened 2020-09-07
[FR] NaN-separated series with changing colors


ignacioq:
I would like to have a NaN separated series for paths, in which color changes for each part of the path.
For instance x, y, z are vectors, with some NaN separating the different lines of the tree; say `x = [0.1, 0.2, 0.3, NaN, 0.1, 0.2]`, `y = [0.5, 0.4, 0.5, NaN, 0.5, 0.6]` and `z = [0.2, 0.8, 0.6, NaN, 0.1, 0.9]`. I would like each of the series to have the corresponding path colored according to `z`. This would work if there are no `NaN` separating the paths, but it does not work when NaN's are included. I raised the issue on slack (https://julialang.slack.com/archives/C8JKM4JQ6/p1599257009014800) and I got an answer to use the `@series` macro, but it becomes prohibitively slow. 
It is fast If I do not put NaN's (though this is not what I want, the diagonal lines should not appear). Note that the color changes along the branches of the tree: 
![t1](https://user-images.githubusercontent.com/5589248/92416816-eae1ba80-f124-11ea-8c50-8f2b52ef8ea9.png)
or if I don't try to color the segments:
![t0](https://user-images.githubusercontent.com/5589248/92416814-e74e3380-f124-11ea-9ef8-73ca38acc816.png)
Any help of making this possible would be very appreciated! Thanks.
My current recipe is (note the commented section which does what I want but takes very long using the `@series` macro):
```
@recipe function f(tree::iTgbm, zfun::Function)

  x = Float64[]
  y = Float64[]
  z = Float64[]
  rplottree(tree, treeheight(tree), 1:sntn(tree), zfun, x, y, z)
  pop!(x)
  pop!(y)
  pop!(z)

  # plot defaults
  line_z          --> z
  linecolor       --> :inferno
  legend          --> :none
  colorbar        --> true
  xguide          --> "time"
  fontfamily      --> font(2, "Helvetica")
  xlims           --> (0, treeheight(tree))
  ylims           --> (0, sntn(tree)+1)
  xflip           --> true
  xtickfont       --> font(8, "Helvetica")
  grid            --> :off
  xtick_direction --> :out
  yshowaxis       --> false

  # nan_inds = findall(x -> isnan(x),z)
  # for i in eachindex(nan_inds)
  #     start = 1 + (i == 1 ? 0 : nan_inds[i - 1])
  #     stop = nan_inds[i] - 1
  #     @series begin
[...]

Issue #3117, opened 2020-11-03
[BUG] Rendering of scatterplot with a vector of colors is really slow after some update


VPetukhov:
After some of the updates, rendering `scatter` with individual color per point got really slow. When I run normal scatter for 20000 points it takes less than a second. But when I add vector with colors it takes 25 seconds.

## Details

The code to reproduce:

```julia
import Plots
n_cols = 20000;
Plots.scatter(1:n_cols, rand(n_cols), format=:png)
```

and 

```julia
Plots.scatter(1:n_cols, rand(n_cols), color=["red" for  i in 1:n_cols], format=:png)
```

It already was on v1.6.10, then I updated to v1.7.3, and it's still there. I don't remember what version I had before (most probably something above 1.3), but there I was able to plot hundreds of thousands of points.

I didn't test it previously on PyPlot backend, but now there is still huge drop in performance after adding `color`. It works almost immediately on 5k points without colors and around 20 seconds with colors.

**UPDATE:** just checked that v1.6.0 works fine on both GR and PyPlot, while 1.6.1 is broken.

### Backends

This bug occurs on ( insert `x` below )

Backend      | yes | no  | untested
-------------|-----|-----|---------
gr (default) |  x  |     |
pyplot       |  x  |     |
plotly       |     |     | x
plotlyjs     |     |     | x
pgfplotsx    |     |     | x
inspectdr    |     |     | x

### Versions

Plots.jl version: v1.6.10 or v1.7.3
Backend  version (`]st -m`): GR v0.52.0, PyPlot v2.9.0
Output of `versioninfo()`:
```
Julia Version 1.5.2
Commit 539f3ce943 (2020-09-23 23:17 UTC)
Platform Info:
  OS: Linux (x86_64-pc-linux-gnu)
  CPU: Intel(R) Core(TM) i7-8850H CPU @ 2.60GHz
  WORD_SIZE: 64
  LIBM: libopenlibm
  LLVM: libLLVM-9.0.1 (ORCJIT, skylake)
```

fedario:
Can confirm, same for me.

Can I do something to help to resolve this?

mkborregaard:
Does the same happen if you wrap the color vector in `cgrad`?

fedario:
> Does the same happen if you wrap the color vector in `cgrad`?

[...]

PR #3320, opened 2021-02-25
Fix for "segmented" attributes with NaNs

### LINKED ISSUE #3258: [BUG] Scatter plots silently plot wrong data when NaNs are present
## Details

When you plot an XY scatter plot with additional dimensions of data in markers or colors, any NaNs in the XY data will shift the markers and colors and cause Plots to display incorrect data.

I hope this gets immediate attention because I think this is a super serious problem. I very nearly submitted a paper with completely messed up results because of this.

Demo (correct): note that the third circle is large and green.

```
julia> scatter([1,2,3], [3,2,1], markersize=[30,10,30], c=[1,2,3], xlims=(0,4), ylims=(0,4), legend=false)
```
newplot (2)

Now change a coordinate for circle 2 to NaN and note what happens with circle 3.

```
julia> scatter([1,2,3], [3,NaN,1], markersize=[30,10,30], c=[1,2,3], xlims=(0,4), ylims=(0,4), legend=false)
```
newplot (3)

### Backends

gr (default) |   X  |     |
pyplot       |     |     | X
plotly       |   X  |     |
plotlyjs     |     |     | X
pgfplotsx    |     |     | X
inspectdr    |     |     | X

### Versions

Plots.jl version:  1.10.2
Backend  version (`]st -m`):
Output of `versioninfo()`:
```
julia> versioninfo()
Julia Version 1.6.0-beta1.0
Commit b84990e1ac (2021-01-08 12:42 UTC)
Platform Info:
  CPU: Intel(R) Core(TM) i5-5200U CPU @ 2.20GHz
  WORD_SIZE: 64
  LIBM: libopenlibm
  LLVM: libLLVM-11.0.0 (ORCJIT, broadwell)
Environment:
  JULIA_EDITOR = "C:\Program Files\Microsoft VS Code\Code.exe"
  JULIA_NUM_THREADS = 4
  JULIA_PKG_DEVDIR = C:/Stuff
```

yha:
Fixes indexing into attributes that vary by series or data point, so that it is not affected by `NaN` values.
Some examples that previously produced either a `BoundsError` or silently wrong output:
```
plot(plot([2 3; 3 4; 1 NaN; 2 1; 4 2], c=:plasma, fill_z=[1 3], fillrange=0, fillalpha=0.5),
   plot([2 3; 3 4; 1 NaN; 2 1; 4 2], c=:plasma, fill_z=1:5, fillrange=0, fillalpha=0.5),
   scatter([1 2; NaN 3; 4 5], marker_z = [1 2]),
   scatter([1 2; NaN 3; 4 5], marker_z =[1 1; 2 2; 3 3]),
   plot([1 2; 2 3; 4 5], line_z=[1 1; 3 4; 5 6]),
   plot([[1,2,NaN,4,5] 2:6], line_z=[1 2]), 
   legend=false)
```
image

Fixes #3258 (maybe also #2959?).
Breaks the current [reference image 48](http://docs.juliaplots.org/latest/generated/gr/#gr-ref48), but I don't think the current behavior is documented other than in this reference image. As I mentioned in https://github.com/JuliaPlots/Plots.jl/issues/3258#issuecomment-785390216, I think this fix is important enough to break it.


daschw:
In general, I'd also find this more consistent. However, I'm not sure if this "just" is a bugfix or breaking.
Consider e.g.
```julia
x = [1, 3, 2, NaN, 4, 6, 5, NaN, 0, 1, 1, 0]
y = [2, 2, 3, NaN, 0, 1, 0, NaN, 0, 0, 1, 1]

plot(x, y, st=:shape, fillcolor=[:red, :blue, :green])
plot(x, y, st=:shape, fill_z=1:3)
```
This PR changes
fillcolor_old

to
fillcolor_new

and
fill_z_old

to
fill_z_new

same results for
```julia
xs = [[1, 3, 2], [4, 6, 5], [0, 1, 1, 0]]
ys = [[2, 2, 3], [0, 1, 0], [0, 0, 1, 1]]
shapes = Shape.(xs, ys)

plot(Shape.(xs, ys), fillcolor=[:red, :blue, :green])
plot(Shape.(xs, ys), fill_z=1:3)
```
I wonder if we can keep the old behavior somehow with the changes in this PR (and still be consistent). I can be quite handy. I used it sometimes for plotting shapefile maps and coloring countries according to some values.

mkborregaard:
I agree @daschw , this is a (not necessarily desirable) behaviour change rather than a fix.

yha:
@daschw, I would say that the "before" plots in your first example are definitely a bug. I think that it's extremely surprising that the way indexes in attributes correspond to indexes in data depends in any way on the contents rather than types and shapes. Real data is eventually going to contain `NaN`s which are "real", and not there intentionally to separate segments.
(incidentally, this is more insidious than the cycling behavior of attributes because a) it can crop up after shapes have been debugged, and b) output are less likely to seem obviously wrong: cycled attributes create repetitive patterns in the plots that usually stand out)

The second example is more compelling, and not something I've tested. It's not consistent with the behavior of a plain `plot(xs, ys,...)` (with or without `st=:shape`), but also doesn't seem worth changing, so I'll fix the PR to preserve the behavior for this case.

To deal with the breakage, perhaps we can add a "soft deprecation", i.e. a warning when sizes of "segmenting" attributes don't match, and an extra warning when `NaN`s are present?

yha:
I fixed the `AVec{Shape}` recipe to give the same output as before.
I also added a warning on attribute length mismatch + a hint when the data contains `NaN`s (which is where the output changes with this PR).
```
x = [1, 3, 2, NaN, 4, 6, 5, NaN, 0, 1, 1, 0]
y = [2, 2, 3, NaN, 0, 1, 0, NaN, 0, 0, 1, 1]

plot(x, y, st=:shape, fillcolor=[:red, :blue, :green])
```
gives
```
┌ Warning: Indices Base.OneTo(3) of attribute `fillcolor` does not match data indices 1:12.
└ @ Plots c:\Users\sternlab\.julia\dev\Plots\src\utils.jl:106
┌ Info: Data contains NaNs or missing values, and indices of `fillcolor` vector do not match data indices.
│ If you intend elements of `fillcolor` to apply to individual NaN-separated segements in the data,
│ pass each segment in a separate vector instead, and use a row vector for `fillcolor`. Legend entries 
│ may be suppressed by passing an empty label.
│ For example,
└     plot([1:2,1:3], [[4,5],[3,4,5]], label=["y" ""], fillcolor=[1 2])
```
An alternative would be to merge these message and show them only when the data has `NaN`s. 
My rationale for choosing to display the warning on any dimension mismatch (even when data does not have `NaN`s) was that the current behavior in this case is very messy anyway: e.g. on GR `plot(1:3,1:5)` and `scatter(1:5, marker_z=1:2)` error, `scatter(1:3, 1:5)` truncates `y`, and `scatter(1:5, c=["blue","red"])` cycles.

mkborregaard:
Hmm, I'm being swayed by the good arguments here, sorry for making noise.

daschw (approved):
Nice, I'm for these changes. Note that this also fixes #3000 and #3117

PR #3324, opened 2021-02-25
[WIP] position legend at angle

gustaphe:
WIP because
* Is only on GR
* No docs
* No tests

but otherwise works about how I want it to.

Adds the option to specify legend position using a single angle:
```julia
plot(randn(5,5), legend=30)
```
will place the legend at 30˚ from the horizon and the center of the plot, as close to the axis edges as possible.

```julia
plot(randn(5,5), legend=(-15, :outer))
```
will place the legend at -15˚, outside the axes.

Is this functionality you guys think we want? It feels pretty good once you have it.


gustaphe:
A demonstration:

```julia
anim = @animate for t = 0:5:719
    plot(A,legend=(t,t>=360 ? :inner : :outer))
end
```

https://imgur.com/a/edYwVBP

isentropic:
While I personally appreciate the effort, I find the need for this feature very hard to justify. +100lines of code that need maintenance in the current GR codebase given the utility this PR provides does not seem good.
Though I'm too opinionated here, perhaps others would have a different opinion

gustaphe:
That's fair. To me, it's a fairly common occurrence that I wish I could just move the legend ever so slightly out of the way, and this would be a pretty low-user-effort way to achieve that. 

Initially I thought about making `:right` go to `0` etc, which would offset the increase by a bit, but of course `:topright` and the likes don't have simple angle equivalences.

You do as you want, it's not me who'll be maintaining it, so consider this a working feature request.

gustaphe:
6193519 is a simple bug fix, btw. I can make that a separate PR if this one is rejected, I just forgot that was in here.

daschw:
Hard to say, I like it, but I agree with @isentropic that it's a lot of added LOC for GR and I won't really find the time to implement it in other backends. Though I guess some of the logic could be extracted to a backend-agnostic function that could be reused in other backends.

gustaphe:
So, I factored out some common factors, and introduced this in two more backends. The diff is still ~100LOC, so there's clearly a lot of synergy in here.

It obviously depends on what the other backends are like, I haven't worked with any of them before, but I'll be happy to keep going for fun.

gustaphe:
That's all of the backends with legends and legend positioning, (`gr`,`pgfplotsx`,`plotly*`,`pyplot`) net + 136 loc. 

In pyplot, the legend position is given in normal coordinates, so the angle is squashed, but I don't think that's a big problem. It also means I could factor out all of the legend keywords to a backend-agnostic angle function (i.e. `:topright => (45,:inner)`). I was expecting more backends to follow this pattern, but unfortunately only pyplot does (so far).

daschw (approved):
Awesome

PR #3332, opened 2021-03-03
Enable `dpi`  option for `pgfplotsx()` PNGs.

korsbo:
Enable `dpi` specification for PGFPlotsX png file generation. 

This bypasses the `Base.show` that PGFPlotsX implements in favour of just creating a file using `pgfsave` and reading that file into IO. It seems to work fine for me and the tests pass but it would be good for someone who knows the plumbing of Plots.jl to weigh in on whether this is fine.


daschw:
Nice, thanks!

Issue #3174, opened 2020-12-01
[FR] Colorbar properties


rafaqz:
It would be useful to be able to style additional attributes of colorbars, as discussed recently on slack in plotting

~- [ ] tick size~
~- [ ] tick color~
- [x] ticks
- [x] tick scale
- [x] tick font
- [x] tick font size
- [x] tick font color
- [ ] border color
- [ ] border thickness
- [ ] bar width
- [ ] bar height
- [ ] other things?

My use-case is cleaner map-plotting recipes in GeoData.jl. Currently, if you stray too far from the defaults the colorbar starts to look out of place.

This could start with just GR?

related, but just for ticks: https://github.com/JuliaPlots/Plots.jl/issues/2308

mkborregaard:
It's tricky because Plots development aims to keep keywords few(-ish), in order to keep maintainability. In general all new keywords should preferably be applied across all backends to keep interchangeability of backends. One possibility could be to adapt `extra_kwargs` to be able to pass colorbar kwargs, but this wouldn't work in recipes. I guess this requires a bit of discussion.

@rafaqz do you happen to know how many of these attributes are supported natively by e.g. GR?

hervasa2:
I think the colorbar should have the same attributes as an axis. For example just like you have xticks and xscale, the colorbar should be able to support cticks and cscale. I came across this limitation while attempting cscale = :log10 (also tried zscale = :log10) on a heatmap.

rafaqz:
I can see how keeping the keyword scope down would be important! especially given the breadth of use cases plots has to deal with. 

If we can use the same set of keywords (maybe with `colorbar_` prefixed?) it would at least add less overhead in terms of new names for things.

I'm not sure what is possible with GR, the `colorbar` method doesn't have many arguments, but there is lot of global state that is incorporated somehow. It does seem possible to do with matplotlib, plotly, and pgfplots.

mkborregaard:
So having a new keyword type `colorbar` in addition to `Plot`, `Subplot`, `Axis` and `Series`, and reusing `Axis` keywords while prepending `colorbar_`? That does sound like a neat design. Thoughts @daschw ?

rafaqz:
Also I didn't read @hervas2 comment properly, `cticks` etc is nice too for where x/y is usually used.

daschw:
[...]

PR #3346, opened 2021-03-11
Colorbar enhancements

isentropic:
Minimal working colorbar enhancements, pyplot only. Response to https://github.com/JuliaPlots/Plots.jl/issues/3174
Colorbars  need some :heart: too )

```julia
heatmap(rand(10,10), colorbar_ticks=[0.3,0.6])
```
image


src/args.jl
```diff
@@ -151,54 +151,54 @@ const _shape_keys = Symbol[
 
 const _allMarkers = vcat(:none, :auto, _shape_keys) #sort(collect(keys(_shapes))))
 const _markerAliases = Dict{Symbol,Symbol}(
```
isentropic, line 153: Aligning the dict properly

src/args.jl
```diff
@@ -363,104 +363,115 @@ const _plot_defaults = KW(
...
+    :projection                 => :none,             # can also be :polar or :3d
+    :aspect_ratio               => :auto,             # choose from :none or :equal
+    :margin                     => 1mm,
+    :left_margin                => :match,
+    :top_margin                 => :match,
+    :right_margin               => :match,
+    :bottom_margin              => :match,
+    :subplot_index              => -1,
+    :colorbar_title             => "",
+    :colorbar_titlefontsize     => 10,
```
isentropic, line 419: New colorbar args

isentropic:
Most importantly I need some comments about the Plots backend restructure and creation of the colobars.jl file.
currently only colorbar_ticks implemented, but the rest of the arguments should be easy to implement in the backend code only. Just want to make sure that the Plots pipeline altering is done fine here.
This does not break any previous behavior, like usual it would be up to the backend to support the these colorbar arguments. `colorbar_ticks=:native`should default back the original behavior for all backends.

src/colorbars.jl
```diff
@@ -0,0 +1,209 @@
...
+        cbar_gradient
+    else
+        nothing
+    end
+end
+
+hascolorbar(series::Series) = colorbar_style(series) !== nothing
+hascolorbar(sp::Subplot) = sp[:colorbar] != :none && any(hascolorbar(s) for s in series_list(sp))
+
+function optimal_colorbar_ticks_and_labels(sp::Subplot, ticks = nothing)
```
daschw, line 74: Do we really need to duplicate this code from axes.jl?

isentropic: the reason for duplication is that ive decided that colorbar is a subplot arg, hence it does not have an axis (x, y, or z)

isentropic: axis is not defined, hence the colorbar args need different treatment

isentropic: perhaps it could just overload the method, but is it better?

daschw: maybe we should have the main logic of `optimal_ticks_and_labels` in axes.jl in a method
```julia
optimal_ticks_and_labels(amin, amax, scale, formatter, ticks)
```
Then we can define the existing version as
```julia
optimal_ticks_and_labels(sp::Subplot, axis::Axis, ticks = nothing) =
    optimal_ticks_and_labels(axis_limits(sp, axis[:letter])..., axis[:scale], axis[:formatter], ticks)
```
and use the first version also for colorbars. I think it's easier to maintain if we do not have duplicate code.

isentropic: yes this sounds much better

src/colorbars.jl
```diff
@@ -0,0 +1,209 @@
...
+    #
+    # If the axis input was a Date or DateTime use a special logic to find
+    # "round" Date(Time)s as ticks
+    # This bypasses the rest of optimal_ticks_and_labels, because
+    # optimize_datetime_ticks returns ticks AND labels: the label format (Date
+    # or DateTime) is chosen based on the time span between amin and amax
+    # rather than on the input format
+    # TODO: maybe: non-trivial scale (:ln, :log2, :log10) for date/datetime
+    if ticks === nothing && scale == :identity
+        if sp[:colorbar_formatter] == RecipesPipeline.dateformatter
```
isentropic, line 91: what about here? sp[:colorbar formatter] is not part of axis

src/colorbars.jl
```diff
@@ -0,0 +1,209 @@
...
+    end
+
+    # @show unscaled_ticks labels
+    # labels = Showoff.showoff(unscaled_ticks, scale == :log10 ? :scientific : :auto)
+    unscaled_ticks, labels
+end
+
+# return (continuous_values, discrete_values) for the ticks on this axis
+function get_colorbar_ticks(sp::Subplot; update = true)
+    if update || !haskey(sp.attr, :colorbar_optimized_ticks)
```
isentropic, line 159: Also this case

src/colorbars.jl
daschw: I made a PR for this on the colorbar_enhancements branch of your fork.

isentropic (commented):
Done

src/args.jl
```diff
@@ -363,104 +363,115 @@ const _plot_defaults = KW(
...
+    :colorbar_title_location    => :center,           # also :left or :right
+    :colorbar_fontfamily        => :match,
```
isentropic, line 421: Where do I do the matching the colorbar_fontfamily, colorbartitlefont and colorbar_titlefontcolor to the rest of the fonts?

isentropic (commented):
I cannot find the colorbar font matching (setting sp[:colorbarfontcolor] =:black for example?). Currently all of the items have :match but I could not find where the calculation is made. 

Other than that, now the colorbarscale works too, with the PyPlot backend

isentropic:
I checked recipespipeline, but cannot find it there too

src/args.jl
daschw: I think you are looking for https://github.com/JuliaPlots/Plots.jl/blob/f76390c20d188a35f517a8cf467c061325d1c22e/src/args.jl#L1342-L1344

isentropic:
Nvrmind, i got it figured out, the matching happens in the `args.jl`. I'd say it is working pretty good now

```julia
plot = histogram2d(rand(1000,1000), colorbar_tickfontfamily="Serif", colorbar_tickfontcolor="Red", colorbar_scale=:log10)
```
image

isentropic:
@daschw could you perhaps follow along with gr implementation for this? 

@BeastyBlacksmith pgfplots could also benefit too. 

I believe that most of the pipeline processing has been accomplished, there should not be a lot backend code to be implemented. If anything, it should make construction of the colorbar easier as relevant kws are introduced.

We also need to think about additional kws like: colorbar width, height, tick direction, extend and what not

isentropic:
A lot of ideas were mentioned in #3174, but I believe we can't blindly accept all requests and should decide on the reasonable scope to cover. I personally think that we should cover as many as the backends can easily handle. 

Perhaps even run a poll for the needed args and decide on the scope.

mkborregaard:
> I personally think that we should cover as many as the backends can easily handle.

Sounds like a good pragmatic approach to me

PR #3351, opened 2021-03-14
Move `@df` macro to Plots

### LINKED ISSUE #3336: Move df macro here?
I think this has been discussed several times on StatsPlots, but let's move it here. My suggestion is to simply move the `@df` macro from StatsPlots to here. Having data in dataframes is a common use case for plotting, and the placement on the macro in StatsPlots is a legacy from julia 0.5 and before, where DataFrames was a massive dependency. With the Tables interface this issue is resolved. @daschw @piever

daschw:
fix #3336

### DIFF
--- a/src/df.jl
+++ b/src/df.jl
@@ -1,0 +1,208 @@
+"""
+    `@df d x`
+
+Convert every symbol in the expression `x` with the respective column in `d` if it exists.
+
+If you want to avoid replacing the symbol, escape it with `^`.
+
+`NA` values are replaced with `NaN` for columns of `Float64` and `""` or `Symbol()`
+for strings and symbols respectively.
+
+`x` can be either a plot command or a block of plot commands.
+"""
+macro df(d, x)
+    esc(Expr(:call, df_helper(x), d))
+end
+
+"""
+    `@df x`
+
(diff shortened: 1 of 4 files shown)


src/df.jl
```diff
@@ -0,0 +1,208 @@
+"""
```
isentropic, line 1: Perhaps this code should instead go to recipes.jl?

getzze:
Is there anything in the code that is specific to plotting? Otherwise I would say this should be defined in Tables.jl, no?
I always have a hard time remembering which package imports @df, as there is nothing in the name or the functionality that makes me think of Plots (or StatsPlots).

Ickaser:
It's been a while since any discussion happened on this; to summarize the discussion at https://github.com/JuliaData/Tables.jl/issues/287, it looks like they already have a macro `@with` that is "better" but relies on some DataFrames functionality so that it is not obvious how to implement for general tables, as works for the `@df` macro. That being the case, I as a user would be in favor of moving `@df` to Plots, since that is the only thing that I use from StatsPlots. I have been using it for `TypedTables`, rather than DataFrames, so if it were to be renamed to `@table` (or something similar) I would be on board.

PR #3352, opened 2021-03-14
new release

daschw:
Since we agreed that https://github.com/JuliaPlots/Plots.jl/pull/3320 was a bugfix I would only use a bugfix release here.
Any objections @yha @mkborregaard?


yha:
Maybe it's better to be more conservative and use a minor release? The previous behavior was documented in one example and has been around for a few months.
What's the downside to having a minor release instead?

PR #3374, opened 2021-03-25
Conform to InspectDR's new layout API in v0.4.0.

ma-laforge:
Was not practical to write out coordinates in pixels.
Changed API to accept coordinates relative to w & h (known as "percent" in Plots.jl).


daschw:
Nice, thanks

PR #3377, opened 2021-03-25
guide label alignment

mmikhasenko:
any preference for the function names?
any tests?


daschw:
Thanks for this! The naming is fine. I'll do a follow-up PR to fix `guideposition` and add a new test example there that also covers guide font alignment.

PR #3382, opened 2021-03-25
Add support for `seriestype=:stepmid`

david-macmahon:
This PR adds support for `seriestype=:stepmid`.  Like `:steppre` and `steppost`, the path of a `:stepmid` series is composed of vertical and horizontal segments that pass through the data points.  Unlike the other two step plots, the X coordinates of the vertical segments of a `:steamid` series are positioned half way between the X coordinates of consecutive data points and the data points themselves are on the midpoints of the horizontal segments (except for the first and last point which are endpoints of a horizontal segment.

Here is an example comparing the three step types:

```julia
x=1:5
y = [1, 2, 3, 2, 1]
default(shape=:circle)
plot(
    plot(x, y, seriestype=:steppre, label="steppre"),
    plot(x, y, seriestype=:stepmid, label="stepmid"),
    plot(x, y, seriestype=:steppost, label="steppost"),
    layout=(3,1)
)
```

image

### DIFF
--- a/src/args.jl
+++ b/src/args.jl
@@ -54,7 +54,7 @@ const _3dTypes = [
     :path3d, :scatter3d, :surface, :wireframe, :contour3d, :volume, :mesh3d
 ]
 const _allTypes = vcat([
-    :none, :line, :path, :steppre, :steppost, :sticks, :scatter,
+    :none, :line, :path, :steppre, :stepmid, :steppost, :sticks, :scatter,
     :heatmap, :hexbin, :barbins, :barhist, :histogram, :scatterbins,
     :scatterhist, :stepbins, :stephist, :bins2d, :histogram2d, :histogram3d,
     :density, :bar, :hline, :vline,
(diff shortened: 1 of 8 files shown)


david-macmahon:
CI is failing because I added another series to test/example 11 so the image generated by the test no longer matches the reference image in PlotReferenceImages.  I could dev that package and update the reference image locally, but how do I make the Plots CI happy?  I could do a PR on PlotReferenceImages for the new reference image, but then that would break Plots CI for everyone else.  It seems like a chicken and egg problem.  What to do?

david-macmahon:
Looks like GitHub was faster than me at connecting the dots, but there is now a corresponding pull request on PlotReferenceImages.

daschw (approved):
Nice work! I'll merge the PlotReferenceImages PR and restart CI here. After that I think we can merge.

daschw:
Alternatively, you could add your example from above as an additional reference image and leave ref11 as it ist. Then we would not break CI for previous Plots versions and we'd have a nice image showing the difference between steppre, stepmid and steppost in the Examples section of the docs. What's your opinion on that?

david-macmahon:
> Alternatively, you could add your example from above as an additional reference image and leave ref11 as it ist. Then we would not break CI for previous Plots versions and we'd have a nice image showing the difference between steppre, stepmid and steppost in the Examples section of the docs. What's your opinion on that?

I like that idea!  I will update the PR shortly.

david-macmahon:
Looks like Guthub beat me to it again! :)  I assume it will fail because the new reference plot is not yet merged on JuliaReferencePlots, but once that's done CI should pass OK.

daschw:
Awesome, thanks!

daschw:
Tests are passing now, thanks again @david-macmahon !

david-macmahon:
👍

PR #3387, opened 2021-03-27
Use PlotlyJS's integration with WebIO

olegshtch:
See http://juliaplots.org/PlotlyJS.jl/stable/syncplots/#SyncPlots


daschw:
Thanks a lot!

Issue #3397, opened 2021-03-29
[BUG] Undefined variable, 3d plot with framestyle = :origin


hustf:
### Reproduce
```
using Plots
# Don't trigger
plot(-1:1, -1:1, -1:1)
# Trigger
plot(-1:1, -1:1, -1:1; framestyle = :origin)
```
### Error message
```
Error showing value of type Plots.Plot{Plots.GRBackend}:
ERROR: UndefVarError: ind not defined
Stacktrace:
  [1] axis_drawing_info_3d(sp::Plots.Subplot{Plots.GRBackend}, letter::Symbol)
    @ Plots ~\.julia\packages\Plots\z5Msu\src\axes.jl:761
```

### Backends

This bug occurs in Plots.

### Versions
Plots.jl version: Plots v1.11.0  
Backend  version (`]st -m`):
Output of `versioninfo()`:
```
Julia Version 1.6.0-rc3
Commit 23267f0d46 (2021-03-16 17:04 UTC)
Platform Info:
  OS: Windows (x86_64-w64-mingw32)
  CPU: AMD Ryzen 5 3600X 6-Core Processor
  WORD_SIZE: 64
  LIBM: libopenlibm
  LLVM: libLLVM-11.0.1 (ORCJIT, znver2)
```
### Cause
'ind' was probably in scope during development, as it's a variable name used in the recipes pipeline.

### Proposed fix

1) Change axes.jl:760
`                    if ind !== nothing`
to
`                    if i0 !== nothing
`

2) Include code in CI tests.

### Workaround
Use the backend directly, change code locally, or don't use framestyle.

PR #3398, opened 2021-03-29
Bugfix and test, issue #3397

hustf:
modified:   src/axes.jl  #3397 fix varname
modified:   test/test_recipes.jl #3397 test all framestyles effect


daschw:
Thanks!

Issue #1495, opened 2018-04-23
primary attribute not taking effect in nested series


juliohm:
Could you please confirm that the following should produce two series with the same color?

```julia
@userplot FooPlot

@recipe function f(foo::FooPlot)
  y = foo.args[1]
  x = 1:length(y)

  @series begin
    seriestype := :path
    primary := false

    x, y
  end

  seriestype --> :scatter
  
  x, 2y
end

fooplot(rand(3))
```
![screenshot from 2018-04-23 14-53-28](https://user-images.githubusercontent.com/3345261/39155264-230b2d18-4706-11e8-880b-ceadbf7f57f8.png)

I remember using it in the past, but I am not sure.

joshday:
I believe `primary=false` is used for removing a series from the legend, which is what is happening in your example.  Questions like this may be better suited for slack or discourse.

juliohm:
Thank you @joshday , I think this one is a real issue. I used to make use of the `primary` attribute for exactly that: maintain the previous style from the last plot command. It is not only about the legend. Check the example below, which still works as of today:

```julia
using Plots

x = linspace(0,1,10)

scatter(x)
plot!(x, primary=false)
```
![screenshot from 2018-04-24 07-37-52](https://user-images.githubusercontent.com/3345261/39194380-7174557e-4792-11e8-846f-75ca1d6fe697.png)

The color is preserved with built-in plot commands. The issue is in defining nested series in plot recipes. If there is any other way to circumvent this, I'd be happy to learn.

mkborregaard:
You're right @juliohm . I think this was possibly introduced by https://github.com/JuliaPlots/Plots.jl/pull/1011 , but it's certainly an unwelcome side effect. Could you use this workaround?
```julia
@recipe function f(foo::FooPlot)
         y = foo.args[1]
         x = 1:length(y)

       @series begin
         seriestype --> :scatter
         x, 2y
       end


       @series begin
           seriestype := :path
           primary := false
           x, y
        end

         [],[]
       end
```

juliohm:
Thank you @mkborregaard, that is a very nice solution.

juliohm:
@mkborregaard the solution works nicely, except that the legend is affected. Even though the second series copies the color and style, the legend shows a different series (for the [], [] pair?) in the final plot.

mkborregaard:
Yes it seems to do that in some backends but not others. Just adding `primary := false` to the outermost series (`[],[]') should solve this.

[...]

PR #3423, opened 2021-04-07
correct scatterpath + add linearfit

matthieugomez:
correct scatterpath so that color does not change (see https://github.com/JuliaPlots/Plots.jl/issues/1495). Add linearfit.
@mkborregaard and @martinbiel since they prob. understand better than me what needs to be fixed.


daschw:
Awesome, thanks!

PR #3432, opened 2021-04-12
fixing :log10 colorbar scale

isentropic:
Cannot fix log scale colorbars in pyplot backend. The issue is that the marker fill color is calculated somewhere in PlotUtils colors and it is not consistent with the colorbar,

```
plot(rand(100), zcolor=10 .^(4 .+rand(100)), clim=(1,1e6), colorbar_scale=:log10, m=:c, msw=0)
```

image


src/backends/pyplot.jl
```diff
@@ -937,13 +937,21 @@ function _before_layout_calcs(plt::Plot{PyPlotBackend})
                 kw[:boundaries] = vcat(0, kw[:values] + 0.5)
             elseif any(colorbar_series[attr] !== nothing for attr in (:line_z, :fill_z, :marker_z))
                 cmin, cmax = get_clims(sp)
-                norm = pycolors."Normalize"(vmin = cmin, vmax = cmax)
+                norm = if sp[:colorbar_scale] === :identity
```
isentropic, line 940: This makes the matplotlib draw the colorbar properly (or the color gradient strip does not look linear)

src/backends/pyplot.jl
```diff
@@ -937,13 +937,21 @@ function _before_layout_calcs(plt::Plot{PyPlotBackend})
...
+
                 f = if colorbar_series[:line_z] !== nothing
-                    py_linecolormap
+                    x -> py_linecolormap(x, scale=sp[:colorbar_scale])
                 elseif colorbar_series[:fill_z] !== nothing
-                    py_fillcolormap
+                    x -> py_fillcolormap(x, scale=sp[:colorbar_scale])
                 else
-                    py_markercolormap
+                    x -> py_markercolormap(x, scale=sp[:colorbar_scale])
```
isentropic, line 954: I thought the marker color would be calculated here, but it presumably gets calculated somewhere early in the pipeline and I do not know where.

isentropic:
I'm also partially regretting the support for `colorbar_scale` option as it is really not trivial to implement. There are a few layers where log (and perhaps other) need to be implemented: backend specific scaling, Plots color calculation, JuliaGraphics/ColorSchemes? 

The easiest way is for the user to log the colorscale data beforehand and just plot normally; next, label the colorbar as log(Quantity).

I also feel like any scale could be supported if Plots preemptively scales the data accordingly. Plots could just provide appropriate decorations and ticks in mind with the input scale. So under the hood everything is linear scale, but the data would scaled first (with any function of users liking) and then mapped linearly. It feels what we do (:log10, :log2) is too specific and introduces bugs with backends not supporting a needed scale and the need to account for the nonlinearity. Eventually, even the backends map everything back to linear (as our displays are linear),

isentropic:
This is also not so much relevant to this issue, but why does Plots even try to support the scales? Can't Plots just provide axis formatters? 
For example:
I am plotting a 10^2, 10^3, 10^4 (three points) in log scale,
Then, Plots would just log it to get -> (2, 3, 4)
Then, just decorate the axis labels as (10^2, 10^3, 10^4) (instead of natural 2,3,4)

This would really avoid a lot shortcomings of backends that have trouble supporting logscale simply (matplotlib for example)

BeastyBlacksmith:
If we can get al the transformations working on the Plots side, I'm all for it.
I think delegating to the backend was the quick way to get something. The tricky part is to get the correct tick positions, lets say for minor ticks, since they are not linear on a log plot for example.

BeastyBlacksmith:
The issue from the OP seems to be a backend issue to me, that somehow is connected to specifying `clims`, since this
```
scatter(rand(100), zcolor=10 .^(4 .+rand(100)), colorbar_scale=:log10, m=:c, msw=0)
```
gives correct colors

isentropic:
I'm confused about mutual relationship between cscale clim and markerz

This PR makes it work without clims. It feels like clims is a Plots hack that changes the colors of the markers. I'm not sure though

BeastyBlacksmith:
I dont think Plots does any more than forward these to the backen.

PR #3435, opened 2021-04-14
Add ticks getter functions

briochemc:
I'm not sure why these don't already exist (surely there's a reason) but I thought that it would be nice to have a simple convenient way to get tick values and labels of a given subplot. This small PR adds that through functions called `xticks`, `yticks` and `zticks`. MWE:

```julia
julia> using Plots

julia> plt = plot(plot(rand(10)), plot(rand(10), yticks=([0.5, 0.6], ["foo", "bar"])))

julia> xticks(plt[1])
([2.0, 4.0, 6.0, 8.0, 10.0], ["2", "4", "6", "8", "10"])

julia> yticks(plt[2])
([0.5, 0.6], ["foo", "bar"])
```

where `plt` looked like this for reference:

Screen Shot 2021-04-14 at 10 25 05 am

Maybe this is a bad idea, or implemented the wrong way, and there are no tests, so please let me know what you think before
I spent more time in this :)


isentropic:
Great idea, I also think there should be a method for doing the same thing but on a plot object (not subplot) in case there is only one subplot. In case of multiple subplots perhaps return a list of all of them?

briochemc:
Oh yes, that's a good idea, I tried adding that in the latest commit. But for simplicity and for type-stability, when called on a `Plot` object, I thought a list (a `Vector`) should always be returned, even for plots with a single subplot. Maybe that's no good? MWEs with latest commit:

- Mutiple subplots:

    ```julia
    julia> plt = plot(plot(rand(10)), plot(rand(10), yticks=([0.5, 0.6], ["foo", "bar"])))
    
    julia> yticks(plt)
    2-element Vector{Tuple{Vector{Float64}, Vector{String}}}:
     ([0.0, 0.25, 0.5, 0.75, 1.0], ["0.00", "0.25", "0.50", "0.75", "1.00"])
     ([0.5, 0.6], ["foo", "bar"])
    ```

- Single subplot:
    
    ```julia
    julia> plt = plot(rand(10), yticks=([0.5, 0.6], ["foo", "bar"]))
    
    julia> yticks(plt)
    1-element Vector{Tuple{Vector{Float64}, Vector{String}}}:
     ([0.5, 0.6], ["foo", "bar"])
    ```

isentropic:
I did not notice https://github.com/JuliaPlots/Plots.jl/blob/0a7035979f9f03eb65823deaf8aec3b844725a7d/src/axes.jl#L220
before, perhaps this function is sufficient enough already?

briochemc:
It's sufficient in the sense that `xticks` just calls it, but the goal of this PR is to expose shortcut functions to users for convenience. For instance, compare

```julia
xticks(plt)
```

vs

```julia
[Plots.get_ticks(sp, sp[:xaxis]) for sp in plt.subplots]
```

isentropic:
Let's have someone else to review and merge this. At this point I'm not so sure

daschw (approved):
I think there's no harm in adding these. This requires a minor release, right?

briochemc:
> I think there's no harm in adding these. This requires a minor release, right?

I think so, so I bumped the version number to v1.12.0. I also added a simple test in `test/test_axes.jl`.

briochemc:
Tiny usage example here that may be useful for the documentation:

Given a simple plot,

```julia
julia> p = plot(rand(5))
```

Screen Shot 2021-04-15 at 11 44 03 am


one can get the ticks of the x-axis with the `xticks` function

```julia
julia> t = xticks(p)[1] # need the [1] because xticks returns a vector of the xticks of the subplots
([1.0, 2.0, 3.0, 4.0, 5.0], ["1", "2", "3", "4", "5"])
```

and use them to modify the ticks of `p`, e.g., to add a custom tick

```julia
julia> xticks!(([t[1]; 3.5], [t[2]; "α"]))
```

Screen Shot 2021-04-15 at 11 44 27 am

or replace an existing one

```julia
julia> xticks!((t[1], replace(t[2], "3"=>"α=3")))
```

Screen Shot 2021-04-15 at 11 44 19 am

briochemc:
Speaking of which, where do we write docstrings for Plots.jl? I started writing a docstring in `src/axes.jl` (see snippet of what `help?> xticks` would return below) but then I realized that there were no docstrings in that file...

Screen Shot 2021-04-15 at 11 53 43 am

src/axes.jl
```diff
@@ -235,6 +235,17 @@ function get_ticks(sp::Subplot, axis::Axis; update = true)
...
+# Ticks getter functions
+get_ticks(sp::Subplot, s::Symbol) = get_ticks(sp, sp[Symbol(s, :axis)])
+xticks(sp::Subplot) = get_ticks(sp, :x)
+yticks(sp::Subplot) = get_ticks(sp, :y)
+zticks(sp::Subplot) = get_ticks(sp, :z)
+get_ticks(p::Plot, s::Symbol) = [get_ticks(sp, s) for sp in p.subplots]
+xticks(p::Plot) = get_ticks(p, :x)
+yticks(p::Plot) = get_ticks(p, :y)
+zticks(p::Plot) = get_ticks(p, :z)
+export xticks, yticks, zticks
```
BeastyBlacksmith, line 247:
```suggestion
```
There is no official guideline on this, but as far as I can see, we don't export "getters" (like `Plots.xlims` for example)

briochemc: Maybe I'm misunderstanding something, but isn't `xlims` exported there?

https://github.com/JuliaPlots/Plots.jl/blob/1ceb3a561f5e9f20471636865180d603d09753ed/src/Plots.jl#L68-L70

Maybe it's just a matter of moving these exports in that long list?

briochemc: Yep, this works for me with the current version (v1.11.2):
```julia
julia> using Plots

julia> xlims(plot())
(0.0, 1.0)
```

BeastyBlacksmith: good Point, I didn't know. There is nothing to be seen here then ^^;

briochemc: But should I move the export of `xticks` to where that of `xlims` is?

briochemc:
> Speaking of which, where do we write docstrings for Plots.jl? I started writing a docstring in `src/axes.jl` (see snippet of what `help?> xticks` would return below) but then I realized that there were no docstrings in that file...
> 
> 
> 
> Screen Shot 2021-04-15 at 12 33 18 pm
> 
> 
> 
> 

@daschw or @BeastyBlacksmith could you comment on that? I.e., Where should the docstrings be written?

briochemc:
Is there anything else I can do for this PR?

BeastyBlacksmith:
All good, thank you :)

PR #3441, opened 2021-04-16
colorbar improvements for pgfplotsx

BeastyBlacksmith:
catching up the changes from the PR of @isentropic #3346

fix https://github.com/JuliaPlots/Plots.jl/issues/3449


BeastyBlacksmith:
If tests pass this should be ready to merge. I don't think it got the `colorbar_scale`, but setting ticks independently of axis is an improvement already I'd say

PR #3463, opened 2021-05-01
Add annotation font defaults/settings

gustaphe:
The font family for annotations were inconsistent with other text objects, if `default(fontfamily)` was set. I figured there was a place for defaults for all the annotation font options:

```julia
default(fontfamily="Computer Modern", annotationfontsize=16)
anns=[(0.1,0.1,"String"), (0.2,0.2,("String",45.0)), (0.3,0.3,("String",colorant"red")), (0.5,0.5,("String", 18, "Computer Modern"))]
plot(layout=2)
annotate!(anns; subplot=1, title="Defaults")
annotate!(anns; subplot=2, title="Others", annotationrotation=-45.0, foreground_color_subplot=colorant"green")
```

fontdefaults

Any thoughts?


daschw (approved):
Just a general remark on all the font attributes, not just for annotations: In the long run we might think about replacing all these `fontfamily`, `fontsize`, ... attributes with just `font` attributes holding a `Font` type, that may be set by the user with NamedTuples like e.g. `xguidefont = (family="Computer Modern", size=14, halign=:right)`

gustaphe:
I agree. Just one issue with that. Ideally you'd want it to be possible for `default` to set a fontsize, a type recipe to set a valign, the plot call kwargs to set a family and individual annotation tuples to set a color. Extreme example, but still.

It could work, but `font=tuple` must then not be an alias for `font=Font(tuple...)` but for `font=Font(sp[:font], tuple...)` or something.

PR #3473, opened 2021-05-04
WIP: Fonts

gustaphe:
I started fiddling with a redefinition of the fonts system (as discussed in #3463)

It's currently in a state where it compiles and plots, but I've only done just enough adjustment in the `gr` backend to get it in "working" order. But I think it illustrates the idea.

Maybe the gains aren't that big, and there's probably quite a bit of work left, but the shortening of the attribute list feels pretty good.

```julia
plot(randn(10); font=(:green, 45.0))
```
image

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,7 +1,7 @@
 name = "Plots"
 uuid = "91a5bcdd-55d7-5caf-9e0b-520d859cae80"
 author = ["Tom Breloff (@tbreloff)"]
-version = "1.13.2"
+version = "1.13.3"
 
 [deps]
 Base64 = "2a0f44e3-6c83-55bd-87e4-b1978d98bd5f"
--- a/src/args.jl
+++ b/src/args.jl
@@ -320,6 +320,7 @@ const _series_defaults = KW(
     :contour_labels     => false,
     :subplot            => :auto,     # which subplot(s) does this series belong to?
     :series_annotations => nothing,       # a list of annotations which apply to the coordinates of this series
+    :series_annotationfont => nothing,
     :primary            => true,     # when true, this "counts" as a series for color selection, etc.  the main use is to allow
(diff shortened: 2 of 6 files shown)


isentropic:
Perhaps I'm a little late to this discussion, but what are the clear advantages of separating font attrs into a new named tuple therefore breaking a lot of existing recipes (perhaps most of them)? This is a massive breaking change, but what are the clear advantages here?

Additionally, this has to be implemented for all backends or else other backends would just stop working

gustaphe:
Not at all, it's an infant idea, but I wanted to see what it would look like.

Advantages I see:
* Shortening attribute list -- each text object currently has 6 attributes by a specific pattern (unless control of any of them is missing): `*family`, `*pointsize`, `*rotation`, `*color`, `*halign` and `*valign`. All of these attributes are already encapsulated in the `Font` type
* Guaranteed uniformity -- those text objects that currently *are* missing certain controls won't, and any text objects added through a scheme like this won't either
* Clarifying and simplifying font dependencies -- instead of specifying where `xguidefontfamily`, `xguidefontsize`... are inherited from separately, one simply specifies that `guidefont` is `font`, but with a specific `pointsize`, etc.

That said, it *is* a breaking change, and quite a bit of details to implement. It might be possible to implement it in a non-breaking fashion, where `font=(45.0, "Computer Modern")` is a convenience for `fontrotation=45.0, fontfamily="Computer Modern"`, or vice versa.

BeastyBlacksmith:
You may also look at https://github.com/JuliaPlots/Plots.jl/pull/2854 where I wrote a macro to traverse into any Plot specific types (including Font) and creating the attributes automatically (incl. aliases).
I think it could also work non-breaking if we can make sure, that aliases work for all kinds of recipes.

PR #3494, opened 2021-05-12
Create docs on pushing to master

BeastyBlacksmith:
This is an experiment if we can build the docs from here, such that we don't get out of touch so easily


BeastyBlacksmith:
It's done, our docs are alive again https://docs.juliaplots.org/latest/

PR #3497, opened 2021-05-13
Allow `while`-block in at-animate macro

giordano:
The `@animate` example in [documentation](https://docs.juliaplots.org/dev/animations/#Convenience-macros) just works for me on this branch with
```julia
anim = @animate while i ≤ n
    circleplot(x, y, i)
    i += 1
end
```

For reference: https://discourse.julialang.org/t/animate-with-while-loop-in-julia/61069


daschw:
Nice, thanks!

Issue #3379, opened 2021-03-25
[FR] Adding Histogram3d/Lego-Plot


bernd1995:
After trying to find a nice way of creating a 3D-Histogram it turned out that there is no such feature yet. (Although it is reported as possible `:seriestype` in the [Docs/attribute_series](http://docs.juliaplots.org/latest/generated/attributes_series/))

So I came up with a very hacky way of doing it using `wireframe()` and I thought the Code may be useful in the future.

```julia
function histogram3d(x::AbstractArray{T,1}, y::AbstractArray{T,1}; xlims::Tuple{Number,Number}=(0,0), ylims::Tuple{Number,Number}=(0,0), bins::Int64, kws...) where {T}

    # input check, if not given take extrema
    if all(xlims == (0,0))
        xlims = extrema(x)
    end
    if all(ylims == (0,0))
        ylims = extrema(y)
    end

    # create the x-y-grid to check values
    xgrid = range(xlims[1],xlims[end],length=bins+1)
    ygrid = range(ylims[1],ylims[end],length=bins+1)

    # count all x-y-values within a specific x-y-grid-cell (histcount may be nicer)
    z = [sum(all(((x[i],y[i]).>=(xgrid[itx],ygrid[ity])) .& ((x[i],y[i]).<(xgrid[itx+1],ygrid[ity+1]))) for i in eachindex(x)) for ity = 1:bins, itx = 1:bins]

    # Helpers for nice bars
    xdat = repeateps(xgrid,inner=(2));
    ydat = repeateps(ygrid,inner=(2));

    zdat = zeros(length(xdat),length(ydat));
    zdat[2:end-1,2:end-1] .= repeat(z,inner=(2,2));

    wireframe(xdat,ydat,zdat,xlims=xlims,ylims=ylims;kws...)
end

repeateps(x;kws...) = repeat(x;kws...) .- [k%2 == 1 ? 1e-15 : 0 for k = 1:2*length(x)];
```

The code basically adds additional points for the `wireframe` such that the resulting graph contains something that looks like a `bar`.
However I tried this so far only with the GR- and the plotly-backend.
GR produced quite decent results (apart from the known issue #1911 which makes it hard to look at it from all angles) :
![7-7-0 7-0 5_sig38normlego](https://user-images.githubusercontent.com/20151553/112508367-c16e4080-8d8f-11eb-8b9f-f1f85e9aea4b.png)

[...]

PR #3507, opened 2021-05-17
First attempt at histogram3d/legoplot

bernd1995:
As already mentioned in #3379 I wanted to add a `histogram3d()` function for plotting. 
Therefore I took over the ides from the issue and translated it in a recipe, however now I am a bit stuck.

The main problems + possible solutions are:
  1. `wireframe` is not reliable across the different backends (very different behaviours) so I switched to `surface`
  2. `surface` does not provide a mesh option (like `surface` + `wireframe`) and so one does not get a lot of information from the plots. 
     - solution 1: adding all the lines manually? may be very resource intensive
     - solution 2: Adding a wireframe? does not provide the correct overlay properties depending on the backend
  3. color: Should it be one color like it is done for usual histograms or would it be nicer to have colored bars depending on their height (like a heatmap not seen from above)?
  4. Additionally I wanted to add the same possibility as in 2d-histograms to use the attribute `show_empty_bins` which turned out to be really hard. Especially the `plotly()` backend seems to have a lot of trouble with a mixture of `NaN` and numeric-values at the same x-y position.
     - Additionally this needs the `_allneighbours` function which checks if all neighbours in a matrix are the same value, which I need to set the `NaN`-values for the plot. I really dislike this function a lot but I couldn't wrap my head around another idea to resolve this

# Figures

All plots are created by
```julia
x=rand(100)
y=rand(100)
histogram3d(x,y,bins=(10,10),camera=(20,60),xlabel="x")
```

### One-coloured surface plot
- Good: Shows bins as expected
- Bad: Not very useful without lines at the edges (Problem 2 from above)
hist3d_blue_gr

### Multi-coloured surface plot
- Good: Shows bins as expected, able to separate different bins by color
- Bad: Color gradient messes here with the plot, would be better to have one fixed color for each bar instead of a gradient (How to resolve this?)
hist3d_colored_gr

### Single-coloured surface + wireframe plot
Not possible with the current solution, wince `wireframe` in GR needs x and y values in ascending (i.e. repeated values are forbidden). Therefore one would need again the workaround with `repeateps()` in #3379 which I have discarded for now.

### One-coloured surface plot
- Good: Bins separable by eye due to automatic shadows in plotly
- Bad: lines at the edges would still be helpful (Problem 2 from above), `NaN`-values are not completely correct rendered as can be seen at for instance `x=0.2` and `y = 0.0 to 0.1` by a missing part of the wall.
hist3d_blue_plotly

### Multi-coloured surface plot
- Good: Shows bins as expected, able to separate different bins by color
- Bad: Color gradient messes here with the plot, would be better to have one fixed color for each bar instead of a gradient (How to resolve this?), also here problem with `NaN`-values e.g. at `x=0.4`
hist3d_colored_plotly_2
hist3d_colored_plotly

### Single-coloured surface + wireframe plot
- Good: Bins separable by eye due to automatic shadows in plotly
- Bad: Wireframe lines are definitely not displayed as expected
hist3d_colored_wf_plotly

## Pyplot

Already worked yesterday, but somehow is not compiling at the moment on my PC. I'll try to add figures as soon as possible.

# Further info

I am totally willing to expand this myself but I also definitely need some help (and also style-decisions) here. So let me know if you have any idea. Also @dorn-gerhard it would be really nice if your colleagues could add their solutions as proposed in #3379

Thanks already and Cheers.

### DIFF
--- a/src/recipes.jl
+++ b/src/recipes.jl
@@ -988,6 +988,74 @@ end
 
 
 # ---------------------------------------------------------------------------
+# Histogram 3D
+
+# Check if all neighbours have the same value
+# TODO: Improve this somehow
+function _allneighbours(arr, I::CartesianIndex, val = 0.0)
+    arr[I] == val &&
+    arr[I + CartesianIndex(1,0)]  == val &&
+    arr[I + CartesianIndex(0,1)]  == val &&
+    arr[I - CartesianIndex(1,0)]  == val &&
+    arr[I - CartesianIndex(0,1)]  == val &&
+    arr[I + CartesianIndex(1,1)]  == val &&
+    arr[I + CartesianIndex(1,-1)] == val &&
+    arr[I - CartesianIndex(1,1)]  == val &&
+    arr[I - CartesianIndex(1,-1)] == val
+end
+
+@recipe function f(::Type{Val{:bins3d}}, x, y, z)
+    edge_x, edge_y, weights = x, y, z.surf
+
+    if !plotattributes[:show_empty_bins]
+        edge_x = repeat(edge_x,inner=(4))[2:end-1]
+        edge_y = repeat(edge_y,inner=(4))[2:end-1]
+        float_weights = spzeros(length(edge_x),length(edge_y))
+        float_weights[2:end-1,2:end-1] = repeat(float(weights),inner=(4,4))
+        m = falses(size(float_weights))
+        for I in CartesianIndices((2:length(edge_x)-1,2:length(edge_y)-1))
+            m[I] = _allneighbours(float_weights,I,0)
+        end
+        float_weights[m] .= NaN
+    else
+        edge_x = repeat(edge_x,inner=(2))
+        edge_y = repeat(edge_y,inner=(2))
+        float_weights = spzeros(length(edge_x),length(edge_y))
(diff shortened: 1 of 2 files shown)


bernd1995:
And please could someone add the **help wanted** label since I do 't habe permissions to add labels 😅

bernd1995:
Added now another possibility to tackle this problem, by really plotting 3d-bars at each x-y-spot as a separate series. 
This works more reliable across backends but becomes very unfeasible for too much bins i.e. `bins=(10,10)` already creates 100 bars which uses a lot of memory.

Also in `gr()` the order of the series matters because otherwise a bar in the background overlay one in the front. 

I will try to resolve this by not plotting one bar at each x-y-spot but instead plot one surface for each connected area (e.g. one slice from `x[i]` to `x[i+1]` for all y-values, or really trying to identify isolated areas). This should reduce the amount of unnecessary planes by a huge amount.

bernd1995:
> I will try to resolve this by not plotting one bar at each x-y-spot but instead plot one surface for each connected area (e.g. one slice from `x[i]` to `x[i+1]` for all y-values, or really trying to identify isolated areas). This should reduce the amount of unnecessary planes by a huge amount.

However this will make it harder again to have differently colored bars.

Issue #3503, opened 2021-05-17
How to get acknowledged


BeastyBlacksmith:
As Plots.jl is a community maintained project, it lives from its contributors.
To give proper credit to all our contributors we created a `.zenodo.json` file in https://github.com/JuliaPlots/Plots.jl/pull/3482.

If you are a new contributor, please add your information to that file.

All contributors who didn't comment on https://github.com/JuliaPlots/Plots.jl/pull/3482 may double check their entry.
I only added names of people and sometimes I didn't even find those.

Thank you for your work <3

BeastyBlacksmith:
Partly a reason for doing this was to include this data as a table for a Paper on Plots.

We now have a preprint for this up at https://arxiv.org/abs/2204.08775. The coming weeks are the time to give it a review and get in contact with me if you want your name on there.
I am tagging all those poeple that are in the code ownership graphics anyway and that I haven't had contact with until now: @ma-laforge, @yha, @jheinen, @t-bltg, @apalugniok, @piever, @JackDevine, @isentropic

PR #3510, opened 2021-05-18
Minor cleanup plotlyjs

fonsp:
Adding a `var` to avoid ["sloppy mode" JS](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Strict_mode).


BeastyBlacksmith:
Thanks! Would you mind also updating your information in https://github.com/JuliaPlots/Plots.jl/issues/3503?

PR #3514, opened 2021-05-19
NaN checking in backends/gr.jl

bernd1995:
Although it seems that the function `gr_nans_to_infs!()` is not used anywhere (maybe in some other Package?) I guess it is still better to have it working if it is necessary some day.

Also updated my infos in the .zenodo.json file. Thanks for including me 🥰

### DIFF
--- a/.zenodo.json
+++ b/.zenodo.json
@@ -452,8 +452,10 @@
             "type": "Other"
         },
         {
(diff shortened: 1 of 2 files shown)


BeastyBlacksmith:
Thank you for contributing! :)
( and for updating your entry )

Issue #2579, opened 2020-04-14
twiny() ?


jenfenghsu:
Is it possible to have similar hack as twinx() to have secondary x-axis on top of the plot? This can be useful when displaying the same curve with alternative units. See below for example. Or, any suggestion to achieve the same in Julia Plots? Thanks!
![image](https://user-images.githubusercontent.com/12201322/79241438-0e882980-7e41-11ea-9d4d-8cfef7e3a4bf.png)

My minimal example code with PyPlot package (but I'd very much love to continue using Plots):
```julia
fig, ax1 = PyPlot.subplots(num="Temperature Dependence",figsize=(10,6),facecolor="0.98")

ax1.set_xlabel("Temperature (°C)", fontsize=20)
ax1.set_ylabel("Response", fontsize =20)
x = collect(0:0.1:100)
y = 0.1 .* x .^ 2 - 5 .* x .+ 8
ax1.plot(x, y, lw=1., label="Object 1") 
ax1.legend(loc="best")
ax1.grid(true)

ax2 = ax1.twiny()
ax2.set_xlabel("Temperature (°F)", fontsize =20)
ax2.plot(x .* 9/5 .+ 32, y, lw=0.) # Ghost plot to show scale

PyPlot.show()
```

feanor12:
I found some code for twiny here :
https://stackoverflow.com/questions/64176617/julia-two-x-axes-for-plot-of-same-data

PR #3532, opened 2021-06-02
Twiny

mmikhasenko:
adds `twiny()` function that is analogous to `twinx()`
Close https://github.com/JuliaPlots/Plots.jl/issues/2579

### DIFF
--- a/src/Plots.jl
+++ b/src/Plots.jl
@@ -43,6 +43,7 @@ export
     default,
     with,
     twinx,
+    twiny,
 
     pie,
     pie!,
--- a/src/layouts.jl
+++ b/src/layouts.jl
@@ -785,3 +785,16 @@ function twinx(sp::Subplot)
 end
 
 twinx(plt::Plot = current()) = twinx(plt[1])
+
+function twiny(sp::Subplot)
+    sp[:top_margin] = max(sp[:top_margin], 20px)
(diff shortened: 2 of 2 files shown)


isentropic:
I think we can merge this right away if we have the accompanying docs. Since `twinx`is not even oficially documented, would you please add a new subsection under http://docs.juliaplots.org/latest/layouts/#Adding-Subplots-incrementally and document a brief usage of `twinx, twiny`? Once the corresponing PR with docs lands to PlotDocs repo we can merge both right away.

BeastyBlacksmith:
I am actually a bit hesitant here, because `twinx` is causing a lot of trouble (which may be why it is undocumented so far). I'd prefer having the "providing additional scales" sorted out before adding more of them

isentropic:
That's fair, let's hold on this then

mmikhasenko:
@BeastyBlacksmith 
Are issues with `twinx` collected somewhere?
 - I noticed that it does not work correctly with `pgfplotsx()`, particularly with the `heatmap`

BeastyBlacksmith:
https://github.com/JuliaPlots/Plots.jl/issues?q=is%3Aissue+is%3Aopen+twinx

t-bltg:
Superseded by https://github.com/JuliaPlots/Plots.jl/pull/4488.

Issue #3318, opened 2021-02-23
[FR] Fancy plotting of the ticks in log scale


getzze:
I have been annoyed by the ticks in logscale for some time now, so I finally wrote a `fancylogscale!` function to improve it and I wanted to share it with you all in case somebody needs it.
Tested with PyPlot and GR.

The function:

```julia
"""
    get_tickslogscale(lims; skiplog=false)
Return a tuple (ticks, ticklabels) for the axis limit `lims`
where multiples of 10 are major ticks with label and minor ticks have no label
skiplog argument should be set to true if `lims` is already in log scale.
"""
function get_tickslogscale(lims::Tuple{T, T}; skiplog::Bool=false) where {T<:AbstractFloat}
    mags = if skiplog
        # if the limits are already in log scale
        floor.(lims)
    else
        floor.(log10.(lims))
    end
    rlims = if skiplog; 10 .^(lims) else lims end

    total_tickvalues = []
    total_ticknames = []

    rgs = range(mags..., step=1)
    for (i, m) in enumerate(rgs)
        if m >= 0
            tickvalues = range(Int(10^m), Int(10^(m+1)); step=Int(10^m))
            ticknames  = vcat([string(round(Int, 10^(m)))],
                              ["" for i in 2:9],
                              [string(round(Int, 10^(m+1)))])
        else
            tickvalues = range(10^m, 10^(m+1); step=10^m)
            ticknames  = vcat([string(10^(m))], ["" for i in 2:9], [string(10^(m+1))])
        end

        if i==1
            # lower bound
            indexlb = findlast(x->x<rlims[1], tickvalues)
            if isnothing(indexlb); indexlb=1 end
        else
            indexlb = 1
        end
        if i==length(rgs)
            # higher bound
            indexhb = findfirst(x->x>rlims[2], tickvalues)
            if isnothing(indexhb); indexhb=10 end
        else
            # do not take the last index if not the last magnitude
            indexhb = 9
        end

        total_tickvalues = vcat(total_tickvalues, tickvalues[indexlb:indexhb])
        total_ticknames = vcat(total_ticknames, ticknames[indexlb:indexhb])
    end
    return (total_tickvalues, total_ticknames)
end

"""
    fancylogscale!(p; forcex=false, forcey=false)
Transform the ticks to log scale for the axis with scale=:log10.
forcex and forcey can be set to true to force the transformation
if the variable is already expressed in log10 units.
"""
[...]

Issue #3502, opened 2021-05-15
[BUG] xscale=:ln not working as intended


gertvv:
## Details

![image](https://user-images.githubusercontent.com/80785/118374126-b9c25f80-b5ba-11eb-8e49-63a742bdba25.png)

I don't think xscale=:ln is working as expected, see the above annotated screenshot. The first plot is use a linear scale, the second I have manually logged the points, the third is using xscale=:ln, and the fourth xscale=:log.

I would have expected the tickmarks on the :ln plot to be evenly spaced, as they are on the :log plot.

I brought this up for discussion on Gitter a few days ago, but it seems completely dead. If so, you might want to remove the link from the documentation.

### Backends

This bug occurs on ( insert `x` below )

Backend      | yes | no  | untested
-------------|-----|-----|---------
gr (default) | x   |     |
pyplot       |     |     | x
plotly       |     |     | x
plotlyjs     |     |     | x
pgfplotsx    |     |     | x
inspectdr    |     |     | x

### Versions

Plots.jl version: Plots v1.13.2
Backend  version (`]st -m`): GR v0.57.4
Output of `versioninfo()`:
Julia Version 1.6.1
Commit 6aaedecc44 (2021-04-23 05:59 UTC)
Platform Info:
  OS: Linux (x86_64-pc-linux-gnu)
  CPU: Intel(R) Core(TM) i5-5200U CPU @ 2.20GHz
  WORD_SIZE: 64
  LIBM: libopenlibm
  LLVM: libLLVM-11.0.1 (ORCJIT, broadwell)

t-bltg:
@gertvv , `:ln` is not supported by `gr` (see https://docs.juliaplots.org/latest/generated/supported/#Scales), do you get this warning ?

```julia
┌ Warning: scale ln is unsupported with Plots.GRBackend().  Choose from: [:identity, :log10]
└ @ Plots [...]/src/args.jl:1260
```

gertvv:
I didn't think I did, but... I did this in a Pluto notebook, and I've just re-run it, and turns out the warning went to the Julia console where Pluto was started rather than appearing in the notebook. I wonder if that's by design (I would certainly prefer if the warning had appeared in the notebook). In any case looks like this is not an issue with Plots.jl - thanks!

PR #3547, opened 2021-06-04
Rework major/minor ticks for log scales

### LINKED ISSUE #3303: Setting custom ticks in log plots with minor ticks enabled has confusing behavior
Say, for example, that I want to make a simple log-log plot.

```
x = range(0.1, 100, length=1000);
y = x.^2
```
And then plot it on a log log scale
```
p = plot(x, y, axis=:log, minorticks=10, minorgrid=true)
```
This will work, but produces a plot whose major ticks may not be where I want them (I.E not whole numbers, or whatever)

So if I go and add in x/y tick marks to correspond to the major locations, I.E
```
p = plot( ... , xticks=[5e-1, 5e0, 5e1, 5e2])
```
This works fine. But let's say now I want to just plot a sub-range of that data.
```
p = plot(x, y, axis=:log, minorticks=10, minorgrid=true, xticks=[5e-1, 5e0, 5e1, 5e2])
xlims!(1,10)
```
This will produce the following error:
```
julia> PlotTest.test_plot()
ERROR: MethodError: no method matching iterate(::Nothing)
Closest candidates are:
  iterate(::ExponentialBackOff) at error.jl:252
  iterate(::ExponentialBackOff, ::Any) at error.jl:252
  iterate(::Base.AsyncGenerator, ::Base.AsyncGeneratorState) at asyncmap.jl:382
  ...
Stacktrace:
 [1] axis_drawing_info(::Plots.Subplot{Plots.GRBackend}, ::Symbol) at C:\Users\fergu\.julia\packages\Plots\5ItHH\src\axes.jl:668
 [2] gr_draw_axis(::Plots.Subplot{Plots.GRBackend}, ::Symbol, ::Array{Float64,1}) at C:\Users\fergu\.julia\packages\Plots\5ItHH\src\backends\gr.jl:1222
 [3] gr_draw_axes(::Plots.Subplot{Plots.GRBackend}, ::Array{Float64,1}) at C:\Users\fergu\.julia\packages\Plots\5ItHH\src\backends\gr.jl:1216
 [4] gr_display(::Plots.Subplot{Plots.GRBackend}, ::Measures.Length{:mm,Float64}, ::Measures.Length{:mm,Float64}, ::Array{Float64,1}) at C:\Users\fergu\.julia\packages\Plots\5ItHH\src\backends\gr.jl:881
 [5] gr_display(::Plots.Plot{Plots.GRBackend}, ::String) at C:\Users\fergu\.julia\packages\Plots\5ItHH\src\backends\gr.jl:596
 [6] _show(::Base.GenericIOBuffer{Array{UInt8,1}}, ::MIME{Symbol("image/svg+xml")}, ::Plots.Plot{Plots.GRBackend}) at C:\Users\fergu\.julia\packages\Plots\5ItHH\src\backends\gr.jl:1777
 [7] display(::VSCodeServer.InlineDisplay, ::MIME{Symbol("image/svg+xml")}, ::Plots.Plot{Plots.GRBackend}) at c:\Users\fergu\.vscode\extensions\julialang.language-julia-1.1.17\scripts\packages\VSCodeServer\src\display.jl:0   
 [8] display(::VSCodeServer.InlineDisplay, ::String, ::Any) at .\multimedia.jl:216
 [9] display(::VSCodeServer.InlineDisplay, ::Plots.Plot{Plots.GRBackend}) at c:\Users\fergu\.vscode\extensions\julialang.language-julia-1.1.17\scripts\packages\VSCodeServer\src\display.jl:108
 [10] display(::Any) at .\multimedia.jl:328
 [11] test_plot() at K:\data\ShockDt\PlotTest.jl:8
 [12] top-level scope at REPL[102]:1
```

Likewise, if I comment out the `xlims!` line and instead do something else like...

```
# Note that the ticks have been changed to span the range of x that I want to restrict to via xlims!
p = plot(x, y, axis=:log, minorticks=10, minorgrid=true, xticks=[1e-1, 5e0, 1e1])
```
This works fine, but if I reintroduce that `xlims!` call I get the same error, even though the the ticks span the whole range I want to view. (And, in fact, it seems that the ticks don't even have to span the range of the data)

Additionally, if I remove the `minorticks` and `minorgrid` arguments, this works fine. It also works as expected if I keep the minor axis arguments but remove the axis scale specification.

I would expect that setting ticks and lims to not have a dependence on each other, or on the scale type or whether a minor grid exists. I.E - if a tick falls outside of the range specified, it would just not be plotted. If the limits are greater than the specified ticks there are either no ticks plotted (which seems to already be the case), or an error indicating that the ticks do not span the plotted range to be emitted.

t-bltg:
This PR is a proposal for generating automatic ticks on `log` based scale family (`log`, `log2`, `log10`).
Currently, exponents are not integer based.
Furthermore, minor ticks are generated using arbitrary (`5`) intervals for both linear and log scales.

This PR:
- adds automatic computation of major ticks based on integer exponents for log scales (decades for `log10`);
- computes sub minor ticks when transitioning from non unit order of magnitude,  e.g.  `10¹ -> 10³` instead of `10¹ -> 10² -> 10³` on major ticks for a `log10` based scale (see below for an example).

Example:
```julia
using Plots; gr()
x = y = Float64[1]
for i ∈ range(-5, 6, step=1)
  append!(x, 2 * 10.0^i:10.0^i:10.0^(i + 1) |> collect)
end
plot(
  x, y, axis=:log10, minorgrid=true, legend=false,
  minorticks=true, shape=:cross, ms=2, lw=.5, dpi=300
)
png("loglog")
```
Without PR (open the link for high res minorticks grid):
loglog-cmp-NOK

With PR (open the link for high res minorticks grid):
loglog-cmp-OK

Closes https://github.com/JuliaPlots/Plots.jl/issues/3318
Fixes #3303

### DIFF
--- a/src/axes.jl
+++ b/src/axes.jl
@@ -136,6 +136,23 @@ const _label_func_tex = Dict{Symbol,Function}(
 )
 labelfunc_tex(scale::Symbol) = get(_label_func_tex, scale, convert_sci_unicode)
 
+function scaled_log_ticks(amin, amax, scale, max_ticks=8, strict=true)
+  base = _logScaleBases[scale]
+  log_amin = log(amin) / log(base)
+  log_amax = log(amax) / log(base)
+
+  # number of decades
+  num_decs = floor(Int, log_amax) - ceil(Int, log_amin)
+  step = max(1, min(round(Int, num_decs / max_ticks), num_decs))
+
+  start = (strict ? ceil : floor)(Int, log_amin)
+  stop = (strict ? floor : ceil)(Int, log_amax)
+  exponents = range(start, stop, step=step) |> collect
+
+  # majorticks = base .^ exponents
+  # return exponents, majorticks
+  return exponents
+end
 
 function optimal_ticks_and_labels(ticks, alims, scale, formatter)
     amin, amax = alims
@@ -169,12 +186,16 @@ function optimal_ticks_and_labels(ticks, alims, scale, formatter)
 
     # get a list of well-laid-out ticks
     if ticks === nothing
-        scaled_ticks = optimize_ticks(
-            sf(amin),
-            sf(amax);
-            k_min = 4, # minimum number of ticks
-            k_max = 8, # maximum number of ticks
-        )[1]
+        if scale in _logScales
+            scaled_ticks = scaled_log_ticks(amin, amax, scale)
+        else
+            scaled_ticks = optimize_ticks(
+                sf(amin),
+                sf(amax);
+                k_min = 4, # minimum number of ticks
+                k_max = 8, # maximum number of ticks
+            )[1]
+        end
     elseif typeof(ticks) <: Int
         scaled_ticks, viewmin, viewmax = optimize_ticks(
             sf(amin),
@@ -335,11 +356,12 @@ function get_minor_ticks(sp, axis, ticks)
     ticks = ticks[1]
     length(ticks) < 2 && return nothing
 
+    scale = axis[:scale]
     amin, amax = axis_limits(sp, axis[:letter])
-    #Add one phantom tick either side of the ticks to ensure minor ticks extend to the axis limits
+    # Add one phantom tick either side of the ticks to ensure minor ticks extend to the axis limits
     if length(ticks) > 2
         ratio = (ticks[3] - ticks[2])/(ticks[2] - ticks[1])
-    elseif axis[:scale] in (:none, :identity)
+    elseif scale in (:none, :identity)
         ratio = 1
     else
         return nothing
@@ -348,13 +370,33 @@ function get_minor_ticks(sp, axis, ticks)
     last_step = ticks[end] - ticks[end-1]
     ticks =  [ticks[1] - first_step/ratio; ticks; ticks[end] + last_step*ratio]
 
-    #Default to 5 intervals between major ticks
-    n = typeof(axis[:minorticks]) <: Integer && axis[:minorticks] > 1 ? axis[:minorticks] : 5
+    # Default to 5 intervals between major ticks on a linear scale,
+    # base - 1 for integral based log scale and 0 for ℯ based log scale
+    integral_log = scale in (:log2, :log10)
+    base = scale in _logScales ? _logScaleBases[scale] : nothing
+    n_default = integral_log ? base - 1 : (scale == :ln ? 0 : 5)
+    n = typeof(axis[:minorticks]) <: Integer && axis[:minorticks] > 1 ? axis[:minorticks] : n_default
     minorticks = typeof(ticks[1])[]
-    for (i,hi) in enumerate(ticks[2:end])
-        lo = ticks[i]
-        if isfinite(lo) && isfinite(hi) && hi > lo
-            append!(minorticks,collect(lo + (hi-lo)/n :(hi-lo)/n: hi - (hi-lo)/2n))
+
+    if n > 0
+        sub = integral_log ? round(Int, log(ratio) / log(base)) : nothing
+        for (i,hi) in enumerate(ticks[2:end])
+            lo = ticks[i]
+            if isfinite(lo) && isfinite(hi) && hi > lo
+                if integral_log
+                    for j in 1:sub
+                        lo_ = lo * base^(j - 1)
+                        hi_ = lo_ * base
+                        step = (hi_ - lo_) / n
+                        append!(minorticks, collect(
+                            lo_ + (j > 1 ? 0 : step) : step : hi_ - (j < sub ? 0 : step / 2)
+                        ))
+                    end
+                else
+                    step = (hi - lo) / n
+                    append!(minorticks, collect(lo + step : step : hi - step / 2))
+                end
+            end
         end
     end
     minorticks[amin .<= minorticks .<= amax]
@@ -701,7 +743,7 @@ function axis_drawing_info(sp, letter)
                 )
             end
         end
-        if !(ax[:ticks] in (:none, nothing, false))
+        if ax[:ticks] ∉ (:none, nothing, false)
             f = RecipesPipeline.scale_func(oax[:scale])
             invf = RecipesPipeline.inverse_scale_func(oax[:scale])
             if ax[:tick_direction] !== :none
@@ -732,7 +774,10 @@ function axis_drawing_info(sp, letter)
                 end
             end
 
-            if !(ax[:minorticks] in (:none, nothing, false)) || ax[:minorgrid]
+            if (
+                (ax[:minorticks] ∉ (:none, nothing, false) || ax[:minorgrid]) &&
+                minor_ticks !== nothing
+            )
                 tick_start, tick_stop = if sp[:framestyle] == :origin
                     t = invf(f(0) + 0.006 * (f(oamax) - f(oamin)))
                     (-t, t)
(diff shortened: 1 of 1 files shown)


src/axes.jl
```diff
@@ -169,12 +185,17 @@ function optimal_ticks_and_labels(ticks, alims, scale, formatter)
 
     # get a list of well-laid-out ticks
     if ticks === nothing
-        scaled_ticks = optimize_ticks(
-            sf(amin),
-            sf(amax);
-            k_min = 4, # minimum number of ticks
-            k_max = 8, # maximum number of ticks
-        )[1]
+        if scale in (:ln, :log2, :log10)
```
BeastyBlacksmith, line 188:
```suggestion
        if scale in _logScales
```

src/axes.jl
```diff
@@ -169,12 +185,17 @@ function optimal_ticks_and_labels(ticks, alims, scale, formatter)
...
+            base = scale == :ln ? ℯ : (scale == :log2 ? 2. : 10.)
```
BeastyBlacksmith, line 189:
```suggestion
            base = _logScaleBases[scale]
```

src/axes.jl
```diff
@@ -348,13 +370,33 @@ function get_minor_ticks(sp, axis, ticks)
     last_step = ticks[end] - ticks[end-1]
     ticks =  [ticks[1] - first_step/ratio; ticks; ticks[end] + last_step*ratio]
 
-    #Default to 5 intervals between major ticks
-    n = typeof(axis[:minorticks]) <: Integer && axis[:minorticks] > 1 ? axis[:minorticks] : 5
+    # Default to 5 intervals between major ticks on a linear scale,
+    # base - 1 for integral based log scale and 0 for ℯ based log scale
+    integral_log = scale in (:log2, :log10)
+    base = scale == :log2 ? 2 : 10
```
BeastyBlacksmith, line 376:
```suggestion
    base = _logScaleBases[scale]
```

t-bltg:
@BeastyBlacksmith thanks for the review, I will apply the changes right away.

In the mean time, the CI build failed (e.g. https://github.com/JuliaPlots/PlotReferenceImages.jl/blob/master/Plots/gr/1.11.2/ref5.png), can you please suggest a procedure to regenerate the reference images ?

BeastyBlacksmith:
You need to create the image locally and then provide a PR against PlotReferenceImages.jl, like https://github.com/JuliaPlots/PlotReferenceImages.jl/pull/81

t-bltg:
> You need to create the image locally and then provide a PR against PlotReferenceImages.jl, like [JuliaPlots/PlotReferenceImages.jl#81](https://github.com/JuliaPlots/PlotReferenceImages.jl/pull/81)

Thanks, will try.
> How do the ticks look like if I only plot in a narrow range, lets say two order of magnitudes?

Here is a narrower example:
loglog-cmp

BeastyBlacksmith (approved):
I think this looks good, but would be good to have some other eyes on this since ticks are an opinionated topic

src/axes.jl
```diff
@@ -732,7 +774,10 @@ function axis_drawing_info(sp, letter)
                 end
             end
 
-            if !(ax[:minorticks] in (:none, nothing, false)) || ax[:minorgrid]
+            if (
+                (!(ax[:minorticks] in (:none, nothing, false)) || ax[:minorgrid]) &&
+                !isnothing(minor_ticks)
```
t-bltg, line 779: Added a guard against minor_ticks === nothing.

BeastyBlacksmith:
```suggestion
                minor_ticks !== nothing
```

t-bltg:
> I think this looks good, but would be good to have some other eyes on this since ticks are an opinionated topic

Yes, better have other reviews (not bug 100% free e.g. when  `get_minor_ticks` returns nothing https://github.com/JuliaPlots/Plots.jl/blob/master/src/axes.jl#L345).

src/axes.jl
t-bltg: Is there a reason for not using the builtin `isnothing` ?

BeastyBlacksmith:
Would be also good if you can check what is the influence on https://github.com/JuliaPlots/Plots.jl/issues/3303 of this PR

src/axes.jl
BeastyBlacksmith: the compiler loves `===` checks

t-bltg:
> Would be also good if you can check what is the influence on #3303 of this PR

```julia
using Plots; gr()

x = range(0.1, 100, length=1000)
y = x.^2

p = plot(x, y, axis=:log, minorticks=10, minorgrid=true, xticks=[5e-1, 5e0, 5e1, 5e2])
xlims!(1,10)
png("test0.png")

p = plot(x, y, axis=:log, minorticks=10, minorgrid=true, xticks=[1e-1, 5e0, 1e1])
png("test1.png")
```

No errors, ticks correct.

src/axes.jl
t-bltg: Updated.

t-bltg:
It totally missed https://github.com/JuliaPlots/Plots.jl/issues/3318 when I wrote this PR. Maybe we can merge the best of both worlds and create something even better ?

BeastyBlacksmith:
Sure, if you find inspiration there to improve this PR. Just indicate when you think it is done.

BeastyBlacksmith:
Another one you might check is https://github.com/JuliaPlots/Plots.jl/issues/3502

t-bltg:
> Another one you might check is #3502

I think that nothing can be done here since gr does not support `:ln` scale:
```julia
┌ Warning: scale ln is unsupported with Plots.GRBackend().  Choose from: [:identity, :log10]
└ @ Plots [...]/src/args.jl:1260
```
Also, there is no MWE to reproduce (and the screenshot is too small, unreadable).


> Sure, if you find inspiration there to improve this PR. Just indicate when you think it is done.

I've looked into https://github.com/JuliaPlots/Plots.jl/issues/3318, I think nothing can be added here, so I think I'm done.

BeastyBlacksmith:
@daschw, @yha could you have a look at this?

isentropic:
I think the changes should go to Plot utils under `optimize_ticks` function instead. Perhaps overload the function or create `optimize_ticks_log`

isentropic:
Getting better tick locations is the job of the `optimize_ticks` function. Let us keep it like that instead of expanding the Plots.jl codebase more. Delegation to PlotUtils would lead to easier testing too.

isentropic:
Better yet, is to find optimum weights and params to the [optimize_ticks](function optimize_ticks(https://github.com/JuliaPlots/PlotUtils.jl/blob/4007370cd3c3b9a6cdc62871687dac21364190ff/src/ticks.jl#L130) function so that the returned ticks prefer "ten divisible ticks". Current solution feels like a hack. I believe by using the optimize_ticks (which is smart, perhaps too much even) function properly we can achieve the same result.

daschw:
Looks really great, but I agree with @isentropic that the code for finding tick locations should be in PlotUtils.jl.

t-bltg:
Hi all,
Sorry for the late reply.
I'm OK with rewriting stuff to go into https://github.com/JuliaPlots/PlotUtils.jl by the end of the week.

However I think we should bypass `optimize_ticks`'s logic rather that try to bury every functionality into one method.

The essence of this PR is to have `9` intervals per decade between each major tick for a `log10` scale e.g. for two major ticks at `10^0`  and `10^1` we should get in between `2, 3, 4, 5, 6, 7, 8, 9`. For other scales (`log2` and `log` the choice is arbitrary and can be left to the default 5 intervals.

t-bltg:
Superseded by https://github.com/JuliaPlots/Plots.jl/pull/3577, closing.

PR #3552, opened 2021-06-07
Respect :widen when passing :lims

### LINKED ISSUE #3339: [FR] Option to widen manual limits
Sorry if I missed a previous issue.

Plots.jl version: 1.10.6

In this case I tried using `ylims=(0,Inf)` to indicate that i want the y-axis to start at 0 (because here the area under the blue curve somewhat matters):

This results in _points_ at the maximum being cut off in the middle (and even the zero at the bottom isn't that pretty)! Because I couln't figure out another way except to calculate the maximum and use that times some factor (which I couldn't even do easily since they are a result of `scatterhist()`), I added `plot!([0],[0], markersize=0, primary=false)` to trick Plots.jl into thinking there was a point at y=0. This results in a somewhat nicer plot like this:

I get that `lims` are supposed to specify _exact_ limits but I would think that the `widen` parameter would accomplish exactly what I want. Instead it gets ignored when custom limits are passed since it defaults to `true` (whereas if you did want the current behaviour you would also have had the option to set `widen=false`). Since it's probably a bit late to change that, could there at least be an option like `widen = :force`? If that's an option, it seems simple enough that I could implement it.

Another workaround I just found is using `ylims()` - adjusting that and passing it back in - but even that seems a bit complicated and hard to discover (is there a reason why there can't be some page just documenting all Plots.jl functions, or have I just somehow missed it? Examples are great but `{x,y}lims()` and e.g. `mp4()` don't seem to be mentioned anywhere).

Easy way to reproduce (with any backend):
```jl
scatter([1,2,3], [0,2,1], ylims=(0,Inf), legend=:none)
```

Edit: fwiw, this seems to work: `ylims!(Plots.widen(ylims()...))` (unless you're using an logarithmic scale)

fredcallaway:
Previously, the limits were not widened when passing a tuple or
:round to lims (ignoring the widen argument). Fixes #3339

### DIFF
--- a/src/axes.jl
+++ b/src/axes.jl
@@ -504,6 +504,11 @@ end
 const _widen_seriestypes = (:line, :path, :steppre, :stepmid, :steppost, :sticks, :scatter, :barbins, :barhist, :histogram, :scatterbins, :scatterhist, :stepbins, :stephist, :bins2d, :histogram2d, :bar, :shape, :path3d, :scatter3d)
 
 function default_should_widen(axis::Axis)
+    if axis[:widen] isa Bool
+        return axis[:widen]
+    end
+    # automatic behavior: widen if limits aren't specified and series type is appropriate
+    (is_2tuple(axis[:lims]) || axis[:lims] == :round) && return false
     for sp in axis.sps
         for series in series_list(sp)
             if series.plotattributes[:seriestype] in _widen_seriestypes
@@ -559,7 +564,7 @@ function axis_limits(sp, letter, should_widen = default_should_widen(sp[Symbol(l
         else
             amin, amax
         end
-    elseif should_widen && axis[:widen]
+    elseif should_widen
         widen(amin, amax, axis[:scale])
     elseif lims == :round
         round_limits(amin,amax)
--- a/src/axes.jl
+++ b/src/axes.jl
@@ -504,17 +504,14 @@ end
 const _widen_seriestypes = (:line, :path, :steppre, :stepmid, :steppost, :sticks, :scatter, :barbins, :barhist, :histogram, :scatterbins, :scatterhist, :stepbins, :stephist, :bins2d, :histogram2d, :bar, :shape, :path3d, :scatter3d)
 
 function default_should_widen(axis::Axis)
-    should_widen = false
-    if !(is_2tuple(axis[:lims]) || axis[:lims] == :round)
-        for sp in axis.sps
-            for series in series_list(sp)
-                if series.plotattributes[:seriestype] in _widen_seriestypes
-                    should_widen = true
-                end
+    for sp in axis.sps
+        for series in series_list(sp)
+            if series.plotattributes[:seriestype] in _widen_seriestypes
+                return true
             end
         end
     end
-    should_widen
+    false
 end
 
 function round_limits(amin,amax)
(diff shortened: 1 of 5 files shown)


src/axes.jl
```diff
@@ -504,17 +504,14 @@ end
 const _widen_seriestypes = (:line, :path, :steppre, :stepmid, :steppost, :sticks, :scatter, :barbins, :barhist, :histogram, :scatterbins, :scatterhist, :stepbins, :stephist, :bins2d, :histogram2d, :bar, :shape, :path3d, :scatter3d)
 
 function default_should_widen(axis::Axis)
-    should_widen = false
-    if !(is_2tuple(axis[:lims]) || axis[:lims] == :round)
```
BeastyBlacksmith, line 508: I do think that we want to keep `:round` having precedence here

fredcallaway: Really? I think the result of `ylim=(0,1)` and `ylim=:round` should be the same if the data fall within 0 < y < 1. I would rather take the widen=:force route than to have different behavior for different ways of specifying the limits.

FWIW, `ylim=:round` has no effect for me (on gr but I don't think it should matter), i.e. the axis just ends wherever the data ends.

fredcallaway:
Note: I will fix the tests when we agree on the expected behavior.

src/axes.jl
BeastyBlacksmith: Here is one example for `round` vs `widen`
```julia
julia> plot([0.1,0.5,3.0,3.6], rand(4)) |> Plots.xlims
(-0.0049999999999999906, 3.705)

julia> plot([0.1,0.5,3.0,3.6], rand(4), xlims = :round) |> Plots.xlims
(0.0, 4.0)
```

test/test_axes.jl
```diff
@@ -35,6 +35,18 @@ end
 @testset "Axis limits" begin
     pl = plot(1:5, xlims=:symmetric, widen = false)
     @test Plots.xlims(pl) == (-5, 5)
+    
+    pl = plot(1:3)
+    @test Plots.xlims(pl) == Plots.widen(1,3)
+    
+    pl = plot(1:3, xlims=:round)
```
BeastyBlacksmith, line 42:
```suggestion
    pl = plot(0.5:0.5:2.5, xlims=:round)
```

test/test_axes.jl
```diff
@@ -35,6 +35,18 @@ end
 @testset "Axis limits" begin
     pl = plot(1:5, xlims=:symmetric, widen = false)
     @test Plots.xlims(pl) == (-5, 5)
+    
+    pl = plot(1:3)
+    @test Plots.xlims(pl) == Plots.widen(1,3)
+    
+    pl = plot(0.5:0.5:2.5, xlims=:round)
+    @test Plots.xlims(pl) == (1, 3)
```
BeastyBlacksmith, line 43: I did that wrong, sorry
```suggestion
    pl = plot([1.05,2.0,2.95], ylims=:round)
    @test Plots.ylims(pl) == (1, 3)
```

BeastyBlacksmith:
Also make sure to add your data to our `.zenodo.json` file to get proper credit!

test/test_axes.jl
fredcallaway: Slipped by me as well!

fredcallaway:
Sorry for the merge commit, forgot to rebase. Should I rewrite?

BeastyBlacksmith:
I'll squash anyways no need to bother

BeastyBlacksmith:
Thank you for this contribution!

Issue #2071, opened 2019-06-24
twinx() not working with PGFPlots


dancsi:
Hello, 
It seems to me that PGFPlots gives an incorrect output when trying to share the same x axis between two plots with `twinx()`. Here is the sample code
```julia
using Plots
pgfplots()
plot(1:10); plot!(twinx(), -(1:10))
```
Here is the output of PGFPlots:
![pgfplots](https://user-images.githubusercontent.com/405156/60018588-109fb280-968c-11e9-8372-3c7dcc0222f3.png)
Here is the generated .tex: [pgfplot.txt](https://github.com/JuliaPlots/Plots.jl/files/3320775/pgfplot.txt)
For reference, here is the correct output from GR:
![gr](https://user-images.githubusercontent.com/405156/60018773-75f3a380-968c-11e9-9533-23846242380b.png)

daschw:
That's messed up. Thanks for reporting!

tzxiong:
I have experienced the same problem today. `twinx()` works fine for `pyplot()` and `gr()` backends, but will black-out the entire background under pgfplots.

BeastyBlacksmith:
This does not become black, but still is messed up in the `pgfplotsx`-backend

pjentsch0:
Any workarounds for this?

matthieugomez:
I'd be interested in a solution too! The `twinx()` issue seems to be checked in the PGFPlotsX meta-issue, but it's still not working for me.

IljaK91:
I have the same problem with the x-axis being not identical for the second plot. Would greatly appreciate a fix :)

PythonNut:
> Any workarounds for this?

I know this is terrible but in the meantime, I just save the `tex` and edit it by hand to make the axes match.

BeastyBlacksmith:
The relevant part for layouting is here
https://github.com/JuliaPlots/Plots.jl/blob/57ba6aa5a99f595db153faecc576b652d58329ca/src/backends/pgfplotsx.jl#L120-L185

bhc1010:
Is there a reason why the right margin of the subplot is being set to `max(sp[:right_margin], 30px)` and not `max(sp[:right_margin], 0px)` ? Changing this fixed the problem for me, but I'm not sure if that creates other issues.

https://github.com/JuliaPlots/Plots.jl/blob/57ba6aa5a99f595db153faecc576b652d58329ca/src/layouts.jl#L776-L787

[...]

PR #3564, opened 2021-06-10
Better twinx

BeastyBlacksmith:
This solves https://github.com/JuliaPlots/Plots.jl/issues/2071 for pgfplotsx, but not for plotly.
GR seems unaffected by this.

Still we'd need a seperate mechanism for sharing legends or colorbars.


rvignoloccm:
is there a workaround for plotly? I would like to use a right y-axis using plotly

PR #3567, opened 2021-06-10
show extrakwargs on `show`

BeastyBlacksmith:
Sometimes accidently mistyped variables get interpreted as extra_kwargs.

This PR would allow to discover that by using `show` on the Plot object.

```julia
julia> plot(1:5, margines = 2Plots.cm) |> show
Plot{Plots.GRBackend() n=1}
Captured extra kwargs:
  Series{1}:
    margines: 20.0mm
```

### DIFF
--- a/src/plot.jl
+++ b/src/plot.jl
@@ -23,7 +23,38 @@ current(plot::AbstractPlot) = (CURRENT_PLOT.nullableplot = plot)
 
 Base.string(plt::Plot) = "Plot{$(plt.backend) n=$(plt.n)}"
 Base.print(io::IO, plt::Plot) = print(io, string(plt))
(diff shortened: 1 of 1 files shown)


isentropic:
Does this mean that the captured extrakws would be displayed with a typical repl plotting?

isentropic:
If so, I love this pr. More often than not I make a typo and wonder whys something off.

isentropic:
Perhaps we can even run a word distance with all supported args and print possible fixes to the typos

BeastyBlacksmith:
Normal Repl Exekution calls the 3 Argument show. So no, you'd get it for vectors of Plots though.
But of something seems off you could call show to check

daschw (approved):

PR #3570, opened 2021-06-11
change input to `optimize_ticks` for log scales

daschw:
Alternative implementation to @t-bltg's awesome initiative in https://github.com/JuliaPlots/Plots.jl/pull/3547 based on @isentropic's suggestion in https://github.com/JuliaPlots/Plots.jl/pull/3547#issuecomment-857329723
```julia
using Plots; gr()
x = y = Float64[1]
for i ∈ range(-5, 6, step=1)
  append!(x, 2 * 10.0^i:10.0^i:10.0^(i + 1) |> collect)
end
plot(
  x, y, axis=:log10, minorgrid=true, legend=false,
  minorticks=true, shape=:cross, ms=2, lw=.5, dpi=300
)
```
test

### DIFF
--- a/src/axes.jl
+++ b/src/axes.jl
@@ -174,6 +174,11 @@ function optimal_ticks_and_labels(ticks, alims, scale, formatter)
             sf(amax);
             k_min = 4, # minimum number of ticks
             k_max = 8, # maximum number of ticks
+            Q = if scale in _logScales
+                [(1.0,1.0), (5.0, 0.9), (2.0, 0.7)]
+            else
+                [(1.0,1.0), (5.0, 0.9), (2.0, 0.7), (2.5, 0.5), (3.0, 0.2)]
+            end,
         )[1]
     elseif typeof(ticks) <: Int
         scaled_ticks, viewmin, viewmax = optimize_ticks(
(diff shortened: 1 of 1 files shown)


isentropic:
This looks good and without hacks I like it

daschw:
Looking at the plot above I thought that for log scales multiples of three would be the better choices for ticks (representing ..., nano, micro, kilo, mega, ...). So I played around locally a little bit and found that this is not true if the minimum and maximum power are close to each other (e.g. `(10^0, 10^0.5, 10^1)` is probably better than `(10^0, 10^0.3, 10^0.6, 10^0.9)`). Hence, for log scales I added a check if the scale is `:log10` and the difference in the powers is sufficiently high to decide if we want to consider multiples of 3 in the tick optimization. With the latest changes, the example above results in
test

BeastyBlacksmith:
what I liked about #3547 is that you still got minor ticks for the omitted powers, which doesn't let the plot look so sparse

daschw:
That's a good point, I'm open for either implementation, I just wanted to check here how we could use the existing functionality in optimize_ticks.

t-bltg:
Hi @daschw ,
Thank for the proposition. However, I think that the minor ticks here are not consistent with a `log10` scale, please see https://github.com/JuliaPlots/Plots.jl/pull/3547#issuecomment-860622934

Issue #696, opened 2017-02-22
y axis inside margins are too small when `y_scale = :log10`


mkborregaard:
It seems that the mechanism padding the lower and upper values of the y axis is thrown off when y_scale is logged:

![skaermbillede 2017-02-22 kl 10 59 41](https://cloud.githubusercontent.com/assets/8429802/23206477/0c08f7f6-f8ee-11e6-9e05-80f4d5f64f20.png)

Issue #458, opened 2016-08-26
Alternative ticks for log-plots and minor tick marks


mfalt:
It it possible in the current framework to show minor ticks in log plots as 2_10^x, 3_10^x .. 9*10^x as is common in for example bode plots:
![bode](https://upload.wikimedia.org/wikipedia/commons/f/fd/Bode-pt2.png)
or is there a simple way to add it in this package or in the user code?
I was able to create something similar with the following (ugly) code

```
function getLogTicks(x)
    min = ceil(log10(minimum(x)))
    max = floor(log10(maximum(x)))
    major = 10.^collect(min:max)
    majorText = ["\$10^{$(round(Int64,i))}\$" for i=min:max]
    minor = [j*10^i for i=(min-1):(max+1) for j=2:9]
    minor = minor[find(minimum(x) .<= minor .<= maximum(x))]
    ([major; minor], [majorText; fill("", length(minor))])
end
```

but I can't find a command for having different styles on the different tick lines. And when using these ticks with something like
`Plots.plot!(x, y, xscale=:log10, ticks=getLogTicks(x))`
the old ticks are overwritten when adding more plots.

tbreloff:
The quick answer is: no, you'll need to do some hacking like you're already
doing. But you might improve the hack by using the 'formatter' attribute
and some trickery that returns an alternative format if the value is not
approximately a major tick mark.

It also might make sense to let 'ticks' accept a function mapping axis
limits to a list of ticks, so you could compute them dynamically based on
the current axis limits. This should be easy.

mfalt:
Thanks, I will look into the formatter function. Although I think accepting a function to `ticks` is the best way to go. This would enable showing text on my "minor" tick in case the range is very small for example, or adding even more "minor" ticks based on the range. I would definitely write such a function for my toolbox and would be happy to write one for Plots.jl too if you have interest in supporting some natively in Plots.jl.

mfalt:
A question related to the formatter, it seems like not all back-ends support latex in the tick text, is there a good way of getting around this, either by a common formatting rule, or formatting in different ways for different back-ends?

tbreloff:
I'm open to suggestions on generic latex handling.

mkborregaard:
@mfalt currently Latex formatting is done with the package LaTeXStrings. Can this be closed?

mfalt:
[...]

PR #3577, opened 2021-06-18
GR: rework automatic major/minor ticks for log scales

t-bltg:
Superseding https://github.com/JuliaPlots/Plots.jl/pull/3547, required patch: https://github.com/JuliaPlots/PlotUtils.jl/pull/115.

Test cases:
```julia
using Plots; gr()

for (i, (start, stop)) ∈ enumerate([(-1, 2), (-6, 5)])
    x = y = Float64[1]
    for j ∈ start:stop
      append!(x, 2 * 10.0^j:10.0^j:10.0^(j + 1) |> collect)
    end
    plot(
      x, y, axis=:log10, minorgrid=true, legend=:none,
      shape=:cross, ms=2, lw=.5, dpi=200
    )
    png("loglog-$i")
end
```

Output:
loglog-1
loglog-2

Fix https://github.com/JuliaPlots/Plots.jl/issues/3318
Fix https://github.com/JuliaPlots/Plots.jl/issues/696
Fix https://github.com/JuliaPlots/Plots.jl/issues/458

### DIFF
--- a/src/axes.jl
+++ b/src/axes.jl
@@ -172,8 +172,9 @@ function optimal_ticks_and_labels(ticks, alims, scale, formatter)
         scaled_ticks = optimize_ticks(
             sf(amin),
             sf(amax);
-            k_min = 4, # minimum number of ticks
+            k_min = scale in _logScales ? 2 : 4, # minimum number of ticks
             k_max = 8, # maximum number of ticks
+            scale = scale,
         )[1]
     elseif typeof(ticks) <: Int
         scaled_ticks, viewmin, viewmax = optimize_ticks(
@@ -185,6 +186,7 @@ function optimal_ticks_and_labels(ticks, alims, scale, formatter)
             # `strict_span = false` rewards cases where the span of the
             # chosen  ticks is not too much bigger than amin - amax:
             strict_span = false,
+            scale = scale,
         )
         # axis[:lims] = map(RecipesPipeline.inverse_scale_func(scale), (viewmin, viewmax))
     else
@@ -336,10 +338,11 @@ function get_minor_ticks(sp, axis, ticks)
     length(ticks) < 2 && return nothing
 
     amin, amax = axis_limits(sp, axis[:letter])
-    #Add one phantom tick either side of the ticks to ensure minor ticks extend to the axis limits
+    scale = axis[:scale]
+    # Add one phantom tick either side of the ticks to ensure minor ticks extend to the axis limits
     if length(ticks) > 2
         ratio = (ticks[3] - ticks[2])/(ticks[2] - ticks[1])
-    elseif axis[:scale] in (:none, :identity)
+    elseif scale ∈ (:none, :identity)
         ratio = 1
     else
         return nothing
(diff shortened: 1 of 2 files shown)


t-bltg:
@BeastyBlacksmith, pinging you one this one for review / merge, if we could move forward that'll be great ;)

t-bltg:
Fixed the only minor issue with ref1:
aaa

t-bltg:
Ok, so after merging https://github.com/JuliaPlots/PlotUtils.jl/pull/115 and triggering a new release of `PlotUtils`, the only failing test is `ref5` as expected.

@BeastyBlacksmith or @daschw if I could have your approval, we can merge https://github.com/JuliaPlots/PlotReferenceImages.jl/pull/97 and this PR.

BeastyBlacksmith:
how did ref5 look like with minmal number of ticks = 4?

t-bltg:
> how did ref5 look like with minmal number of ticks = 4?

```bash
$ julia ref5.jl
┌ Warning: No strict ticks found
└ @ PlotUtils [...]/PlotUtils/n4I4Q/src/ticks.jl:295
┌ Warning: No strict ticks found
└ @ PlotUtils [...]/PlotUtils/n4I4Q/src/ticks.jl:295
```

ref5

BeastyBlacksmith:
I suppose this is because it tries to find integer exponents. Can we have some kind of look for integer exponents -> if not enough ticks present look for mulitples of 1/2 -> if not enough ticks found look for multiples of 1/4 etc. ?

t-bltg:
> I suppose this is because it tries to find integer exponents. Can we have some kind of look for integer exponents -> if not enough ticks present look for mulitples of 1/2 -> if not enough ticks found look for multiples of 1/4 etc. ?

What's wrong with requiring less major ticks ? It is justified by the choice of log scaling.

Non-integer powers do not bring information to the plot or the user: `10^-.5` ? you need a calculator, I am not in favor of it.

BeastyBlacksmith:
I think the plot looks a bit lost with only two ticks. But surely that is subjective, probably would look okay with minorticks. Overall I am okay with the changes

BeastyBlacksmith (approved):

t-bltg:
> I think the plot looks a bit lost with only two ticks. But surely that is subjective, probably would look okay with minorticks. Overall I am okay with the changes

It is subjective, see also the examples in https://fr.mathworks.com/help/matlab/ref/loglog.html.

Don't count on minor-ticks there, we need at least 3 major ticks because of this https://github.com/JuliaPlots/Plots.jl/blob/6f4400411ee68ade1f1d916a5e70be58d628e894/src/axes.jl#L340.

t-bltg:
@BeastyBlacksmith 

Ok this has be bothering me not to be able to have minogrids if we have 2 major ticks, I dislike un-finished things.
I reworked the logic and here is the result:
bbb

Note that we have `3` major ticks not because of `optimize_ticks` in PlotsUtils which returned `2` but because of the phantom ticks added in `get_minor_ticks`.

Would that be ok ?

PR #3586, opened 2021-06-25
Simplify major / minor grid segments

t-bltg:
As per https://github.com/JuliaPlots/Plots.jl/blob/master/src/axes.jl#L844:
>         # TODO this can be simplified, we do almost the same thing twice for grid and minorgrid

Simplify duplicate code for 2d and 3d.

### DIFF
--- a/src/axes.jl
+++ b/src/axes.jl
@@ -703,61 +703,46 @@ function axis_drawing_info(sp, letter)
                 )
             end
         end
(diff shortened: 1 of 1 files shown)


t-bltg:
@BeastyBlacksmith, if you could double check that I did not break anything, this should reduce lots of duplicate code ...

Issue #3581, opened 2021-06-22
[BUG] rotated label are not at correct position


ValentinKaisermayer:
## Details
If the xtick labels are not rotated the appear centered at the tick marks.
```julia
import Plots

Plots.plot(xrotation=0, dpi=100, minorticks=false)
Plots.scatter!(rand(10), rand(10))
```
![test](https://user-images.githubusercontent.com/50108075/122873301-e09a5100-d331-11eb-9f90-794a361e248a.png)

However, if they are rotated the are not centered.
```julia
import Plots

Plots.plot(xrotation=90, dpi=100, minorticks=false)
Plots.scatter!(rand(10), rand(10))
```
![test](https://user-images.githubusercontent.com/50108075/122873204-c52f4600-d331-11eb-9a2c-82bcb538d418.png)

### Backends

This bug occurs on ( insert `x` below )

Backend      | yes | no  | untested
-------------|-----|-----|---------
gr (default) |   x  |     |
pyplot       |     |   x  |
plotly       |     |  x   | 
plotlyjs     |     |     | x
pgfplotsx    |     |     | x
inspectdr    |     |     | x

### Versions

Plots.jl version: 1.16.5
Backend  version (`]st -m <backend(s)>`):  
[28b8d3ca] GR v0.57.5
[d330b81b] PyPlot v2.9.0
Output of `versioninfo()`:
```julia
julia> versioninfo()
Julia Version 1.6.1
Commit 6aaedecc44 (2021-04-23 05:59 UTC)
Platform Info:
  OS: Windows (x86_64-w64-mingw32)
  CPU: Intel(R) Core(TM) i5-6200U CPU @ 2.30GHz
  WORD_SIZE: 64
  LIBM: libopenlibm
  LLVM: libLLVM-11.0.1 (ORCJIT, skylake)
Environment:
  JULIA_EDITOR = code
  JULIA_NUM_THREADS =
```

BeastyBlacksmith:
link to the PR that implemented this: https://github.com/JuliaPlots/Plots.jl/pull/3529

t-bltg:
> link to the PR that implemented this: #3529

no, #3529 is not responsible here

BeastyBlacksmith:
Sorry I mixed that up

PR #3587, opened 2021-06-28
GR: fix alignment of ticks labels when rotating

t-bltg:
Fixes https://github.com/JuliaPlots/Plots.jl/issues/3581.

Reproducer:
```julia
using Plots; gr()

for r in (0, 20, 70, 90)
    ttl = "bug-rot-xr=$r-yr=$r"
    @show ttl
    plot(xrotation=r, yrotation=r, dpi=100, minorticks=false, title=ttl)
    scatter!(rand(10), rand(10))
    png(ttl)
  end
```

Without PR
---------------
out

With PR
-----------
out_PR

### DIFF
--- a/src/backends/gr.jl
+++ b/src/backends/gr.jl
@@ -651,12 +651,17 @@ end
 
 function gr_set_tickfont(sp, letter)
     axis = sp[Symbol(letter, :axis)]
+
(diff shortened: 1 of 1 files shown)


t-bltg:
Thanks for merging :+1:  !

BeastyBlacksmith:
The tests already did that

Issue #3445, opened 2021-04-20
[BUG] camera on GR doesn't work properly


hurricane007:
seems I hit the same issue as befor : https://github.com/JuliaPlots/Plots.jl/issues/1911

## Details
camera doesn't work properly in GR. 
![image](https://user-images.githubusercontent.com/16286632/115284948-96d49500-a14d-11eb-80a3-4399e62e830b.png)
![image](https://user-images.githubusercontent.com/16286632/115284989-a6ec7480-a14d-11eb-8823-b7e8e03a9295.png)

### Backends

This bug occurs on ( insert `x` below )

Backend      | yes | no  | untested
-------------|-----|-----|---------
gr (default) |   x  |     |
pyplot       |     |   x   |
plotly       |     |  x   |
plotlyjs     |     |     |x
pgfplotsx    |     |     |x
inspectdr    |     |     |x

### Versions

Plots.jl version:  [91a5bcdd] Plots v1.11.2
Backend  version (`]st -m`):  GR v0.57.4
Output of `versioninfo()`:
```
Julia Version 1.6.0
Commit f9720dc2eb (2021-03-24 12:55 UTC)
Platform Info:
  OS: Windows (x86_64-w64-mingw32)
  CPU: Intel(R) Core(TM) i7-8665U CPU @ 1.90GHz
  WORD_SIZE: 64
  LIBM: libopenlibm
  LLVM: libLLVM-11.0.1 (ORCJIT, skylake)
```

Issue #1911, opened 2019-02-03
GR: camera angles outside [0,90] do not work


yha:
Camera angles outside of the range 0..90 are not plotted correctly.
If it's the first plot, the z axis scaling is wrong:
```
using Plots
wireframe( 0:10, 0:10, (x,y)->x*y, camera=(130,20) )
```
![image](https://user-images.githubusercontent.com/4170948/52184253-81e3e380-2819-11e9-905d-2fe49c5b2f8e.png)

If it's not the first plot, the last used camera angle is repeated:
```
plts = [wireframe( 0:10, 0:10, (x,y)->x*y, camera=(x,20), title="angle=$x" ) 
        for x in [30,120,60,120,30,-30]]
plot(plts...)
```
![image](https://user-images.githubusercontent.com/4170948/52184262-b2c41880-2819-11e9-8623-5abd2da2c53b.png)

The problem also occurs with the vertical angle, though I suppose angles outside of 0..90 are less useful there.

mkborregaard:
Thanks for reporting! I had noticed this before but embarassingly it seems as if I didn't open an issue. This also means that the "simple is beautiful" gif on the front page of Plots' docs doesn't work with GR!

aramirezreyes:
I think this still happens

RodolfoFigueroa:
Can confirm it still happens.

BoundaryValueProblems:
Yes, this problem has not been resolved. On the other hand, it works under the pyplot backend.

Issue #3541, opened 2021-06-04
camera keyword not working on pgfplotsx (and not properly on GR)


floswald:
## Details

```julia
using Plots
ϕs = range(-π, π, length = 50)
θs = range(0, π, length = 25)
θqs = range(1, π - 1, length = 25)
x = vec([sin(θ) * cos(ϕ) for (ϕ, θ) = Iterators.product(ϕs, θs)])
y = vec([sin(θ) * sin(ϕ) for (ϕ, θ) = Iterators.product(ϕs, θs)])
z = vec([cos(θ) for (ϕ, θ) = Iterators.product(ϕs, θs)])
u = 0.1 * vec([sin(θ) * cos(ϕ) for (ϕ, θ) = Iterators.product(ϕs, θqs)])
v = 0.1 * vec([sin(θ) * sin(ϕ) for (ϕ, θ) = Iterators.product(ϕs, θqs)])
w = 0.1 * vec([cos(θ) for (ϕ, θ) = Iterators.product(ϕs, θqs)])
quiver(x, y, z, quiver = (u, v, w))
```

<img width="715" alt="Screenshot 2021-06-04 at 12 29 49" src="https://user-images.githubusercontent.com/1021558/120788245-9000b680-c530-11eb-95f8-264ec38eeb6a.png">

### GR: camera only works in [1-90]

I cannot look at plot from _behind_. poor example (looks same from all sides), but quite common to look at surfaces which need to be looked at from a certain angle.

```
julia> quiver(x, y, z, quiver = (u, v, w), camera = (80,10))
```

<img width="706" alt="Screenshot 2021-06-04 at 12 30 26" src="https://user-images.githubusercontent.com/1021558/120788455-a6a70d80-c530-11eb-88a0-cfb4a5e81467.png">

```
julia> quiver(x, y, z, quiver = (u, v, w), camera = (130,10))
```

picture does not change. same result for `camera = (-10,10)`.

## PGFPlotsX `camera` does not work

```
julia> pgfplotsx()
Plots.PGFPlotsXBackend()

julia> quiver(x, y, z, quiver = (u, v, w))
```

<img width="673" alt="Screenshot 2021-06-04 at 12 32 57" src="https://user-images.githubusercontent.com/1021558/120788771-000f3c80-c531-11eb-9ee8-a88b30a94fd0.png">

```
julia> quiver(x, y, z, quiver = (u, v, w), camera = (80,10))
```

same image. also for negative values.

### Backends

[...]

PR #3590, opened 2021-07-01
GR: 3D, warn on invalid rotation or tilt

t-bltg:
Invalid rotation or tilt are currently silently ignored in GR, I believe the user should be warned.

This is a wontfix from `Plots.jl`'s side, users should open an issue or PR in https://github.com/sciapp/gr for supporting `θ < 0` or `θ > 90`.

See https://gr-framework.org/julia-gr.html#GR.setspace-0880646251e9db0f5d16f4369795d314:
> Angles of rotation and viewing angle must be specified between 0° and 90°.

Linked https://github.com/JuliaPlots/Plots.jl/issues/3541.
Fixes https://github.com/JuliaPlots/Plots.jl/issues/3445.
Fixes https://github.com/JuliaPlots/Plots.jl/issues/1911.

### DIFF
--- a/src/backends/gr.jl
+++ b/src/backends/gr.jl
@@ -1326,7 +1326,13 @@ function gr_draw_axes(sp, viewport_plotarea)
         # set space
         xmin, xmax, ymin, ymax = gr_xy_axislims(sp)
         zmin, zmax = gr_z_axislims(sp)
-        GR.setspace(zmin, zmax, round.(Int, sp[:camera])...)
+
+        camera = round.(Int, sp[:camera])
+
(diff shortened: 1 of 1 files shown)


BeastyBlacksmith:
what if we project the value in the feasible interval?

t-bltg:
> what if we project the value in the feasible interval?

We could, that would require flipping & mirroring logic, which I suspect to be :exploding_head: .
Maybe I'll do that when I'll have time :wink:, although this is low right now on my todo list.
In the mean time, with this PR, we close 2 issues and we avoid potential piling up new issues on that matter.

daschw (approved):
I think it's fine to have this warning. We can think about projecting in a separate PR.

hurricane007:
so basically this "issue" is due to the limitation of GR backend

PR #3591, opened 2021-07-01
GR: expose surface(...) display options, fix hardcoded constants

t-bltg:
This PR add:
1. Expose the options described in [GR surface's table](https://gr-framework.org/julia-gr.html#GR.surface-e3e6f234cc6cd4713b8727c874a5f331);
2. Add the possibility to override hardcoded `nx` and `ny` when re-griding.

NOTES:
- name used: `:display_option` is arbitrary, maybe there is a better naming ?
- using `series[:extra_kwargs]` seems appropriate since this is specific to the GR backend
- should we document this or not ?

mwe
------
```julia
using Plots; gr()
import GR

main() = begin
  meshgrid(x, y) = (ones(eltype(y), length(y)) * x', y * ones(eltype(x), length(x))')

  x, y = meshgrid(-2:.25:2, -2:.25:2)
  f = x .* exp.(-x.^2 - y.^2)

  surface(x[:], y[:], f[:], display_option=GR.OPTION_SHADED_MESH, nx=50, ny=50, legend=:none, c=:viridis)
  return
end

main()
```

without PR
--------------
pr2_ex-WITHOUT

with PR
----------
pr2_ex-WITH

### DIFF
--- a/src/backends/gr.jl
+++ b/src/backends/gr.jl
@@ -1845,8 +1845,10 @@ function gr_draw_surface(series, x, y, z, clims)
     if series[:seriestype] === :surface
         fillalpha = get_fillalpha(series)
         fillcolor = get_fillcolor(series)
-        if length(x) == length(y) == length(z)
-            x, y, z = GR.gridit(x, y, z, 200, 200)
+        # NOTE: setting nx = 0 or ny = 0 disables GR.gridit interpolation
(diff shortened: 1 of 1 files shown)


BeastyBlacksmith:
Thats all fine and should definitely be documented in the GR backend section.

Issue #3558, opened 2021-06-07
[FR] control padding inside legend


ptoche:
Feature request: add the option to increase padding inside legend, particularly to avoid labels overlap. Fine control over horizontal/vertical spacing would be even better, but a general padding argument would probably cover most cases.

The issue is described [here](https://discourse.julialang.org/t/increase-space-between-legend-labels-in-plots/62474/3) with a workaround. Code copied for convenience. 
```julia
using LaTeXStrings, Plots; gr()

y1, y2 = rand(100), rand(100)
histogram(y1, ylims = (0,50), label = L"\left\Vert\beta_{1}^{2}-\beta_{2}^{2}\right\Vert")
histogram!(y2, label = L"\left\Vert\beta_{2}^{2}-\beta_{3}^{2}\right\Vert", legendfontsize=10)
```

<img width="671" alt="a" src="https://user-images.githubusercontent.com/6339274/121086053-8a58da00-c797-11eb-82df-2e56a22c99de.png">

![b](https://user-images.githubusercontent.com/6339274/121086066-8e84f780-c797-11eb-9b58-e3712b840fd3.png)

PR #3598, opened 2021-07-02
Proposed fix for Latex sub/super-scripts in legends

t-bltg:
Fixes https://github.com/JuliaPlots/Plots.jl/issues/3558:
1. Fix height computation by retaining the maximum legend height, for each serie;
2. Allow changing the height factor via the `extra_kwargs` mecanism.

Discussion:
Change [1.] does modify some reference images, but I think the fix is justified. Only applying [2.] allows fixing the OP issue, without changing the default legend sizes.

MWE
-------
```julia
using LaTeXStrings
using Plots; gr()

main() = begin
  y1, y2 = rand(100), rand(100)
  histogram(y1, ylims=(0, 50), label=L"\left\Vert\beta_{1}^{2}-\beta_{2}^{2}\right\Vert", extra_kwargs=Dict(:subplot=>Dict(:legend_hfactor=>1.25)))
  histogram!(y2, label=L"\left\Vert\beta_{2}^{2}-\beta_{3}^{2}\right\Vert", legendfontsize=10)
  png("foo")
end

main()
```

With PR
-----------
Without using `extra_kwargs`:
pr3

Using `:legend_hfactor=>1.25`:
pr3

Without PR
---------------
pr3

Default behavior modification (example # 23)
==================================

With PR
-----------
ref23

Without PR
---------------
ref23_old

### DIFF
--- a/src/backends/gr.jl
+++ b/src/backends/gr.jl
@@ -1162,25 +1162,25 @@ function gr_legend_pos(theta::Real, leg, viewport_plotarea; axisclearance=nothin
 end
 
 function gr_get_legend_geometry(viewport_plotarea, sp)
-    legendn = 0
-    legendw = 0
+    legendn = legendw = 0; dy = 0.
     if sp[:legend] != :none
         GR.savestate()
         GR.selntran(0)
         GR.setscale(0)
         if sp[:legendtitle] !== nothing
             gr_set_font(legendtitlefont(sp), sp)
+            legendn += 1
             tbx, tby = gr_inqtext(0, 0, string(sp[:legendtitle]))
             legendw = tbx[3] - tbx[1]
-            legendn += 1
+            dy = tby[3] - tby[1]
         end
         gr_set_font(legendfont(sp), sp)
         for series in series_list(sp)
             should_add_to_legend(series) || continue
             legendn += 1
-            lab = series[:label]
-            tbx, tby = gr_inqtext(0, 0, string(lab))
+            tbx, tby = gr_inqtext(0, 0, string(series[:label]))
             legendw = max(legendw, tbx[3] - tbx[1]) # Holds text width right now
+            dy = max(dy, tby[3] - tby[1])
         end
 
         GR.setscale(1)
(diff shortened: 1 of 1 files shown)


daschw (approved):
Looks great!

t-bltg:
@ptoche, is this approach ok for you ?

src/backends/gr.jl
```diff
@@ -1162,25 +1162,25 @@ function gr_legend_pos(theta::Real, leg, viewport_plotarea; axisclearance=nothin
 end
 
 function gr_get_legend_geometry(viewport_plotarea, sp)
-    legendn = 0
-    legendw = 0
+    legendn = legendw = 0; dy = 0.
```
BeastyBlacksmith, line 1165:
```suggestion
    legendn = legendw = 0; dy = 0
```

src/backends/gr.jl
```diff
@@ -1197,7 +1197,8 @@ function gr_get_legend_geometry(viewport_plotarea, sp)
     x_legend_offset = (viewport_plotarea[2] - viewport_plotarea[1]) / 30
     y_legend_offset = (viewport_plotarea[4] - viewport_plotarea[3]) / 30
 
-    dy = gr_point_mult(sp) * sp[:legendfontsize] * 1.75
+    dy *= get(sp[:extra_kwargs], :legend_hfactor, 1.)
```
BeastyBlacksmith, line 1200:
```suggestion
    dy *= get(sp[:extra_kwargs], :legend_hfactor, 1)
```
If these don't have to be floats for a good reason, integer literals are preferred.

ptoche:
This looks great, thanks for putting this together! :-)

t-bltg:
You can use [`legend_hfactor`](https://github.com/t-bltg/Plots.jl/blob/293b43013bd0fd9ac022c706025b877c86a9ca5a/src/backends/gr.jl#L1200) > 1 to fine tune the spacing.

lmiq:
That does not seem to have any effect on my plot. Am I doing something wrong?

image

Issue #3594, opened 2021-07-02
[BUG] GR error message when y=0 and yaxis=:log10


mortenpi:
The following snippet

```julia
plot(
    1:5, [1,2,0,4,5],
    m = [:utriangle, :dtriangle, :utriangle, :dtriangle, :dtriangle],
    yaxis = :log10, ylims = (1e-1, 1e1),
)
```

will produce a warning in the terminal when rendering with GR:

```
GKS: Number of points is invalid in routine POLYLINE
```

The plot is fine (with a missing data point at `x = 3`, where the `y` value is `0` and therefore can't be displayed on a log scale). At a glance, looks similar to #2735, but I think it's unrelated to that.

### Backends

This bug occurs on ( insert `x` below )

Backend      | yes | no  | untested
-------------|-----|-----|---------
gr (default) |  x   |     |
pyplot       |     |  x   |
plotly       |     |     | x
plotlyjs     |     |     | x
pgfplotsx    |     |     | x
inspectdr    |     |     | x

### Versions

I tested the latest master for both Plots and GR, but it is also present in recent releases. Full MWE environment [in this gist](https://gist.github.com/mortenpi/e272da48dd248f3ea0891bb4c03f3289).

Plots.jl version: v1.16.7
Backend  version (`]st -m <backend(s)>`): GR v0.57.5
Output of `versioninfo()`:

t-bltg:
Happens also on octave with zero valued data:
```bash
$ octave --eval 'loglog(1:5, [1 2 0 4 5]); ylim([1e-1 1e1]); saveas(1, "a.png")'
warning: axis: omitting non-positive data in log plot
warning: called from
    __plt__>__plt2vv__ at line 500 column 10
    __plt__>__plt2__ at line 247 column 14
    __plt__ at line 112 column 18
    loglog at line 65 column 10
```
![foo](https://user-images.githubusercontent.com/13423344/124275607-0cdc7b80-db43-11eb-9657-da5ae5f3d694.png)

The question is, what should to do with undefined behavior ?

mortenpi:
It's a fair point that this is a case of slightly garbage input. I would make the argument that Plots should check for this, issue its own warning (which would actually say what the problem is, something like "`0 value passed for a log plot, removing points`") and then sanitize the data so that GR wouldn't print any warnings of its own.

[...]

PR #3601, opened 2021-07-03
Warn on invalid value for log scale based series

t-bltg:
With this PR, the example in https://github.com/JuliaPlots/Plots.jl/issues/3594 now shows a warning or a full stacktrace controlled via the `JULIA_DEBUG` environment variable.

@mortenpi, is this what you expect ?

~~If someone has a tip for logging the messages only once, I'm all ears ...~~
==> now warning only once
~~Not sure about the potential performance penalty traversing series though.~~
==> now traversing only once

Fix https://github.com/JuliaPlots/Plots.jl/issues/3594.

```julia
$ julia bug.jl
┌ Warning: Invalid negative or zero value 0.0 found at series index 3 for log10 based yscale
└ @ Plots [...]/Plots/Bfn5f/src/utils.jl:92

$ JULIA_DEBUG=Plots julia bug.jl
┌ Warning: Invalid negative or zero value 0.0 found at series index 3 for log10 based yscale
└ @ Plots [...]/Bfn5f/src/utils.jl:92
┌ Debug: 
│   exception =
│    DomainError with 0.0:
│    
│    Stacktrace:
│      [1] macro expansion
│        @ ./logging.jl:341 [inlined]
│      [2] series_segments(series::Plots.Series, seriestype::Symbol; check::Bool)
│        @ Plots [...]/Bfn5f/src/utils.jl:93
│      [3] pgfx_add_series!(#unused#::Val{:path}, axis::PGFPlotsX.Axis, series_opt::PGFPlotsX.Options, series::Plots.Series, series_func::Type{PGFPlotsX.Plot}, opt::RecipesPipeline.DefaultsDict)
│        @ Plots [...]/Bfn5f/src/backends/pgfplotsx.jl:334
│      [4] (::Plots.PGFPlotsXPlot)(plt::Plots.Plot{Plots.PGFPlotsXBackend})
│        @ Plots [...]/Bfn5f/src/backends/pgfplotsx.jl:287
│      [5] _update_plot_object(plt::Plots.Plot{Plots.PGFPlotsXBackend})
│        @ Plots [...]/Bfn5f/src/backends/pgfplotsx.jl:1415
│      [6] prepare_output(plt::Plots.Plot{Plots.PGFPlotsXBackend})
│        @ Plots [...]/Bfn5f/src/plot.jl:245
│      [7] show
│        @ [...]/Bfn5f/src/output.jl:214 [inlined]
│      [8] png(plt::Plots.Plot{Plots.PGFPlotsXBackend}, fn::String)
│        @ Plots [...]/Bfn5f/src/output.jl:7
│      [9] png(fn::String)
│        @ Plots [...]/Bfn5f/src/output.jl:10
│     [10] main(i::Int64)
│        @ Main [...]/bug.jl:10
│     [11] top-level scope
│        @ [...]/bug.jl:16
│     [12] include(mod::Module, _path::String)
│        @ Base ./Base.jl:386
│     [13] exec_options(opts::Base.JLOptions)
│        @ Base ./client.jl:285
│     [14] _start()
│        @ Base ./client.jl:485
└ @ Plots [...]/Bfn5f/src/utils.jl:93
```

### DIFF
--- a/src/utils.jl
+++ b/src/utils.jl
@@ -75,22 +75,23 @@ function iter_segments(args...)
     NaNSegmentsIterator(tup, n1, n2)
 end
 
-function series_segments(series::Series, seriestype::Symbol = :path)
+function series_segments(series::Series, seriestype::Symbol=:path; check=false)
     x, y, z = series[:x], series[:y], series[:z]
     (x === nothing || isempty(x)) && return UnitRange{Int}[]
 
     args = RecipesPipeline.is3d(series) ? (x, y, z) : (x, y)
     nan_segments = collect(iter_segments(args...))
 
-    scales = :xscale, :yscale, :zscale
-    for (n, s) ∈ enumerate(args)
-        scale = get(series, scales[n], :identity)
-        if scale ∈ _logScales
-            for (i, v) ∈ enumerate(s)
-                if v <= 0
-                    msg = "Invalid negative or zero value $v found at serie index $i for $(scale) based $(scales[n])"
-                    @warn msg
-                    @debug msg exception=(DomainError(v), stacktrace())
+    if check
+        scales = :xscale, :yscale, :zscale
+        for (n, s) ∈ enumerate(args)
+            scale = get(series, scales[n], :identity)
+            if scale ∈ _logScales
+                for (i, v) ∈ enumerate(s)
+                    if v <= 0
+                        @warn "Invalid negative or zero value $v found at series index $i for $(scale) based $(scales[n])"
+                        @debug "" exception=(DomainError(v), stacktrace())
+                    end
                 end
             end
         end
--- a/src/utils.jl
+++ b/src/utils.jl
@@ -82,6 +82,20 @@ function series_segments(series::Series, seriestype::Symbol = :path)
     args = RecipesPipeline.is3d(series) ? (x, y, z) : (x, y)
     nan_segments = collect(iter_segments(args...))
 
+    scales = :xscale, :yscale, :zscale
+    for (n, s) ∈ enumerate(args)
+        scale = get(series, scales[n], :identity)
+        if scale ∈ _logScales
+            for (i, v) ∈ enumerate(s)
+                if v <= 0
+                    msg = "Invalid negative or zero value $v found at serie index $i for $(scale) based $(scales[n])"
+                    @warn msg
+                    @debug msg exception=(DomainError(v), stacktrace())
+                end
+            end
+        end
+    end
+
     segments = if has_attribute_segments(series)
         Iterators.flatten(map(nan_segments) do r
             if seriestype in (:scatter, :scatter3d)
--- a/src/backends/gr.jl
+++ b/src/backends/gr.jl
@@ -355,9 +355,9 @@ end
 
 # draw ONE symbol marker
 function gr_draw_marker(series, xi, yi, clims, i, msize, strokewidth, shape::Symbol)
-    GR.setborderwidth(strokewidth);
-    gr_set_bordercolor(get_markerstrokecolor(series, i));
-    gr_set_markercolor(get_markercolor(series, clims, i));
+    GR.setborderwidth(strokewidth)
+    gr_set_bordercolor(get_markerstrokecolor(series, i))
+    gr_set_markercolor(get_markercolor(series, clims, i))
     gr_set_transparency(get_markeralpha(series, i))
     GR.setmarkertype(gr_markertype(shape))
     GR.setmarkersize(0.3msize / gr_nominal_size(series))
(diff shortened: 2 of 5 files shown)


src/utils.jl
```diff
@@ -82,6 +82,20 @@ function series_segments(series::Series, seriestype::Symbol = :path)
     args = RecipesPipeline.is3d(series) ? (x, y, z) : (x, y)
     nan_segments = collect(iter_segments(args...))
 
+    scales = :xscale, :yscale, :zscale
+    for (n, s) ∈ enumerate(args)
+        scale = get(series, scales[n], :identity)
+        if scale ∈ _logScales
+            for (i, v) ∈ enumerate(s)
+                if v <= 0
+                    msg = "Invalid negative or zero value $v found at serie index $i for $(scale) based $(scales[n])"
```
mortenpi, line 91:
```suggestion
                    msg = "Invalid negative or zero value $v found at series index $i for $(scale) based $(scales[n])"
```

mortenpi (commented):
This is awesome -- thanks for looking into this! I think it's a great UX improvement.

> However, I don't think that we should sanitize the data: it is the responsibility of the user not to feed invalid data (since again this is undefined behavior, and fixing undef is arbitrary and thus subject to debate).

The GR warning still showing up would be my only minor complaint, but I do see your point. And since are printing the bigger warning anyway, I think it's fine.

My only concern would be that passing broken data on to the backend may break it. I do manage to break GR on a regular basis (as in, independent follow-up plots in the same session fail to work render, and I have to either change backend or restart Julia).

On another note, for some reason, for the original MWE it does print the new warning three times:

```
julia> plot(
           1:5, [1,2,0,4,5],
           m = [:utriangle, :dtriangle, :utriangle, :dtriangle, :dtriangle],
           yaxis = :log10, ylims = (1e-1, 1e1),
       )
┌ Warning: Invalid negative or zero value 0.0 found at serie index 3 for log10 based yscale
└ @ Plots ~/.julia/packages/Plots/7T4MM/src/utils.jl:92
GKS: Number of points is invalid in routine POLYLINE
┌ Warning: Invalid negative or zero value 0.0 found at serie index 3 for log10 based yscale
└ @ Plots ~/.julia/packages/Plots/7T4MM/src/utils.jl:92
┌ Warning: Invalid negative or zero value 0.0 found at serie index 3 for log10 based yscale
└ @ Plots ~/.julia/packages/Plots/7T4MM/src/utils.jl:92
```

It's in the REPL.. so perhaps it is because when it opens the GKS QtTerm window, it renders it multiple times or something along those lines?

t-bltg:
> This is awesome -- thanks for looking into this! I think it's a great UX improvement.
Thanks for the feedback !

> The GR warning still showing up would be my only minor complaint, but I do see your point. And since are printing the bigger warning anyway, I think it's fine.

> My only concern would be that passing broken data on to the backend may break it. I do manage to break GR on a regular basis (as in, independent follow-up plots in the same session fail to work render, and I have to either change backend or restart Julia).
In the end the plots appears or is saved to file, so this doesn't "break" GR in the running julia session, does it ?

> On another note, for some reason, for the original MWE it does print the new warning three times:
Yes, I've seen that. By inspecting the new debug traces, I found why it happened and pushed a fix. Can you try with the latest commit ?
On my machine I now have:
```bash
$ julia bug.jl
┌ Warning: Invalid negative or zero value 0.0 found at series index 3 for log10 based yscale
└ @ Plots [...]/Plots/Bfn5f/src/utils.jl:93
GKS: Number of points is invalid in routine POLYLINE
```

t-bltg:
We hit https://github.com/JuliaStats/Distributions.jl/issues/1358.

mortenpi:
> Yes, I've seen that. By inspecting the new debug traces, I found why it happened and pushed a fix. Can you try with the latest commit ?

Yup, can confirm that it's now down to just one warning!

```
julia> include("mwe.jl")
  Activating environment at `~/Julia/test/plots-number-of-points/Project.toml`
┌ Warning: Invalid negative or zero value 0.0 found at series index 3 for log10 based yscale
└ @ Plots ~/.julia/packages/Plots/empMW/src/utils.jl:92
GKS: Number of points is invalid in routine POLYLINE
```

> In the end the plots appears or is saved to file, so this doesn't "break" GR in the running julia session, does it ?

Ah, I did not mean that this particular case breaks GR. Even if GR emits the `GKS: Number of points is invalid in routine POLYLINE`, it looks like this has no effect on follow-up plots (or even the current plot).

However, in the past I have managed to break GR to the point where even follow-up plots fail to render properly by (accidentally) passing bad data to `plot`. So I was thinking that, as a general principle, it might not be a good idea to pass on bad data to the backend. However, I don't have an MWE at hand though, and it's always possible that I was running into some one-off bug that has already been fixed.

BeastyBlacksmith (approved):

PR #3608, opened 2021-07-03
Update plot_title

t-bltg:
Reproducer
---------------
```julia
using Plots; gr()

main() = begin
  plot(
    (plot(rand(3), legend=:none) for _ ∈ 1:7)...,
    layout=(@layout [_ ° _; ° ° °; ° ° °]), plot_title="Complex layout", plot_titlefontcolor=:red
  )
  plot(rand(3), plot_title="Simple layout", plot_titlevspan=.25)
  plot(rand(4, 2), layout=2, plot_title="Very nice layout", plot_titlelocation=:left)
  plot!(plot(1:5), plot_title="Title", plot_titlefontrotation=10)
end

main()
```
It's feels hacky (I'm sure we can do better), but it does the job.

Output
---------
layouts-1
layouts-2
layouts-3
layouts-4


BeastyBlacksmith:
What case does this cover?

PR #3612, opened 2021-07-04
GR: add support for mesh3d

t-bltg:
Add GR support for mesh3d since, for now, [GR](https://github.com/sciapp/gr) does not support drawing arbitrary points using a triangle connectivity array.
The error check code was copied from the `pgfplotsx` backend: https://github.com/JuliaPlots/Plots.jl/blob/master/src/backends/pgfplotsx.jl 

The test for this case already exists [here](https://github.com/JuliaPlots/Plots.jl/blob/ef93aa87be11f30837c87089d9f8e5c1f73c1baf/src/examples.jl#L1001) as example `47`, and is run e.g. for `pgfplotsx`: https://docs.juliaplots.org/latest/generated/pgfplotsx/#pgfplotsx-ref47.

Needs https://github.com/JuliaPlots/RecipesPipeline.jl/pull/90 patch.

Output of example 47
----------------------------
ref47

- [x] generate `ref47.png` for GR 
https://github.com/JuliaPlots/PlotReferenceImages.jl/pull/101

Linked https://github.com/JuliaPlots/Plots.jl/issues/392

### DIFF
--- a/src/backends.jl
+++ b/src/backends.jl
@@ -352,6 +352,7 @@ const _gr_seriestype = [
     :scatter3d,
     :surface,
     :wireframe,
+    :mesh3d,
     :volume,
     :shape,
 ]
--- a/src/backends/gr.jl
+++ b/src/backends/gr.jl
@@ -524,7 +524,7 @@ function gr_draw_colorbar(cbar::GRColorbar, sp::Subplot, clims, viewport_plotare
         levels = _cbar_unique(contour_levels.(series, Ref(clims)), "levels")
         # GR implicitly uses the maximal z value as the highest level
         if levels[end] < clims[2]
-            @warn("GR: highest contour level less than maximal z value is not supported.")
+            @warn "GR: highest contour level less than maximal z value is not supported."
             # replace levels, rather than assign to levels[end], to ensure type
             # promotion in case levels is an integer array
             levels = [levels[1:end-1]; clims[2]]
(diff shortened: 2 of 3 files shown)


BeastyBlacksmith:
This is missing the faces of the triangles though. Also some legend figure would be good.

t-bltg:
> This is missing the faces of the triangles though. Also some legend figure would be good.

OK I will change the example then.

BeastyBlacksmith:
You could also add a warning, that the GR backend doesn't support 3D mesh's with faces at the moment

t-bltg:
New output

ref47

bernd1995:
Quick question.
Could this `mesh3d` function be used to get a nicely working `histogram3d` (#3379)?
I started a PR (#3507) a few months ago but currently don't have any time to work on that 😕

So maybe my initial work on that could be finished with that PR being merged.

t-bltg:
> Quick question.
> Could this `mesh3d` function be used to get a nicely working `histogram3d` (#3379)?
> I started a PR (#3507) a few months ago but currently don't have any time to work on that confused
> 
> So maybe my initial work on that could be finished with that PR being merged.

I don't think so since this is only for triangles.

However you mentioned in https://github.com/JuliaPlots/Plots.jl/pull/3507#issue-645996631 that surface didn't have a mesh option:

I've pushed this PR which allows to use GR options when using surface https://github.com/JuliaPlots/Plots.jl/pull/3591

Is that of any help ?

bernd1995:
Thanks for the suggestions.
I somehow completely missed that this PR is only for triangles. 😅

I'll think about your suggestions, and if you happen to have another good idea comment on my PR #3507.

t-bltg:
@BeastyBlacksmith, if you don't see anything against, can I merge this ?

BeastyBlacksmith (approved):

PR #3628, opened 2021-07-07
Document use of tuples in annotations attribute

anowacki:
The `annotations` attribute (and `annotate!` function) has supported
the use of plain tuples containing arguments which are passed to
`text` since v0.22.2.  Document this option in the list of SubPlot
attributes and in example 20.

Closes (mostly) https://github.com/JuliaPlots/RecipesBase.jl/issues/72
but note that `series_annotations` does not yet support passing tuples
of arguments for `Plots.text`.


mkborregaard (approved):
nice

t-bltg:
Looks like the CI fails on this.

anowacki:
Whoops, yes—I included some unquoted `"`s in the string.  Let's see if this fixes things.

anowacki:
I have just added another commit which adds information about this to the `annotate!` docstring, but if tests pass now then I think this is finished.

PR #3632, opened 2021-07-08
Add :margins as an alias for :margin

t-bltg:
I find (and it's subjective) that writing `margins=...` in a plot is more intuitive than `margin=...`, so an alias might be handy here.

What do you think ?

### DIFF
--- a/src/args.jl
+++ b/src/args.jl
@@ -577,7 +577,8 @@ end
 
 
 # margin
-add_aliases(:left_margin , :leftmargin )
+add_aliases(:margins)
+add_aliases(:left_margin , :leftmargin)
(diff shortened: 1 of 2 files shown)


BeastyBlacksmith:
That already works, since we have https://github.com/JuliaPlots/Plots.jl/blob/4a180989447d7a4e5961ae8b0bdd5bcea02d7f91/src/args.jl#L1-L7

t-bltg:
> That already works, since we have
> 
> https://github.com/JuliaPlots/Plots.jl/blob/4a180989447d7a4e5961ae8b0bdd5bcea02d7f91/src/args.jl#L1-L7

Damn, I had the impression a savefig was failing the margins ...
Closing.

Issue #2728, opened 2020-05-22
[FR] Relative coordinates for annotate


mmikhasenko:
It would be extremely useful to have a way to specify annotation coordinates in percent.
```julia
annotate!( [ (relativex(x), relativey(y), text) ] )
```
I have several plots with different coordinates but the same additional labels added by `annotate!`,
it is quite tedious to align them manually.

Relative coordinates work nicely already for `inset=(1,bbox(x,y,wx,wy))`.
It uses `make_measure_hor/vert` [function](https://github.com/JuliaPlots/Plots.jl/blob/master/src/layouts.jl#L125)
I could not identify quickly where `w` and `h` are coming from there.

mmikhasenko:
Is there a way to access the extrema of the x and y-axis before the backend is called?
```
 xlims = extrema(plot!()[1][1][:x])
 ylims = extrema(plot!()[1][1][:y])
``` 
works only alter the plot has been displayed

mmikhasenko:
there is also `axis_limits`
```julia
    xlims = extrema(Plots.axis_limits(subplot, :x))
    ylims = extrema(Plots.axis_limits(subplot, :y))
```
but I do not know how to provide the `subplot`. `plot!()[1]` would not work

BeastyBlacksmith:
there is `Plots.ylims` and `Plots.xlims`, but I also found that sometimes it is neccessary to display or save the plot before I got the correct value.

mmikhasenko:
> there is `Plots.ylims` and `Plots.xlims`
it is not clear to which subplot it belongs to. Is there a way to return a specific subplot?

BeastyBlacksmith:
you can just pass the subplot that you want the limits of

mmikhasenko:
> you can just pass the subplot that you want the limits of

You mean to `axis_limits`? for sure, that is clear. What is still unclear to me is which method would return me a subplot I need. (before displaying)

BeastyBlacksmith:
```julia
julia> plot(1:5);

julia> p = plot(1:5);

julia> Plots.xlims(p)
(0.88, 5.12)

julia> Plots.ylims(p[1]) # limits of subplot
(0.88, 5.12)
```

mmikhasenko:
thanks
```julia
let
    p = plot(
        plot(rand(100)),
        plot(rand(200))
        )
    Plots.xlims(p[1]),  Plots.xlims(p[2])
[...]

Issue #3633, opened 2021-07-08
x/y-lims are not updated when fitting subplots into layout


anneaux:
Hey!
When I fit multiple subplots with equal aspect ratios into a layout, then their axis may be extended in order to fit the layout. However, the values for xlims and ylims are not being updated. Here's a MWE:

```julia
using Plots; gr()
using Random; Random.seed!(1234)

p = plot(layout = grid(1,2), frame=:box)

y1 = rand(10) .* 10
p1 = plot!(sp=1, y1, aspect_ratio=:equal)

y2 = rand(5) .* 10
p2 = plot!(sp=2, y2, aspect_ratio=:equal)
```
This produces this:
![plot1](https://user-images.githubusercontent.com/50271588/124911182-c1b3e400-dfec-11eb-9e0f-449dabaf1516.png)

and I get the following values:
```
extrema(y1) = (2.005860349338411, 8.541465903790503)
Plots.xlims(p1) = (0.73, 10.27)
Plots.ylims(p1) = (1.809792182704848, 8.737534070424065)
extrema(y2) = (0.10905889635595356, 9.567533636029237)
Plots.xlims(p2) = (0.73, 10.27)
Plots.ylims(p2) = (1.809792182704848, 8.737534070424065)
```
You see, for example, that the y axis of subplot 1 (the left one) goes from somewhere -1 to ~12 (ignore the cutted ticks label....), but the according ylims(p1) gives roughly (1.8,8.7). 

So... how can one get the actual limits of the axis after they have been fitted into the layout?

Thanks in advance!

(I'm using Julia v1.6.1, Plots v1.18.0)

t-bltg:
Could you please re-run (for the values and images: I've modified your example to use a fixed seed).

t-bltg:
Stack triggering incorrect values:
```
┌ Debug: 
│   exception =
│    14-element Vector{Base.StackTraces.StackFrame}:
│     macro expansion at logging.jl:341 [inlined]
│     axis_limits(sp::Plots.Subplot{Plots.GRBackend}, letter::Symbol, should_widen::Bool, consider_aspect::Bool) at axes.jl:557
│     axis_limits(sp::Plots.Subplot{Plots.GRBackend}, letter::Symbol) at axes.jl:557
│     get_sp_lims at utils.jl:420 [inlined]
│     ylims(sp::Plots.Subplot{Plots.GRBackend}) at utils.jl:435
[...]

PR #3634, opened 2021-07-08
Allow relative positioning of annotations

t-bltg:
Allow using `pct` for relative positioning of annotations.

Note that this also fixes https://github.com/JuliaPlots/Plots.jl/issues/2728#issuecomment-875432563, but NOT https://github.com/JuliaPlots/Plots.jl/issues/3633.

~~Before going forward with this, I think we should discuss https://github.com/JuliaPlots/Plots.jl/issues/3635 (I would prefer writing `3Plots.pct` instead of `.03Plots.pct`).~~

Fix https://github.com/JuliaPlots/Plots.jl/issues/2728.

MWE
----
```julia
using Plots; gr()

main() = begin
  p = plot(
    plot(rand(10) .* 15, aspect_ratio=:equal),
    plot(rand(10) .* 10, aspect_ratio=:equal),
    legend=false, layout=grid(1, 2), frame=:box
  )
  annotate!(sp=1, (.03, .95), text("Cats&Dogs", :left))
  annotate!(sp=2, (.03, .95), text("Cats&Dogs", :left))
  png("annotate")
end

main()
```

Output
----
annotate

### DIFF
--- a/src/components.jl
+++ b/src/components.jl
@@ -573,64 +573,52 @@ function Base.iterate(ea::EachAnn, i = 1)
     ((_cycle(ea.x,i), _cycle(ea.y,i), str, fnt), i+1)
 end
 
+# -----------------------------------------------------------------------
 annotations(::Nothing) = []
 annotations(anns::AVec) = anns
 annotations(anns::AMat) = map(annotations, anns)
 annotations(anns) = Any[anns]
 annotations(sa::SeriesAnnotations) = sa
 
+_annotationfont(sp::Subplot) = Plots.font(;
+  family=sp[:annotationfontfamily],
+  pointsize=sp[:annotationfontsize],
+  halign=sp[:annotationhalign],
+  valign=sp[:annotationvalign],
+  rotation=sp[:annotationrotation],
+  color=sp[:annotationcolor],
+)
+
+_annotation(sp, font, lab, pos...; alphabet="abcdefghijklmnopqrstuvwxyz") = (
+  if lab == :auto
+    (pos..., text("($(alphabet[sp[:subplot_index]]))", font))
+  else
+    (pos..., isa(lab, PlotText) ? lab : isa(lab, Tuple) ? text(lab[1], font, lab[2:end]...) : text(lab, font))
+  end
+)
+
 # Expand arrays of coordinates, positions and labels into induvidual annotations
 # and make sure labels are of type PlotText
-function process_annotation(sp::Subplot, xs, ys, labs, font = nothing)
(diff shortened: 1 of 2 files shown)


BeastyBlacksmith:
I'd prefer not to expose `Plots.pct` to the user, this makes using it in recipes hard

t-bltg:
> I'd prefer not to expose `Plots.pct` to the user, this makes using it in recipes hard

How ?

BeastyBlacksmith:
> How ?

You dispatch on some type in Base or RecipesBase first in this case maybe `Tuple{<:Number, <:Number}` if it unambiguous.

t-bltg:
> I'd prefer not to expose Plots.pct to the user, this makes using it in recipes hard

I'm not following you: there is no change in exposing `pct` here. It remains un-exported.

> You dispatch on some type in Base or RecipesBase first in this case maybe `Tuple{<:Number, <:Number}` if it unambiguous.

There is no dispatch in `RecipesBase` nor `Base` here: dispatching occurs in `Plots.jl/src/compnents.jl`, in `locate_annotation`, which is perfectly unambiguous here.

BeastyBlacksmith:
> > I'd prefer not to expose Plots.pct to the user, this makes using it in recipes hard
> 
> I'm not following you: there is no change in exposing `pct` here. It remains un-exported.

You are right with that, but a user is required to use `Plots.pct` for this feature. And since recipes are required to work without loading Plots, this feature cannot be used in recipes. Thats why you need a different user facing type for the call.

For example 
```julia
annotate!(sp=1, (.03, .95), text("Cats&Dogs", :left))
```

t-bltg:
> And since recipes are required to work without loading Plots, this feature cannot be used in recipes. Thats why you need a different user facing type for the call.

Thanks for the explanation, that makes sense.

I've reworked the PR to work with `Tuple`, and simplified `process_annotation`, and updated the [MWE](https://github.com/JuliaPlots/Plots.jl/pull/3634#issue-686183860): let's see if this does break `Plots` or if the CI passes.

BeastyBlacksmith:
Would be good to add a test for this otherwise LGTM

BeastyBlacksmith (approved):

mmikhasenko:
Many thanks, this extremely useful feature!

drandran12:
Is there a way we could implement the same feature for 3 dimensional plots?

BeastyBlacksmith:
Its usually better to open a feature request issue rather than comment on a closed PR, since this will get lost more easily

SteffenMuehle:
Alright, will do!

Issue #3641, opened 2021-07-14
[BUG] plot_title does not work with pyplot backend


BeastyBlacksmith:
## Details

```julia
using Plots
pyplot()
plot(rand(3,3), layout = 3, plot_title = "THE_Title")
```

### Backends

This bug occurs on ( insert `x` below )

Backend      | yes | no  | untested
-------------|-----|-----|---------
gr (default) |     |   x  |
pyplot       |   x  |     |
plotly       |     |     |x
plotlyjs     |     |     |x
pgfplotsx    |     |    x |
inspectdr    |     |     |x

### Versions

Plots.jl version: 1.19.0

PR #3642, opened 2021-07-14
Fix plot_title for pyplot

t-bltg:
Fix https://github.com/JuliaPlots/Plots.jl/issues/3641

ttl


BeastyBlacksmith:
Great, thank you!

Issue #3640, opened 2021-07-14
[BUG] [GR] Custom `markershape` causes `scale = :log10` to misinterpret coordinates. 


toma-vlad:
## Details
If one is using a custom marker `markershape = Shape(Plots.partialcircle(0, 2π))` in GR, then the axis scale `xscale = :log10` misplaces the series markers.

```julia
using Plots

circ = Shape(Plots.partialcircle(0, 2π))

y = -8:0.2:-6
x = [10^i for i in y]

default  = scatter(x,y);
logscale = scatter(x,y, xscale = :log10);
marker   = scatter(x,y, markershape = circ, seriescolor = :red);
both     = scatter(x,y, markershape = circ, seriescolor = :red, xscale = :log10);

p = plot(default, logscale, marker, both, title = ["default" "logscale" "marker" "both"], dpi = 360)
savefig(p, "test.png")
```
which produces

![test](https://user-images.githubusercontent.com/77381952/125616598-db82f7e2-2574-4bdc-b145-81b29eed5537.png)

and has some artifacts on the far left side which I assume is where the markers are starting to be drawn?

### Backends

This bug occurs on ( insert `x` below )

Backend      | yes | no  | untested
-------------|-----|-----|---------
gr (default) | x    |     |
pyplot       |     |     |x
plotly       |     | unsupported  |
plotlyjs     |     |     |x
pgfplotsx    |     | unsupported? |
inspectdr    |     |     |x

### Versions
Plots.jl version: 
```
v1.18.2
```
Backend  version (`]st -m <backend(s)>`):
```
GR v0.57.5
GR_jll v0.57.3+0
```
Output of `versioninfo()`: 
```
Julia Version 1.6.0
Commit f9720dc2eb (2021-03-24 12:55 UTC)
Platform Info:
  OS: Linux (x86_64-pc-linux-gnu)
  CPU: Intel(R) Xeon(R) CPU E5-2640 v4 @ 2.40GHz
  WORD_SIZE: 64
  LIBM: libopenlibm
  LLVM: libLLVM-11.0.1 (ORCJIT, broadwell)
```

t-bltg:
Thanks for the bug report.
I've played around in `gr.jl` trying to fix that, but from what I can tell, this might come from GR's side.

@jheinen might now better.

jheinen:
If you use GR's marker routines, it works. That's why I would expect a bug within the GR backend code (`gr.jl`) for shaped markers:
```julia
[...]

PR #3644, opened 2021-07-15
Fix computing marker coordinates with log scales

t-bltg:
Fix https://github.com/JuliaPlots/Plots.jl/issues/3640

Output:
test


BeastyBlacksmith (approved):
I think this is fine

PR #3659, opened 2021-07-21
Make tickmarks the same length for both axes in GR 2D plots

daschw:
Avoid too short ticks for one axis in very wide or narrow plots. This changes
```
plot(rand(10), size=(200, 400))
```
from
portrait_old

to
portrait_new

and
```julia
plot(rand(10), size=(600, 100))
```
from
wide_old

to
wide_new

### DIFF
--- a/src/axes.jl
+++ b/src/axes.jl
@@ -731,10 +731,6 @@ function axis_drawing_info(sp, letter)
             f = RecipesPipeline.scale_func(oax[:scale])
             invf = RecipesPipeline.inverse_scale_func(oax[:scale])
 
-            w = width(sp.plotarea).value
-            h = height(sp.plotarea).value
-            tick_factor = (letter === :x ? w : h) / min(w, h)
-
             add_major_or_minor_segments(ticks, grid, segments, factor, cond) = begin
                 ticks === nothing && return
                 if cond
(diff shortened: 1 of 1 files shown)


BeastyBlacksmith:
Thats definetly an improvement, I would argue that the ticklength should be the same irrespective of the size of the window.
For example are the ticks on the second figure much longer than on the first

isentropic:
Oh my god, this been bothering me since forever. I could never get it fixed myself

isentropic (commented):
looks really good to me

isentropic:
I actually think it should be like 
```
max(ticksize_x_axis, ticksize_y_axis)

```
where ticksize_y_axis is the old way of calculating it. So it would scale as before

daschw:
OK, how about this?
regular
minorticks
portrait_update
wide_update
tiny

isentropic:
I think it looks a little to long compared to pyplot, maybe scale down by factor of 1.25 or 1.5?

isentropic:
Overall, whatever you decide i support it. This is already much better than before

daschw:
Thanks for the feedback! 
next suggestion:

regular
minorticks
portrait
wide
tiny
big

isentropic:
This looks perfect to me.

t-bltg (approved):

PR #3695, opened 2021-07-27
Added colorbar font settings for GR

bernd1995:
I guess the title says it all.
Without this PR `colorbar_font*`-properties are completely ignored. 
Now one can either set the `colorbar_font*`-properties directly in the plotting functions or pass a `Plots.PlotText` object created with `text(string, args...; kw...)` to the `colorbar_title` keyword.


t-bltg:
Thanks !

PR #3713, opened 2021-07-28
Added colorbar font attributes for GR

bernd1995:
Fixup to #3695.
Added the new keyword arguments to `_gr_attr`. This is mainly to fix the docs/help to reflect the supported keyword arguments.


t-bltg:
Fine.

Issue #3691, opened 2021-07-27
[FR] legend=false by default for single series plots


isentropic:
I suggest for Plots 2.0 (maybe earlier) to turn off legends by default if:

1. There is only one series
2. If no user label has been provided

I believe this makes sense.

isentropic:
shoutout and thank to @t-bltg for the PR

PR #3732, opened 2021-07-31
Rework adding legends

t-bltg:
Disable legend entry when a single series is present and the label has been auto-generated for that series.

Fix https://github.com/JuliaPlots/Plots.jl/issues/3691.

- [x] generate ref images → [here](https://github.com/JuliaPlots/PlotReferenceImages.jl/pull/108)

### DIFF
--- a/src/args.jl
+++ b/src/args.jl
@@ -1656,12 +1656,13 @@ function _slice_series_args!(plotattributes::AKW, plt::Plot, sp::Subplot, comman
     return plotattributes
 end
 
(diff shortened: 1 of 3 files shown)


BeastyBlacksmith:
I'd vote to do this in 2.0

isentropic:
This is great, I certainly love it, but I do agree we should keep unmerged before 2.0. 
That's to say, we should keep the PR as default breaking change and ship all similar PRs together with 2.0

t-bltg:
All right, let's reserve this for `2.0`.

Issue #3629, opened 2021-07-07
[FR] Allow tuples of Plots.text arguments to series_annotations


anowacki:
[Since v0.22.2](https://github.com/JuliaPlots/Plots.jl/commit/7acb5a4bdb0497877b4d214c0cc1a05a8d9bfcf7) Plots has allowed the user to pass tuples of arguments to `Plots.text` as part of the `annotations` attribute.  However, this is not possible for `series_attributes`, which requires either `String`s or `Plots.PlotText` objects:

```julia
julia> using Plots

julia> plot([1, 2], [3, 4], series_annotations=[("A", :red), ("B", :blue)])
```
![text](https://user-images.githubusercontent.com/1231209/125435398-709e8a1b-1322-45ef-950e-7a6443ac232b.png)

I suggest that it would be useful to mirror the support for tuples of arguments to `text`.

Discussed in https://github.com/JuliaPlots/RecipesBase.jl/issues/72

Also relevant: https://github.com/JuliaPlots/Plots.jl/pull/3473

PR #3743, opened 2021-08-02
Allow passing collection of tuples to series_annotations

t-bltg:
Fix https://github.com/JuliaPlots/Plots.jl/issues/3629. And some code cleanup.

Example
---
```julia
using Plots
plot([1, 2], [3, 4], series_annotations=[("A", :red), ("B", :blue)])
plot([1, 2], [3, 4], series_annotations=(["A", "B"], [:red, :blue]))  # also supported
plot([1, 2], [3, 4], series_annotations=[["A", "B"], [:red, :blue]])  # same as previous
```
Output
---
rel1

### DIFF
--- a/src/components.jl
+++ b/src/components.jl
@@ -4,10 +4,10 @@ const P3 = GeometryBasics.Point3{Float64}
 const _haligns = :hcenter, :left, :right
 const _valigns = :vcenter, :top, :bottom
 
-nanpush!(a::AbstractVector{P2}, b) = (push!(a, P2(NaN, NaN)); push!(a, b))
-nanappend!(a::AbstractVector{P2}, b) = (push!(a, P2(NaN, NaN)); append!(a, b))
-nanpush!(a::AbstractVector{P3}, b) = (push!(a, P3(NaN, NaN, NaN)); push!(a, b))
-nanappend!(a::AbstractVector{P3}, b) = (push!(a, P3(NaN, NaN, NaN)); append!(a, b))
+nanpush!(a::AVec{P2}, b) = (push!(a, P2(NaN, NaN)); push!(a, b))
+nanappend!(a::AVec{P2}, b) = (push!(a, P2(NaN, NaN)); append!(a, b))
+nanpush!(a::AVec{P3}, b) = (push!(a, P3(NaN, NaN, NaN)); push!(a, b))
+nanappend!(a::AVec{P3}, b) = (push!(a, P3(NaN, NaN, NaN)); append!(a, b))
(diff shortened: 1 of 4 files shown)


mkborregaard:
Given that this is series-annotations, would it be able to allow passing of a tuple of equal-length vectors, e.g.
```julia
plot([1, 2], [3, 4], series_annotations=(["A", "B"], [:red, :blue]))
```

src/components.jl
```diff
@@ -774,16 +749,16 @@ end
...
+  curve, Base.range(first(range), stop=last(range), length=n)
+)
 
 # build a BezierCurve which leaves point p vertically upwards and arrives point q vertically upwards.
-# may create a loop if necessary.  Assumes the view is [0,1]
-function directed_curve(args...; kw...)
-    error("directed_curve has been moved to PlotRecipes")
-end
+# may create a loop if necessary.  Assumes the view is [0, 1]
+directed_curve(args...; kw...) = error("directed_curve has been moved to PlotRecipes")
```
BeastyBlacksmith, line 758:
```suggestion
```
It might make sense to remove this function here if it isn't exported. At leas `PlotRecipes` needs to be changed to `GraphRecipes`

t-bltg:
> Given that this is series-annotations, would it be make sense to allow passing of a tuple of equal-length vectors, e.g.
> 
> ```julia
> plot([1, 2], [3, 4], series_annotations=(["A", "B"], [:red, :blue]))
> ```

I find this less readable than using `plot(..., series_annotations=[("A", :red), ("B", :blue)])`.

t-bltg:
That PR was unreadable sorry, I've split it into two commits, the first one relative to the fix of the issue, and the second to code cleanup.

mkborregaard:
> find this less readable than using 

Yes sure. But most of the time you'll be plotting data from tables, and they will have that format, e.g. `plot(height, shoes, seriesannotations(names, textcolor)`. For `series_annotations`(as opposed to `annotations`) It's very rare to type them out in practice, because you'd tend to have more than 2-3 points.

t-bltg:
> > find this less readable than using
> 
> Yes sure. But most of the time you'll be plotting data from tables, and they will have that format, e.g. `plot(height, shoes, seriesannotations = (names, textcolor)`. For `series_annotations`(as opposed to `annotations`) It's very rare to type them out in practice, because you'd tend to have more than 2-3 points. I guess accepting a `zip` is an easy workaround.

Ha, I think I misunderstood your first comment.
I've implemented what you expected.

PR #3749, opened 2021-08-07
Code style - format

t-bltg:
I think it's time to have a consistent code style / format in Plots.jl.

The formatter used is https://github.com/domluna/JuliaFormatter.jl.

A test is added in the CI, running `JuliaFormatter` on PRs.

`julia -e 'using JuliaFormatter; format(["src", "test"])'` has been applied, using the newly added `.JuliaFormatter.toml` configuration file.

Format style is subjective: https://github.com/invenia/BlueStyle seems quite popular, I don't have a strong opinion on this, as long as there is consistency within the `Plots.jl` repo, so comments are welcome.


t-bltg:
Waiting for a solution to https://github.com/domluna/JuliaFormatter.jl/issues/453 ...

BeastyBlacksmith:
Is there an option to exclude certain parts from being formatted? There are sections where there is some meaningful whitespace which is actually helpful, like in the export section, that I would like to keep

t-bltg:
> Is there an option to exclude certain parts from being formatted? There are sections where there is some meaningful whitespace which is actually helpful, like in the export section, that I would like to keep

Yes, there is: https://domluna.github.io/JuliaFormatter.jl/dev/skipping_formatting/

I've applied it, and modified the PR.

src/Plots.jl
```diff
@@ -219,33 +231,48 @@ include("backends/web.jl")
...
-    global annotate!(plt::PlotOrSubplot, anns...; kw...)                         = plot!(plt; annotation = anns, kw...)
-    global annotate!(plt::PlotOrSubplot, anns::AVec{T}; kw...) where {T<:Tuple}         = plot!(plt; annotation = anns, kw...)
-    global xflip!(plt::PlotOrSubplot, flip::Bool = true; kw...)                  = plot!(plt; xflip = flip, kw...)
-    global yflip!(plt::PlotOrSubplot, flip::Bool = true; kw...)                  = plot!(plt; yflip = flip, kw...)
-    global xaxis!(plt::PlotOrSubplot, args...; kw...)                            = plot!(plt; xaxis = args, kw...)
-    global yaxis!(plt::PlotOrSubplot, args...; kw...)                            = plot!(plt; yaxis = args, kw...)
+let PlotOrSubplot = Union{Plot,Subplot}
+    global title!(plt::PlotOrSubplot, s::AbstractString; kw...) = plot!(plt; title=s, kw...)
+    global xlabel!(plt::PlotOrSubplot, s::AbstractString; kw...) =
+        plot!(plt; xlabel=s, kw...)
```
BeastyBlacksmith, line 237: I am not a fan on having this on an extra line

src/arg_desc.jl
```diff
@@ -1,187 +1,187 @@
...
-:contour_labels              => "Bool. Show labels at the contour lines?",
-:match_dimensions            => "Bool. For heatmap types... should the first dimension of a matrix (rows) correspond to the first dimension of the plot (x-axis)?  The default is false, which matches the behavior of Matplotlib, Plotly, and others.  Note: when passing a function for z, the function should still map `(x,y) -> z`.",
-:subplot                     => "Integer (subplot index) or Subplot object.  The subplot that this series belongs to.",
-:series_annotations          => "AbstractVector of String or PlotText.  These are annotations which are mapped to data points/positions.",
-:primary                     => "Bool.  Does this count as a 'real series'?  For example, you could have a path (primary), and a scatter (secondary) as 2 separate series, maybe with different data (see sticks recipe for an example).  The secondary series will get the same color, etc as the primary.",
-:hover                       => "nothing or vector of strings. Text to display when hovering over each data point.",
-:colorbar_entry              => "Bool.  Include this series in the color bar?  Set to `false` to exclude.",
```
BeastyBlacksmith, line 53: I think we can skip formatting for this whole file also

src/axes.jl
```diff
@@ -431,9 +433,18 @@ function expand_extrema!(sp::Subplot, plotattributes::AKW)
         data = plotattributes[if vert
             letter
         else
-            letter == :x ? :y : letter == :y ? :x : :z
```
BeastyBlacksmith, line 434: I'd rather have the ternary operator here

src/axes.jl
```diff
@@ -898,26 +954,37 @@ function axis_drawing_info_3d(sp, letter)
...
         end
     end
 
     return (
-        ticks = ticks,
-        segments = segments,
-        tick_segments = tick_segments,
-        grid_segments = grid_segments,
-        minorgrid_segments = minorgrid_segments,
-        border_segments = border_segments
```
BeastyBlacksmith, line 916: I think I'm in general for keeping whitespace around `=`

src/axes.jl
```diff
@@ -898,26 +954,37 @@ function axis_drawing_info_3d(sp, letter)
...
+        grid_segments=grid_segments,
+        minorgrid_segments=minorgrid_segments,
+        border_segments=border_segments,
     )
 end
 
 reverse_if(x, cond) = cond ? reverse(x) : x
 axis_tuple(x, y, letter) = reverse_if((x, y), letter === :y)
 
-axes_shift(t, i) = i % 3 == 0 ? t : i % 3 == 1 ? (t[3], t[1], t[2]) : (t[2], t[3], t[1])
```
BeastyBlacksmith, line 923: again ternary

src/components.jl
```diff
@@ -52,75 +52,81 @@ function coords(shapes::AVec{<:Shape})
...
-    :ltriangle  => makeshape(3, offset=1.0),
-    :pentagon   => makeshape(5),
-    :hexagon    => makeshape(6),
-    :heptagon   => makeshape(7),
-    :octagon    => makeshape(8),
-    :cross      => makecross(offset=-0.25),
-    :xcross     => makecross(),
-    :vline      => Shape([(0, 1), (0, -1)]),
-    :hline      => Shape([(1, 0), (-1, 0)]),
-  )
```
BeastyBlacksmith, line 120: should keep the alignment here

BeastyBlacksmith (changes requested):
Overall I am fine with this, but there are a few points where I think it can be improved.
I didn't mark all spots for the items (e.g. there are more of the ternaries, than I commented on)

src/components.jl
t-bltg: ok, should be the setting `align_pair_arrow`

src/Plots.jl
t-bltg: yes, because of margin set at 92 chars, not sure we can do much on this, maybe `... = plot(` and add e newline here.

src/arg_desc.jl
t-bltg: better use `align_pair_arrow = true`

src/axes.jl
t-bltg: odd, let me see what I can do on this.

src/axes.jl
t-bltg: Let me explain my position on this: we are limited by the screen width, and it is thus normal and excepted to try to respect a margin limit, especially for people working on editors which do not support line wrapping. Thus, in order to minimize the need for splitting over multiple lines, we should reduce the number of characters. One way to do that is to remove spaces around `=` sign, especially in function calls, where multiple kwargs induce multiple unnecessary spaces. So my position on this is to remove those spaces.

src/axes.jl
t-bltg: ok, that is because `BlueStyle` does not allow chaining, I've reverted it to using `DefaultStyle`.

src/Plots.jl
t-bltg: Fixed: using `align_assignment = true`.

src/axes.jl
BeastyBlacksmith: I don't see the benefit to be honest. The formatter will split too long lines into horizontal layout anyways and thus there is text should always fit a reasonable screen. On the other hand removing spaces here makes things hard to read even there is plenty of space left and right.

t-bltg: I disagree, especially on named tuples, where the name is identical, I find it easier to read when the `lhs` and the `rhs` are glued to the equal sign. I also find it easier to read for named arguments in methods definitions. But anyway this is very opinionated so I do not really care as long as there is consistency. I've reverted to using `whitespace_in_kwargs = true`.

BeastyBlacksmith (approved):

PR #3751, opened 2021-08-15
Fix vector attributes to bar plots

yha:
Fix JuliaPlots/StatsPlots.jl#442.
New example still requires a reference image.


BeastyBlacksmith:
The limits for the first plot looked a bit different on my machine, than what shows up in the docs: https://docs.juliaplots.org/latest/generated/gr/#gr-ref56.
Also it would be good to analyze why GR is the only backend where this works.

PR #3755, opened 2021-08-18
Implementation of filling the space between and under curves

LukasKrumwiede:
It it now possible to fill the space between a parametric function or under (between graph and x-axis) several curves. In addition one can pass a vector as argument to fill the space between altering y-values.
The plots:

``
plot(sin, (x->begin
            sin(2x)
        end), 0, 2π, line = 4, leg = false, fill = (0, :orange))
``

and 

``
y = rand(100)
plot(0:10:100, rand(11, 4), lab = "lines", w = 3, palette = cgrad(:grays), fill = 0, α = 0.6)
``
and 

``
plot(1:5, fillrange = [1,2])
``

are now possible and the output is:

Parametric_plot
several_filledcurves
vector_input

Possibility for adding markers is still missing when using the `fill` option.

### DIFF
--- a/src/backends/gaston.jl
+++ b/src/backends/gaston.jl
@@ -303,10 +303,14 @@ function gaston_seriesconf!(
         lc, dt, lw = gaston_lc_ls_lw(series, clims, i)
         pt, ps, mc = gaston_mk_ms_mc(series, clims, i)
         push!(curveconf, "w points pt $pt ps $ps lc $mc")
-    elseif st ∈ (:path, :straightline, :path3d)
+    elseif st ∈ (:path, :straightline, :path3d)      
+        fr = series[:fillrange]
+        fc = gaston_color(get_fillcolor(series, i), get_fillalpha(series, i))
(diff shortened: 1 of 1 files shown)


src/backends/gaston.jl
```diff
@@ -225,13 +225,13 @@ function gaston_add_series(plt::Plot{GastonBackend}, series::Series)
     gsp = sp.o
     x, y, z = series[:x], series[:y], series[:z]
     st = series[:seriestype]
-
     curves = []
     if gsp.dims == 2 && z === nothing
         for (n, seg) in enumerate(series_segments(series, st; check = true))
             i, rng = seg.attr_index, seg.range
+            fr =_cycle(series[:fillrange], 1:length(x[rng])) ####
```
t-bltg, line 232:
```suggestion
            fr =_cycle(series[:fillrange], 1:length(x[rng]))
```

t-bltg (changes requested):

Issue #3356, opened 2021-03-18
[BUG] Plotly backend does not support passing arrays as levels in contour plot


melonedo:
### Backends

This bug occurs on ( insert `x` below )

Backend      | yes | no  | untested
-------------|-----|-----|---------
gr (default) |     |    x |
pyplot       |     |     |
plotly       |   x  |     |
plotlyjs     |   x  |     |
pgfplotsx    |     |     |
inspectdr    |     |     |

### Versions

Plots.jl version: v1.10.6
Backend  version (`]st -m`): `[f0f68f2c] PlotlyJS v0.14.1`
Output of `versioninfo()`: 
```
Julia Version 1.5.4
Commit 69fcb5745b (2021-03-11 19:13 UTC)
Platform Info:
  OS: Windows (x86_64-w64-mingw32)
  CPU: Intel(R) Core(TM) i5-9300H CPU @ 2.40GHz
  WORD_SIZE: 64
  LIBM: libopenlibm
  LLVM: libLLVM-9.0.1 (ORCJIT, skylake)
```
MWE:
```julia
using Plots
plotlyjs()
contour(0:1:10, 0:1:10, (x,y)->x*y, levels=[1,2,3])
```
Result:
```julia
julia> contour(0:1:10, 0:1:10, (x,y)->x*y, levels=[1,2,3])
Error showing value of type Plots.Plot{Plots.PlotlyJSBackend}:
ERROR: MethodError: no method matching +(::Array{Int64,1}, ::Int64)
For element-wise addition, use broadcasting with dot syntax: array .+ scalar
Closest candidates are:
  +(::Any, ::Any, ::Any, ::Any...) at operators.jl:538
  +(::Missing, ::Number) at missing.jl:115
  +(::Base.CoreLogging.LogLevel, ::Integer) at logging.jl:116
  ...
Stacktrace:
 [1] plotly_series(::Plots.Plot{Plots.PlotlyJSBackend}, ::Plots.Series) at C:\Users\melonedo\.julia\packages\Plots\SjqWU\src\backends\plotly.jl:570
 [2] plotly_series(::Plots.Plot{Plots.PlotlyJSBackend}) at C:\Users\melonedo\.julia\packages\Plots\SjqWU\src\backends\plotly.jl:873
[...]

PR #3757, opened 2021-08-19
Plotly: Enable specified contour values for ranges; warn otherwise

anowacki:
This PR is meant to help address #3356.  It only does partially.

Note that there are no tests yet, though things seem to be
working locally for me.  Guidance on how to add tests would
be greatly appreciated.

Plotly does not support arbitrary contour values, but can plot
specific contours if they are an equally-spaced range.  However,
the current behaviour is to simply error if a range or collection
of values are passed via the `levels` keyword, since the current
implementation (correctly) adds two to whatever is passed in

This commit implements the plotting of contours if the `levels`
keyword argument is passed an `AbstractRange`, or if a set
of arbitrary values are passed.  In the latter case, however,
since this is not supported by Plotly, a range based on the
first and last values of the collection passed in is created
and used to define the contours.  A warning is then issued to
the user.

Otherwise, any other types are assumed to be number-like and
adjusted as previously.  Note that Plotly does not guarantee
the exact number of contours will be used.

See https://github.com/plotly/plotly.js/issues/4503 (Plotly issue
tracking the ability to set arbitrary contours).

Possible things to look at are:
- How we warn the user.  I couldn't find any mechanism in to which
  to hook to do this properly, since this is a partial limitation
  of a backend, not a complete absence of a feature.
- Whether changing a user's set of contour values for them is the
  right thing to do, even with a warning, or we should instead
  error when passed a tuple or vector.  Or should we check that
  the values therein are actually equally spaced, in which case
  we could plot those exact contours?

### DIFF
--- a/src/arg_desc.jl
+++ b/src/arg_desc.jl
@@ -30,7 +30,7 @@ const _arg_desc = KW(
     :marker_z           => "AbstractVector, Function `f(x,y,z) -> z_value`, or Function `f(x,y) -> z_value`, or nothing. z-values for each series data point, which correspond to the color to be used from a markercolor gradient.",
     :line_z             => "AbstractVector, Function `f(x,y,z) -> z_value`, or Function `f(x,y) -> z_value`, or nothing. z-values for each series line segment, which correspond to the color to be used from a linecolor gradient.  Note that for N points, only the first N-1 values are used (one per line-segment).",
     :fill_z             => "Matrix{Float64} of the same size as z matrix, which specifies the color of the 3D surface; the default value is `nothing`.",
-    :levels             => "Integer, NTuple{2,Integer}, or AbstractVector. Levels or number of levels (or x-levels/y-levels) for a contour type.",
+    :levels             => "Integer (number of contours) or AbstractVector (contour values). Determines contour levels for a contour type.",
     :orientation        => "Symbol.  Horizontal or vertical orientation for bar types.  Values `:h`, `:hor`, `:horizontal` correspond to horizontal (sideways, anchored to y-axis), and `:v`, `:vert`, and `:vertical` correspond to vertical (the default).",
     :bar_position       => "Symbol.  Choose from `:overlay` (default), `:stack`.  (warning: May not be implemented fully)",
     :bar_width          => "nothing or Number. Width of bars in data coordinates. When nothing, chooses based on x (or y when `orientation = :h`).",
--- a/src/args.jl
+++ b/src/args.jl
@@ -1502,6 +1502,11 @@ function RecipesPipeline.preprocess_attributes!(plotattributes::AKW)
         plotattributes[:framestyle] = _framestyleAliases[plotattributes[:framestyle]]
     end
 
+    # contours
+    if haskey(plotattributes, :levels)
+        check_contour_levels(plotattributes[:levels])
+    end
+
     # warnings for moved recipes
     st = get(plotattributes, :seriestype, :path)
     if st in (:boxplot, :violin, :density) && !isdefined(Main, :StatsPlots)
@@ -1626,6 +1631,26 @@ convertLegendValue(v::Tuple{S,T}) where {S<:Real,T<:Real} = v
 convertLegendValue(v::Tuple{<:Real,Symbol}) = v
 convertLegendValue(v::Real) = v
 convertLegendValue(v::AbstractArray) = map(convertLegendValue, v)
+
+# -----------------------------------------------------------------------------
+
+"""Throw an error if the `levels` keyword argument is not of the correct type
+or `levels` is less than 1"""
+function check_contour_levels(levels)
+    if !(levels isa Union{Integer,AVec})
+        throw(
+            ArgumentError(
+                "the levels keyword argument must be an integer or AbstractVector",
+            ),
+        )
+    elseif levels isa Integer && levels <= 0
+        throw(
+            ArgumentError(
+                "must pass a positive number of contours to the levels keyword argument",
+            ),
+        )
+    end
+end
 
 # -----------------------------------------------------------------------------
 
--- a/src/backends/plotly.jl
+++ b/src/backends/plotly.jl
@@ -617,11 +617,34 @@ function plotly_series(plt::Plot, series::Series)
         filled = isfilledcontour(series)
         plotattributes_out[:type] = "contour"
         plotattributes_out[:x], plotattributes_out[:y], plotattributes_out[:z] = x, y, z
-        plotattributes_out[:ncontours] = series[:levels] + 2
         plotattributes_out[:contours] = KW(
             :coloring => filled ? "fill" : "lines",
             :showlabels => series[:contour_labels] == true,
         )
+        # Plotly does not support arbitrary sets of contours:
+        # https://github.com/plotly/plotly.js/issues/4503
+        let levels = series[:levels]
+            if levels isa AbstractRange
+                plotattributes_out[:contours][:start] = first(levels)
+                plotattributes_out[:contours][:end] = last(levels)
+                plotattributes_out[:contours][:size] = step(levels)
+            elseif levels isa AVec
+                levels_range =
+                    range(first(levels), stop = last(levels), length = length(levels))
+                plotattributes_out[:contours][:start] = first(levels_range)
+                plotattributes_out[:contours][:end] = last(levels_range)
+                plotattributes_out[:contours][:size] = step(levels_range)
+                @warn(
+                    "setting arbitrary contour levels with Plotly backend is not supported; " *
+                    "use a range to set equally-spaced contours or an integer to set the " *
+                    "approximate number of contours with the keyword `levels`. " *
+                    "Using levels $(levels_range)"
+                )
+            elseif isinteger(levels)
+                # Assume this is a number of levels
+                plotattributes_out[:ncontours] = levels + 2
+            end
+        end
         plotattributes_out[:colorscale] =
             plotly_colorscale(series[:linecolor], series[:linealpha])
         plotattributes_out[:showscale] = hascolorbar(sp) && hascolorbar(series)
--- a/src/backends/plotly.jl
+++ b/src/backends/plotly.jl
@@ -621,8 +621,9 @@ function plotly_series(plt::Plot, series::Series)
             :coloring => filled ? "fill" : "lines",
             :showlabels => series[:contour_labels] == true,
         )
-        # Plotly does not support arbitrary sets of contours:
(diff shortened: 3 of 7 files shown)


anowacki:
With this PR:

```julia
x = y = -π:0.01:π
z = 1.01.*sin.(x).*cos.(y)'
levels = -1:0.5:1
plot(
    [contour(x, y, z, levels=l, c=:rainbow)
     for l in (length(levels), levels, collect(levels))]...,
    grid=false
)
```

newplot

t-bltg:
> Note that there are no tests yet, though things seem to be
> working locally for me. Guidance on how to add tests would
> be greatly appreciated.

The are no image comparison test for backends other than `GR` actually.

But you can try running and compare the examples against the [`docs`](http://docs.juliaplots.org/latest/generated/plotlyjs/):
```julia
using Plots; plotlyjs()
for i ∈ 1:length(Plots._examples)
    i ∈ Plots._backend_skips[:plotlyjs] && continue
    try
        Plots.reset_defaults()
        map(x->Base.eval(Main, x), Plots._examples[i].exprs)
        png("plotly-$i")
    catch err
        @error "" bt=(err, stacktrace(catch_backtrace()))
    end
end
```

BeastyBlacksmith:
I think we got 3 different type of tests:
 - Visual regression tests: thats what @t-bltg was referring to, setting these up and keeping up to date is quite involved and consumes a fair bit of CI time, thus we run it only for GR
 - Frontend tests: for example in https://github.com/JuliaPlots/Plots.jl/blob/master/test/test_axes.jl make sure that backends recieve the intended output, these are really cheap and the more the better
 - Backend tests: like the ones in https://github.com/JuliaPlots/Plots.jl/blob/master/test/test_pgfplotsx.jl these test a non graphical backend specific plot representation (say a textfile or backend julia types): these check that the correct backend objects get created, are somewhere in between depending on which format is being tested

For the given case here I'd recommend to add some of the latter

anowacki:
Thanks for the explanations, both.

> But you can try running and compare the examples against the docs:

Okay—I have done this and everything looks the same as in the docs.  I guess this is expected because the only change is to the contour handling for Plotly, and the example does not specify any contour number or levels.

The caveat to this is that examples 32, 47 and 50 produce empty files and the `kaleido plotly` command spawned by Julia hangs for these examples until I kill it.  This happens on my machine without this PR as well, so I don't think it can be related.

> Frontend tests

I'm not totally sure how these tests work, but I guess I should add a file `test/test_contours.jl` which just checks that a backend gets the value passed in via the `levels` keyword argument?

> Backend tests

I can't see any backend tests for Plotly at the moment.  Should I add a `test/test_plotly.jl` file and add tests for the changes there?  I'm afraid I don't have the time to add tests for other functionality.

anowacki:
In starting to write these tests, I realise I don't understand what the `levels` keyword argument is actually meant to do when it gets a `NTuple{2, Integer}`.  The [docs](http://docs.juliaplots.org/latest/generated/attributes_series/#Series) say:

> levels | levs, nlev, nlevels | `15` | Integer, NTuple{2,Integer}, or AbstractVector | Levels or number of levels (or x-levels/y-levels) for a contour type.

What are the number of x-levels and y-levels for a contour plot?  I don't understand what that could mean here.

Trying the example above with `contours(x, y, z, levels=(3, 5))` with GR just throws a `MethodError` in `convert` at [`GR:src/GR.jl:2662`](https://github.com/jheinen/GR.jl/blob/5d4e0aaca46e734d08548c26dbb957e5c9c34b69/src/GR.jl#L2662-L2665) (without this PR and with).  Clearly GR is trying to convert the tuple into a set of contours, which is not intended per the docs.  However I think this is a separate issue.

Just asking since I will need to update this PR to reflect.

t-bltg:
> What are the number of x-levels and y-levels for a contour plot? I don't understand what that could mean here.

Using git blame, traces appear in https://github.com/JuliaPlots/Plots.jl/commit/cf4fcf335124c8884080df2ab1d3e163dbb22c73, https://github.com/JuliaPlots/Plots.jl/pull/99 and https://github.com/JuliaPlots/Plots.jl/pull/101, although I did not find explicit reference to `Tuple`s apart from `arg_desc.jl`.
I'm guessing that it can be removed from the docs ? (needs confirmation though).

t-bltg:
> > What are the number of x-levels and y-levels for a contour plot? I don't understand what that could mean here.
> 
> Using git blame, traces appear in [cf4fcf3](https://github.com/JuliaPlots/Plots.jl/commit/cf4fcf335124c8884080df2ab1d3e163dbb22c73), #99 and #101, although I did not find explicit reference to `Tuple`s apart from `arg_desc.jl`.

I'm guessing that it can be removed from the docs ? (needs confirmation though).

>  Should I add a test/test_plotly.jl

You can add a minimal `test_plotly.jl` if the test is representative and robust.

BeastyBlacksmith:
> > Frontend tests
> 
> I'm not totally sure how these tests work, but I guess I should add a file `test/test_contours.jl` which just checks that a backend gets the value passed in via the `levels` keyword argument?
If there is no processing of the argument on the plots-side than adding a test is not really needed, but wouldn't hurt either.

Those tests really check if the `Plot` object fields have the expected values.
 
> > Backend tests
> 
> I can't see any backend tests for Plotly at the moment. Should I add a `test/test_plotly.jl` file and add tests for the changes there? I'm afraid I don't have the time to add tests for other functionality.

Thats correct and also fine to only test what is done in this PR. Hopefully it will grow over time.

Thanks a lot for your effort!

anowacki:
I have updated the PR with tests and corrected the logic in converting `levels` to Plotly-specific settings.

In looking over this I realised that the different backends do different things if the input to `levels` is not correct.  As it stands, this PR and at least some other backends simply ignore arguments which are not a vector (`isa AVec`) or integer (`isinteger()`) and so the default number of contours are used.  Should we instead throw an `ArgumentError` in general for unsupported arguments?

t-bltg:
> I have updated the PR with tests and corrected the logic in converting `levels` to Plotly-specific settings.

Thanks. Can you also update `arg_desc.jl` by removing the reference to `NTuple{...}` for levels ?

> In looking over this I realised that the different backends do different things if the input to `levels` is not correct. As it stands, this PR and at least some other backends simply ignore arguments which are not a vector (`isa AVec`) or integer (`isinteger()`) and so the default number of contours are used. Should we instead throw an `ArgumentError` in general for unsupported arguments?

I'm supporting unification of the backends on this and error out on unsupported (not Integer nor AVec).

anowacki:
> Can you also update `src/arg_desc.jl` by removing the reference to `NTuple{...}` for levels ?

No problem—I'll update shortly.

> I'm supporting unification of the backends on this and error out on unsupported (not `Integer` nor `AVec`).

Sounds like a a good, user-friendly solution to me and I'm happy to add that here or as a separate PR if that's more appropriate.  Where should the check go?  I haven't found any sort of central argument-checking place in the codebase so far, but likely I have missed it.  Or perhaps it should go in RecipesPipeline?

t-bltg:
> Sounds like a a good, user-friendly solution to me and I'm happy to add that here or as a separate PR if that's more appropriate. Where should the check go? I haven't found any sort of central argument-checking place in the codebase so far, but likely I have missed it. Or perhaps it should go in RecipesPipeline?

I think central error checking is unfeasible since it's too backend dependent.

Thinking out loud maybe you can check it directly in `plotly.jl`:

Something like this:
```julia
if !(typeof(series[:levels]) <: Union{Integer,Avec})
    error("Unsupported message ...")
end
```

anowacki:
> maybe in `src/utils.jl`…

I'd be happy to do that, but that suggestion does then still require quite a lot of code duplication in the backends—each one will separately need to call e.g. `check_supported_levels`.

As a potential alternative, it seems like [`RecipesPipeline.preprocess_attributes!(::KW)`](https://github.com/JuliaPlots/Plots.jl/blob/0742d475bcb788c875022c4606a3f2022eb472e1/src/args.jl#L1330-L1513) (as defined in Plots.jl) seems like a really good candidate for this.  There are already checks for things like [legend placement in `convertLegendValue`](https://github.com/JuliaPlots/Plots.jl/blob/0742d475bcb788c875022c4606a3f2022eb472e1/src/args.jl#L1620) and [label naming in `label_to_string`](https://github.com/JuliaPlots/Plots.jl/blob/0742d475bcb788c875022c4606a3f2022eb472e1/src/args.jl#L2032), both ultimately called from `preprocess_attributes!`.

How about calling the function you posted @t-bltg (or equivalent) in `preprocess_attributes!(::KW)`?

t-bltg:
> > maybe in `src/utils.jl`…
> 
> I'd be happy to do that, but that suggestion does then still require quite a lot of code duplication in the backends—each one will separately need to call e.g. `check_supported_levels`.
> 
> As a potential alternative, it seems like [`RecipesPipeline.preprocess_attributes!(::KW)`](https://github.com/JuliaPlots/Plots.jl/blob/0742d475bcb788c875022c4606a3f2022eb472e1/src/args.jl#L1330-L1513) (as defined in Plots.jl) seems like a really good candidate for this. There are already checks for things like [legend placement in `convertLegendValue`](https://github.com/JuliaPlots/Plots.jl/blob/0742d475bcb788c875022c4606a3f2022eb472e1/src/args.jl#L1620) and [label naming in `label_to_string`](https://github.com/JuliaPlots/Plots.jl/blob/0742d475bcb788c875022c4606a3f2022eb472e1/src/args.jl#L2032), both ultimately called from `preprocess_attributes!`.
> 
> How about calling the function you posted @t-bltg (or equivalent) in `preprocess_attributes!(::KW)`?

Agreed, I haven't checked if all backends behave in the same way on keyword `levels`, but if checking valid kw upstream in the pipeline is feasible I'm all in !

anowacki:
From a survey of the backends, only PyPlot previously checked what `levels` was, and so it was simple enough to remove a couple of lines of code there.  Now anything backend-specific I *think* will have come after `preprocess_attributes!` and so can rely on `levels` being an `Integer` or `AbstractVector`.

anowacki:
Note that I added a check that `levels` is greater than 0 if an integer.  GR, PyPlot and Plotly accepted 1 level but not 0 when I tested, so this seems like it would not break existing code that worked, but I guess it is technically a breaking change.  I can't imagine many people's workflows relied on being able to pass negative numbers of contours.

BeastyBlacksmith (approved):

t-bltg (approved):

t-bltg:
> I can't imagine many people's workflows relied on being able to pass negative numbers of contours.

Safe assumption, we'll see if any issue pops up on this matter.

PR #3765, opened 2021-08-25
No exception thrown in `apply_recipe` fallback

BioTurboNick:
Throwing an exception here can vastly slow down plotting, as it happens frequently. Returning an error value instead, which the calling function can check, can speed up plotting 2x. An accompanying change (https://github.com/JuliaPlots/RecipesPipeline.jl/pull/95) is necessary in RecipesPipeline.jl, assuming this value is suitable.


BioTurboNick:
An alternative would be to have an API to check if a recipe exists prior to applying it, but that would break existing usage.

PR #3775, opened 2021-08-27
Cache axis args in a dictionary

### LINKED ISSUE #3763: [FR] Cut down performance hit of dynamically generating prefix-symbols each time
11% of profiler samples in one of my plots are being spent in `Symbol(x, y)`, which constructs a new symbol from two different symbols. Each time this happens, 3 allocations are required.

This code was run 429,042 times, and consuming 72 MiB, in generating my plot. It thus appears to account for 24% of allocations and 23% of allocation amount of generating my plot, between calls to `_update_axis` and `_update_subplot_args`.

Can this list of attribute combinations be generated once and then read from, rather than being constructed expensively on the fly?

BioTurboNick:
These axis arguments can be generated lots of times for large plots. We can save time by caching them in a dictionary. For my large map test plot:

```
Before: 1.075 s (3393429 allocations: 183.49 MiB)
After:  959.024 ms (3393294 allocations: 181.64 MiB) - 10% improvement in speed, small decrease in allocations

TTFP Before: 7.543192 seconds (26.79 M allocations: 1.544 GiB, 3.89% gc time, 0.06% compilation time)
TTFP After: 6.886222 seconds (23.58 M allocations: 1.355 GiB, 3.63% gc time, 0.07% compilation time) - 8% improvement in speed, 12% fewer allocations, 12% lower allocation amount
```

The cost of this is, I think, just 3-4 kb. Is there a function to read out the size in memory of a dictionary? `sizeof` for the dictionary itself just returns 64.

Resolves #3763


t-bltg:
My mental model on `Symbol` concatenation refuses to see that it can be slower than a `Dict` lookup, but I guess the number are here :wink:.

Could you share a mwe for the TTFP, so that I can also check the numbers on my machine ?

BioTurboNick:
I hear you haha. Each time a symbol is created, it has to make a new string.

This should do it.

U.S. Census county shapefiles: https://www2.census.gov/geo/tiger/GENZ2018/shp/cb_2018_us_county_5m.zip

```
using Shapefile
using Plots

shapepath = joinpath("usgeodata", "cb_2018_us_county_5m.shp")
table = Shapefile.Table(shapepath)
geoms = Shapefile.shapes(table)
ALASKA = 2
HAWAII = 15
PUERTORICO = 72
lower48geoms = geoms[stateids .∉ Ref([ALASKA, HAWAII, PUERTORICO])]

plot(lower48geoms)
```

t-bltg:
> using Shapefile
> using Plots
> 
> shapepath = joinpath("usgeodata", "cb_2018_us_county_5m.shp")
> table = Shapefile.Table(shapepath)
> geoms = Shapefile.shapes(table)
> ALASKA = 2
> HAWAII = 15
> PUERTORICO = 72
> lower48geoms = geoms[stateids .∉ Ref([ALASKA, HAWAII, PUERTORICO])]
> 
> plot(lower48geoms)
> ```

Thanks, but `ERROR: UndefVarError: stateids not defined` ?

BioTurboNick:
Oops, that'll teach me to run on a fresh session. Updated it.

t-bltg:
My numbers:

**without PR**
6.375625 seconds (11.34 M allocations: 661.273 MiB, 4.46% gc time, 0.23% compilation time)

**with PR**
6.193512 seconds (9.50 M allocations: 546.272 MiB, 3.75% gc time, 0.23% compilation time)

t-bltg (approved):

src/utils.jl
```diff
@@ -1214,3 +1214,23 @@ function mesh3d_triangles(x, y, z, cns)
     end
     return X, Y, Z
 end
+
+_axisattrcache = Dict{Symbol, Dict{Symbol, Symbol}}()
```
BeastyBlacksmith, line 1218: shouldn't that be const?

t-bltg:
Re-thinking about this, another possible implementation (not reinventing the wheel) is:
```diff
--- a/Plots.jl/src/utils.jl	2021-08-26 20:05:17.172180754 +0200
+++ b/Plots.jl/src/utils.jl	2021-08-28 21:26:22.900145140 +0200
@@ -1,4 +1,5 @@
 # ---------------------------------------------------------------
+using Memoize
 
 treats_y_as_x(seriestype) =
     seriestype in (:vline, :vspan, :histogram, :barhist, :stephist, :scatterhist)
@@ -1214,3 +1215,6 @@
     end
     return X, Y, Z
 end
+
+@memoize get_axis_attr(letter::Symbol, keyword::String) = get_axis_attr(letter, Symbol(keyword))
+@memoize get_axis_attr(letter::Symbol, keyword::Symbol) = Symbol(letter, keyword)

```

```julia
julia> @time main()
  5.993177 seconds (10.00 M allocations: 562.238 MiB, 4.08% gc time, 0.24% compilation time)
```

BioTurboNick:
Oh, that change to a const may have made TTFP worse, but did further improve repeated plotting.

BioTurboNick:
Memoize.jl version with my own example:

```
TTFP: 6.755638 seconds (23.88 M allocations: 1.360 GiB, 3.66% gc time, 0.08% compilation time)
907.864 ms (3598686 allocations: 187.91 MiB)
```

t-bltg:
> Oh, that change to a const may have made TTFP worse, but did further improve repeated plotting.

On my machine timings improve with const. Maybe a gc effect ?

t-bltg:
There are some missing replacements, e.g. 
https://github.com/JuliaPlots/Plots.jl/blob/8b6073c08806c371e76d065bc7e7da282bbfcb11/src/backends/gr.jl#L1581
https://github.com/JuliaPlots/Plots.jl/blob/8b6073c08806c371e76d065bc7e7da282bbfcb11/src/backends/gr.jl#L1559

Those should be replaced:
```
$ cd Plots.jl
$ grep -ir 'Symbol(.*, :axis)'
```

BioTurboNick:
Are you testing this on Julia master, or 1.6.2?

t-bltg:
> Are you testing this on Julia master, or 1.6.2?

1.6.2, but not the downloaded binaries, local build.

```julia
julia> versioninfo()
Julia Version 1.6.2
Commit 1b93d53fc4b* (2021-07-14 15:36 UTC)
Platform Info:
  OS: Linux (x86_64-unknown-linux-gnu)
  CPU: Intel(R) Xeon(R) CPU E5-2670 0 @ 2.60GHz
  WORD_SIZE: 64
  LIBM: libopenlibm
  LLVM: libLLVM-11.0.1 (ORCJIT, sandybridge)
```

BioTurboNick:
Alright, I've reproduced your numbers. Are there enough other things that need to be cached to justify the extra overhead?

Profiling this example, I don't see many other opportunities. Since a future change shouldn't be hard, would it be better to stay with this and just leave a comment to that effect? Wouldn't be a hard change to make if it became necessary to memoize more generally.

t-bltg:
I'm OK with merging this as is.
Can you confirm @BeastyBlacksmith ?

src/utils.jl
```diff
@@ -1214,3 +1214,23 @@ function mesh3d_triangles(x, y, z, cns)
     end
     return X, Y, Z
 end
+
+const _attrsymbolcache = Dict{Symbol, Dict{Symbol, Symbol}}()
```
BeastyBlacksmith, line 1218:
```suggestion
const _attrsymbolcache = Dict{Symbol, Dict{Symbol, Symbol}}( 
    :x => Dict( k => Symbol("x", String(k)) for k in keys(_axis_defaults)),
    :y => Dict( k => Symbol("y", String(k)) for k in keys(_axis_defaults)),
    :z => Dict( k => Symbol("z", String(k)) for k in keys(_axis_defaults))
)
```

src/utils.jl
```diff
@@ -1214,3 +1214,23 @@ function mesh3d_triangles(x, y, z, cns)
...
+    end
+
+    lk = if haskey(lt, keyword)
+        lt[keyword]
+    else
+        lt[keyword] = Symbol(letter, keyword)
+    end
+
+    return lk
+end
```
BeastyBlacksmith, line 1236:
```suggestion
function get_attr_symbol(letter::Symbol, keyword::Symbol)
    return _attrsymbolcache[letter][keyword]
end
```

BeastyBlacksmith (commented):
I think it we can just generate the whole cache ahead of time to simplify access.

src/utils.jl
BioTurboNick: Good idea. I think my mindset was on avoiding unused symbols to not add to startup costs.

src/utils.jl
```diff
@@ -1218,11 +1218,15 @@ end
 const _attrsymbolcache = Dict{Symbol, Dict{Symbol, Symbol}}( 
     :x => Dict( k => Symbol("x", String(k)) for k in keys(_axis_defaults)),
     :y => Dict( k => Symbol("y", String(k)) for k in keys(_axis_defaults)),
-    :z => Dict( k => Symbol("z", String(k)) for k in keys(_axis_defaults))
+    :z => Dict( k => Symbol("z", String(k)) for k in keys(_axis_defaults)),
+    # for PyPlot
+    :label => Dict( k => Symbol("label", String(k)) for k in (:top, :bottom, :left, :right)),
```
t-bltg, line 1223: IMHO, the PR was more readable before , this is more difficult to maintain.

t-bltg: Note that using `_axis_defaults` in `utils.jl` creates a circular dependency between `utils.jl` and `args.jl`.

BeastyBlacksmith: and if you move this to `args.jl`?

t-bltg: This is horrible, but
```julia
julia> GC.enable(false); @time main(); GC.enable(true); GC.gc()
6.855578 seconds (10.66 M allocations: 627.726 MiB)
julia> GC.enable(false); @time main(); GC.enable(true); GC.gc()
  1.124249 seconds (2.49 M allocations: 150.345 MiB)
```j
First: slower, more allocs. Repeated: similar.

Comment (BeastyBlacksmith) [reply src/utils.jl:1223]: Thats somewhat surprising to me, but then it would be good to revert to the dynamic caching

Comment (t-bltg) [reply src/utils.jl]: Done

**Comment by BioTurboNick** [comment]
I found a good place to populate the dictionary up front. There's a loop at the top level of `args.jl` to create underscore aliases that loops over letters and arguments. I just expanded it. That was probably where most of them were being initialized anyway. Taking a similar approach, I added the PyPlot specific ones to `pyplot.jl`, so they're only loaded when that backend is activated. 

```
Before (dynamic):
julia> GC.enable(false); @time main(); GC.enable(true); GC.gc()
  4.060057 seconds (10.50 M allocations: 558.878 MiB, 0.19% compilation time)
julia> GC.enable(false); @time main(); GC.enable(true); GC.gc()
  0.563965 seconds (3.18 M allocations: 167.640 MiB)

After:
julia> GC.enable(false); @time main(); GC.enable(true); GC.gc()
  3.952987 seconds (10.50 M allocations: 559.009 MiB, 0.19% compilation time)
julia> GC.enable(false); @time main(); GC.enable(true); GC.gc()
  0.546460 seconds (3.18 M allocations: 167.640 MiB)

```

This is with PyPlot:
```
Before (dynamic):
julia> GC.enable(false); @time main(); GC.enable(true); GC.gc()
  9.440951 seconds (35.43 M allocations: 1.878 GiB, 0.08% compilation time)

julia> GC.enable(false); @time main(); GC.enable(true); GC.gc()
  0.519489 seconds (2.81 M allocations: 156.496 MiB)

After:
julia> GC.enable(false); @time main(); GC.enable(true); GC.gc()
  9.663443 seconds (34.90 M allocations: 1.853 GiB, 0.08% compilation time)

julia> GC.enable(false); @time main(); GC.enable(true); GC.gc()
  0.515910 seconds (2.81 M allocations: 156.496 MiB)
```

**Comment by t-bltg** [comment]
Thanks for reworking, I think this looks way neater.

File: src/args.jl [hunk 5]
```diff
@@ -2255,3 +2264,27 @@ function _splitdef!(blk, value_args, key_args)
     end
     blk
 end
+
+
+# const _attrsymbolcache = Dict{Symbol, Dict{Symbol, Symbol}}()
```
t-bltg, line 2269: I think you can remove that commented block ?

BioTurboNick: done!

t-bltg (approved):

BeastyBlacksmith (approved):

PR #3782, opened 2021-08-29
Generalize GR tick label rotations

### LINKED ISSUE #3776: [BUG] Recent update causes angled axis labels to be mis-positioned
## Details
image

It looks like the tick labels are being rotated around their center, but not offset like they used to be.

### Backends

gr (default) |   x  |     |
pyplot       |     |     | x
plotly       |   x*  |     | 
plotlyjs     |     |     | x 
pgfplotsx    |     |     | x
inspectdr    |     |     | x

* plotly crashes in `Plots.tick_padding` with `ERROR: MethodError: no method matching length(::Date)` within a call to `maximum`

### Versions

Plots.jl version: 1.21.1
Backend  version (`]st -m <backend(s)>`): 0.58.1
Output of `versioninfo()`:
```
Julia Version 1.6.2
Commit 1b93d53fc4 (2021-07-14 15:36 UTC)
Platform Info:
  CPU: Intel(R) Core(TM) i7-8700 CPU @ 3.20GHz
  WORD_SIZE: 64
  LIBM: libopenlibm
  LLVM: libLLVM-11.0.1 (ORCJIT, skylake)
Environment:
  JULIA_EDITOR = code
  JULIA_NUM_THREADS =
```

BioTurboNick:
Resolves #3776 

PR for reference images: https://github.com/JuliaPlots/PlotReferenceImages.jl/pull/109


BioTurboNick:
I think I have the offsets working well, let me know what you think.

t-bltg:
> I think I have the offsets working well, let me know what you think.

I have a few simplifications to propose, do you mind if I push the changes here after your commits ?
I don't want to mess things up :wink:.

BioTurboNick:
Sure!

t-bltg:
> Sure!

So I've simplified based on the assumption that the expected for `y` axis is what we do for `x` axis, rotated by `90°`.

Also, I've remove the offsets (sorry about the time you spent on this), based on the look of `ref29` which now looks like:
bug-rot

vs http://docs.juliaplots.org/latest/generated/gr/#gr-ref29.

BioTurboNick:
Yeah, longer labels aren't adjusted the same, I'm seeing. Might be worth adding a graph with a long tick label to the test suite, like a DateTime, which would pick up on that.

Looking into the issue.

t-bltg:
> Yeah, longer labels aren't adjusted the same, I'm seeing. Might be worth adding a graph with a long tick label to the test suite, like a DateTime, which would pick up on that.
> 
> Looking into the issue.

To me, the [rotated mwe](https://github.com/JuliaPlots/Plots.jl/issues/3776#issuecomment-907852894) and `ref29` look OK without the offsets.

BioTurboNick:
You convinced me it looks better with an offset haha. I think I have it working, and this generic positioning code allows the angle-based alignment code to be removed.

Only issue is the 3d plots' ticks aren't adjusted correctly anymore.

BioTurboNick:
And figured out a way to do it for 3d too. Not 100% happy with the offset, but I think a better solution would scale with the size of the plot? Check out the ref images and see what you think.

isentropic:
I'm a little late to the discussion, but why not go with xrotation=:auto

BioTurboNick:
> I'm a little late to the discussion, but why not go with xrotation=:auto
> 
> This seems consistent with other kws like ticks, lims.

Can you clarify? The intent of this PR is just to make sure all rotations of the ticks look reasonable.

src/backends/unicodeplots.jl
```diff
@@ -41,6 +41,7 @@ function unicodeplots_rebuild(plt::Plot{UnicodePlotsBackend})
         end
 
         kw = (
+            compact = true,
```
isentropic, line 44: I was a little confused by this line. I did not look closely, so I thought you are adding a new kw "compact" to rotate the labels automatically (but apparenetly not really)

isentropic (commented):
why is unicode affected

src/backends/unicodeplots.jl
BioTurboNick: It's a merge conflict error, will remove.

isentropic (commented):
Looks fine to me, but I'm not that proficient in GR

t-bltg:
Pros:
- The new images at https://github.com/JuliaPlots/PlotReferenceImages.jl/pull/109 look good.

Cons:
- Pardon me for saying that, but the new code is a mess :laughing: .

BioTurboNick:
Oh it's definitely a mess.

I can try to simplify it, since the output is good.

BioTurboNick:
I'm within inches of a much better positioning logic for especially 3d plots, but I keep tripping up on all the angles involved. I'll come back to this.

BioTurboNick:
Well, I don't know if it is *that* much simpler to look at..., but the 3d plots positioning is now perfect IMO.

BeastyBlacksmith:
@BioTurboNick can you make a PR with the most recent images?

BioTurboNick:
It doesn't seem like there are any changes vs. the last one.

BioTurboNick:
Oh I see, very tiny differences...

BioTurboNick:
Okay I'm not sure what's going on. I keep seeing no difference on my end.

BeastyBlacksmith:
@daschw, @t-bltg  can you have a look if you can reproduce on your machine?

t-bltg:
`1` passes, but I still don't know why `ref40` does not pass on nightly.

BeastyBlacksmith:
Thanks. Then we take it as is for now and need to look into 40 on nightly in another PR

BioTurboNick:
Thanks all!

Issue #3418, opened 2021-04-05
[BUG] ticks won't work correctly with \n with pyplot backend


vancleve:
## Details

With the `pyplot` backend, Plots.jl surrounds tick labels with `\mathregular` which breaks adding `\n` to tick labels. For example,

```
using Plots 
import PyPlot
pyplot()

x = 1:5; 

p = plot(x, x, xticks=(x, ["$i\ntext" for i in x]))
```
produces
![image](https://user-images.githubusercontent.com/1442334/113528042-70c7c600-958d-11eb-9aa5-d8c72eb852ea.png)

whereas I can easily plot the labels correctly directly using `PyPlot.jl`:
```
PyPlot.plot(x, x)
PyPlot.gca()."xaxis"."set_ticks"(1:5)
PyPlot.gca()."xaxis"."set_ticklabels"(["$i\ntext" for i in x])
```
produces
![image](https://user-images.githubusercontent.com/1442334/113528205-eb90e100-958d-11eb-9469-bb60650dbd7f.png)

Should `\mathregular` always be wrapping normal text???

(tagging @isentropic from here: https://github.com/JuliaPlots/Plots.jl/issues/2890#issuecomment-668956523)

### Versions

Plots.jl version: v1.11.2
PyPlot.jl: v2.9.0
Output of `versioninfo()`:
Julia Version 1.6.0
Commit f9720dc2eb (2021-03-24 12:55 UTC)
Platform Info:
  OS: macOS (x86_64-apple-darwin19.6.0)

isentropic:
Perhaps we need an extra kwarg to avoid this surrounding, I added a new subplot extrakw for pyplot for this purpose:
```
x = 1
p = plot(x, x, xticks=(x, ["$i\ntext" for i in x]), rawticklabels=true, extra_kwargs=:subplot)
```
that bypasses the `\mathregular` padding. More about extra_kws here http://docs.juliaplots.org/latest/input_data/#extra_kwargs

isentropic:
Currently, this change is on the master branch only. However, I think this has been a long-desired feature. Would you perhaps mind documenting this to http://docs.juliaplots.org/latest/backends/#[PyPlot](https://github.com/stevengj/PyPlot.jl)?
Basically, create a short section for extra_kwargs that we are tracking for the pyplot backend.

vancleve:
[...]

PR #3793, opened 2021-09-04
handling fonts better, moving to stable 3.4

isentropic:
Better handling fonts overall for pyplot. I've been waiting to to add this only when conda.jl starts providing 3.4 matplotlib (which is already)
Matplotlib introduces sane way of setting math_fontfamily per every text element separately, for the pyplot backend i set it to be inherited from the relevant plots text element. So mathfont will change everywhere it needs to change (per user choosing). Also so no need for latex surrounding anymore. So all in all an overall improvement.

```

julia> 

p = plot(x, 10 .^x, zline=x,xticks=(x, [raw"ttext $\alpha$" for i in x]), title=raw"test $\beta$", xlabel = raw"$\Gamma \Pi$", label = raw"text: $\Gamma \Pi$", fontfamily="serif", thickness_scaling=2)

p = plot(x, 10 .^x, zline=x,xticks=(x, [raw"ttext $\alpha$" for i in x]), title=raw"test $\beta$", xlabel = raw"$\Gamma \Pi$", label = raw"text: $\Gamma \Pi$", fontfamily="sans-serif", thickness_scaling=2)
```
image
image

### DIFF
--- a/src/backends/pyplot.jl
+++ b/src/backends/pyplot.jl
@@ -178,14 +178,19 @@ py_fillstyle(::Nothing) = nothing
 py_fillstyle(fillstyle::Symbol) = string(fillstyle)
 
 function py_get_matching_math_font(parent_fontfamily)
-    math_font_family = if parent_fontfamily == "sans-serif"
-        "dejavusans"
-    elseif parent_fontfamily == "serif"
-        "dejavuserif"
-    else
-        parent_fontfamily
-    end
-    return math_font_family
+    # matplotlib supported math fonts according to
+    # https://matplotlib.org/stable/tutorials/text/mathtext.html
+    py_math_supported_fonts = Dict{String, String}(
+        "sans-serif" => "dejavusans", 
+        "serif" => "dejavuserif", 
+        "cm" => "cm", 
+        "stix" => "stix", 
+        "stixsans" => "stixsans"
+    )
+    # Fallback to "dejavusans" or "dejavuserif" in case the parentfont is different
+    # from supported by matplotlib fonts
+    matching_font(font) = occursin("serif", lowercase(font)) ? "dejavuserif" : "dejavusans"
+    return get(py_math_supported_fonts, parent_fontfamily, matching_font(parent_fontfamily))
 end
 
 # # untested... return a FontProperties object from a Plots.Font
--- a/src/backends/pyplot.jl
+++ b/src/backends/pyplot.jl
@@ -24,11 +24,15 @@ pycollections = PyPlot.pyimport("matplotlib.collections")
 pyart3d = PyPlot.art3D
 pyrcparams = PyPlot.PyDict(PyPlot.matplotlib."rcParams")
 
-# "support" matplotlib v1.5
+
+# "support" matplotlib v3.4
+if PyPlot.version < v"3.4"
+    @warn("You are using Matplotlib $(PyPlot.version), which is no longer
+    officialy supported by the Plots community. To ensure smooth Plots.jl
+    integration update your Matplotlib library to a version >= 3.4.0")
+end
+
 set_facecolor_sym = if PyPlot.version < v"2"
-    @warn(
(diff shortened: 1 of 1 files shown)


isentropic:
Note how math font changes smoothly

isentropic:
After setting, ` Plots.pyrcparams["text.usetex"] = true` everything keeps working as intended too, so it closes the  https://github.com/JuliaPlots/Plots.jl/issues/3418

image

src/backends/pyplot.jl
```diff
@@ -824,16 +839,7 @@ function py_set_lims(ax, sp::Subplot, axis::Axis)
     getproperty(ax, Symbol("set_", letter, "lim"))(lfrom, lto)
 end
 
-function py_surround_latextext(latexstring, env)
-    if !isempty(latexstring) && latexstring[1] == '$' && latexstring[end] == '$'
-        unenclosed = latexstring[2:(end - 1)]
-    else
-        unenclosed = latexstring
-    end
-    PyPlot.LaTeXStrings.latexstring(env, "{", unenclosed, "}")
```
t-bltg, line 833: Will this PR remove support for `LaTeXStrings` ?

src/backends/pyplot.jl
```diff
@@ -173,6 +177,17 @@ end
 py_fillstyle(::Nothing) = nothing
 py_fillstyle(fillstyle::Symbol) = string(fillstyle)
 
+function py_get_matching_math_font(parent_fontfamily)
+    math_font_family = if parent_fontfamily == "sans-serif"
+        "dejavusans"
```
t-bltg, line 182: What if we want to use a font other than `DejaVu` ?

src/backends/pyplot.jl
isentropic: Nope, all the texts could be latex, matplotlib has its own mathjax like text renderer.

src/backends/pyplot.jl
isentropic: Only a few math fonts are supported (as opposed to normal fonts). Its only natural that selection of math fonts is limited

isentropic: Thuogh, if you use something "computer modern" it will try to set it as both
it is limited to "dejavu both serif and sans", stix, cm for math fonts

t-bltg: Can you make it configurable via [`extra_kwargs`](http://docs.juliaplots.org/latest/input_data/#extra_kwargs) ?

src/backends/pyplot.jl
```diff
@@ -178,14 +178,19 @@ py_fillstyle(::Nothing) = nothing
...
+        "serif" => "dejavuserif", 
+        "cm" => "cm", 
+        "stix" => "stix", 
+        "stixsans" => "stixsans"
+    )
+    # Fallback to "dejavusans" or "dejavuserif" in case the parentfont is different
+    # from supported by matplotlib fonts
+    matching_font(font) = occursin("serif", lowercase(font)) ? "dejavuserif" : "dejavusans"
+    return get(py_math_supported_fonts, parent_fontfamily, matching_font(parent_fontfamily))
 end
```
isentropic, line 194: @t-bltg I think the extrakw could be an overkill here. Only 5 mathfonts are supported by matplotlib for math fonts. I think it'd be better to check if the parentfont matches it, if not try to get a matching font based on "serif" substring

For example,
`plot(rand(10), xlabel =raw"abc $\Gamma$", fontfamily="Noto Serif")` will select "dejavuserif" for the math font. I feel this is good enough for almost everyone.

isentropic: We could update this dic further if some users prefer better font matching, i.e. create new entries

isentropic:
The main concern with this PR is the moving to a stable 3.4 version of matplotlib. This is sort of breaking for some people (only in case they'd like to combine some math+normal fonts, which is like 1% of population I think), but I think it is worthwhile overall as it fixes some issues

First class ssupport only  of the version provided by Conda.jl is right move I feel

daschw (approved):
Nice

BeastyBlacksmith (approved):

Issue #3815, opened 2021-09-15
[BUG] warn_on_unsupported does not warn on some attributes 


diegozea:
## Details

The following should warn on `line_z` and `aspect_ratio`; but it is only warning on `linestyle`:

```julia
julia> using Plots

julia> unicodeplots()
Plots.UnicodePlotsBackend()

julia> x = range(0, 2π, length=100);

julia> y = sin.(x);

julia> plot(x, y, line_z=y, aspect_ratio=1, linestyle="-")
┌ Warning: linestyle - is unsupported with Plots.UnicodePlotsBackend().  Choose from: [:auto, :solid]
└ @ Plots ~/.julia/packages/Plots/4UbNP/src/args.jl:1570
           ┌────────────────────────────────────────┐   
    2.2797 │⠀⡇⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀│ y1
           │⠀⡇⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀│   
           │⠀⡇⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀│   
           │⠀⡇⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀│   
           │⠀⡇⠀⠀⠀⠀⣀⠤⠒⠚⠉⠑⠒⠤⡀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀│   
[...]

PR #3816, opened 2021-09-15
Update args.jl

t-bltg:
Condition should be negated (https://github.com/JuliaPlots/Plots.jl/issues/3815).


BeastyBlacksmith:
Do I read it correctly, that this PR turns warnings for unsupported arguments off by default?

t-bltg:
> Do I read it correctly, that this PR turns warnings for unsupported arguments off by default?

It shouldn't. Fixed.

BeastyBlacksmith (approved):

PR #3803, opened 2021-09-14
Fix invalid axes aliases

t-bltg:
Fix https://github.com/JuliaPlots/Plots.jl/issues/3783.

PR #3824, opened 2021-09-17
Fix aliases

### LINKED ISSUE #3823: [BUG] Keyword labelfontsize stopped working in version 1.22
## Details

On version 1.22.0 the keyword "labelfontsize" stopped working and the second line in the MWE errors with `Unknown key: labelfontsize` . Previously (for example in version 1.21.3) the following adjusted the fontsize of the x- and y-labels.
```julia
using Plots
gr(labelfontsize=20)
scatter(rand(10),rand(10),xlabel="test")
```
When the keyword is used like `scatter(rand(10),rand(10),xlabel="test",labelfontsize=20)`, it is just ignored. I assume this is related to the recent changes in #3803 ?

### Backends

gr (default) |  x   |     |
pyplot       |     |     | x
plotly       |     |     | x
plotlyjs     |     |     | x
pgfplotsx    |  x   |     |
inspectdr    |     |     | x

### Versions

Plots.jl version: 1.22.0
Backend  version (`]st -m <backend(s)>`): GR v0.58.1
Output of `versioninfo()`:
Julia Version 1.6.2
Commit 1b93d53fc4 (2021-07-14 15:36 UTC)
Platform Info:
  CPU: Intel(R) Core(TM) i7-7700 CPU @ 3.60GHz
  WORD_SIZE: 64
  LIBM: libopenlibm
  LLVM: libLLVM-11.0.1 (ORCJIT, skylake)
Environment:
  JULIA_EDITOR = atom  -a
  JULIA_NUM_THREADS = 4

bernd1995:
Recent changes to the alias setup removed the possibility to have overall aliases for axis-specific keywords. This should fix this.

Introduced in #3803 and pointed out in #3823.

Fix #3823.


t-bltg:
Can you add tests cover this in https://github.com/JuliaPlots/Plots.jl/blob/82f31c48b7d81e99145e7023a0a2c191128eb892/test/test_axes.jl#L67 ?

bernd1995:
Sure will do in a bit.

bernd1995:
I wanted to add the tests now and realized that there is some problem with my approach. 

The `label` keyword becomes now an alias for `guide` which then sets both axis-labels to the stated string. So this shadows then the labeling of data sets. Therefore I added a boolean keyword to `add_axes_aliases`.

Please let me know @t-bltg if you think this is a bad idea, or we should deal with that in another way.

t-bltg:
> I added the tests now and realized that there is some problem with my approach.
> 
> The `label` keyword becomes now an alias for `guide` which then sets both axis-labels to the stated string. So this shadows then the labeling of data sets. Therefore I added a boolean keyword to `add_axes_aliases`.

This look consistent: `:label` is problematic because is should not alias the [series `:label`](https://github.com/JuliaPlots/Plots.jl/blob/9d84e53bcb5cf23986d143037c4621cafb3c1d5d/src/args.jl#L346), so your boolean kw seems justified.

Looks good to me, but let's wait for other reviews (maybe there's a better approach we don't see yet).

src/args.jl
```diff
@@ -24,8 +24,9 @@ function add_aliases(sym::Symbol, aliases::Symbol...)
     return nothing
 end
 
-function add_axes_aliases(sym::Symbol, aliases::Symbol...)
+function add_axes_aliases(sym::Symbol, aliases::Symbol...; onlyxyz::Bool = false)
     sym in keys(_axis_defaults) || throw(ArgumentError("Invalid `$sym`"))
+    !onlyxyz && add_aliases(sym, aliases...)
```
t-bltg, line 29:
```suggestion
function add_axes_aliases(sym::Symbol, aliases::Symbol...; generic::Bool = true)
    sym in keys(_axis_defaults) || throw(ArgumentError("Invalid `$sym`"))
    generic && add_aliases(sym, aliases...)
```

src/args.jl
```diff
@@ -841,7 +842,7 @@ add_aliases(:markerstrokealpha, :msa, :msalpha, :msα, :markerstrokeopacity, :ms
 add_aliases(:fillalpha, :fa, :falpha, :fα, :fillopacity, :fopacity)
 
 # axes attributes
-add_axes_aliases(:guide, :label, :lab, :l)
+add_axes_aliases(:guide, :label, :lab, :l; onlyxyz = true)
```
t-bltg, line 845:
```suggestion
add_axes_aliases(:guide, :label, :lab, :l; generic = false)
```

bernd1995:
`generic` is a much better keyword :sweat_smile: 

Thanks for that.

test/test_axes.jl
```diff
@@ -88,3 +88,30 @@ end
...
+    p = plot(1:2, gridcolor = :red)
+    @test p[1][:xaxis][:foreground_color_grid] === p[1][:yaxis][:foreground_color_grid] === RGBA{Float64}(1.,0.,0.,1.)
+    p = plot(1:2, minorgridcolor = :red)
+    @test p[1][:xaxis][:foreground_color_minor_grid] === p[1][:yaxis][:foreground_color_minor_grid] === RGBA{Float64}(1.,0.,0.,1.)
+    p = plot(1:2, grid_lw = .01)
+    @test p[1][:xaxis][:gridlinewidth] ≈ p[1][:yaxis][:gridlinewidth] ≈ .01
+    p = plot(1:2, minorgrid_lw = .01)
+    @test p[1][:xaxis][:minorgridlinewidth] ≈ p[1][:yaxis][:minorgridlinewidth] ≈ .01
+    p = plot(1:2, tickor = :out)
+    @test p[1][:xaxis][:tick_direction] === p[1][:yaxis][:tick_direction] === :out
```
BeastyBlacksmith, line 116: Would be good to also have tests for z-Axis

bernd1995: True that.

Is it okay to introduce functions like
```julia
xyzeq(p::Plot, s::Symbol, val) = p[1][:xaxis][s] == p[1][:yaxis][s] == p[1][:zaxis][s] == val
xyzid(p::Plot, s::Symbol, val) = p[1][:xaxis][s] === p[1][:yaxis][s] === p[1][:zaxis][s] === val
xyzsim(p::Plot, s::Symbol, val) = p[1][:xaxis][s] ≈ p[1][:yaxis][s] ≈ p[1][:zaxis][s] ≈ val
```
inside the testset for this purpose or should such a thing be avoided and be added explicitly?

t-bltg: Looks more readable so ok for me.
Note that I've used `≈` for floats (good practice), but it's not mandatory in that case since there are no ops on floats.

bernd1995:
I have added now a function to the tests for comparison that takes the comparison operator as argument.

Let me know what you think. If this is too involved/complicated or whatever, I'll change it to a simpler check.

test/test_axes.jl
```diff
@@ -88,3 +88,33 @@ end
     p = plot(1:2, xtickor = :out)
     @test p[1][:xaxis][:tick_direction] === :out
 end
+
+@testset "aliases" begin
+    compare(p::Plots.Plot, s::Symbol, val, op::Symbol) =
+        eval(Expr(:comparison, Meta.quot(p[1][:xaxis][s]), op, Meta.quot(p[1][:yaxis][s]), op, Meta.quot(p[1][:zaxis][s]), op, Meta.quot(val)))
```
BeastyBlacksmith, line 94:
```suggestion
    compare(p::Plots.Plot, s::Symbol, val, op) =
        op(p[1][:xaxis][s]), va) &&  op(p[1][:yaxis][s]), val) && op(p[1][:zaxis][s], val)
```
I'd avoid `eval` where I can. Seems a bit overkill to me here.

t-bltg: Even better, turn `compare` to `@compare` (that's a macro job).

bernd1995: Indeed this also works, and is probably even slightly faster. 
Will update the PR.

bernd1995: Okay I had quite a deep look into macros and think it is not (easily) possible and the suggestion by @BeastyBlacksmith looks much clearer.
From my side the current state of this PR would be mergeable.

Let me know if you still have any suggestions or an idea how the macro implementation would work.

t-bltg: Writing macros can be tricky. The current implementation is compact and readable, that's all that matters.

t-bltg (approved):

BeastyBlacksmith (approved):

PR #3838, opened 2021-09-22
Fix type inference failure in `get_clims`

### LINKED ISSUE #3837: `get_clims` as bottleneck in plot rendering
In profiling `Plots.frame` for animating a plot, I identified `get_clims` as taking up the vast majority of processing time. 2402 samples out of 2742.

I'm looking into the root cause and will update.

Possibly related: setting `fill_z` instead of `fillcolor` leads to an infinite loop (or at least a very very long run time, where the interrupt tends to break within `get_clims`).

14% of the time is spent looking up dictionary entries via `getindex(dd::RecipesPipeline.DefaultsDict, k::Symbol)` in `RecipesPipeline\src\utils.jl`
19% of the time is spent iterating tuples.
43% of the time is spent in overhead around `get_clims` calls

But most of the time is spent in recursion.

EDIT: Seems to be generic to rendering plots, not just adding to the animation.

BioTurboNick:
Resolves #3837 

The intent is to short-circuit the `clims` calculation with a `shape` series type. Each shape is a series and only has a single color (right?), so all that calculation is unnecessary.

Only issue with the current PR is that using `fill_z` instead of `fillcolor` is broken, plot shapes have no color. Has to be a way to detect this situation to allow `get_clims` to work properly?

### DIFF
--- a/src/colorbars.jl
+++ b/src/colorbars.jl
@@ -8,7 +8,7 @@ process_clims(f) = f
 function get_clims(sp::Subplot, op = process_clims(sp[:clims]))
     zmin, zmax = Inf, -Inf
     for series in series_list(sp)
-        if series[:colorbar_entry] && (series[:seriestype] != :shape || series[:fillcolor] === nothing)
+        if series[:colorbar_entry] && series[:seriestype] != :shape
             zmin, zmax = _update_clims(zmin, zmax, get_clims(series, op)...)
         end
     end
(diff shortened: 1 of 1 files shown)


BeastyBlacksmith:
> Each shape is a series and only has a single color (right?)

I don't think so, you should be able to set the color of the border separate from the fillcolor and while not implemented we shouldn't block the way to fill shapes with gradients

BioTurboNick:
Fair enough.

Taking a closer look, it seems that code is very type unstable, or inference failed. If I enforce the type of the return value as `Tuple{Float64, Float64}`, performance improves almost as much.

t-bltg (approved):
Nice, we should have more of these to be as type stable as possible.

PR #3840, opened 2021-09-23
Legend positioning for 3d plots in plotly

bernd1995:
Disclaimer: This PR is more like a personal aesthetic thing with legends for 3D plots in Plotly.

Personally I got the feeling, that (especially) for 3D-plots with a square size (e.g. `size=(400,400)`), all the legends that are considered `:inner` are a bit too high. This is mainly due to the fact that a 3D plot has for most views a large empty space in the corners.

```julia
p1=scatter(rand(10,3),rand(10,3),rand(10,3),size=(400,400));
p2=scatter(rand(10,3),rand(10,3),rand(10,3),size=(400,400),legend=:top,title="test");
```

Before PR:
newplot(1)newplot(3)

With PR:
newplot(2)newplot(4)

Additionally this resolves the issue that a `:top` label overlaps with the title (although I would discourage `:top` labels at all).

Let me know what you think :smile:

### DIFF
--- a/src/backends/plotly.jl
+++ b/src/backends/plotly.jl
@@ -411,7 +411,7 @@ end
 plotly_legend_pos(v::Tuple{S,T}) where {S<:Real,T<:Real} =
     (coords = v, xanchor = "left", yanchor = "top")
 
-plotly_legend_pos(theta::Real) = plotly_legend_pos((theta, :inner), is3d)
+plotly_legend_pos(theta::Real) = plotly_legend_pos((theta, :inner))
 
 function plotly_legend_pos(v::Tuple{S,Symbol}) where {S<:Real}
     (s, c) = sincosd(v[1])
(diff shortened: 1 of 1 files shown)


BeastyBlacksmith:
wouldn't it would make sense to have it the same for 2D or are there issues? Once putting a 3D figure next to a 2D figure in plotly works it might be difficult to align the legends otherwise

bernd1995:
The alignment is indeed a valid point.

Didn't think about that. The problem with also shifting down in the 2D case is that it then takes up too much space of the plot.

My issue is that I really think that the 3D plots are wasting a lot of space by having such a huge separation between the data and the legend. For instance when I want to use a snapshot of 3D plots in a paper they don't look nicely because of the huge amount of whitespace.

I am completely up to rework this PR but in my opinion there should be a way too (manually?) pull the legend closer to the data in the 3d case.

BeastyBlacksmith:
Both GR and pgfplotsx allow passing a tuple of numbers between 0 and 1 to place the legend absolutely, maybe that would be an option here?

bernd1995:
Indeed this does also work here.

Thanks for the help.

I changed the PR now such that only the hardcoded numbers in the code are not present anymore.

Issue #3851, opened 2021-09-28
[FR] GR legendbox changes


isentropic:
I'm not really sure how to accomplish this cleanly, but I think GR's legend box should not scale relatively as well.
https://github.com/JuliaPlots/Plots.jl/pull/3659 @daschw  now that the ticks are the same lengths, clearly looks better to me, legend box should also retain its proportions. 

Such legendbox scaling does not look good to me.
![image](https://user-images.githubusercontent.com/24591123/135051712-d776d026-3464-4d43-ad42-a30d20aea1d1.png)

isentropic:
I also think it should receive some extra horizontal padding inside. I am partial to how mpl handles this
![image](https://user-images.githubusercontent.com/24591123/135052451-e2abb1ed-be4e-4e29-92cf-fa3aa0af6d7b.png)

PR #3855, opened 2021-09-29
constant gr legendbox size

isentropic:
This is a followup to https://github.com/JuliaPlots/Plots.jl/issues/3851
Setting constant scaling factor renders:
`plot(rand(10), size=(100,400), )`
image
vs old
image


isentropic:
With such change, for example https://docs.juliaplots.org/latest/layouts/#Advanced-Layouts
looks much better:
```
l = @layout [
    a{0.3w} [grid(3,3)
             b{0.2h}  ]
]
plot(
    rand(10, 11),
    layout = l, legend = true, seriestype = [:bar :scatter :path],
    title = ["($i)" for j in 1:1, i in 1:11], titleloc = :right, titlefont = font(8)
)
```
image

vs old
image

isentropic:
By reducing legendfontsize, it looks really good now I would argue

BeastyBlacksmith:
Needs new reference images though

t-bltg:
I agree that in the first examples (https://github.com/JuliaPlots/Plots.jl/pull/3855#issue-1010663664) it looks better.

However, in the second (https://github.com/JuliaPlots/Plots.jl/pull/3855#issuecomment-929970069), the data is unreadable.
It is IMO more important to be able to interpret the data than to see the label.

I believe the old plot:
old
looks better than the new one (unusable as is):
new

BeastyBlacksmith:
I'd say putting labels inside such a tiny plot should not be done in the first place.
We should encourage users to put the legends outside of the plot in this case (if needed).

src/backends/gr.jl
```diff
@@ -1260,7 +1260,10 @@ function gr_get_legend_geometry(viewport_plotarea, sp)
         GR.restorestate()
     end
 
-    legend_width_factor = (viewport_plotarea[2] - viewport_plotarea[1]) / 45 # Determines the width of legend box
+    # legend_width_factor = (viewport_plotarea[2] - viewport_plotarea[1]) / 45 # Determines the width of legend box
+    # Let's prefer constant width instead
+    legend_width_factor = 0.02
```
t-bltg, line 1265: I'm always reticent to go back to hard-coded factors, when a dynamic solution exists.

isentropic:
@BeastyBlacksmith @t-bltg it looks very much fine if you reduce the legendfontsize. Currently however, it looks bad either way

src/backends/gr.jl
isentropic: Here 0.02 is chosen to match the same size under default plot size.
I do not see how 45 is not hardcoded

isentropic:
I just showed you the really bad exaggerated case of course, but on average-sized 1:1 aspect ratio plots, this is a definite improvement

isentropic:
There must be something off with my gr installation, as can't get to pass the tests

isentropic:
Yes I understand that, every time I do it, too many new reference images are spawned (without legend box). I expect only the reference images with the legends to change, but a lot of them change with 1e-4 tolerance, higher than default 1e-5

isentropic:
failed to rebase correctly...

PR #3868, opened 2021-10-07
Use `GR.polygonmesh3d` for `:mesh3d` seriestype

lmh91:
This is basically a followup PR to #3835.

Thanks to @jheinen GR.jl has now the function `polygonmesh3d` to plot a mesh of polygons
which would be now used for the seriestype `:mesh3d` with the GR backend. 

So the example from the [documentation](https://docs.juliaplots.org/latest/generated/plotlyjs/#plotlyjs-ref47) would look like this (with modified facecolors and alpha value):

```julia
using Plots
x = [0, 1, 2, 0]
y = [0, 0, 1, 2]
z = [0, 2, 0, 1]
i = [0, 0, 0, 1]
j = [1, 2, 3, 2]
k = [2, 3, 1, 3]

mesh3d(x, y, z; connections = (i, j, k), title = "triangles", 
    xlabel = "x", ylabel = "y", zlabel = "z", 
    legend = :none, margin = 2 * Plots.mm, fillcolor = [:blue,:red,:green,:yellow], fillalpha = 0.5 )
```
gr_plots_triangle

(The blue face is in front that is why all faces look bluish)


jheinen:
> Is there a way, we can style the edges or would you need to overlay a wireframe for that?

You can use border color (`setbordercolorind`) and border width (`setborderwidth`). So there es no need to make an additional `polyline3d` call.

lmh91:
Thanks for the info :) I added it. 

```julia
mesh3d(x, y, z; connections = (i, j, k), title = "triangles", 
    xlabel = "x", ylabel = "y", zlabel = "z", 
    linecolor = :black, lw = 3,
    legend = :none, margin = 2 * Plots.mm, fillcolor = [:blue,:red,:green,:yellow], fillalpha = 0.50 )
```

gr_plots_triangle

lmh91:
Alright, I will look into it tomorrow.

BeastyBlacksmith:
Nice work! Thank you <3

PR #3874, opened 2021-10-10
Fix for regression with colorbar limits

### LINKED ISSUE #3873: [BUG] Performance regression in colobar limits
Somehow my #3839 undid the benefits of my #3838. That last change to make it update at the right times did it...

🤷‍♂️

I'm looking into it.

BioTurboNick:
#3839 solved a slowdown when re-rendering a plot for an animation, but reversed the benefits of #3838 for rendering the original plot.

This PR does two things:
- Only updates the colorbars once when a plot is added. Previously it was running twice.
- Stores the colorlimit values in a dedicated typed dictionary; storing them in the general properties dictionaries appeared to interfere with efficient code generation.

One area of concern: Are there other times when `_update_sublot_args` gets called outside of the pipeline, when the colorbar would need updating? If so, it would skip the colorbar updating routine.

Resolves #3873

### DIFF
--- a/src/colorbars.jl
+++ b/src/colorbars.jl
@@ -5,8 +5,8 @@ process_clims(s::Union{Symbol,Nothing,Missing}) = ignorenan_extrema
 # don't specialize on ::Function otherwise python functions won't work
 process_clims(f) = f
 
-const sp_clims = Dict{Subplot, Tuple{Float64, Float64}}()
-const series_clims = Dict{Series, Tuple{Float64, Float64}}()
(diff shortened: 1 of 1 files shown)


BioTurboNick:
Though I'm still curious why all it caused a regression at all... with #3838, plotting took less than a second, but it's 3 seconds currently. Because all that first PR did was add type signatures.

BioTurboNick:
Got it down to 2 seconds by refactoring the if/else statement into functions.

BioTurboNick:
Tests are passing on my machine, so not sure what's happening here.

Issue #3884, opened 2021-10-18
[BUG] overlap of labels with LaTeXStrings


lmiq:
## Details

The labels get overlapped here:

```
using Plots, LaTeXStrings
plot(rand(10),label=L"2\pi")
plot!(rand(10),label=L"2\pi")
```
![image](https://user-images.githubusercontent.com/31046348/137723097-13908bca-f7ec-4da9-9e27-45ed9ae71eb2.png)

It is not a GR issue (alone), as with GR alone it works:

```
julia> using GR

julia> plot(randn(10,2),labels=(L"2\pi", L"2\pi"))
```

![image](https://user-images.githubusercontent.com/31046348/137723238-820cf271-afe7-42e7-bb31-9a53040ec4ae.png)

### Backends

This bug occurs on ( insert `x` below )

Backend      | yes | no  | untested
-------------|-----|-----|---------
gr (default) |  x   |     |
pyplot       |     |     |
plotlyjs     |     |     |
pgfplotsx    |     | x    |
unicodeplots |     |     |
inspectdr    |     |     |
gaston       |     |     |

### Versions

Plots.jl version:
Backend  version (`]st -m <backend(s)>`):

```
(@v1.6) pkg> st -m Plots
      Status `~/.julia/environments/v1.6/Manifest.toml`
  [91a5bcdd] Plots v1.22.6

(@v1.6) pkg> st -m GR
      Status `~/.julia/environments/v1.6/Manifest.toml`
  [28b8d3ca] GR v0.61.0

```

Output of `versioninfo()`:
```
julia> versioninfo()
Julia Version 1.6.3
Commit ae8452a9e0 (2021-09-23 17:34 UTC)
Platform Info:
  OS: Linux (x86_64-pc-linux-gnu)
  CPU: Intel(R) Core(TM) i7-8550U CPU @ 1.80GHz
  WORD_SIZE: 64
  LIBM: libopenlibm
  LLVM: libLLVM-11.0.1 (ORCJIT, skylake)

```

lmiq:
Ok, in the MWE they do not overlap, they are very, very close, such that with normal resolutions they seem to overlap.

The `legend_hfactor` works, but it is somewhat inconsistent, because this does not work:

```
julia> plot(rand(10),label=L"2\pi")

julia> plot!(rand(10),label=L"2\pi")

julia> plot!(legend_hfactor=1.5,extra_kwargs=:subplot)

```

[...]

PR #3891, opened 2021-10-20
Change default white space in legend

t-bltg:
Follow up of https://github.com/JuliaPlots/Plots.jl/pull/3598, add `5%` default whitespace.

Fix https://github.com/JuliaPlots/Plots.jl/issues/3884.

```julia
using Plots, LaTeXStrings
plot([rand(10) rand(10)],labels=[L"\textrm{ABC}" L"\textrm{DEF}"])
```
bar

- [ ] new images

### DIFF
--- a/src/backends/gr.jl
+++ b/src/backends/gr.jl
@@ -1271,7 +1271,7 @@ function gr_get_legend_geometry(viewport_plotarea, sp)
     x_legend_offset = (viewport_plotarea[2] - viewport_plotarea[1]) / 30
     y_legend_offset = (viewport_plotarea[4] - viewport_plotarea[3]) / 30
 
-    dy *= get(sp[:extra_kwargs], :legend_hfactor, 1)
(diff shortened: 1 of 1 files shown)


BeastyBlacksmith:
Personally I'd like to see this deferred to GR (since they seem to have an algorithm for this already)

t-bltg:
> Personally I'd like to see this deferred to GR (since they seem to have an algorithm for this already)

Could you point us to that algorithm ?

BeastyBlacksmith:
Well, they do have consistent spacing of their legend entries regardless of the font as seen in
```julia
julia> using Plots, LaTeXStrings
julia> Plots.GR.plot(randn(10,2),labels=(L"2\pi", L"2\pi"))

julia> Plots.GR.plot(randn(10,2),labels=("2pi", "2pi"))
```
using Plots shouldn't be a regression here.

t-bltg:
Probably because `gr_inqmathtex` seems to be provided but not used internally in `GR` (`gr_inqtext` instead).

PR #3915, opened 2021-10-31
Add `:log10`-`colorbar_scale` support in GR

lmh91:
Relates to https://github.com/JuliaPlots/Plots.jl/issues/3174
I added support for log10 scale of the colormap in heatmaps in GR. Please have a look. 

Notes:
* It is not possible that `z` contains negative values.
* In the non-uniform case, "pixels" with `z < clims[1]` are plotted white and not transparent as it seems that this is not yet supported by GR.jl (though I am not sure about that).

Some example: 

```julia
using Plots

x = 0:0.2:1;
# y = [0.0, 0.2, 0.4, 0.6]; # also works for evenly spaced heatmaps
y = [0.0, 0.15, 0.45, 0.6]; 
z = Array{Float64,2}(undef, length(y)-1, length(x)-1);
map(i -> z[i] = 10.0^(i-5), eachindex(z));

plot(
    plot(x, y, z, st=:heatmap, colorbar_scale = :identity),
    plot(x, y, z, st=:heatmap, colorbar_scale = :log10),
    plot(x, y, z, st=:heatmap, colorbar_scale = :log10, clims = (z[4], z[end-3])),
    layout = (3, 1), size = (800, 900)
)

```
clog_heatmaps_gr

### DIFF
--- a/src/backends.jl
+++ b/src/backends.jl
@@ -377,6 +377,8 @@ const _gr_attr = merge_with_base_supported([
     :colorbar,
     :colorbar_title,
     :colorbar_entry,
+    :colorbar_scale,
+    :clims,
     :fill_z,
     :line_z,
     :marker_z,
--- a/src/backends/gr.jl
+++ b/src/backends/gr.jl
@@ -569,8 +569,11 @@ function gr_draw_colorbar(cbar::GRColorbar, sp::Subplot, clims, viewport_plotare
 
     ztick = 0.5 * GR.tick(zmin, zmax)
     gr_set_line(1, :solid, plot_color(:black), sp)
+    if sp[:colorbar_scale] == :log10
+        GR.setscale(2)
+    end
     GR.axes(0, ztick, xmax, zmin, 0, 1, 0.005)
-
+    
     title = if isa(sp[:colorbar_title], PlotText)
         sp[:colorbar_title]
     else
@@ -949,6 +952,9 @@ function get_z_normalized(z, clims...)
 end
 
 function gr_clims(args...)
+    if args[1][:clims] != :auto
+        return args[1][:clims]
+    end
     lo, hi = get_clims(args...)
     if lo == hi
         if lo == 0
@@ -992,7 +998,7 @@ function gr_display(sp::Subplot{GRBackend}, w, h, viewport_canvas)
 
     # init the colorbar
     cbar = GRColorbar()
-
+    
     for series in series_list(sp)
         gr_add_series(sp, series)
         gr_update_colorbar!(cbar, series)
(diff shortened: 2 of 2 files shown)


BeastyBlacksmith:
> "pixels" with z < clims[1] are plotted white and not transparent

can you make that picking up the background color?

lmh91:
> > "pixels" with z < clims[1] are plotted white and not transparent
> 
> can you make that picking up the background color?

I changed it. 

I also rethought this and changed the behavior with `clims`:
Voxels outside the `clims` are plotted with the color corresponding to `cmin` and `cmax`. 
Only voxels where `z is NaN` are plotted transparent or with the background color. 

I think that gives the user more control. What do you think?

Here the updated and extended example:

clog_heatmaps_gr

lmh91:
By the way: I take the `log10` of the data (`z`). 
In principle, I think that is not the right approach (performance wise)
One should make a logarithmic scaled colormap. 
But I don't understand how/where `fillgrad = _as_gradient(series[:fillcolor])` is defined
or actually `series[:fillcolor]` and how it is connected to the colorbar in `gr_draw_colorbar`.

BeastyBlacksmith:
`_as_gradient` is defined here: https://github.com/JuliaPlots/Plots.jl/blob/2e67e843611b9a13c568e7343b119c036ce9bd30/src/utils.jl#L206

and `cgrad` is defined in PlotUtils.jl

oschulz:
Thanks a lot for this @lmh91 !

lmh91:
But where does `series[:fillcolor]` get initialized?
It is already a colormap (`::PlotUtils.ContinuousColorGradient`) inside `gr_draw_heatmap`. 

I would need to change `series[:fillcolor]` depending on `colorbar_scale`.

BeastyBlacksmith:
There is `get_fillcolor` which is hard to find: https://github.com/JuliaPlots/Plots.jl/blob/2e67e843611b9a13c568e7343b119c036ce9bd30/src/utils.jl#L497

and a similar construct in `_update_series_attributes!`: https://github.com/JuliaPlots/Plots.jl/blob/2e67e843611b9a13c568e7343b119c036ce9bd30/src/args.jl#L2162

lmh91:
Thanks. With your help I was able to track it back to 

https://github.com/JuliaPlots/Plots.jl/blob/fdbd2c397c28bd29bfc3bf317ff961cca2185fe7/src/args.jl#L2070

However, it turned out I still have to take the logarithm to map `z` in the `[0,1]`.

Can we take it like it is?

BeastyBlacksmith (approved):
> Pixels outside the clims are plotted with the color corresponding to cmin and cmax.

I can't find references right now, but I remember people asking for having the ability to specify a color outside of the colormap for this. But this can also come in another PR.

I am fine with merging as is.

lmh91:
Alright, then I would leave it like it is for now.

zhiyuanzhai:
Does this only work for `heatmap`? It doesn't seem to work for `contour` in my environment...

PR #3942, opened 2021-11-16
GR margin 0-point set to allow no padding

BioTurboNick:
Currently, a 2mm pad is added to all plots, and the user-provided margin is added to it.

I propose that the 0-point for margins should be the box where all ink is visible and no larger. If a user wants no padding, `0mm` is the natural choice to indicate that.

This change makes the user-provided margins override the default, rather than add to it.

Borderless plots can be generated with: `plot([1, 2, 3], [5, 3, 6], widen = false, axis = false, ticks = false, margin=0(Plots.mm))`

### DIFF
--- a/src/backends/gr.jl
+++ b/src/backends/gr.jl
@@ -780,10 +780,10 @@ function _update_min_padding!(sp::Subplot{GRBackend})
         end
     end
     # Add margin given by the user
-    leftpad   = 2mm + sp[:left_margin]
-    toppad    = 2mm + sp[:top_margin]
-    rightpad  = 2mm + sp[:right_margin]
-    bottompad = 2mm + sp[:bottom_margin]
+    leftpad   = sp[:left_margin] == :match ? 2mm : sp[:left_margin]
+    toppad    = sp[:top_margin] == :match ? 2mm : sp[:top_margin]
+    rightpad  = sp[:right_margin] == :match ? 2mm : sp[:right_margin]
+    bottompad = sp[:bottom_margin] == :match ? 2mm : sp[:bottom_margin]
     # Add margin for title
     if sp[:title] != ""
         gr_set_font(titlefont(sp), sp)


BioTurboNick:
For the tests, I suggest just changing the specified margin rather than changing the test images?

BeastyBlacksmith:
Would be good to make this consistent accross the backends

BioTurboNick:
Checking to see where `plot([1, 2, 3], [5, 3, 6], widen = false, axis = false, ticks = false, margin = 0(Plots.mm), legend = false)` produces a borderless plot:

GR - fixed in this PR
PyPlot - ✅
Plotly - Horizontally borderless, vertically big borders; NOTE: also draws tick marks when it shouldn't.
PlotlyJS - Same as Plotly
InspectDR - Huge borders on all sides
PGFPlotsX - Small borders
UnicodePlots - n/a
Gaston - Extermely slow; maybe hung? Had to terminate. NOTE: warnings for unsupported keyword arguments `fillcolor` and `markerstrokecolor`

BeastyBlacksmith:
I fixed the pgfplotsx case, @ma-laforge can you comment on InspectDR?

ma-laforge:
I just saw this question (I haven't had much time look at Github issues).

The intent of InspectDR wasn't originally to have fully controllable plots visuals, but rather have a Plots.jl compatible interface to a fast, interactive, GUI-enabled plotting tool. So I'm pretty sure I ignore many of the margin parameters.

That said, InspectDR probably ***does*** have the ability to control these margins - so it would be a good idea to support this feature.

I simply do not forsee having the time to fix this in the next few weeks. It would probably have to wait until January.

Sorry that wasn't the answer you were hoping to get - but thanks for pinging me to let me know.

BioTurboNick:
No worries, we can just move forward with the rest.

BeastyBlacksmith:
The more I think about this, the more I think we should keep that for 2.0, since I think there will be many people who currently have negative margins or the like who will get broken plots after this change.

BeastyBlacksmith:
The more I think about this, the more I think we should keep that for 2.0, since I think there will be many people who currently have negative margins or the like who will get broken plots after this change.

BioTurboNick:
Yeah, I suppose that makes sense.

iabraham:
I have spent a couple of days trying to deal with something I suspect is related [(see second image here)](https://stackoverflow.com/questions/75893295/julia-plots-xtick-and-ytick-font-halign-and-valign) but now that I chanced across this issue here; I think I should switch to a different backend (`pyplot` takes 3 minutes to make one of my plots though 😓).

Can someone confirm that the default GR behavior of _not_ truly doing zero-padding or margins (especially visible as shown in gridded layouts) is what this PR is trying to fix?

BioTurboNick:
Yeah, there are hard-coded offsets. You can provide a negative offset to compensate, it's just annoying.

iabraham:
> Yeah, there are hard-coded offsets. You can provide a negative offset to compensate, it's just annoying.

Just to confirm ... is it the `margin` keywords that should be given the negative offsets to reduce spacing? Because that changes the spacing between subplots in a gridded layout but does nothing for the padding/space between an axis and its ticklabels unlike in the case of a single plot.

PR #3943, opened 2021-11-18
rework layout macro to use Base types

BeastyBlacksmith:
this should enable users to use the `@layout` macro in recipes.

Currently this seems to work fine, except for the case of `grid` calls in the argument expression. Which is hardly documented and I am considering ~~removing this option~~ leaving things as they are, since it is usable while having Plots loaded and you can't use that in recipes either way.

Fix https://github.com/JuliaPlots/RecipesBase.jl/issues/15


BeastyBlacksmith:
Here is a case, that currently gives different results:
```julia
using Plots
plot(rand(100, 6), layout = @layout([_ ° _; ° ° °; ° ° °]), title = ["A" "B" "C"])
```
## Master
layout

## PR

plot_2

BeastyBlacksmith:
Tests need JuliaPlots/RecipesBase.jl#85 to pass

PR #3957, opened 2021-11-26
Fix annotations getting drawn under markers in pgfplotsx backend

chwons:
When adding annotations to a scatter plot, the annotations get drawn under the markers.
I added the option `clip mode = individual` to fix this.

annotation


BeastyBlacksmith:
Thanks!

PR #3966, opened 2021-12-01
test on 1.6 also

BeastyBlacksmith:
Now that 1.6 is LTS, we should support that as long as possible, to cut down on CI time, I only test LTS and nightly on ubuntu

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -40,14 +40,11 @@ jobs:
     steps:
       # Setup environment
       - uses: actions/checkout@v2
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -22,6 +22,7 @@ jobs:
       fail-fast: false
       matrix:
         version:
+          - '1.6'
```
giordano, line 25: Why 1.6 and not v1.5, since that's the oldest version you claim to be compatible with?
https://github.com/JuliaPlots/Plots.jl/blob/b1c56126fb777f9f3b09251f745b51565bb627cd/Project.toml#L58

BeastyBlacksmith:
Does anyone know, how to make it, that it only creates the jobs for all OS's on latest stable and runs on ubuntu otherwise?

Issue #3964, opened 2021-12-01
[BUG] Fails to precompile Plots.jl


ValentinKaisermayer:
```julia
julia> import Plots
[ Info: Precompiling Plots [91a5bcdd-55d7-5caf-9e0b-520d859cae80]
ERROR: LoadError: LoadError: too many parameters for type
Stacktrace:
  [1] _precompile_()
    @ Plots ***\.julia\packages\Plots\AGxtZ\deps\SnoopCompile\precompile\precompile_Plots.jl:92
  [2] top-level scope
    @ ***\.julia\packages\Plots\AGxtZ\src\precompile_includer.jl:14
  [3] include(mod::Module, _path::String)
    @ Base .\Base.jl:384
  [4] include(x::String)
    @ Plots ***\.julia\packages\Plots\AGxtZ\src\Plots.jl:1
  [5] top-level scope
    @ ***\.julia\packages\Plots\AGxtZ\src\Plots.jl:267
  [6] include
    @ .\Base.jl:384 [inlined]
  [7] include_package_for_output(pkg::Base.PkgId, input::String, depot_path::Vector{String}, dl_load_path::Vector{String}, load_path::Vector{String}, concrete_deps::Vector{Pair{Base.PkgId, UInt64}}, source::Nothing)
    @ Base .\loading.jl:1235
  [8] top-level scope
    @ none:1
  [9] eval
    @ .\boot.jl:360 [inlined]
 [10] eval(x::Expr)
    @ Base.MainInclude .\client.jl:446
 [11] top-level scope
    @ none:1
in expression starting at ***\.julia\packages\Plots\AGxtZ\src\precompile_includer.jl:9
in expression starting at ***\.julia\packages\Plots\AGxtZ\src\Plots.jl:1
ERROR: Failed to precompile Plots [91a5bcdd-55d7-5caf-9e0b-520d859cae80] to ***\.julia\compiled\v1.6\Plots\jl_9DF5.tmp.
Stacktrace:
  [1] error(s::String)
    @ Base .\error.jl:33
  [2] compilecache(pkg::Base.PkgId, path::String, internal_stderr::Base.TTY, internal_stdout::Base.TTY, ignore_loaded_modules::Bool)
    @ Base .\loading.jl:1385
  [3] compilecache(pkg::Base.PkgId, path::String)
    @ Base .\loading.jl:1329
  [4] _require(pkg::Base.PkgId)
    @ Base .\loading.jl:1043
[...]

PR #3972, opened 2021-12-01
Versioned precompile script - drop 1.5 - support 1.6 and 1.7

t-bltg:
Fix https://github.com/JuliaPlots/Plots.jl/issues/3964.

I've tested bot generated precompilation scripts on ~~`1.5.4`,~~ `1.6.4` and `1.7.0`.

~~I tried to use `yml_path` as explained in https://timholy.github.io/SnoopCompile.jl/v1.5/bot/#Configure-the-bot-to-run-with-a-GitHub-Action-file-1, but it messes with the version in `SnoopCompile.yml` (`1` interpreted as `1.0` instead of `1.7`).~~

~~With `1.6` as LTS, should we drop `1.5` in `Project.toml` ?~~


BeastyBlacksmith:
If we don't have a specific reason to drop 1.5 I wouldn't do so. I now set the tests to run on 1.5, 1.6, 1 and nightly and you may want to rebase this on master.
I'd just try if we can reuse the 1.6 precompile files on 1.5 otherwise we would have a reason.

t-bltg:
> I'd just try if we can reuse the 1.6 precompile files on 1.5 otherwise we would have a reason.

Allright, I'll rework versions.

t-bltg:
@BeastyBlacksmith, I'd prefer to use explicit minor versions instead of the majored `1` to avoid hitting https://github.com/JuliaPlots/Plots.jl/issues/3964 again.

As for reusing `1.6` precompilation files with `1.5` I'm afraid that won't be possible with the current configuration of `CompileBot`, we'd have to generate extrea precompilation files for `1.5`. Are you ok with this ?

src/precompile_includer.jl
```diff
@@ -14,4 +14,16 @@ elseif !ismultios && !ismultiversion
         _precompile_()
     end
 else
+    @static if v"1.6.0-DEV" <= VERSION <= v"1.6.9"
```
BeastyBlacksmith, line 17: Don't you just have to change this line? Or will it be overwritten once the bot runs again?

t-bltg: Yes, it will be overwritten, this file is autogenerated.

deps/SnoopCompile/snoop_bot_config.jl
```diff
@@ -0,0 +1,7 @@
+using CompileBot
+
+botconfig = BotConfig(
+    "Plots",
+    version = [v"1.6"],
+    else_version = v"1.7",
```
BeastyBlacksmith, line 6: I think I am in favor of putting `v"1.7"` also in `version`. IIUC then you might get a slower version of Plots on other versions, but can't get a broken version, or am I wrong?

t-bltg: It looks like a safer approach.

t-bltg:
I've replaced occurences of '1' in `.yml` files with lastest stable `1.7`.

I now need to regenerate `src/precompile_includer.jl`, and I think we'll be ok.

t-bltg:
@BeastyBlacksmith, the failure on `1.5` is related to an old `GR_jll` artifact (missing `gr_polygonmesh3d` symbol in artifact `0.58.1`).

What do we do ?

BeastyBlacksmith:
>  Is this a valid argument for dropping 1.5 ?

I'd say so

t-bltg:
@BeastyBlacksmith, I know we have to spare github actions CI credits, but shouldn't `windows` and `macos` also be tested on `1.6` ? Testing only `ubuntu` for `nightly` is appropriate.

BeastyBlacksmith:
Honestly I can't remember having seen a failure that is dependent on the OS __and__ the julia version.

t-bltg:
> Honestly I can't remember having seen a failure that is dependent on the OS **and** the julia version.

Me neither, however, even if unlikely, it can still happen. Your call.

With `CI` passing, this PR is now ready.

BeastyBlacksmith (approved):
Let's see how it works out

t-bltg:
Good, this looks consistent, and less likely to break `Plots` releases when `1.8.0` is out.

PR #3978, opened 2021-12-02
Fix #3967 Pluto white plot

### LINKED ISSUE #3967: [BUG] Using Plots with plotly backend only shows white.
## Details
When I use Plots with the plotly backend, I get only a white area. It works fine if I use `gr()` instead, see image:

### Backends

gr (default) |     |  x   |
pyplot       |     |     | x
plotlyjs     |  x   |     | x
pgfplotsx    |     |     | x
unicodeplots |     |     | x
inspectdr    |     |     | x
gaston       |     |     | x

### Versions

Plots.jl version: v1.24.4
Backend  version (`]st -m <backend(s)>`): v0.18.8
Output of `versioninfo()`:
```
Julia Version 1.7.0
Commit 3bf9d17731 (2021-11-30 12:12 UTC)
Platform Info:
  OS: macOS (arm64-apple-darwin21.1.0)
  CPU: Apple M1
  WORD_SIZE: 64
  LIBM: libopenlibm
  LLVM: libLLVM-12.0.1 (ORCJIT, cyclone)
```

What can I do to make this work?

jks-liu:
Bug cause:  
In JS code, we use variable of `Plotly` to ref the `plotly.js` lib, this variable has different sources in different environments.
- IJulia: use `require.js` to load `plotly.js`
- Pluto: directly use the `plotly.js`

When updating the `plotly.js` to v2.x.x, I changed the variable name loaded by `required.js` to `Plotly2` which is not the original name in `plotly.js`. So failed in Pluto.

fix #3967

### DIFF
--- a/src/backends/plotly.jl
+++ b/src/backends/plotly.jl
@@ -1079,10 +1079,10 @@ function plotly_html_body(plt, style = nothing)
         requirejs_prefix = """
             requirejs.config({
                 paths: {
-                    Plotly2: '$(plotly_no_ext)'
+                    Plotly: '$(plotly_no_ext)'
                 }
             });
-            require(['Plotly2'], function (Plotly2) {
+            require(['Plotly'], function (Plotly) {
         """
         requirejs_suffix = "});"
     end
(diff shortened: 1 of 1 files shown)


BeastyBlacksmith:
Thank you!

Issue #3986, opened 2021-12-07
[FR] xlims [ ] or () 


bienpierre:
Dear all,

First of all, thanks for the great work about Plots.jl.

I did not find the difference between [ ] or ( ) with xlims. 

I performed the comparison, one plot with [ ] and the second with ( ):

![plots_Fr](https://user-images.githubusercontent.com/50803547/145054958-77871ecb-cf92-47fd-a483-f4a128fe8e52.png)

The results are
p1) xlims = [-2, 2]: ![p1](https://user-images.githubusercontent.com/50803547/145055361-f7ad95ef-8f59-4892-9814-19cd17e8370d.png)

p2) xlims = (-2, 2): ![p2](https://user-images.githubusercontent.com/50803547/145055381-08ed1022-eef5-45e2-b0a1-5eb1564a0444.png)

xlims = [-2, 2] is accepted but the result is vague. 
Why does xlims is allowed with [ ] and not returning an error?

Regards

code: 
```

using Plots
gr()

x = randn(1000);
y = randn(1000);

x1 = sort(x, rev = true);
y1 = sort(y, rev = true);

minimum(x1)
maximum(x1)

p1 = plot(x1,y1, legend = :false, xlims = [-2,2])
	        xgrid!(p1, :on, :black, 5, 0.4)

savefig("./p1.png")


p2 = plot(x1,y1, legend = :false, xlims = (-2,2))
	        xgrid!(p2, :on, :black, 5, 0.4)

savefig("./p2.png")
```

BeastyBlacksmith:
It seems to me like `xlims = [-2,2]` behaves like `xlims = (-2,2), widen = true`

BeastyBlacksmith:
The behaviour should now that https://github.com/JuliaPlots/Plots.jl/pull/3989 is merged be consistent

PR #3989, opened 2021-12-07
Warning on bad lims argument

yha:
Would have caught the problem with boxplot range keyword (JuliaPlots/StatsPlots.jl#473)


BeastyBlacksmith (approved):

t-bltg:
Might be good to add some tests covering all cases in https://github.com/JuliaPlots/Plots.jl/blob/master/test/test_axes.jl.

yha:
Added tests, which caught a bug in `default_should_widen`, now fixed. (default was different depending if `lims` was a tuple or vector)

BeastyBlacksmith:
> (default was different depending if `lims` was a tuple or vector)

As far as I remember this was intentionally (or accidentally and deemed worth keeping), since `(1, 2)` resembles the mathematical notation for an open interval and `[1,2]` for a closed interval.

Related #3986

yha:
Seems like quite odd behavior to me. Even if you interpret `[1,2]` as a closed interval, why would that mean widening it?
The docs for `widen` also say "Defaults to `:auto`, which widens unless limits were manually set."

BeastyBlacksmith:
I am also not too big of a fan of it and if it is not documented, we can ditch it

src/axes.jl
```diff
@@ -587,6 +588,16 @@ function round_limits(amin, amax, scale)
     amin, amax
 end
 
+parse_limits(lims::Tuple{<:Real,<:Real}) = lims
```
BeastyBlacksmith, line 591: Maybe better `validate_limits`? Since we don't actually do any parsing?

yha: It's not just validating, but also translating to a specific representation (tuple of two numbers or symbol or nothing), which is "parsing" in some general sense. Or at least, I can't think of a better word, even if "parsing" is a bit of a stretch. I don't really mind about this name either way.

BeastyBlacksmith: Maybe `normalize` would also fit

PR #4027, opened 2022-01-10
Check if plot title already exists before adding again

pearlzli:
I noticed that `_add_plot_title!` doesn't check if a plot title already exists before adding a new plot title.

Old behavior:
```julia
using Plots
data = rand(10,4)
p = plot(data; layout = grid(2,2), plot_title = "My title here")
plot!(p)
```
two_titles

New behavior:
one_title

### DIFF
--- a/src/pipeline.jl
+++ b/src/pipeline.jl
@@ -277,26 +277,35 @@ end
 
 function _add_plot_title!(plt)
     plot_title = plt[:plot_title]
+
     if plot_title != ""
-        the_layout = plt.layout
-        vspan = plt[:plot_titlevspan]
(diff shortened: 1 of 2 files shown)


t-bltg:
Can you add a test, maybe in https://github.com/JuliaPlots/Plots.jl/blob/master/test/test_layouts.jl checking that `plt[:plot_title] == "My title here"`  ?

BeastyBlacksmith:
Would be alos good to have a test, that you can change the plot title like
```julia
plot(1:5, plot_title = "Wrong title")
plot!(; plot_title = "Correct title")
```

pearlzli:
Done!

BeastyBlacksmith:
Thanks!

PR #4062, opened 2022-01-24
Example for vspan and hspan

### LINKED ISSUE #4051: [FR] Filling horizontal regions across whole vertical axis range
I was looking to achieve something very similar to [this example from matplotlib](https://matplotlib.org/stable/gallery/lines_bars_and_markers/fill_between_demo.html#selectively-marking-horizontal-regions-across-the-whole-axes), whereby one can shade some horizontal extent.

I came up with the following:

```julia
function fill_xrange!(xmin, xmax; kwargs...)
    ymin, ymax = ylims()
    result = plot!([xmin, xmax], [ymin, ymin]; fillrange=[ymax, ymax], lw=0, kwargs...)
    ylims!((ymin, ymax))
    return result
end
```

... but this only is guaranteed to work if it is the last thing to be plotted (and even then it requires the rather dubious call to `ylims!` to undo the automatic zooming-out that will occur)

Is something better possible with the existing `Plots.jl` API, or would additional functionality be required?

(I marked this as a 'feature request', as I think even if it is possible, it'd be a useful example to add to the documentation!)

tpgillam:
Closes #4051.

Provide example of using vspan and hspan.

### DIFF
--- a/.zenodo.json
+++ b/.zenodo.json
@@ -702,6 +702,10 @@
             "affiliation": "Alogus Research Corporation",
             "name": "Lee Phillips",
             "type": "Other"
+        },
+        {
+            "name": "Tom Gillam",
+            "type": "Other"
         }
     ],
     "upload_type": "software"
--- a/src/examples.jl
+++ b/src/examples.jl
@@ -1269,15 +1269,13 @@ const _examples = PlotExample[
         """
         `vspan` and `hspan` can be used to shade horizontal and vertical ranges.
         """,
-        [
-            :(
-                begin
-                    hspan([1, 2, 3, 4]; label="hspan", legend=:topleft)
-                    vspan!([2, 3]; alpha=0.5, label="vspan")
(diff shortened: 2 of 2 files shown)


tpgillam:
Looks like I would need to add a new image to https://github.com/JuliaPlots/PlotReferenceImages.jl — what's the usual workflow for doing a PR like this? I'm happy to make a PR there now if we're happy with the example, which generates the following:
ref57

t-bltg:
> what's the usual workflow for doing a PR like this?

https://docs.juliaplots.org/latest/contributing/#VisualRegressionTests

tpgillam:
I've opened a PR on PlotReferenceImages.jl — but that PR will obviously need tweaking if anyone would like the example tweaked :)

t-bltg:
Looks good.
Please format your pr using `$ julia -e 'using JuliaFormatter; format(["src", "test"])'` in your `Plots.jl` fork.

BeastyBlacksmith:
Also as a first time contributor check out https://github.com/JuliaPlots/Plots.jl/issues/3503 and consider adding your information to our `.zenodo.json` file

PR #4080, opened 2022-01-28
UnicodePlots: use more `extra_kwargs`

t-bltg:
`p = plot(extra_kwargs=Dict(:subplot=>(; canvas=:dot)))`


BeastyBlacksmith:
These should be documented somewhere

PR #4081, opened 2022-01-28
Rework warn_on_unsupported

t-bltg:
I need this to go in for `extra_kwargs` to work properly in `UnicodePlots`.
`warn_on_unsupported` should be evaluated per backend.

### DIFF
--- a/src/args.jl
+++ b/src/args.jl
@@ -1588,6 +1588,8 @@ end
 const _already_warned = Dict{Symbol,Set{Symbol}}()
 const _to_warn = Set{Symbol}()
 
+should_warn_on_unsupported(::AbstractBackend) = _plot_defaults[:warn_on_unsupported]
+
 function warn_on_unsupported_args(pkg::AbstractBackend, plotattributes)
     empty!(_to_warn)
     bend = backend_name(pkg)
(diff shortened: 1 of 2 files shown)


BeastyBlacksmith:
Why is that needed? Extra keyword arguments shouldn't be treated as unsupportet arguments. Or do you have keywords that conflict with Plots default keywords?

t-bltg:
Because https://github.com/JuliaPlots/Plots.jl/blob/5facd2bf55cd165d45cdb12c466143548d34fa15/src/backends/unicodeplots.jl#L4 was a cheap way to remove unsupported arguments warnings, basically disables `extra_kwargs` by returning `nothing`.

BeastyBlacksmith:
Can't we remove that line then?

t-bltg:
> Can't we remove that line then?

I already dit it in https://github.com/JuliaPlots/Plots.jl/commit/3e8db4532bfd8f885dee8fb3a6e363aec6695bc2, replaced it with `plt.attr[:warn_on_unsupported] = false`, but that doesn't work (too late in the pipeline), and thus tons of warnings appearing again.

BeastyBlacksmith (approved):
I see, makes sense

Issue #4079, opened 2022-01-28
[FR] Enable precompilation for nightly 


KristofferC:
Developers of Julia use the time to load and create the first plot as a way of measuring latency. If the precompilation is not active on nightly that means that it will look like something has regressed even when this would not be the case if precompilation was available.

t-bltg:
I'll have a look into that.

t-bltg:
I am reopening this since `CompileBot` fails to update the precompilation files on nightly (job hangs), although I can generate them locally.

timholy:
One option is to just skip the bot and move https://github.com/JuliaPlots/Plots.jl/blob/master/deps/SnoopCompile/precompile_script.jl into the definition of `Plots` itself.

PR #4082, opened 2022-01-28
Precompilation statements for `nightly`

t-bltg:
- add nightly precompile statements (currently `1.8)`.
- update `SnoopCompile` bot to generate `nightly` statements.

Close https://github.com/JuliaPlots/Plots.jl/issues/4079.


deps/SnoopCompile/snoop_bot_config.jl
```diff
@@ -3,5 +3,5 @@ using CompileBot
 botconfig = BotConfig(
     "Plots",
     version = [v"1.6", v"1.7"],  # <<< keep versions in sync with .github/workflows/SnoopCompile.yml
-    # else_version = v"1.8",
+    else_version = "nightly",
```
BeastyBlacksmith, line 6: Would this mean, that people running julia 1.2 get the nightly precompile statements? If so, we should add `nightly` to `version`.

t-bltg: Ha, that's a good catch.

BeastyBlacksmith (approved):

Issue #4087, opened 2022-02-01
[BUG] image tick positions drifting/ misplaced


sefffal:
## Details
Tick positions for images are not placed correctly, at least with GR. The first pixel has a tick on the left, but the last pixel has it's tick on the right.  Between them, the ticks transition from one side of the pixels to the other.

First, here is a plot using heatmap that has the correct pixel locations:
```julia
heatmap(randn(10,15),ticks=1:15,color=:grey)
```
![image](https://user-images.githubusercontent.com/7330605/152051838-935a7ac2-ed3a-403f-a874-951ba0fb8639.png)

And an image series that does not:
```julia
plot(Gray.(randn(10,15)),ticks=1:15)
```
![image](https://user-images.githubusercontent.com/7330605/152051897-4de06a12-54ac-43b6-bee8-3bac2fa19a32.png)

Finally, here is an image of the same code using plotly thanks to @BioTurboNick where everything works as expected:
![image](https://user-images.githubusercontent.com/7330605/152052310-b86f8042-64e1-4679-bac3-cccc5c538f7f.png)

I came across this while working on AstroImages.jl. For us, getting ticks in the right portion of the pixel is important.
Happy to look into this myself if someone can point me in the right direction.

Thanks!

### Backends

This bug occurs on ( insert `x` below )

Backend      | yes | no  | untested
-------------|-----|-----|---------
gr (default) |  x   |     |
pyplot       |     |     | x
plotlyjs     |     | x    |
pgfplotsx    |     |     | x
unicodeplots |     |     | x
inspectdr    |     |     |x
gaston       |     |     |x

### Versions

Plots.jl version: v1.25.4
Backend  version (`]st -m <backend(s)>`): GR v0.63.1
Output of `versioninfo()`:
Julia Version 1.7.1
Commit ac5cc99908 (2021-12-22 19:35 UTC)
Platform Info:
  OS: macOS (x86_64-apple-darwin19.5.0)
  CPU: Intel(R) Core(TM) i5-8259U CPU @ 2.30GHz
  WORD_SIZE: 64
  LIBM: libopenlibm
  LLVM: libLLVM-12.0.1 (ORCJIT, skylake)
Environment:
  JULIA_EDITOR = code
  JULIA_NUM_THREADS = 4

BeastyBlacksmith:
Probably plotly is using the Plots recipe to use a heatmap for images, while GR has a dedicated implementation

t-bltg:
[...]

PR #4088, opened 2022-02-02
Fix img axes

t-bltg:
Fix https://github.com/JuliaPlots/Plots.jl/issues/4087.

```julia
using Plots

png(plot(Gray.(randn(10,15))), "foo")
```
foo

### DIFF
--- a/src/axes.jl
+++ b/src/axes.jl
@@ -677,7 +677,6 @@ end
 # we return (continuous_value, discrete_index)
 function discrete_value!(axis::Axis, dv)
     cv_idx = get(axis[:discrete_map], dv, -1)
-    # @show axis[:discrete_map], axis[:discrete_values], dv
     if cv_idx == -1
         ex = axis[:extrema]
         cv = NaNMath.max(0.5, ex.emax + 1.0)
--- a/src/examples.jl
+++ b/src/examples.jl
@@ -849,7 +849,6 @@ const _examples = PlotExample[
         [
             quote
                 begin
-                    using Plots
                     default(
                         titlefont = (20, "times"),
                         legendfontsize = 18,
@@ -1071,7 +1070,6 @@ const _examples = PlotExample[
         "",
         [
             quote
-                using Plots
                 using TestImages
                 img = testimage("lighthouse")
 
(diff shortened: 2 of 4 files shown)


t-bltg:
@sefffal, I assumed that the labels will suit you, the alternatives are.

- label vertices 0:15 (chosen here)
- labels vertices 1:16
- labels cells 1:15

Labelling `n` cells centers is strictly equivalent to labelling `n+1` vertices, so this was the easiest fix. But if that doesn't suit you, we should find a way to label cell centers as with plotly(to me this is only a matter of personal taste), if that's even possible (not sure it's doable easily).

sefffal:
Thanks @t-bltg for looking at this!
My preference would definitely be to label the cell centres, 1:15. At least for most applications in astronomy we like to think of the indices as pointing to the centres of the pixels.
But if there is another convention that is more widely used, I can work around whatever you choose. Thank you!

t-bltg:
Thanks for your reply.

What we could do right now is fix the current which is obviously wrong with `0:15`, but keep https://github.com/JuliaPlots/Plots.jl/issues/4087 opened until someone finds a more decent solution or a hack for centering ticks.

t-bltg:
@BeastyBlacksmith, whats is your opinion on this ? There is no point opening a PR in https://github.com/JuliaPlots/PlotReferenceImages.jl if this is not the fix which is going to get in.

sefffal:
A second reason to prefer labeling cells starting at 1 is that it matches nicely with Julia indexing. For example, you can access the cell at position 1,1 just by accessing `arr[1,1]`. Since Julia uses 1 based indexing I think this is more natural.

As an aside, how does this work for heatmaps? Can the tick positioning code from there be used for images as well?

t-bltg:
> As an aside, how does this work for heatmaps? Can the tick positioning code from there be used for images as well?

Looks like heatmap is centered at least for `GR`: https://docs.juliaplots.org/stable/generated/gr/#gr-ref43

BeastyBlacksmith:
I think we should make this consistent with heatmap

t-bltg:
```julia
using Plots
img = randn(10, 15)
png(plot(Gray.(img), xticks=1:15, yticks=1:10), "foo")
```

hm_img_2

@sefffal, @BeastyBlacksmith, ok now ?

sefffal:
Beautiful!

PR #4097, opened 2022-02-08
UnicodePlots: add more `extra_kwargs`

t-bltg:
After discussion on slack with @AlexisRenchon.

### DIFF
--- a/src/backends/unicodeplots.jl
+++ b/src/backends/unicodeplots.jl
@@ -73,6 +73,8 @@ function unicodeplots_rebuild(plt::Plot{UnicodePlotsBackend})
         blend &= !(quiver || contour)
 
         plot_3d && (xlim = ylim = (0, 0))  # determined using projection
+        azimuth, elevation = sp[:camera]  # PyPlot: azimuth = -60 & elevation = 30
+        projection = plot_3d ? get(sp[:extra_kwargs], :projection, :orthographic) : nothing
 
         kw = (
             compact = true,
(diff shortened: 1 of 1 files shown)


t-bltg:
@AlexisRenchon, the `extra_kwargs` syntax is a bit odd, this is how it works in `Plots`:
```julia
surface(
  -8:.5:8, -8:.5:8, (x, y) -> 15sinc(√(x^2 + y^2) / π),
  camera = (30, 30),
  extra_kwargs = Dict(:subplot => (; up = :mx, border=:bold), :series => (; colormap=:jet, lines=true))
)
```

You can try out `] dev https://github.com/JuliaPlots/Plots.jl`.

src/backends/unicodeplots.jl
```diff
@@ -184,13 +185,16 @@ function addUnicodeSeries!(
         end
     elseif st in (:surface, :wireframe)  # 3D
         colormap = get(series[:extra_kwargs], :colormap, :none)
```
BeastyBlacksmith, line 187: why do you need an extra keyword for that instead of using `get_colorgradient(series)` ?

src/backends/unicodeplots.jl
```diff
@@ -184,13 +185,16 @@ function addUnicodeSeries!(
...
+        lines = get(series[:extra_kwargs], :lines, st === :wireframe)
+        zscale = get(series[:extra_kwargs], :zscale, :identity)
```
BeastyBlacksmith, line 189: Same here, we have `sp[:zaxis][:scale]` as well as `colorbar_scale` or is it about something else?

src/backends/unicodeplots.jl
t-bltg: Because it would allow to use the colormaps defined in `UnicodePlots` which are not available in `Plots` (and which render better, because they are more suited to ansi 256 color codes (e.g. jet) on a terminal).

Issue #4092, opened 2022-02-03
Scaling of legends when composing plots into layout, v1.25


mcabbott:
Something has changed with the scaling of subplots within a layout, which causes the legend to become very wide. 

Previously, they retained the size they had within individual plots. Now, the legend within one subplot seems to depend on the size of the overall layout. Especially for scatter plots, this causes the point to be marooned out in space, far from the label. Nothing else (that I know of) scales like this -- other features of a plot remain the same when it is placed into a layout.

Example:
```julia
julia> using Plots  # with GR default

julia> p1 = plot(cumsum(rand(30, 5), dims=1); size=(300,400)) # normal short legend lines

julia> plot(p1, p1, p1, p1; size=(1200, 400), layout=(1,4))  # much longer lines in legend

julia> plot!(title="Plots v1.25.7")
```
![legend_125](https://user-images.githubusercontent.com/32575566/152367327-9331e3a0-1007-4f1b-8f52-10088bdced7a.png)

![legend_124](https://user-images.githubusercontent.com/32575566/152364867-52d78334-2168-488c-ad05-3b20609b48c8.png)

Individual plots do not seem to be as strongly affected:

![scatter300_25](https://user-images.githubusercontent.com/32575566/152462617-eba21e45-8e77-411b-8996-04c3ea3a789d.png)

![scatter300_24](https://user-images.githubusercontent.com/32575566/152462613-1f2e2575-6558-4ed3-a1d1-366474ae064d.png)

[Edit - subplot sizes corrected so that they are exactly a multiple of the layout size.]

### Backends

This bug occurs on ( insert `x` below )

Backend      | yes | no  | untested
-------------|-----|-----|---------
gr (default) |  X   |     |
pyplot       |     |     | X
plotlyjs     |     |  X   |
pgfplotsx    |     |     | X
unicodeplots |     |     | X
inspectdr    |     |     | X
gaston       |     |     | X

### Versions

```
(jl_A9O98Y) pkg> st
Status `/private/var/folders/yq/4p2zwd614y59gszh7y9ypyhh0000gn/T/jl_A9O98Y/Project.toml`
  [28b8d3ca] GR v0.63.1
  [58dd65bb] Plotly v0.4.1
  [a03496cd] PlotlyBase v0.8.18
  [91a5bcdd] Plots v1.25.7

julia> versioninfo()
Julia Version 1.8.0-DEV.1440
Commit e3b681c687 (2022-02-02 01:39 UTC)
[...]

PR #4117, opened 2022-02-23
Revert constant legend box

t-bltg:
Revert https://github.com/JuliaPlots/Plots.jl/pull/3855 (through https://github.com/JuliaPlots/Plots.jl/pull/4055), until an acceptable solution is found.

Ref: https://github.com/JuliaPlots/Plots.jl/issues/4092.

### DIFF
--- a/src/backends/gr.jl
+++ b/src/backends/gr.jl
@@ -1272,7 +1272,7 @@ function gr_get_legend_geometry(viewport_plotarea, sp)
         GR.restorestate()
     end
 
-    legend_width_factor = 0.02
+    legend_width_factor = (viewport_plotarea[2] - viewport_plotarea[1]) / 45 # Determines the width of legend box
     legend_textw = legendw
     legend_rightw = legend_width_factor
     legend_leftw = legend_width_factor * 4


BeastyBlacksmith (approved):

isentropic:
I think we should finish the discussion in the linked issue, prior to merging.

isentropic (changes requested):
Let's agree on what all people think first. This is not urgent

t-bltg:
> I think we should finish the discussion in the linked issue, prior to merging.

I don't want to throw oil on the fire, but the discussion had to be made in https://github.com/JuliaPlots/Plots.jl/pull/3855, before being merged to master and released.

> Let's agree on what all people think first. This is not urgent

I disagree, since users seem to complain about the changes, making the reversal a wished for thing .

BeastyBlacksmith:
> we had the discussion in Plots maintainers call...

Well that doesn't mean that decision is set in stone. We all make wrong decisions from time to time and it's easy to miss cases when making a change, since one is biased on the things one whish to improve.
So we did some tests in the call and it turns out we missed some pretty bad ones, especially with different plot sizes.
I know it's a bad feeling when your changes need to get reverted, but in this case it's the right thing to do, especially since the user doesn't have any way to compensate (other than not using a legend at all).
We'll find another way to improve this, heads up.

isentropic:
Alright, I have no problem with that. Im sorry I just got emotional when I felt that I'm being blamed for this. After all this is all volunteering work. I guess it got personal to me. But hey at least we fixed the testing suite

isentropic:
Let's keep the issue open, we should arrive at a better solution somehow.

t-bltg:
> Let's keep the issue open, we should arrive at a better solution somehow.

Agreed.

@BeastyBlacksmith, given the friction in the discussions, I leave the decision of merging this PR up to you. The reference images [have been updated](https://github.com/JuliaPlots/PlotReferenceImages.jl/pull/115) against this PR. Feel free to merge when ready.

Issue #4105, opened 2022-02-17
[BUG] savefig function and legend argument don't work with Plotly


nathanrboyer:
## Details
```julia
julia> using Plots

julia> plotly()
Plots.PlotlyBackend()

julia> plot(1:3, 4:6, label="y", legend=:left)

julia> savefig("test.png")
ERROR: UndefVarError: savefig not defined
Stacktrace:
 [1] _show(io::IOStream, #unused#::MIME{Symbol("image/png")}, plt::Plots.Plot{Plots.PlotlyBackend})
   @ Plots C:\Users\nboyer.AIP\.julia\packages\Plots\hxZ0l\src\backends\plotlybase.jl:21
 [2] show
   @ C:\Users\nboyer.AIP\.julia\packages\Plots\hxZ0l\src\output.jl:214 [inlined]
 [3] png(plt::Plots.Plot{Plots.PlotlyBackend}, fn::String)
   @ Plots C:\Users\nboyer.AIP\.julia\packages\Plots\hxZ0l\src\output.jl:7
 [4] savefig(plt::Plots.Plot{Plots.PlotlyBackend}, fn::String)
   @ Plots C:\Users\nboyer.AIP\.julia\packages\Plots\hxZ0l\src\output.jl:123
 [5] savefig(fn::String)
   @ Plots C:\Users\nboyer.AIP\.julia\packages\Plots\hxZ0l\src\output.jl:128
 [6] top-level scope
   @ REPL[4]:1
```
![image](https://user-images.githubusercontent.com/65452054/154568537-7666d5a4-913b-4b19-9906-96b0efacf7e4.png)

The legend keyword argument is ignored, and the `savefig` function cannot be found. `savefig` works on PlotlyJS backend, but `legend` still doesn't.

### Backends

This bug occurs on ( insert `x` below )

Backend      | yes | no  | untested
-------------|-----|-----|---------
gr (default) |     |    x |
pyplot       |     |  x   |
plotlyjs     |   x  |     |
pgfplotsx    |     |     | x
unicodeplots |     |     | x
inspectdr    |     |     | x
gaston       |     |     | x

### Versions

Plots.jl version: v1.25.10
Backend  version (`]st -m <backend(s)>`): v0.4.1
Output of `versioninfo()`:
```
julia> versioninfo()
Julia Version 1.7.1
Commit ac5cc99908 (2021-12-22 19:35 UTC)
Platform Info:
[...]

PR #4122, opened 2022-02-26
add compat entries for backends

BeastyBlacksmith:
This adds compat entries for PyPlot, PGFPlotsX, PlotlyJS and PlotlyBase.

Actually a reaction to #4105, since we can then tag a version that requires PlotlyBase 0.7 and make a version that is compatible with 0.8.

The pitfall here is that people will complain if their dependencies get hold back by compat requirements of a backend that they don't use.


BeastyBlacksmith:
@t-bltg you should know best, what the lower bound for Unicode plots is these days

t-bltg:
> @t-bltg you should know best, what the lower bound for Unicode plots is these days

The current mentioned version (`2.8`) is fine.

Regarding this PR, I guess some user won't be happy about added dependencies (especially, if I remember about `Plotly`) there are some pretty big dependencies (Kaleido_jll ? https://juliahub.com/ui/Packages/Kaleido_jll/8Y3bT/0.1.0+0) which blows up the download / update / install time for not very fast internet connections. But I understand the need for narrowing versions ...

BeastyBlacksmith:
We don't actually add the dependencies. They are added to the `test` target and only added when testing Plots.

t-bltg:
Ha sorry, I misread the changes.

PR #4148, opened 2022-03-23
Implement `aspect_ratio = :equal` for 3D Plots

### LINKED ISSUE #1949: aspect_ratio=:equal in 3d plots doesn't always work?
I have been trying to plot spheres but there are cases when the `aspect_ratio=:equal` command just doesn't work, example:
using Plots
```julia
function plotCC(pos, rad, n = 16) # Function to spheres
    u = collect(range(0,stop=2,length=n));
    v = collect(range(0,stop=1,length=n));
    x = pos * ones(n) * ones(n)' + rad * cospi.(u) * sinpi.(v)';
    y = rad * sinpi.(u) * sinpi.(v)';
    z = rad * repeat(cospi.(v)',outer=[n, 1])
    surface!(x,y,z, aspect_ratio=:equal)
end
plotlyjs(dpi=1500)
x = -15:0.1:10
y = -2:0.1:2
plot(x,y,(x,y) -> sin(x)*cos(y)*0.5, linetype=:surface, aspect_ratio=:equal, color=:magma)
plotCC(1,2,32)
plotCC(-15,3,32)
gui()
```

which will give me this:
no_AspectRatio

but if I plot the second sphere with `rad=5` then this is the result:
AspectRatio

Is this some kind of limitation of Plots.jl ? It looks like it is deciding not to do `aspect_ratio=:equal` just because the data is too stretched, but I need to plot it with equally scaled axis (or the spheres won't look like it). Any advise?

BeastyBlacksmith:
I started this some time ago with @rafael-guerra-www and got it wrapped up now.

Fix #1949


BeastyBlacksmith:
One issue here is that this doesn't work for backends that don't use a square box for 3D Plots like pgfplotsx.

lmiq:
Any chance of merging this soon? Anything that has to be done?

BeastyBlacksmith:
hmm.. its not a universal solution. So there are definitely cases where this doesn't work. On the other hand it may be better than the current status quo..

rafael-guerra-www:
Hello.
This doesn't seem to be implemented yet. Please check this recent [thread ](https://discourse.julialang.org/t/plotting-surface-without-interpolating-at-discontinuity/95279/6?u=rafael.guerra) in discourse and the MWE.

PR #4155, opened 2022-03-26
Improve coverage

t-bltg:
Add more unit tests (`+15%` coverage).

Something we should have done a long time ago ...

### DIFF
--- a/src/examples.jl
+++ b/src/examples.jl
@@ -169,7 +169,7 @@ const _examples = PlotExample[
         )],
     ),
     PlotExample( # 9
-        "",
+        "Build plot in pieces (continued)",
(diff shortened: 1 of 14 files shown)


src/examples.jl
```diff
@@ -169,7 +169,7 @@ const _examples = PlotExample[
         )],
     ),
     PlotExample( # 9
-        "",
+        "Build plot in pieces (continued)",
```
BeastyBlacksmith, line 172: Those headers will make this extra cards, I don't think there is much benefit to that

t-bltg: I didnt think of that, you're right.

PR #4156, opened 2022-03-28
Sort supported args

IanButterworth:
Adds sorting to this tip because the unsorted options are hard to scan.

Screenshot from 2022-03-27 20-02-44

### DIFF
--- a/src/args.jl
+++ b/src/args.jl
@@ -1626,7 +1626,7 @@ function warn_on_unsupported_args(pkg::AbstractBackend, plotattributes)
         for k in sort(collect(_to_warn))
             push!(already_warned, k)
             @warn(
-                "Keyword argument $k not supported with $pkg.  Choose from: $(sort(collect(supported_attrs(pkg))))"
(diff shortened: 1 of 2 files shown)


src/args.jl
```diff
@@ -1626,7 +1626,7 @@ function warn_on_unsupported_args(pkg::AbstractBackend, plotattributes)
         for k in sort(collect(_to_warn))
             push!(already_warned, k)
             @warn(
-                "Keyword argument $k not supported with $pkg.  Choose from: $(supported_attrs(pkg))"
+                "Keyword argument $k not supported with $pkg.  Choose from: $(sort(collect(supported_attrs(pkg))))"
```
IanButterworth, line 1629: Maybe it doesn't need the type info for the string array, so join as a single string?
```suggestion
                "Keyword argument $k not supported with $pkg.  Choose from: $(join(sort(collect(supported_attrs(pkg))), ", "))"
```

BeastyBlacksmith:
I think `supported_attrs` could already return the sorted vector

PR #4168, opened 2022-04-04
backends: fix Main.Plots dependency

fonsp:
I did not test this code, but this should fix https://github.com/fonsp/Pluto.jl/issues/1982


t-bltg:
Thanks !

PR #4172, opened 2022-04-05
Merge `extra_kwargs` in plotly series

antholzer:
Also merge `extra_kwargs` dict of series to `plotattributes_out` since at the moment this is only done in `plotly_layout`.

At the moment it is not possible to add e.g. [customdata](https://plotly.com/javascript/reference/scatter/#scatter-customdata) information to plotly plots without manually manipulating the returns of `plotly_series`.

I just merge all the entries of the `extra_kwargs` dict from the series into the output attributes, but I am not sure if it is alright.

### DIFF
--- a/src/backends/plotly.jl
+++ b/src/backends/plotly.jl
@@ -753,7 +753,8 @@ function plotly_series(plt::Plot, series::Series)
 
     plotly_polar!(plotattributes_out, series)
     plotly_adjust_hover_label!(plotattributes_out, series[:hover])
-    return [plotattributes_out]
+
+    return [merge(plotattributes_out, series[:extra_kwargs])]
 end
 
 function plotly_series_shapes(plt::Plot, series::Series, clims)
@@ -807,7 +808,7 @@ function plotly_series_shapes(plt::Plot, series::Series, clims)
         plotattributes_out[:showlegend] = k == 1 ? should_add_to_legend(series) : false
         plotly_polar!(plotattributes_out, series)
         plotly_adjust_hover_label!(plotattributes_out, _cycle(series[:hover], i))
-        plotattributes_outs[k] = plotattributes_out
+        plotattributes_outs[k] = merge(plotattributes_out, series[:extra_kwargs])
     end
     if series[:fill_z] !== nothing
         push!(plotattributes_outs, plotly_colorbar_hack(series, plotattributes_base, :fill))
@@ -968,6 +969,7 @@ function plotly_series_segments(series::Series, plotattributes_base::KW, x, y, z
         else
             plotattributes_outs[k] = plotattributes_out
         end
+        plotattributes_outs[k] = merge(plotattributes_outs[k], series[:extra_kwargs])
     end
 
     if series[:line_z] !== nothing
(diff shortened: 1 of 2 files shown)


BeastyBlacksmith:
Could you give an example of how you would pass the customdata information now and what the result looks like?

This should also get documented in the backends section of JuliaPlots/PlotDocs.jl

antholzer:
Now I would pass it like
```
customdata = rand(3)
p = scatter([1,2,3], [1,2,3], hover=["a", "b", "c"], 
        extra_kwargs=Dict(:series=>Dict(:customdata=>customdata)))
```
Then `Plots.plotly_series(p)[1][:customdata] == customdata`

Before one would have to manually manipulate the dict returned by `plotly_series` to insert `customdata` which can get complicated when more than one series is present.

Sadly I only now how to use click-events inside Dash. 
Example (with this PR): https://gist.github.com/antholzer/3d6237b4c505d000017cbb93060c9ce3
same example without this PR: https://gist.github.com/antholzer/e19d5ba0270616243331100cee60e06b

BeastyBlacksmith:
> In the following example the number of elements of `customdata` should be the same as the number of series (unless I made a mistake), but this is not the case (see line 43).

 [This line](https://gist.github.com/antholzer/4095cff82bf2755ebe8ea39bfb8ab792#file-plot_example-jl-L23) is appending nothing since that range is going from 1:0, as far as I can see.
 
 Since you opened this as a draft PR, what do you intend to change until this gets ready?

antholzer:
Yes the plots with fill create more series in plotly. However, even when getting the length correct, the ordering is not the same, since the plotly series are ordered w.r.t. the subplots first.

I think this PR still has a problem that one can overwrite existing values by passing `extra_kwargs` e.g. with

`scatter(1:5, label="a", extra_kwargs=Dict(:series=>Dict(:customdata => [1,2,3], :name=>nothing)))`

we overwrite the existing label. Just passing it as extra keywords should already work without a problem.
So what I plan to do is restrict it to `:customdata`.

`scatter( ...; customdata = customdata)` should only work with this PR, right?

BeastyBlacksmith:
> we overwrite the existing label. Just passing it as extra keywords should already work without a problem. So what I plan to do is restrict it to `:customdata`.

I think having a general mechamism to add extra_keywords would be very nice. This is also how it was implemented in the PGFPlotsX backend in #2200.
Of course, its possible to overwrite things, but I expect users using the extra_keywords mechanism to be somewhat experiencend and know what they do. We could add a note to the Documentation though.

However adding some tests, would be a good idea.

> `scatter( ...; customdata = customdata)` should only work with this PR, right?
Thats correct.

antholzer:
I added a similar mechanism than in https://github.com/JuliaPlots/Plots.jl/pull/2200 not sure if it is what you had in mind.

I also added some simple tests.

If it is ok, I will make a pull request to add a description on how to pass extra keywords to the series to the documentation.

BeastyBlacksmith:
That looks fine from my point of view. If you rebase this on master and resolve any remaining conflicts, you can mark this as ready for review.

Don't hesitate to ping me, if you need help with that.

antholzer:
@BeastyBlacksmith I resolved the conflict and squashed it to a single commit.

BeastyBlacksmith:
Thanks!

PR #4178, opened 2022-04-11
Add areaplot example to gallery

briochemc:
Not sure this is the right way to add example to the gallery but hopefully this is helpful :)

### DIFF
--- a/src/examples.jl
+++ b/src/examples.jl
@@ -1273,6 +1273,17 @@ const _examples = PlotExample[
             end
         )],
     ),
+    PlotExample( # 58
+        "Stacked area chart",
+        """
+        `areaplot` draws stacked area plots.
+        """,
+        [:(
+            begin
+                areaplot(1:3, [1 2 3; 7 8 9; 4 5 6], seriescolor = [:red :green :blue], fillalpha = [0.2 0.3 0.4])
+            end
+        )],
+    ),
 ]
 
 # Some constants for PlotDocs and PlotReferenceImages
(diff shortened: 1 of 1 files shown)


t-bltg:
@BeastyBlacksmith, what is the criterion for orienting contributions to the new `gallery` (old examples) or the `user gallery` ?

johnnychen94:
Here's my understanding:

- For examples that can be fit into the docstring, it goes here (the old examples); these examples have no differences from the examples listed in `?startswith` and it simply indicates that these functionalities work or don't work for certain backends.
- The new gallery is more of a "how to make a beautiful plot result" collection; it doesn't necessarily work for every backend.

briochemc:
FWIW, I submitted this because [someone on StackOverflow](https://stackoverflow.com/questions/71794565/how-to-draw-a-proportion-graph-in-julia/71828494#71828494) "looked at all of the examples in the documentation and could not find a similar graph" (similar to an `areaplot`).

t-bltg:
Good point for the docstring. We could say that the examples working with all major backends `GR`, `Plotly`, `PyPlot`, `PGFPlotsX` are allowed to go into the regular `Gallery`.

So @briochemc for this PR to go in you'd need to:
1. stick to the format (run `using JuliaFormatter; format(["src", "test"])`)
2. generate the new reference images for `gr` by running the test suite, and submitting a pr to https://github.com/JuliaPlots/PlotReferenceImages.jl (see also https://docs.juliaplots.org/stable/contributing/#VisualRegressionTests).

Issue #4183, opened 2022-04-17
[BUG]Colorbar title remains vertical with colorbar on the top


zhiyuanzhai:
## Details

Even with horizontal top colorbar `colorbar=:top`, the colorbar title is still vertical.

<img width="632" alt="image" src="https://user-images.githubusercontent.com/37478391/163706275-78408629-84ce-4044-90d2-15bf323b6547.png">

### Backends

This bug occurs on ( insert `x` below )

Backend      | yes | no  | untested
-------------|-----|-----|---------
gr (default) |     |     | x
pyplot       |     |     | x
plotlyjs     |     |     | x
pgfplotsx    |  x   |     |
unicodeplots |     |     |x
inspectdr    |     |     |x
gaston       |     |     |x

### Versions

Plots.jl version: v1.27.4
Backend  version (`]st -m <backend(s)>`): PGFPlotsX v1.4.1
Output of `versioninfo()`:
```
Julia Version 1.6.6
Commit b8708f954a (2022-03-28 07:17 UTC)
Platform Info:
  OS: macOS (x86_64-apple-darwin18.7.0)
  CPU: Intel(R) Core(TM) i7-9750H CPU @ 2.60GHz
  WORD_SIZE: 64
  LIBM: libopenlibm
  LLVM: libLLVM-11.0.1 (ORCJIT, skylake)
Environment:
  JULIA_NUM_THREADS = 
  JULIA_EDITOR = code
  JULIA_PKG_SERVER = https://pkg.julialang.org
```

zhiyuanzhai:
OK I have a question with this:
https://github.com/JuliaPlots/Plots.jl/blob/06220420a3456b690323e80891c857f7d8867681/src/backends/pgfplotsx.jl#L210
What does "triple equal sign" (`===`) here mean?

PR #4191, opened 2022-05-02
Fixing the colorbar title for colorbars on the top.

zhiyuanzhai:
Related: #4183.

For the following test code:
```julia
using Plots
pgfplotsx()
plot(rand(10, 2), marker_z=rand(10), seriestype=:scatter, cb=:top, cb_title="mark")
```
before this fix: 
截屏2022-05-02 11 31 19
while after this fix:
mark

### DIFF
--- a/.zenodo.json
+++ b/.zenodo.json
@@ -712,6 +712,10 @@
         {
             "name": "Steve Leung",
             "type": "Other"
+        },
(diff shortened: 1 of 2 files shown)


BeastyBlacksmith:
Thanks!

PR #4210, opened 2022-05-21
Replace a few uses of `1:length` idiom with `eachindex`

brenhinkeller:
Not exhaustive, but I think this catches the only places this idiom was used together with `@inbounds`. Should be a simple swap out.

### DIFF
--- a/src/utils.jl
+++ b/src/utils.jl
@@ -1192,7 +1192,7 @@ function mesh3d_triangles(x, y, z, cns::Tuple{Array,Array,Array})
     X = zeros(eltype(x), 4length(ci))
     Y = zeros(eltype(y), 4length(cj))
     Z = zeros(eltype(z), 4length(ck))
-    @inbounds for I in 1:length(ci)
+    @inbounds for I in eachindex(ci)
         i = ci[I] + 1  # connections are 0-based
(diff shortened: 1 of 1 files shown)


isentropic:
Let's merge one the tests pass

mkborregaard:
Weird - these must have been added after https://github.com/JuliaPlots/Plots.jl/pull/2304 ? Implementing the general array interface back then was a lot of effort

brenhinkeller:
Huh. Oh I'll bet.. There were a few others that popped up as well when searching https://pkg.juliahub.com/ui/Search?type=code&q=1%3Alength&r=true&p=Plots%3A91a5bcdd-55d7-5caf-9e0b-520d859cae80
but not with `@inbounds` and many not in `for` loops so I was less sure what the best replacement would be; some of those may actually be OK.

brenhinkeller:
Sure!

isentropic:
That's a little strange, not sure why such things could happen

BeastyBlacksmith:
Thanks for diggin in and since this is your first contribution, you might want to take a look at #3503 .

brenhinkeller:
Oh cool! Thanks!

PR #4216, opened 2022-05-26
CI: use julia-actions/cache

fonsp:
https://github.com/julia-actions/cache


BeastyBlacksmith:
I had to revert it for the docs build, since it did break PyPlot setup

PR #4225, opened 2022-06-02
Fix handling of multi-series `series_annotations`

### LINKED ISSUE #4073: [BUG] series_annotation does not slice correctly
## Details

This shouldn't happen
```julia
using Plots
plot(rand(4,4), layout = 2, series_annotations = [1 2; 3 4])
ERROR: type Array has no field baseshape
Stacktrace:
  [1] getproperty(x::Vector{Plots.SeriesAnnotations}, f::Symbol)
    @ Base ./Base.jl:42
  [2] series_annotations_shapes!(series::Plots.Series, scaletype::Symbol)
    @ Plots ~/.julia/packages/Plots/YAlrZ/src/components.jl:507
```

### Backends

gr (default) |   x  |     |
pyplot       |     |     | x
plotlyjs     |     |     | x
pgfplotsx    |     |     | x
unicodeplots |     |     | x
inspectdr    |     |     | x
gaston       |     |     | x

### Versions

Plots.jl version: 1.25.7
Backend  version (`]st -m <backend(s)>`): 0.63.1

nick4f42:
Currently, values for `series_annotations` can be:
- Vector of text e.g. `["foo", "bar"]`
- `(x, args...)` where `x` is a vector of text and `args` are the font/marker options
- 1-row matrix of vectors of text e.g. `permutedims([["foo"], ["bar"]])`

Changes were made for specifying annotations for multiple series:
- Matrix of text (one column for each series)
- 1-row matrix where each element is `x` or `(x, args...)` where `x` is a vector of text and `args` are the font/marker options
- A tuple `(x, args...)` where `x` is one of the previous two
  - The `args` in the inner tuples override the `args` in the outer tuple

Fixes #4073

### DIFF
--- a/.zenodo.json
+++ b/.zenodo.json
@@ -716,6 +716,10 @@
         {
             "name": "Xu Zhi-Yuan",
             "type": "Other"
+        },
+        {
+            "name": "Nick OBrien",
+            "type": "Other"
         }
     ],
     "upload_type": "software"
--- a/src/components.jl
+++ b/src/components.jl
@@ -467,10 +467,38 @@ _text_label(lab::Tuple, font) = text(lab[1], font, lab[2:end]...)
 _text_label(lab::PlotText, font) = lab
 _text_label(lab, font) = text(lab, font)
 
-series_annotations(anns::AMat) = map(series_annotations, anns)
 series_annotations(scalar) = series_annotations([scalar])
 series_annotations(anns::SeriesAnnotations) = anns
 series_annotations(::Nothing) = nothing
+
+function series_annotations(anns::AMat{SeriesAnnotations})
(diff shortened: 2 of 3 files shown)


BeastyBlacksmith:
That looks great, thank you. Can we add one more test for the matrix case?

nick4f42:
> That looks great, thank you. Can we add one more test for the matrix case?

There was already one for the matrix case but I just changed it to use `[a c; b d]` instead of `[[a, b] [c, d]]` so it's more clear that it's a matrix of scalars and not a row matrix of vectors. I also added another test case for the `tuple(text_matrix, ...)` case.

Let me know if there's anything else that should be changed!

BeastyBlacksmith:
Great, thank you!

Issue #3485, opened 2021-05-11
[BUG] aspect_ratio=:equal not respected when saving to .tex or .tikz


avonmoll:
## Details

Using the pgfplotsx backend, it appears that setting `aspect_ratio=:equal` does nothing when the plot is exported via `savefig("file.tikz")`. Saving to PDF retains the expected aspect ratio.

### Suggestion

When exporting to TiKZ or TeX, the pgfplots key `axis equal` (or equivalently `unit vector=1 1 1`) should be inserted in the `axis` options list. E.g., `\begin{axis}[unit vector=1 1 1, ...`. This probably has a lot to do with PGFPlotsX's export functionality as well.

### Backends

This bug occurs on ( insert `x` below )

Backend      | yes | no  | untested
-------------|-----|-----|---------
gr (default) |     |     | x
pyplot       |     |     | x
plotly       |     |     | x
plotlyjs     |     |     | x
pgfplotsx    |   x  |     |
inspectdr    |     |     | x

### Versions

Plots.jl version: 1.14.0
Backend  version (`]st -m`): PGFPlotsX 1.2.10
Output of `versioninfo()`:
```
Julia Version 1.6.0
Commit f9720dc2eb (2021-03-24 12:55 UTC)
Platform Info:
  OS: macOS (x86_64-apple-darwin19.6.0)
  CPU: Intel(R) Core(TM) i7-8750H CPU @ 2.20GHz
  WORD_SIZE: 64
  LIBM: libopenlibm
  LLVM: libLLVM-11.0.1 (ORCJIT, skylake)
```

PR #4231, opened 2022-06-07
forward aspect ratio for 3D plots in pgfplotsx

BeastyBlacksmith:
In 2D plots this yields sometimes undesired results as shown in https://github.com/JuliaPlots/Plots.jl/issues/3485, but in 3D its the best we have.


t-bltg:
Is it really something we wish ? 3D plots really look distorted now ...

Before PR: https://docs.juliaplots.org/v1.29/gallery/pgfplotsx
After PR: https://docs.juliaplots.org/v1.30/gallery/pgfplotsx

BeastyBlacksmith:
its a pity, those who are not messed up, look correct. Maybe we should disable it for `:auto` and only apply it on user request.

PR #4234, opened 2022-06-10
Auto-layout mode that prevents plots from shrinking beyond readability

BioTurboNick:
This is a first pass that probably doesn't account for everything needed.

However, what it does is it simply computes a row and column count, the widths and sizes of them, passes the work down to the normal layout method, then the calling plot code reads the height and width attributes from the layout and overrides `size` if the total size needed is larger than the set size.

```
plot([plot(rand(3)) for x ∈ 1:20]..., layout = :auto, legend = :none)
```
Current:
image

This PR:
image

### DIFF
--- a/src/args.jl
+++ b/src/args.jl
@@ -434,6 +434,7 @@ const _plot_defaults = KW(
     :thickness_scaling   => 1,
     :display_type        => :auto,
     :warn_on_unsupported => true,
+    :plotarea => :auto,
     :extra_plot_kwargs   => Dict(),
     :extra_kwargs        => :series,    # directs collection of extra_kwargs
 )
--- a/src/layouts.jl
+++ b/src/layouts.jl
@@ -516,7 +516,9 @@ end
 
 layout_args(n_override::Integer, layout::Union{AbstractVecOrMat,GridLayout}) =
     layout_args(layout)
-
+layout_args(n_override::Integer, mode::Symbol) =
+    mode === :auto ? (mode, n_override) : error("unhandled layout mode $(typeof(mode)): $mode") # pass-through mode
+layout_args(mode::Symbol) = mode === :auto ? (mode, 1) : error("unhandled layout mode $(typeof(mode)): $mode") # pass-through mode
 layout_args(huh) = error("unhandled layout type $(typeof(huh)): $huh")
 
 # ----------------------------------------------------------------------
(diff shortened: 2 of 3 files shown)


BioTurboNick:
Currently has a bias for taller over wider, but I can change that.

src/args.jl
```diff
@@ -434,6 +434,7 @@ const _plot_defaults = KW(
     :thickness_scaling   => 1,
     :display_type        => :auto,
     :warn_on_unsupported => true,
+    :plotarea => :auto,
```
BeastyBlacksmith, line 437: We already have a notion for plotarea with a slightly different meaning
https://github.com/JuliaPlots/Plots.jl/blob/6266ff8d08e98f2f3980731174cb1e2a631fd49a/src/types.jl#L39

IIIUC, your plotarea is the whole area covered by one subplot, whereas the subplot field is the area of the axes and data without labels and margins and stuff

BeastyBlacksmith:
In general this is a different thing, than I had in mind with `size = :auto` from our conversation on slack, but having a way to say "I don't care about the specific layout" seems useful to me nonetheless.

src/args.jl
BioTurboNick: Ah hmm. What would you recommend?

BioTurboNick:
Hmm ok. Two of my pain points I'd like to resolve is being able to specify rows/columns for a grid without having to provide an exact number of plots to fill it, and prioritizing visibility of many plots over a fixed overall plot area.

How can this fit in with what you have in mind, so that they don't conflict?

BeastyBlacksmith:
> being able to specify rows/columns for a grid without having to provide an exact number of plots to fill it,

can you give an example of how you would use it, if it where implemented?

What I don't like here is that the plot size changes without the user explicitly requesting it.
Here is one idea how we could handle things:
- `layout = :auto` creates a layout that fits all plots "optimally" in the given size
- `size = :auto` indicates that the overall plot size is allowed to change

Then you could have an automatic layout that changes plot size with `layout = :auto` and `size = :auto`. It might also be an option to check whether we can use `GridLayoutBase.jl`.

BioTurboNick:
> > being able to specify rows/columns for a grid without having to provide an exact number of plots to fill it,
> 
> can you give an example of how you would use it, if it where implemented?

I'm building Pluto notebooks for our analysis pipeline, and there can be a variable number of samples per experiment. I may know that I can comfortably fit 3 plots across but otherwise don't care how many rows it takes. Or vice versa for columns. A more specific example might be a linear trace for each event. They can get pretty short and still be legible because the length matters more.

BeastyBlacksmith:
> > > being able to specify rows/columns for a grid without having to provide an exact number of plots to fill it,
> > 
> > 
> > can you give an example of how you would use it, if it where implemented?
> 
> I'm building Pluto notebooks for our analysis pipeline, and there can be a variable number of samples per experiment. I may know that I can comfortably fit 3 plots across but otherwise don't care how many rows it takes. Or vice versa for columns. A more specific example might be a linear trace for each event. They can get pretty short and still be legible because the length matters more.

I meant in terms of function calls.

t-bltg:
This won't be breaking would it ?

BioTurboNick:
I think perhaps this might be better considered as part of a revamped layout system for 2.0 than to solve here.

PR #4235, opened 2022-06-11
Even faster series adding

BioTurboNick:
This eliminates the last bit of unnecessary churn in the clims updating code.

Previous work ensured the updates only occur at series addition or other parameter change, rather than every time they need to be read.

This work ensures the other series are only updated if a new series would change the existing color limits.

Currently my US county plot takes `1.569 s (12306720 allocations: 342.66 MiB)`

With this PR: `553.153 ms (2647060 allocations: 121.55 MiB)`

FINALLY,

The profiler bottleneck is no longer colorbar updates, but is now something in argument processing:
```
Overhead ╎ [+additional indent] Count File:Line; Function
=========================================================
  ╎330 @Base\task.jl:482; (::VSCodeServer.var"#60#61")()
  ╎ 330 @VSCodeServer\src\eval.jl:34; macro expansion
  ╎  330 @Base\essentials.jl:726; invokelatest(::Any)
  ╎   330 @Base\essentials.jl:729; #invokelatest#2
  ╎    330 @VSCodeServer\src\repl.jl:124; (::VSCodeServer.var"#77#79"{Module, Expr, REPL.LineEdit...
  ╎     330 @Base\logging.jl:623; with_logger
  ╎    ╎ 330 @Base\logging.jl:511; with_logstate(f::Function, logstate::Any)
  ╎    ╎  330 @VSCodeServer\src\repl.jl:123; (::VSCodeServer.var"#78#80"{Module, Expr, REPL.LineEdi...
  ╎    ╎   330 @VSCodeServer\src\repl.jl:157; repleval(m::Module, code::Expr, #unused#::String)
  ╎    ╎    330 @Base\Base.jl:65; eval
  ╎    ╎     330 @Base\boot.jl:368; eval
  ╎    ╎    ╎ 330 @Plots\src\plot.jl:82; (::RecipesBase.var"#plot##kw")(::NamedTuple{(:fillco...
  ╎    ╎    ╎  330 @Plots\src\plot.jl:91; #plot#143
  ╎    ╎    ╎   330 @Plots\src\plot.jl:208; _plot!(plt::Plots.Plot, plotattributes::Any, args:...
  ╎    ╎    ╎    284 ...sPipeline\src\RecipesPipeline.jl:97; recipe_pipeline!(plt::Any, plotattributes::Any, ar...        
  ╎    ╎    ╎     277 ...pesPipeline\src\series_recipe.jl:27; _process_seriesrecipes!(plt::Any, kw_list::Any)
[...]

PR #4236, opened 2022-06-12
Even fasterer series adding

BioTurboNick:
While this isn't dependent on #4235 , I will describe its benefits based on that PR.

Currently, every time the series index is needed, it is looping over all series. This gets prohibitive for many series.

This PR's goal is to store the index and only recompute it when something changes.

~~A change would generally only occur when the `:primary` attribute of a series is changed, but I'm currently unsure of how to detect that and where to place the update call.~~

~~Adding the update code will add a bit more time, but just by eliminating the looping:~~

With #4235: `613.605 ms (2640843 allocations: 121.40 MiB)`

With this PR: `344.295 ms (2655830 allocations: 122.06 MiB)`

This PR accomplishes making series-adding an O(n) operation, rather than O(n^2):

```
1.29.0:
julia> @btime plot(rand(2, 1000))
  181.186 ms (1437315 allocations: 53.25 MiB)

julia> @btime plot(rand(2, 10000))
  13.825 s (59428901 allocations: 1.86 GiB)

Soon:
julia> @btime plot(rand(2, 1000))
  99.954 ms (950287 allocations: 38.42 MiB)

julia> @btime plot(rand(2, 10000))
  1.104 s (9572373 allocations: 385.65 MiB)
```

### DIFF
--- a/src/args.jl
+++ b/src/args.jl
@@ -2164,7 +2164,6 @@ function _update_series_attributes!(plotattributes::AKW, plt::Plot, sp::Subplot)
     pkg = plt.backend
     globalIndex = plotattributes[:series_plotindex]
     plotIndex = _series_index(plotattributes, sp)
-    println(plotIndex)
 
     aliasesAndAutopick(
         plotattributes,
--- a/src/args.jl
+++ b/src/args.jl
@@ -2161,7 +2161,7 @@ label_to_string(label, series_plotindex) = string(label)  # Fallback to string p
 function _update_series_attributes!(plotattributes::AKW, plt::Plot, sp::Subplot)
     pkg = plt.backend
     globalIndex = plotattributes[:series_plotindex]
-    plotIndex = _series_index(plotattributes, sp)
+    plotIndex = _series_index(plotattributes)
 
     aliasesAndAutopick(
         plotattributes,
@@ -2244,20 +2244,18 @@ function _update_series_attributes!(plotattributes::AKW, plt::Plot, sp::Subplot)
     plotattributes
 end
 
-function _series_index(plotattributes, sp)
+function update_series_indexes(sp)
     idx = 0
     for series in series_list(sp)
         if series[:primary]
             idx += 1
-        end
-        if series == plotattributes
-            return idx
+            series[:series_index] = idx
         end
     end
-    if get(plotattributes, :primary, true)
-        idx += 1
-    end
-    return idx
+end
+
+function _series_index(plotattributes)
+    return plotattributes[:series_index]
 end
 
 #--------------------------------------------------
--- a/src/args.jl
+++ b/src/args.jl
@@ -2073,7 +2073,6 @@ function _update_subplot_args(
 
     _update_subplot_periphery(sp, anns)
     _update_subplot_colors(sp)
-    #_update_subplot_series_indexes(sp)
 
     lims_warned = false
     for letter in (:x, :y, :z)
--- a/src/args.jl
+++ b/src/args.jl
@@ -2250,10 +2250,9 @@ function _series_index(plotattributes, sp)
     if haskey(plotattributes, :series_index)
         return plotattributes[:series_index]::Int
     elseif get(plotattributes, :primary, true)
-        return sp[:max_series_index] =
-            plotattributes[:series_index] = sp[:max_series_index]::Int + 1
+        return plotattributes[:series_index] = sp.primary_series_count += 1
     else
-        return plotattributes[:series_index] = sp[:max_series_index]::Int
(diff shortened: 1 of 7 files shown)


src/args.jl
```diff
@@ -2161,7 +2161,7 @@ label_to_string(label, series_plotindex) = string(label)  # Fallback to string p
 function _update_series_attributes!(plotattributes::AKW, plt::Plot, sp::Subplot)
     pkg = plt.backend
     globalIndex = plotattributes[:series_plotindex]
-    plotIndex = _series_index(plotattributes, sp)
```
BeastyBlacksmith, line 2164: Something is not cohesive here: You delete `sp` here, but `plotattributes` in the definition

BioTurboNick: I think perhaps you misread? I split updating from getting the index. Currently not sure where to put the update call.

BeastyBlacksmith: now I see, thanks for clarifying

src/pipeline.jl
```diff
@@ -363,6 +363,7 @@ function RecipesPipeline.add_series!(plt::Plot, plotattributes)
             plotattributes[letter2], plotattributes[letter1]
     end
     _expand_subplot_extrema(sp, plotattributes, plotattributes[:seriestype])
+    plotattributes[:series_index] = 1
```
BeastyBlacksmith, line 366: This gets caught as an extra keyword argument

BioTurboNick: Thanks, this was really just a placeholder to make things not break. What does that mean here, though? Would I just have to register it in `args.jl` somewhere? And at what point does it process that? (I ask in part because in another PR I previously added, I have a `:clims_calculated` key used internally, but I don't think I added it there.)

BioTurboNick:
Thought I had the right spot, but didn't. Seems like I might have to store a primary series count in the subplot to refer back to?

An underlying reason for slowness in many of these cases though is looking things up in the attributes dictionary, I think.

BioTurboNick:
New version: maximum is stored in subplot attributes, incremented every time a new `:primary` series is added. No update loops needed. Series index is assigned the maximum value. I believe this matches the original behavior, where non-`:primary` series following a `primary` series all have the same index.

`_series_index` was only ever read when series were being added. So I think this design is safe?

BioTurboNick:
> I added some tests that are currently failing. I might have misunderstood how the new attributes should work, in that case it would be good, if you altered the tests otherwise those would need to get fixed.
> 
> In general its advisable to add tests, if you are doing performance improvements (which I really appreciate), but its easy to break things doing so.

Absolutely, I'm just not familiar enough to know what else to test. As I'm trying to replicate the existing behavior but more efficient, I'm relying on the current tests to capture that.

BeastyBlacksmith:
This is actually a fairly good test case:
```julia
using Plots
plot(rand(4,4), layout=2)
```
This shows two plots having blue and orange curves, but on this branch all lines are blue.

BioTurboNick:
I'm confused, e96ce83438aba68ec0c683cf6d85789f7014a8ad removed the offending line, but I still see it here.

BioTurboNick:
I think the merge you did resurrected some old state.

BeastyBlacksmith:
Almost all attributes are persistent. One of the few examples is `:letter` that gets added before the recipe pipeline and removed afterwards, so you can know which axis is being processed for certain recipes.

https://github.com/JuliaPlots/RecipesPipeline.jl/blob/00ed890cb63f5d65183f0a10f4172f5124a2f746/src/api.jl#L112

BeastyBlacksmith:
I think it makes more sense to store these values as fields of `Subplot` and `Series` objects rather than register them as attributes.
Otherwise you are able to do things like
```julia
using Plots
plot(rand(4,4), layout=2)
plot!(rand(4), max_series_index = 5)
plot!(rand(4), subplot=2, series_index = 15)
```

src/args.jl
```diff
@@ -2245,19 +2247,14 @@ function _update_series_attributes!(plotattributes::AKW, plt::Plot, sp::Subplot)
...
-            return idx
-        end
-    end
-    if get(plotattributes, :primary, true)
-        idx += 1
+    if haskey(plotattributes, :series_index)
+        return plotattributes[:series_index]::Int
+    elseif get(plotattributes, :primary, true)
+        return sp[:max_series_index] =
+            plotattributes[:series_index] = sp[:max_series_index]::Int + 1
```
BeastyBlacksmith, line 2254: Do we actually need `max_series_index` computing it should be as fast, if I'm not mistaken.
```suggestion
        return plotattributes[:series_index] = length(sp.series_list)::Int + 1
```

PR #4247, opened 2022-06-22
Make layouts tolerant of having too-few items

BioTurboNick:
A big headache I have is that you currently have to have exactly the right number of layout spots for the number of plots you want to render. It makes rendering variable numbers of plots a bit of a headache.

This PR loosens the constraints so that fewer plots are acceptable, while too many plots still errors.
```
plot(plot(), plot(), plot(), layout = grid(2, 2))
```
image


BioTurboNick:
I see it passes all current tests, so I'll add one to enforce the permissive behavior if it otherwise checks out.

PR #4254, opened 2022-06-28
Deal with unknown themes

diegozea:
This will make `theme` to fallback to `:default` If the requested theme is unknown, in that case, it will give a proper warning: https://github.com/JuliaPlots/PlotThemes.jl/issues/68


t-bltg (commented):
Maybe add some tests in https://github.com/JuliaPlots/Plots.jl/blob/2ab3e9b3cb655b4c8dc283fb496e51a1d7167fe3/test/test_misc.jl#L62 ?

PR #4269, opened 2022-07-10
add plot-to-png-within-IOBuffer method

wherrera10:
This patch is for situations where it is desired to share a plot image to another function, such as:
```julia
using Cairo
using Gtk
using Plots

const io = IOBuffer()

function plotincanvas(h = 900, w = 800)
    win = GtkWindow("Normal Histogram Widget", h, w) |> (vbox = GtkBox(:v) |> (slide = GtkScale(false, 1:500)))
    Gtk.G_.value(slide, 250.0)
    can = GtkCanvas()
    push!(vbox, can)
    set_gtk_property!(vbox, :expand, can, true)
    @guarded draw(can) do widget
        ctx = getgc(can)
        n = Int(Gtk.GAccessor.value(slide))
        png(histogram(randn(n)), io)
        img = read_from_png(io)
        set_source_surface(ctx, img, 0, 0)
        paint(ctx)
    end
    id = signal_connect((w) -> draw(can), slide, "value-changed")
    showall(win)
    show(can)
end

plotincanvas()
```

### DIFF
--- a/src/output.jl
+++ b/src/output.jl
@@ -17,12 +17,16 @@ svg(plt::Plot, fn::AbstractString) =
 
 svg(fn::AbstractString) = svg(current(), fn)
 
+svg(plt::Plot, io::IOBuffer) = (seekstart(io); show(io, MIME("image/svg"), plt); seekstart(io))
+
 pdf(plt::Plot, fn::AbstractString) =
     open(addExtension(fn, "pdf"), "w") do io
         show(io, MIME("application/pdf"), plt)
     end
 
 pdf(fn::AbstractString) = pdf(current(), fn)
+
+pdf(plt::Plot, io::IOBuffer) = (seekstart(io); show(io, MIME("application/pdf"), plt); seekstart(io))
 
 ps(plt::Plot, fn::AbstractString) =
     open(addExtension(fn, "ps"), "w") do io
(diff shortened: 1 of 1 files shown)


src/output.jl
```diff
@@ -8,6 +8,8 @@ png(plt::Plot, fn::AbstractString) =
 
 png(fn::AbstractString) = png(current(), fn)
 
+png(plt::Plot, io::IOBuffer) = (seekstart(io); show(io, MIME("image/png"), plt); seekstart(io))
```
t-bltg, line 11: Why are the `seekstart(io)` required ?

I'd rather have something like:
```julia
png(io::IO, plt::Plot) = show(io, MIME("image/png"), plt)
```

wherrera10: IOBuffers are strange. Sequential reads following writes and subsequent writes all start at buffer's end. The seekstarts are to make sure the read_from_png() gets the proper data.

I agree we could add for all. I did not know for sure which backends supported what types.  Would a PR to output.jl be able to assume that all types were supported?

wherrera10: In order to have Cairo's read_from_png() work with an IO, the png must be the entire contents of the IO, as far as I can see.

This means that if we generalize the methods to work with any IO we are going to clobber the contents of a file handle, if that is the IO the function gets from user. I wonder if we may prevent mistakes if we keep the IO as an IOBuffer. I could rename the `io` term to `buffer` to emphasize this. Or do you think that write image to any stream (IO) is a feature we should want, and users should be able to decide if they want to clobber an existing stream?

t-bltg: > IOBuffers are strange. Sequential reads following writes and subsequent writes all start at buffer's end. The seekstarts are to make sure the read_from_png() gets the proper data.

Did you try using a [`PipeBuffer`](https://docs.julialang.org/en/v1/base/io-network/#Base.PipeBuffer) instead ?

wherrera10: Hm, a PipeBuffer works without a seek.  So either we could specify a PipeBuffer and not use seekstart, or keep the seeks and use a generic IO (a rw file handle would need the seeks).

t-bltg: I'd prefer if we used the generic `io::IO`, without any `seekstart` and it would be up to the user to either use a `PipeBuffer`or use a `IOBuffer` and surround the `png` call with `seekstart` if necessary.

t-bltg: Corresponding tests should be added to https://github.com/JuliaPlots/Plots.jl/blob/6087ab960fac0587c9e8ccca3c5a89b639a106ac/test/test_output.jl#L5.

t-bltg:
Superseded by https://github.com/JuliaPlots/Plots.jl/pull/4301.

PR #4272, opened 2022-07-11
Add output to IO methods

wherrera10:
This was discussed as PR #4269 and was revised accordingly.

### DIFF
--- a/src/output.jl
+++ b/src/output.jl
@@ -8,12 +8,18 @@ png(plt::Plot, fn::AbstractString) =
 
 png(fn::AbstractString) = png(current(), fn)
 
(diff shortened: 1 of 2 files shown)


t-bltg:
Why did you close and not edit https://github.com/JuliaPlots/Plots.jl/pull/4269 ?

wherrera10:
Oh, I had been editing in Github and the number of revisions at my fork seemed to be a lot, so I thought it would look better to have fewer updates to review.

t-bltg:
It is better to keep the discussions in a single PR for history.
We have tools to squash all the commits into one on merge when the PR will be finalized and approved.
Let's close this one and continue in https://github.com/JuliaPlots/Plots.jl/pull/4269.

Issue #4273, opened 2022-07-12
[BUG] 4243 severely impacted contour subplots


charleskawczynski:
## Details

#4243 made my plots in [TurbulenceConvection.jl](https://github.com/CliMA/TurbulenceConvection.jl) go from 

<img width="1684" alt="Screen Shot 2022-07-12 at 7 17 52 AM" src="https://user-images.githubusercontent.com/1880641/178520629-9077160b-8b2a-4179-ae6f-b1d81e832a16.png">

to 

<img width="1197" alt="Screen Shot 2022-07-12 at 7 18 04 AM" src="https://user-images.githubusercontent.com/1880641/178520656-5ef11452-32ae-49ad-8f3f-236753d09583.png">

It's not clear how to fix this based on the PR comments / documentation. Can someone provide some guidance?

### Backends

This bug occurs on ( insert `x` below )

Backend      | yes | no  | untested
-------------|-----|-----|---------
gr (default) |  x  |     |
pyplot       |     |     | x
plotlyjs     |     |     | x
pgfplotsx    |     |     | x
unicodeplots |     |     | x
inspectdr    |     |     | x
gaston       |     |     | x

### Versions

Plots.jl version:
Backend  version (`]st -m <backend(s)>`):
Output of `versioninfo()`:

```
julia> versioninfo()
Julia Version 1.7.3
Commit 742b9abb4d (2022-05-06 12:58 UTC)
Platform Info:
  OS: macOS (x86_64-apple-darwin21.4.0)
  CPU: Intel(R) Core(TM) i5-7500 CPU @ 3.40GHz
  WORD_SIZE: 64
  LIBM: libopenlibm
  LLVM: libLLVM-12.0.1 (ORCJIT, skylake)
Environment:
  JULIA_EDITOR = subl
  JULIA_PKG_DEVDIR = /dev/julia
```

BeastyBlacksmith:
Do you have a code to run to reproduce a similar figure?

charleskawczynski:
I don't really have an MWE, but this should work:


Before #4243
```
git clone https://github.com/CliMA/TurbulenceConvection.jl
cd TurbulenceConvection.jl
git checkout ca7bd7108a8c5df7e89e9a7ce479a7fc9c2a71ca
julia --project=integration_tests/
include(joinpath("integration_tests", "driver.jl"))
```

After #4243
```
git clone https://github.com/CliMA/TurbulenceConvection.jl
[...]

PR #4275, opened 2022-07-15
keep largest size of subplots when plotting plots

BeastyBlacksmith:
Another option would be to actually add sizes when plotting plots, but I think thats breaking.

Fix: #4273


t-bltg:
Could we add a non-regression test for that ?

PR #4287, opened 2022-07-27
Remove GeometryBasics/StaticArrays

sjkelly:
I realize GeometryBasics is barely used here. It is a quite costly dependency, and is only used in a few places, where a simple Tuple seems to suffice. Removing it as a dependency completely removes the need for Plots to depend on StaticArrays, shaving a substantial amount of time off `using`:

Before:
```
  3.803301 seconds (8.99 M allocations: 601.643 MiB, 10.07% gc time, 20.31% compilation time)
elapsed time (ns): 3803300770
gc time (ns):      382894522
bytes allocated:   630868505
pool allocs:       8986011
non-pool GC allocs:1970
malloc() calls:    201
realloc() calls:   4
GC pauses:         7
full collections:  2

---

julia> @timev using Plots
  4.005920 seconds (8.99 M allocations: 602.993 MiB, 10.34% gc time, 21.30% compilation time)
elapsed time (ns): 4005919780
gc time (ns):      414263105
bytes allocated:   632284409
pool allocs:       8990504
non-pool GC allocs:2083
malloc() calls:    201
realloc() calls:   4
GC pauses:         6
full collections:  1
```

After:
```
  2.442962 seconds (5.31 M allocations: 341.848 MiB, 6.64% gc time, 32.28% compilation time)
elapsed time (ns): 2442961858
gc time (ns):      162169900
bytes allocated:   358453201
pool allocs:       5304483
non-pool GC allocs:1820
malloc() calls:    133
realloc() calls:   4
GC pauses:         6
full collections:  1

----

julia> @timev using Plots
  2.842176 seconds (5.31 M allocations: 342.093 MiB, 14.46% gc time, 27.25% compilation time)
elapsed time (ns): 2842176139
gc time (ns):      411022416
bytes allocated:   358710657
pool allocs:       5304476
non-pool GC allocs:1834
malloc() calls:    133
realloc() calls:   4
GC pauses:         6
full collections:  2
```


src/examples.jl
```diff
@@ -950,17 +950,6 @@ const _examples = PlotExample[
             end,
         ],
     ),
-    PlotExample( # 46
```
sjkelly, line 953: I think removing this completely breaks the reference images...

src/utils.jl
```diff
@@ -209,11 +209,6 @@ maketuple(x::Real) = (x, x)
 maketuple(x::Tuple{T,S}) where {T,S} = x
 
 RecipesPipeline.unzip(v) = unzip(v)
-RecipesPipeline.unzip(points::AbstractVector{<:GeometryBasics.Point}) =
```
sjkelly, line 212: On second thought, we can probably switch to `StaticArraysCore` to maintain this functionality with most of the load time benefits.

sjkelly:
Any reason why Benchmarks is failing? Is it caching a Manifest?

BeastyBlacksmith:
> Any reason why Benchmarks is failing? Is it caching a Manifest?

Seems it is looking for `GeometryBasics` for the run of the master version and can't find it, since it is not in the Project.toml anymore...

sjkelly:
I have not much time to look into BenchmarksCI more. I am wondering if this were merged it would then run okay?

sjkelly:
@BeastyBlacksmith thanks for looking at the BenchmarksCI!

BeastyBlacksmith:
> @BeastyBlacksmith thanks for looking at the BenchmarksCI!

Yeah, well.. this is a mess though. It ought to be easier...

BeastyBlacksmith:
So benchmarks are green now. Not sure how to go about ci, looks kind of unrelated

sjkelly:
Amazing! It seems CI has been failing a few days now?

BeastyBlacksmith:
> So benchmarks are green now.

Looking at the results though.. I question the fidelity of the benchmark

BeastyBlacksmith:
Better now

t-bltg:
CI was fixed in https://github.com/JuliaPlots/Plots.jl/pull/4319.

I will merge this PR when ci is green.

BeastyBlacksmith:
I'd wait with merging until we checked whether we can keep existing functionality with StaticArraysCore (see https://github.com/JuliaPlots/Plots.jl/pull/4287#discussion_r931254890)

t-bltg:
Noted, postponed.

sjkelly:
StaticArraysCore does not cover that dispatch case I commented on. Any GeometryBasics and StaticArrays dep would end up in downstream plotting recipes packages instead. Though the StaticArraysCore dependency is cheap, it isn't going to help here.

t-bltg:
@BeastyBlacksmith, merge ?

src/utils.jl
sjkelly: Edit: this is not going to help with GeometryBasics.

src/examples.jl
```diff
@@ -949,13 +949,29 @@ const _examples = PlotExample[
...
+
+                @recipe f(v::Plots.AVec{<:GeometryBasics.Point}) =
+                    RecipesPipeline.unzip(v)
+                @recipe f(p::GeometryBasics.Point) = [p]
+
+                d = MvNormal([1.0 0.75; 0.75 2.0])
+                plot([(1, 2), (3, 2), (2, 1), (2, 3)])
+                scatter!(Point2.(eachcol(rand(d, 1000))), alpha = 0.25)
+            end,
+        ],
```
BeastyBlacksmith, line 974: I'd change this to use tuples in both cases

BeastyBlacksmith (commented):
Also note, that the Batman example in the documentation is using `P2`s and might need to be updated. Other than these two points this looks good to me.

src/utils.jl
BeastyBlacksmith: I think we can keep this behind a required block to not break to much things

sjkelly:
I have updated the docs here: https://github.com/JuliaPlots/PlotDocs.jl/pull/300

src/examples.jl
BeastyBlacksmith: Thats why I suggested to put these definition in a Requires.jl block, then it should still work if load `GeometryBasics`

j-fu:
:+1:

Issue #3626, opened 2021-07-07
Plots exports too many names


tfburns:
I don't often use too many packages at once, but I've just found that Plots.jl exports names such as `Point`, `grid`, etc.
See 
https://github.com/JuliaGraphs/LightGraphs.jl/issues/1574 for background.

Occupying such a broad namespace for plotting seems a tad overzealous and makes it confusing for new users. I request/suggest thinking about carving out a more specific namespace which minimally conflicts with most non-competing plotting packages.

PR #4295, opened 2022-07-29
Add APNG animation support

### LINKED ISSUE #4293: [FR] Add an Option to Generate Animated PNG (APNG)
Add a function to generate APNG (Animated PNG) as like in the function `gif()`.

The FFMPEG operation is `ffmpeg -i <files> -plays 0 -f apng output.png` where the `-plays` sets the number of loops to play. `0` means infinite.

curio-sitas:
This PR may close #4293 

* [x] Add the support for APNG (Animated PNG) animations
* [x] Add HTML integration (png animation shows in vscode plot pane)
* [x] Add tests in `test_animations.jl` for coverage and consistency
* [x] Add @apng macro

---

The following FFMPEG command is called as proposed in #4293 :

```julia
[...]
elseif file_extension(fn)=="apng"
        # FFMPEG specific command for APNG (Animated PNG) animations
        ffmpeg_exe(
           `-v $verbose_level -framerate $framerate -i $(animdir)/%06d.png -plays $loop -f apng  -y $fn`,
        )
[...]
```

To create a PNG animation simply use the `apng` function : 
```julia
using Plots, Random

animation = @animate for i in 1:50
    Random.seed!(123)
    scatter(cumsum(randn(i)), ms=i, lab="", alpha = 1 - i/50,
        xlim=(0,50), ylim=(-5, 7))
end

animated_png = apng(animation)
```

or using the `@apng` macro : 
```julia
using Plots, Random

animated_png = @apng for i in 1:50
    Random.seed!(123)
    scatter(cumsum(randn(i)), ms=i, lab="", alpha = 1 - i/50,
        xlim=(0,50), ylim=(-5, 7))
end
```

### DIFF
--- a/src/Plots.jl
+++ b/src/Plots.jl
@@ -122,9 +122,11 @@ export
     mov,
     mp4,
     webm,
+    apng,
     animate,
     @animate,
     @gif,
+    @apng,
 
     test_examples,
     iter_segments,
--- a/src/Plots.jl
+++ b/src/Plots.jl
@@ -122,11 +122,9 @@ export
     mov,
     mp4,
     webm,
-    apng,
     animate,
     @animate,
     @gif,
-    @apng,
 
     test_examples,
     iter_segments,
--- a/src/animation.jl
+++ b/src/animation.jl
@@ -126,7 +126,7 @@ function buildanimation(
     elseif file_extension(fn)=="png"
         # FFMPEG specific command for APNG (Animated PNG) animations
         ffmpeg_exe(
-            `-v $verbose_level -framerate $framerate -i $(animdir)/%06d.png -f apng -loop $loop -y $fn`,
+            `-v $verbose_level -framerate $framerate -i $(animdir)/%06d.png -plays $loop -f apng  -y $fn`,
         )
     else
         ffmpeg_exe(
@@ -177,7 +177,7 @@ Base.show(io::IO, ::MIME"image/png", agif::AnimatedGif) =
 
 # -----------------------------------------------
 
-function _animate(forloop::Expr, args...; callgif = false)
+function _animate(forloop::Expr, args...; type::Symbol = :none)
     if forloop.head ∉ (:for, :while)
         error("@animate macro expects a for- or while-block. got: $(forloop.head)")
     end
@@ -218,7 +218,14 @@ function _animate(forloop::Expr, args...; callgif = false)
     push!(block.args, :($countersym += 1))
 
     # add a final call to `gif(anim)`?
-    retval = callgif ? :(Plots.gif($animsym)) : animsym
+    if type == :gif
+        retval = :(Plots.gif($animsym))
+    elseif type == :apng
+        retval = :(Plots.apng($animsym))
+    else
+        retval = animsym
+    end
+
 
     # full expression:
     esc(quote
@@ -244,7 +251,24 @@ Example:
 ```
 """
 macro gif(forloop::Expr, args...)
-    _animate(forloop, args...; callgif = true)
+    _animate(forloop, args...; type=:gif)
+end
+
+"""
+Builds an `Animation` using one frame per loop iteration, then create an animated PNG (APNG).
+
+Example:
+
+```
+  p = plot(1)
+  @apng for x=0:0.1:5
+    push!(p, 1, sin(x))
+  end
+```
+"""
+macro apng(forloop::Expr, args...)
+    _animate(forloop, args...; type=:apng)
+
 end
 
 """
@@ -261,5 +285,5 @@ gif(anim)
 ```
 """
 macro animate(forloop::Expr, args...)
-    _animate(forloop, args...)
+    _animate(forloop, args...; type=:none)
 end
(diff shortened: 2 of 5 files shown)


src/animation.jl
```diff
@@ -22,6 +22,7 @@ giffn() = (isijulia() ? "tmp.gif" : tempname() * ".gif")
 movfn() = (isijulia() ? "tmp.mov" : tempname() * ".mov")
 mp4fn() = (isijulia() ? "tmp.mp4" : tempname() * ".mp4")
 webmfn() = (isijulia() ? "tmp.webm" : tempname() * ".webm")
+apngfn() = (isijulia() ? "tmp.png" : tempname() * ".png")
```
t-bltg, line 25: Why the `.png` extension over `.apng` ?

src/animation.jl
```diff
@@ -198,7 +218,14 @@ function _animate(forloop::Expr, args...; callgif = false)
...
     # add a final call to `gif(anim)`?
-    retval = callgif ? :(Plots.gif($animsym)) : animsym
+    if type == :gif
+        retval = :(Plots.gif($animsym))
+    elseif type == :apng
+        retval = :(Plots.apng($animsym))
+    else
+        retval = animsym
+    end
+
```
t-bltg, line 228:
```suggestion
    retval = if type == :gif
        :(Plots.gif($animsym))
    elseif type == :apng
        :(Plots.apng($animsym))
    else
        animsym
    end

```

src/animation.jl
curio-sitas: As i saw during my researches, APNG files commonly use the standard ".png" extension. Nevertheless the function should be named apng for consistency.

t-bltg: According to https://en.wikipedia.org/wiki/APNG, both `.png` and `.apng` are possible, so I'd rather use the dedicated `.apng` for animations (better be explicit than implicit, consistent with the function name `apngfn`) unless there is an argument against that ?

t-bltg:
Also, may I suggest running `JuliaFormatter` (see https://docs.juliaplots.org/stable/contributing/#Write-code,-and-format) for the `format/code-style` CI check to pass.

src/animation.jl
curio-sitas: I agree with you, i just believed that people used .png for both comonly. But using `.apng` makes more sense obvisouly. I change that

curio-sitas: Well, i tested this change. `.apng` files are not recognized by vscode, while `.png` animated files are.

t-bltg: https://github.com/microsoft/vscode/blob/main/src/vs/base/common/mime.ts#L70: `png`, `gif`, `webp` but no `.apng`. I don't use vscode, so If that doesn't work, I'd suggest reverting to `.png` and add an explicit note to explain why we are forced to use `.png` over `.apng` for compat with vscode.

src/animation.jl
```diff
@@ -198,7 +219,13 @@ function _animate(forloop::Expr, args...; callgif = false)
...
 
     # add a final call to `gif(anim)`?
-    retval = callgif ? :(Plots.gif($animsym)) : animsym
+    retval = if type == :gif
+        :(Plots.gif($animsym))
+    elseif type == :apng
+        :(Plots.apng($animsym))
+    else
+        animsym
+    end
```
t-bltg, line 228:
```suggestion
    retval = if type === :gif
        :(Plots.gif($animsym))
    elseif type === :apng
        :(Plots.apng($animsym))
    else
        animsym
    end
```

I missed that, symbol comparisons should be made with `===`.

src/animation.jl
```diff
@@ -241,5 +284,5 @@ gif(anim)
 ```
 """
 macro animate(forloop::Expr, args...)
-    _animate(forloop, args...)
+    _animate(forloop, args...; type = :none)
```
t-bltg, line 287:
```suggestion
    _animate(forloop, args...)
```

No need to specify the `kwarg` since `type` defaults to `:none`.

src/animation.jl
```diff
@@ -77,6 +78,11 @@ mp4(anim::Animation, fn = mp4fn(); kw...) = buildanimation(anim, fn, false; kw..
 Creates an .webm-file from an `Animation` object.
 """
 webm(anim::Animation, fn = webmfn(); kw...) = buildanimation(anim, fn, false; kw...)
+"""
+    apng(animation[, filename]; fps=20, loop=0, variable_palette=false, verbose=false, show_msg=true)
```
t-bltg, line 82: `variable_palette` is no used for `apng`s, please add only the used keyword args to the signature.

src/animation.jl
t-bltg: @curio-sitas, in the end, is this solved for `vscode` ?

src/Plots.jl
```diff
@@ -122,9 +122,11 @@ export
     mov,
     mp4,
     webm,
+    apng,
```
t-bltg, line 125: I'm puzzled: not all file methods are exported, and exporting lots of symbols from a package is pretty bad (see https://github.com/JuliaPlots/Plots.jl/issues/3626).

@BeastyBlacksmith, what do you think ?

curio-sitas: I think this new feature may not be used by a vast majority of users. We may leave it unexported ?

t-bltg: Yes, I think it would be better that way.

curio-sitas: Done

t-bltg (approved):
LGTM. Ready to merge ?

PR #4300, opened 2022-08-04
added colorbartitle attributes for pgfplotsx

louisponet:
Hi,

I think this PR fixes the ignoring of colorbar_title*** attributes.

Cheers,
Louis

### DIFF
--- a/src/backends/pgfplotsx.jl
+++ b/src/backends/pgfplotsx.jl
@@ -200,6 +200,7 @@ function (pgfx_plot::PGFPlotsXPlot)(plt::Plot{PGFPlotsXBackend})
                 colorbar_style = if sp[:colorbar] === :top
                     push!(
                         Options("xlabel" => sp[:colorbar_title]),
+                        "xlabel style" => pgfx_get_colorbar_title_style(sp),
                         "at" => "(0.5, 1.05)",
                         "anchor" => "south",
                         "xtick" => cticks,
@@ -209,6 +210,7 @@ function (pgfx_plot::PGFPlotsXPlot)(plt::Plot{PGFPlotsXBackend})
                 else
                     push!(
                         Options("ylabel" => sp[:colorbar_title]),
+                        "ylabel style" => pgfx_get_colorbar_title_style(sp),
                         "ytick" => cticks,
                         "yticklabel style" => pgfx_get_colorbar_ticklabel_style(sp),
                     )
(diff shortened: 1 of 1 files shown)


t-bltg:
What was wrong before ?

The tests seem to indicate that `colorbar_title***` attributes are already taken into account: https://github.com/JuliaPlots/Plots.jl/blob/9397db443e113b1f947b1dd48ece3acded9e0a48/test/test_pgfplotsx.jl#L433.

louisponet:
They aren't actually added to the plot

t-bltg:
Thanks for clarifying. Could you post a mwe with / without this PR showing the visual improvements ?

louisponet:
So I've attached the created .tex files (as .txt because github doesn't like .txt) for
`scatter(0:1:1, 0:1:1, zcolor = 0:1:1, colorbar_title="test", colorbar_titlefontsize=50)` 
For both the master and this PR. Focus on the `colorbar style` `ylabel style` fields as in the screenshots.

[PR.txt](https://github.com/JuliaPlots/Plots.jl/files/9268325/PR.txt)
[master.txt](https://github.com/JuliaPlots/Plots.jl/files/9268328/master.txt)


master
PR

t-bltg:
Here is what a mwe would look like:

```julia
using Plots; pgfplotsx()

main() = begin
  pl = scatter(0:1:1, 0:1:1, zcolor = 0:1:1, colorbar_title="test", colorbar_titlefontsize=50, tex_output_standalone=true)
  Plots.png(pl, "mwe.png")
  Plots.tex(pl, "mwe.tex")
  return
end

main()
```

You can then post the `.png` with and without the patch here in github.

louisponet:
Well, I don't understand what the issue is with what I've posted, it's clearly not taken into account as the .tex file suggests... 
Anyway
master

PR

t-bltg:
> Well, I don't understand what the issue is with what I've posted, it's clearly not taken into account as the .tex file suggests...

Nothing wrong, but when you submit a PR, you have to make it as clear as possible for others to review.
I shouldn't have to dig into `.tex` files and mentally draw what the result would be.
Posting a before/after `.png` remains the simplest.
I've edited https://github.com/JuliaPlots/Plots.jl/pull/4300#issuecomment-1206466532 to add a leading exclamation mark to your urls for embedding images.

louisponet:
Since it's the tex file that's the direct result of Plots code, not the png image that gets produced it shouldn't matter. I had the exclamation marks initially but removed them so that it was clear which png was for the PR and which for the master branch. 

Anyway, I hope this has cleared it up.

Cheers!

t-bltg:
>  I had the exclamation marks initially but removed them so that it was clear which png was for the PR and which for the master branch.

I've added a bold label ;).

t-bltg (approved):

t-bltg:
Thanks for this contribution !

PR #4309, opened 2022-08-13
Make `savefig` filepath-type agnostic

### LINKED ISSUE #4279: [FR] Support `FilePathsBase.AbstractPath` in `savefig`
Currently, `savefig` only supports a filename that is a string, referring to the local filesystem.

[FilePathsBase.jl](https://github.com/rofinn/FilePathsBase.jl) provides the `AbstractPath` abstraction, which allows for using more generic paths in I/O functions such as `open`. One concrete example is `S3Path <: AbstractPath` from [AWSS3.jl](https://github.com/JuliaCloud/AWSS3.jl).

It is possible to write a plot directly to S3 as-is, but it requires some hoop-jumping:

```julia
using AWSS3
using FilePathsBase
using Plots

plot(rand(10))

path = p"s3://my-bucket-name/moo.png"
open(path, "w") do io
    show(io, MIME("image/png"), Plots.current())
end
cp(path, p"moo.png")  # To inspect that the file has rendered correctly
```

---

From a quick play around with the code in `src/output.jl`, I think the only changes that need to be made are:

1) Change `AbstractString` -> `Union{AbstractString,AbstractPath}`
2) Adapt this:

https://github.com/JuliaPlots/Plots.jl/blob/439c3c6b45c9b0a7ce9c51e4c46e139e03bd4495/src/output.jl#L91-L95

to replace `string(fn, ".", ext)` with a more generic method to add an extension & return the appropriate type. I suspect @rofinn will know if such a thing exists & where it can be found :)

curio-sitas:
~~This PR may close this issue :~~ Fix #4279 .

I changed the save functions (html, png, pdf ...) and `savefig` in `output.jl` so that the filepath is type agnostic.
The `addExtension` has been changed too according to #4279.

```julia
function addExtension(fp, ext::AbstractString)
     parent = dirname(fp)
     fn = basename(fp)
     oldfn, oldext = splitext(fn) 
     oldext = chop(oldext, head = 1, tail = 0) 
     get(_extension_map, oldext, oldext) == ext ? fp : joinpath(parent, string(fn, ".", ext)) 
 end 

```
Output tests are passing ✅ 

---

Example :
```julia
using Plots
using FilePathsBase

fn = p"C:/Users/someone/image.png" # WindowsPath type

x = 0:0.1:2pi
y = sin.(x).
p = plot(x,y)

# Writing directly plot output in fn
open(fn, "w") do io
    show(io, MIME("image/png"), p)
end


savefig(fn) # Same result despite the fact that filepath is not of type AbstractString
```

### DIFF
--- a/src/output.jl
+++ b/src/output.jl
@@ -112,11 +112,10 @@ const _extension_map = Dict("tikz" => "tex")
 Change filepath extension according to the extension map
 """
 function addExtension(fp, ext::AbstractString)
-    parent = dirname(fp)
-    fn = basename(fp)
-    oldfn, oldext = splitext(fn)
+    dn, fn = splitdir(fp)
+    _, oldext = splitext(fn)
     oldext = chop(oldext, head = 1, tail = 0)
-    get(_extension_map, oldext, oldext) == ext ? fp : joinpath(parent, string(fn, ".", ext))
+    get(_extension_map, oldext, oldext) == ext ? fp : joinpath(dn, string(fn, ".", ext))
 end
 
 """
@@ -126,7 +125,7 @@ Save a Plot (the current plot if `plot` is not passed) to file. The file
 type is inferred from the file extension. All backends support png and pdf
 file types, some also support svg, ps, eps, html and tex.
 """
-function savefig(plt::Plot, fn)
+function savefig(plt::Plot, fn) # fn might be an `AbstractString` or an `AbstractPath` from `FilePaths.jl`
     fn = abspath(expanduser(fn))
 
     # get the extension
--- a/src/output.jl
+++ b/src/output.jl
@@ -1,83 +1,83 @@
 
 defaultOutputFormat(plt::Plot) = "png"
 
-png(plt::Plot, fn::AbstractString) =
+png(plt::Plot, fn) =
     open(addExtension(fn, "png"), "w") do io
         show(io, MIME("image/png"), plt)
     end
-png(fn::AbstractString) = png(current(), fn)
+png(fn) = png(current(), fn)
 
 png(plt::Plot, io::IO) = show(io, MIME("image/png"), plt)
 png(io::IO) = png(current(), io)
 
-svg(plt::Plot, fn::AbstractString) =
+svg(plt::Plot, fn) =
     open(addExtension(fn, "svg"), "w") do io
         show(io, MIME("image/svg+xml"), plt)
     end
-svg(fn::AbstractString) = svg(current(), fn)
+svg(fn) = svg(current(), fn)
 
 svg(plt::Plot, io::IO) = show(io, MIME("image/svg+xml"), plt)
 svg(io::IO) = svg(current(), io)
 
-pdf(plt::Plot, fn::AbstractString) =
+pdf(plt::Plot, fn) =
     open(addExtension(fn, "pdf"), "w") do io
         show(io, MIME("application/pdf"), plt)
     end
-pdf(fn::AbstractString) = pdf(current(), fn)
+pdf(fn) = pdf(current(), fn)
 
 pdf(plt::Plot, io::IO) = show(io, MIME("application/pdf"), plt)
 pdf(io::IO) = pdf(current(), io)
 
-ps(plt::Plot, fn::AbstractString) =
+ps(plt::Plot, fn) =
     open(addExtension(fn, "ps"), "w") do io
         show(io, MIME("application/postscript"), plt)
     end
-ps(fn::AbstractString) = ps(current(), fn)
+ps(fn) = ps(current(), fn)
 
 ps(plt::Plot, io::IO) = show(io, MIME("application/postscript"), plt)
 ps(io::IO) = ps(current(), io)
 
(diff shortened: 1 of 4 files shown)


src/output.jl
```diff
@@ -106,10 +106,17 @@ end
...
+
+Change filepath extension according to the extension map
+"""
+function addExtension(fp, ext::AbstractString)
+    parent = dirname(fp)
+    fn = basename(fp)
     oldfn, oldext = splitext(fn)
     oldext = chop(oldext, head = 1, tail = 0)
-    get(_extension_map, oldext, oldext) == ext ? fn : string(fn, ".", ext)
+    get(_extension_map, oldext, oldext) == ext ? fp : joinpath(parent, string(fn, ".", ext))
```
t-bltg, line 119: Why is an absolute path needed  here ?

t-bltg:
It this were to support `FilePathsBase` under the hood, I think we should test is as such, and add effective tests in `test/output.jl`.

Add `FilePathsBase = "48062228-2e41-5def-b9a4-89aafe57970f"` around [here](https://github.com/JuliaPlots/Plots.jl/blob/ec9b7e89d7af4222c714ee39d944a53771404ebb/Project.toml#L79) in `[extras]` section and to `[targets]` section for testing.

src/output.jl
curio-sitas: Some package may work with absolute filepath no ?

t-bltg: Yes, but it is a bit odd to force an absolute path in the second branch. Anyway I might be overthinking here.

src/output.jl
```diff
@@ -106,10 +106,17 @@ end
...
-function addExtension(fn::AbstractString, ext::AbstractString)
+"""
+    addExtension(filepath, extension)
+
+Change filepath extension according to the extension map
+"""
+function addExtension(fp, ext::AbstractString)
+    parent = dirname(fp)
+    fn = basename(fp)
     oldfn, oldext = splitext(fn)
```
t-bltg, line 117:
```suggestion
    _, oldext = splitext(fn)
```

See https://docs.julialang.org/en/v1/manual/variables/#man-allowed-variable-names (underscore convention when [`l-value`](https://en.wikipedia.org/wiki/Value_(computer_science)#lrvalue))

src/output.jl
```diff
@@ -106,10 +106,17 @@ end
...
 
-function addExtension(fn::AbstractString, ext::AbstractString)
+"""
+    addExtension(filepath, extension)
+
+Change filepath extension according to the extension map
+"""
+function addExtension(fp, ext::AbstractString)
+    parent = dirname(fp)
+    fn = basename(fp)
```
t-bltg, line 116:
```suggestion
    parent, fn = splitdir(fp)
```

t-bltg: + change `parent` to `dn` (more consistent with `fn`)

src/output.jl
```diff
@@ -119,7 +126,7 @@ Save a Plot (the current plot if `plot` is not passed) to file. The file
 type is inferred from the file extension. All backends support png and pdf
 file types, some also support svg, ps, eps, html and tex.
 """
-function savefig(plt::Plot, fn::AbstractString)
+function savefig(plt::Plot, fn)
```
t-bltg, line 129:
```suggestion
function savefig(plt::Plot, fn)  # fn might be an `AbstractString` or an `AbstractPath` from `FilePaths.jl`
```

test/test_output.jl
```diff
@@ -1,6 +1,6 @@
 macro test_save(fmt)
     quote
-        let pl = plot(1:10), fn = tempname()
+        let pl = plot(1:10), fn = tmpname() # fn is an AbstractPath from FilePathsBase.jl
             getfield(Plots, $fmt)(pl, fn)
             getfield(Plots, $fmt)(fn)
```
t-bltg, line 5:
```suggestion
        let pl = plot(1:10), fn = tempname(), fp = tmpname() # fp is an AbstractPath from FilePathsBase.jl
            getfield(Plots, $fmt)(pl, fn)
            getfield(Plots, $fmt)(fn)
            getfield(Plots, $fmt)(fp)
            # ...
```

Let's duplicate, not replace the tests here.

test/test_output.jl
```diff
@@ -1,8 +1,10 @@
 macro test_save(fmt)
     quote
-        let pl = plot(1:10), fn = tmpname() # fn is an AbstractPath from FilePathsBase.jl
+        let pl = plot(1:10), fn = tempname(), fp = tmpname() # fp is an AbstractPath from FilePathsBase.jl
             getfield(Plots, $fmt)(pl, fn)
             getfield(Plots, $fmt)(fn)
+            getfield(Plots, $fmt)(fp)
+            # ...
```
t-bltg, line 7: To be finished ...

curio-sitas: Oh yeah, my bad

t-bltg: Maybe I wasn't that clear enough (reviewing on my phone is less easier than on a pc ;))

t-bltg (approved):
LGTM, pending CI.

Issue #4045, opened 2022-01-16
How to extract subplot from a plot?


JianghuiDu:
I have a plot `p` containing 2 subplots. I want to modify the second subplot later in a function. How can I extract and display the subplot? I tried `p[2]` which shows `Subplot{2}`, but I don't know how to extract and modify it.

iabraham:
I was wondering if this feature was ever added (and is maybe undocumented as of yet?). There are use cases: [for example](https://stackoverflow.com/q/75843396/6395452).

t-bltg:
Is https://github.com/JuliaPlots/Plots.jl/blob/master/src/plot.jl#L274-L283 sufficient ?

iabraham:
Indeed! Can't believe didn't think of trying that.

PR #4320, opened 2022-08-21
3d: add `projection_type`

t-bltg:
- add `projection_type` = `:auto` (default to `:ortho` since supported by all backends), `:ortho(graphic)` or `:persp(ective)`;
- normalize `unicodeplots` azimuth with other backends;
- fix `PyPlot` 3d aspect ratio and padding for `z` axis in 3d;
- fix `GR` `cam` and `fov` when using `ortho` projection;
- fix https://github.com/JuliaPlots/Plots.jl/issues/4045: rework `Plot` and `SubPlot` constructors, move `subplots.jl` code into `types.jl`.

Tested under `GR`, `PyPlot`, `PlotlyJS` and `UnicodePlots`.

`PGFPlots`, `Gnuplot` do **not** support a `perspective` projection, hence the logical choice is to use an orthographic projection on `:auto` (consistency across backends in high level `Plots``).

Example:
```julia
using Plots

main() = begin
  for be in (:unicodeplots, :pyplot, :gr, :plotlyjs)
    p = Plots.test_examples(be, 60)
    png(p, "proj_type-$be")
  end
end

main()
```

**UnicodePlots**
proj_type-unicodeplots

**PyPlot**
proj_type-pyplot

**GR**
proj_type-gr

**PlotlyJS**
proj_type-plotlyjs

~~@jheinen, I have some trouble drawing correct axes in `GR` using an `orthographic` projection.
Margins also seem affected, and the expected unit `aspect_ratio` is totally wrong (use lims instead, should be fixed by https://github.com/JuliaPlots/Plots.jl/pull/4148/files).
Do you know how to fix that in GR ?~~

EDIT: fixed `gr` `cam` and `fov` when using `orthographic` transformation.

- [x] new ref imgs (https://github.com/JuliaPlots/PlotReferenceImages.jl/pull/120).


t-bltg:
MacOS 1.6 error unrelated (`libgit2`).

src/types.jl
```diff
@@ -56,15 +53,12 @@ end
 mutable struct Extrema
     emin::Float64
     emax::Float64
-end
-Extrema() = Extrema(Inf, -Inf)
 
-# -----------------------------------------------------------
+    Extrema() = new(Inf, -Inf)
```
BeastyBlacksmith, line 57: Did you really meant to override the default constructor here?
That results in `Extrema(1,2)` to throw a `MethodError`

t-bltg: It see it now in the logs with the `hdf5` backend. And of course these test were so badly written that they only warn and don't throw errors. I will fix this (and enhance the tests).

PR #4319, opened 2022-08-20
`ci`: try `1.8.0` stable release

t-bltg:
- try `1.8` on all OS;
- keep `1.7` on `ubuntu-latest` only to spare credits;
- ~~try to fix ci by bumping `PlotlyBase` to `0.8`~~;
- add [invalidations action check](https://discourse.julialang.org/t/potential-performance-regressions-in-julia-1-8-for-special-un-precompiled-type-dispatches-and-how-to-fix-them/86359/11);
- fix missing `fgcolor` introduced by https://github.com/JuliaPlots/Plots.jl/pull/2621;
- bump `Unzip`.

PR #4321, opened 2022-08-21
Truncate markersize in legends to not hide linestyle

### LINKED ISSUE #4312: [BUG] Markers in legend obscure linestyles in subplots
## Details
 
See https://github.com/JuliaPlots/Plots.jl/issues/4092#issuecomment-1216744812.

I will look into making a PR for the scaling proposal, but it would be helpful if someone can point to where to look as I'm very unfamiliar with the codebase. 

### Backends

gr (default) |  X   |     |
pyplot       |     |     | X
plotlyjs     |     |     | X
pgfplotsx    |     |     | X
unicodeplots |     |     | X
inspectdr    |     |     |  X
gaston       |     |     | X

### Versions

Plots.jl version: 1.31.7
Backend  version (`]st -m <backend(s)>`): GR v0.66.0
Output of `versioninfo()`:
```julia
julia> versioninfo()
Julia Version 1.7.3
Commit 742b9abb4d (2022-05-06 12:58 UTC)
Platform Info:
  CPU: 11th Gen Intel(R) Core(TM) i5-1145G7 @ 2.60GHz
  WORD_SIZE: 64
  LIBM: libopenlibm
  LLVM: libLLVM-12.0.1 (ORCJIT, tigerlake)
Environment:
  JULIA_EDITOR = code
  JULIA_NUM_THREADS = 4
```

DrChainsaw:
Fixes #4312

Basic idea is to limit the marker size in legends to approximately half the length of the line in the legend. I'm not sure I managed to do this in a clean way but the rationale is described in a comment.

I have only done this for the GR backend, but should this approach turn out to be acceptable I can verify and implement for the other backends (where needed).

If there is no line then there is also no limit to the markersize.

I think the limitations/drawbacks of this approach are quite well summarized by the visual regressiontest nr 48:
PlotsMarkerSizePrRef48

Firstly (and this is ofc easy to fix) there is some inconsistency between marker only and marker + line. 

My thinking was that users might want to convey information with marker sizes and then the truncation might destroy that information. Conveying information with both markersize and linestyle however is not something I would expect anyone to do so then truncation should be ok.

An option to trunction is just to scale everything, but then I guess one must first look through the series and find the largest markersize. Users might still get annoyed that their markers are shrinked in the legend though... 

Secondly, when there are four columns the legend tend to be so small that neither linestyles nor markers are very readable. 

Perhaps one should just leave more room for the lines and markers?

Here is the testcode I used to manually verify whether this is an improvment:
```julia
function legsweep(labsuffix="")
    for m in [:none, :circle, :rect, :star5, :diamond, :hexagon, :cross, :xcross, :utriangle, :dtriangle, :rtriangle, :ltriangle, :pentagon, :heptagon, :octagon, :star4, :star6, :star7, :star8, :vline, :hline, :+, :x]
        linestyles = [:solid ;; :dash ;; :dot ;; :dashdot ;; :dashdot ;; :dashdotdot]
        p = plot(map(k -> fill(k, 10), eachindex(linestyles)); c=:blue, linestyle=linestyles, marker=m, label=string.(linestyles).* labsuffix, title=m)
        display(p)
        display(plot(p,p))   
        display(plot(p,p; size=(1200, 400)))
        display(plot(p,p,p,p; layout=(1, 4)))
        display(plot(p,p,p,p; layout=(1, 4), size=(2400, 400)))
    end
end
```

### DIFF
--- a/src/backends/gr.jl
+++ b/src/backends/gr.jl
@@ -1096,10 +1096,11 @@ function gr_add_legend(sp, leg, viewport_plotarea)
                 if st in (:path, :straightline, :path3d)
                     gr_set_transparency(lc, get_linealpha(series))
                     # This is to prevent that linestyle is obscured by large markers. 
-                    # We are trying to get markers to not be larger than half the line width. 
-                    # 1 / leg.dy translates width_factor to width units (important in the context of size kwarg)
+                    # We are trying to get markers to not be larger than half the line length. 
+                    # 1 / leg.dy translates width_factor to line length units (important in the context of size kwarg)
                     # 4 is an empirical constant to translate between line length unit and marker size unit
-                    maxmarkersize = (leg.width_factor * 3.5 - leg.width_factor / 2) * 0.5 / leg.dy * 4
+                    maxmarkersize =
+                        (leg.width_factor * 3.5 - leg.width_factor / 2) * 0.5 / leg.dy * 4
                     if series[:fillrange] === nothing || series[:ribbon] !== nothing
                         GR.polyline(
                             [xpos - leg.width_factor * 3.5, xpos - leg.width_factor / 2],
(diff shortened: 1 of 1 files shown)


DrChainsaw:
<details>
  <summary>Some before/after examples</summary>
Single plots (minor difference expected)

baseline_single
new_single

Two plots (difference expected)
baseline_two
new_two

Two plots with double x-size (minor difference expected):
baseline_two_size
new_two_size

Four plots (difference expected):
baseline_four
new_four

Four plots with four times x-size (minor difference expected):
baseline_four_size
new_four_size

</details>

BeastyBlacksmith (approved):

t-bltg (approved):
Thanks for this PR.

I think this is overall an enhancement, and since comments are added, we can always rework if needed.

Please run [`JuliaFormatter`](https://docs.juliaplots.org/stable/contributing/#Write-code,-and-format), and open a PR with modified reference files at https://github.com/JuliaPlots/PlotReferenceImages.jl (https://docs.juliaplots.org/stable/contributing/#VisualRegressionTests).

DrChainsaw:
Formatter modification pushed and here is the reference images pr: https://github.com/JuliaPlots/PlotReferenceImages.jl/pull/119

t-bltg:
Sorry for the delay, fixing `ci` in https://github.com/JuliaPlots/Plots.jl/pull/4319 took quite some time.

Thanks for the contribution !

DrChainsaw:
Thanks for the pull! Very smooth.

PR #4326, opened 2022-09-01
cleanup: drop deprecated `pgfplots`

t-bltg:
More than 2 years with a deprecation (https://github.com/JuliaPlots/Plots.jl/pull/2711).


t-bltg:
Can't we really add this to future `1.32` ? Deadcode is a pain for development.

PR #4328, opened 2022-09-01
Fixed missing marker in legend when plotting a series whose first entry is NaN

### LINKED ISSUE #3996: Legend marker does not appear when first y value is NaN on scatter
## Details
For example, `scatter(1:3, [NaN, 2, 3])` produces 
Screen Shot 2021-12-11 at 11 37 28 AM
While `scatter(1:3, [1, NaN, 3])` produces 
Screen Shot 2021-12-11 at 11 38 34 AM
in Atom.

### Backends

gr (default) |  x  |     |
pyplot       |     |     | x
plotlyjs     |     |     | x
pgfplotsx    |     |     |  x
unicodeplots |     |     | x
inspectdr    |     |     | x
gaston       |     |     | x

### Versions

Plots.jl version:   `[91a5bcdd] Plots v1.23.6`
pkg> st -m gr
  No Matches in `~/.julia/dev/all/Manifest.toml`
Output of `versioninfo()`:
```
Julia Version 1.7.0
Commit 3bf9d17731 (2021-11-30 12:12 UTC)
Platform Info:
  OS: macOS (x86_64-apple-darwin19.5.0)
  CPU: Intel(R) Core(TM) i5-8210Y CPU @ 1.60GHz
  WORD_SIZE: 64
  LIBM: libopenlibm
  LLVM: libLLVM-12.0.1 (ORCJIT, skylake)
Environment:
  JULIA_EDITOR = atom  -a
  JULIA_NUM_THREADS = 2
```

leespen1:
Changed function `gr_add_legend` to call `gr_draw_marker` instead of `gr_draw_markers`, since a legend entry only has one marker (if any).

Fix #3996 

I explained the issue and solution in greater detail in this post: https://discourse.julialang.org/t/fixing-missing-marker-in-plot-legend-when-first-value-in-series-is-nan/86658

In summary, the example
```
using Plots; gr()
plot([NaN,2,3], markershape=:circle)
```
produced a graph with no marker in the legend:
NaN_legend_marker_issue

This only happens when the first entry in the series is NaN. If only the second entry is NaN, we get:
```
using Plots; gr()
plot([1,NaN,3], markershape=:circle)
```
NaN_not_first_legend_marker

Morever, this behavior does not happen for the PlotlyJS backend (I did not test backends besides GR and PlotlyJS):
he second entry is NaN, we get:
```
using Plots; plotlyjs()
plot([NaN,2,3], markershape=:circle)
```
NaN_legend_marker_plotlyjs

By changing the `gr_draw_markers` call to a `gr_draw_marker` call, this behavior was fixed. Now the example
```
using Plots; gr()
plot([NaN,2,3], markershape=:circle)
```
produces the following graph:
NaN_legend_marker_fixed

And it works for a single series with multiple marker shapes as well:
```
using Plots; gr()
plot([1,2,3], markershape=[:circle, :square, :diamond])
```
NaN_legend_marker_fixed_multiple_labels

The only downside I can see is the following example:
```
using Plots; gr()
plot([NaN,2,3], markershape=[:circle, :square, :diamond])
```
NaN_legend_marker_fixed_multiple_labels_first_NaN

The marker in the legend is a circle, although no circle appears in the series because the first entry in the series is not plotted (because it's NaN). Still, the previous behavior was to have no marker at all, which I think is worse.


t-bltg:
You might want to read https://docs.juliaplots.org/stable/contributing/#Write-code,-and-format.

t-bltg:
This is just a spurious random error related to network failure : `GitError(Code:ERROR, Class:Net, SecureTransport error: connection closed via error)`.

I'm re-runing the failed job, but this is not blocking.

t-bltg (approved):
I think that this PR is well explained (here and on discourse), and I see no regressions in the gallery, so I don't see any reason why this shouldn't go in.

BeastyBlacksmith (approved):
Thank you for this fix! You might want to check out #3503.

leespen1:
You're welcome, thanks for the review!

PR #4329, opened 2022-09-02
format: function declaration (nfc)

t-bltg:
Saves about 30 lines.


t-bltg:
`GraphRecipes` test failures will be fixed in another PR.

Issue #1486, opened 2018-04-15
Grid lines may show up too weak


t37w:
Using Plots with GR as backend the grid lines show up too week.
See example in ticket #1483 .
In the examples in ticket #743 it looks much better.  I don't know whether it depends on the resolution of the screen where very thin lines may disappear.  In case it matters: I have 1920x1080

t37w:
To be more specific, one has to look at grid lines in 2D plots and 3D plots.
In 2D plots, major grid lines are weak but visible (thin light gray lines on white background)
In 3D plots, the major grid lines are much weaker and depending on background color, more or less invisible.  Minor grid lines are not visible.

t37w:
Using vanilla GR for a 3D plot, with the default window, the major and minor grid lines are very good visible as gray and light gray lines.

Issue #1483, opened 2018-04-13
Wrong lables when doing a 3D plot


t37w:
When creating a 3D plot either with plot() or scatter() graphics with 3 axes and the data is created, but the labels are wrong.  First, there are only 2 and their position look like for a 2d plot as the ylabel is near the z-axis. Found with julia-0.6.2 on Windows running in a shell window.
Small example:
using Plots; gr();
x=rand(101); y=rand(101)+10; z=rand(101)+100;
scatter(x,y,z,xlabel="red", ylabel="green", zlabel="blue")
      or
plot(x,y,z,xlabel="red", ylabel="green", zlabel="blue")
same with plot3D() and scatter3D().

By the way:
grid lines are not drawn.
scatter() and plot() have different tics.  The range is scaled differently.

Please let me know in case you need another ticket for them.

Best regards,
Thomas

t37w:
Sorry, I have overlooked the CR you referenced. I will write tickets for the other 2 topics.

PR #4335, opened 2022-09-05
increase default `(minor)gridalpha`

t-bltg:
The grid lines are barely visible (if not invisible), where they should be (and they were before according to https://github.com/JuliaPlots/Plots.jl/issues/1486), because of the very low transparency level:
https://docs.juliaplots.org/stable/gallery/gr/generated/gr-ref1/#gr_demo_1
https://docs.juliaplots.org/stable/gallery/gr/generated/gr-ref24/#gr_demo_24

Fix https://github.com/JuliaPlots/Plots.jl/issues/1486.

PR: arbitrary increase of `alpha` channel (might need new ref imgs).

- [x] ref imgs (https://github.com/JuliaPlots/PlotReferenceImages.jl/pull/122).


BeastyBlacksmith (commented):
It seems the situation has improved already since 2018 (ref: https://github.com/JuliaPlots/Plots.jl/issues/1483#issuecomment-381418743) and in my opinion we could close #1486 without this.

t-bltg:
Frankly, between `gr`, `pgfplotsx` and `pyplot`:

https://docs.juliaplots.org/stable/gallery/gr/generated/gr-ref1/#gr_demo_1,
https://docs.juliaplots.org/stable/gallery/pyplot/generated/pyplot-ref1/#pyplot_demo_1,
https://docs.juliaplots.org/stable/gallery/pgfplotsx/generated/pgfplotsx-ref1/#pgfplotsx_demo_1,

`gr` is the only backend where you cannot distinguish the grid.

BeastyBlacksmith:
I am fine with increasing the default gridline intensity, despite changing default values being a breaking change, since the difference is barely visible and on average an improvement.
Note, that if there are differences between backends increasing the default value for all backends will magnify the difference. I consider it harmless in this case though.

t-bltg:
Thanks, let's test this now.
If the docs build for `1.32.0` is not satisfactory, the PR and ref images is trivial to revert.

t-bltg:
I think all the examples for all backends on https://docs.juliaplots.org/dev now have discrete but visible grids.

BeastyBlacksmith:
> Why did the docs not deploy to `1.32` ?

You did write `skip ci` on the commit. You need to run the workflow manually on the tag now.

PR #4343, opened 2022-09-09
Extended help to show all attributes

BeastyBlacksmith:
Not sure how useful this is


t-bltg (approved):
I'd say better more docs than less, and it doesn't pollute the help since it appears only in extended mode, nice trick.

Issue #4337, opened 2022-09-06
[BUG] Unknown key: legend_fontrotation


maxchendt:
## Details

``` Julia
using Plots
default(legend=:right, background_color_legend="#ffffff00", legendfont=font(5), grid=nothing)  
display(plot(1:10, rand(10)))
```

### Backends

This bug occurs on ( insert `x` below )

Backend      | yes | no  | untested
-------------|-----|-----|---------
gr (default) |  x   |     |
pyplot       |     |     | x
plotlyjs     |     |     | x
pgfplotsx    |     |     | x
unicodeplots |     |     | x
inspectdr    |     |     | x
gaston       |     |     | x

### Versions

Plots.jl version: [91a5bcdd] Plots v1.32.0
Backend  version (`]st -m <backend(s)>`): [28b8d3ca] GR v0.66.2
Output of `versioninfo()`:
Julia Version 1.8.0
Commit 5544a0fab76 (2022-08-17 13:38 UTC)
Platform Info:
  OS: Linux (x86_64-linux-gnu)
  CPU: 12 × AMD Ryzen 5 3600 6-Core Processor
  WORD_SIZE: 64
  LIBM: libopenlibm
  LLVM: libLLVM-13.0.1 (ORCJIT, znver2)
  Threads: 1 on 12 virtual cores

PR #4344, opened 2022-09-09
fix setting default font via `font`

BeastyBlacksmith:
Fix https://github.com/JuliaPlots/Plots.jl/issues/4337


t-bltg (approved):
I'm a bit lost with the `Plots` <-> `RecipesPipeline` <-> `RecipesBase` changes, but CI passes so I trust your judgment on this one.

Issue #4347, opened 2022-09-15
[BUG] `sp_width` undefined for pgfplotsx backend when setting plot_titles


JamieMair:
## Details
Issue when setting a title for a subplot using `Plots.jl`, MWE:

```julia
using Plots
pgfplotsx()

plt1 = plot(rand(100, 5))
plt2 = plot(rand(100))
plt3 = plot(rand(100))

plot(plt1, plt2, plt3, layout=(1,3), plot_titles=["(a)" "(b)" "(c)"])
```

Error stacktrace:

```julia
ERROR: UndefVarError: sp_width not defined
Stacktrace:
  [1] (::Plots.PGFPlotsXPlot)(plt::Plots.Plot{Plots.PGFPlotsXBackend})
    @ Plots ~/.julia/packages/Plots/lW9ll/src/backends/pgfplotsx.jl:120
  [2] _update_plot_object(plt::Plots.Plot{Plots.PGFPlotsXBackend})
    @ Plots ~/.julia/packages/Plots/lW9ll/src/backends/pgfplotsx.jl:1359
```

Proposed fix in `pgfplotsx.jl`:
```julia
sp_w, sp_h = width(bb2), height(bb2)
if sp[:subplot_index] == plt[:plot_titleindex]
    x = dx + sp_width / 2 - 10mm  # FIXME: get rid of magic constant
    y = dy + sp_height / 2
   # ..etc
```
to:

```julia
sp_w, sp_h = width(bb2), height(bb2)
if sp[:subplot_index] == plt[:plot_titleindex]
    x = dx + sp_w/ 2 - 10mm  # FIXME: get rid of magic constant
    y = dy + sp_h/ 2
   # ..etc
```

### Backends

This bug occurs on ( insert `x` below )

Backend      | yes | no  | untested
-------------|-----|-----|---------
gr (default) |     |     | x
pyplot       |     |     | x
plotlyjs     |     |     | x
pgfplotsx    |  x  |     |
unicodeplots |     |     |x
inspectdr    |     |     |x
gaston       |     |     |x

### Versions
Plots.jl version: Plots v1.31.7
Backend  version (`]st -m <backend(s)>`):  PGFPlotsX v1.5.0
Output of `versioninfo()`:
```
Julia Version 1.8.1
Commit afb6c60d69* (2022-09-06 15:09 UTC)
Platform Info:
  OS: Linux (x86_64-redhat-linux)
[...]

PR #4348, opened 2022-09-15
Updated `pgfplotsx.jl` to use correct variable names

JamieMair:
Should fix https://github.com/JuliaPlots/Plots.jl/issues/4347


BeastyBlacksmith:
Would you mind adding a test for that?

JamieMair:
> Would you mind adding a test for that?

Should be added now. Not the best test, but fails without the fix.

BeastyBlacksmith:
Thanks! You might want to checkout #3503

PR #4360, opened 2022-09-20
Add option for error bar styles

JamieMair:
Currently, the error bar lines follow the same line style as the series:

```julia
using Plots
using Random

Random.seed!(1234)
x = collect(1:10)
y = randn(length(x))
yerr = x ./ x .* 1.0
xerr = yerr ./ 5

plot(x, y, yerror=yerr, xerror=xerr, linestyle=:dash)
```
image

However, error bars should usually be solid, or at least customisable. In this PR, I have added options `:yerrorstyle` and `:xerrorstyle`, along with aliases, to control this style working on the recipe level, and as such should apply to all/most backends.

The result is changing the plot to:
```julia
plot(x, y, yerror=yerr, xerror=xerr, linestyle=:dash, yerrorstyle=:solid, xerrorstyle=:dot)
```
image

I haven't added any tests, as I am not sure the best way to do this. Any feedback would be appreciated!


BeastyBlacksmith:
How do you feel about making work like
```julia
using Plots
using Random

Random.seed!(1234)
x = collect(1:10)
y = randn(length(x))
yerr = x ./ x .* 1.0
xerr = yerr ./ 5

plot(x, y, linestyle=:dash)
yerror!(x,y, yerror=yerr, linestyle = :solid)
xerror!(x,y, xerror-xerr, linestyle = :dot)
```
?
Then you would get control of all series attributes for free and still can use the one function call API for the simple styling.

If not, we'd had to at least rename `xerrorstyle` to `x_error_linestyle`, since there are more things to style then only linetype.

BeastyBlacksmith:
I am actually against introducing a new attribute for that. What would be reasonable is, since `markerstrokecolor` is affecting the color of the whole errorbars, that `markerstrokestyle` sets the style for the whole errorbar.

JamieMair:
> I am actually against introducing a new attribute for that. What would be reasonable is, since `markerstrokecolor` is affecting the color of the whole errorbars, that `markerstrokestyle` sets the style for the whole errorbar.

This is good enough for me. Note, that since the default for `:markerstrokestyle` is `:solid`, this will change the default behaviour, as now the code:
```julia
plot(x, y, yerror=yerr, xerror=xerr, linestyle=:dash)
```
will produce:
image
whose style can be changed via:
```julia
plot(x, y, yerror=yerr, xerror=xerr, linestyle=:dash, markerstrokestyle=:dash)
```
which gives:
image
The main issue is that `:markerstrokestyle` is not supported in all backends, and this gives a warning for `Plots.GRBackend()`. How would you feel about having a single `:errorbarstrokestyle` for this instead that can be supported by default in all backends?

t-bltg:
Can this be merged ?

BeastyBlacksmith:
In principle yes, we still have the issue that GR does not support markerstrokestyle.. I opened an issue at GR.jl

t-bltg:
I see, https://github.com/jheinen/GR.jl/issues/470.
Let's delay the merge until addressed then.

PR #4367, opened 2022-09-21
Scaling axis limits by manual factor

yha:
Adds official API for specifying scaling factor in `widen`, or scaling axis limits in existing plots, which I've often found myself doing with various hacks.

### DIFF
--- a/src/axes.jl
+++ b/src/axes.jl
@@ -592,9 +592,9 @@ const _widen_seriestypes = (
 const default_widen_factor = 1.06
 
 # facor to widen axis limits by, or `nothing` if axis widening should be skipped
-function widen_factor(axis::Axis)
+function widen_factor(axis::Axis; factor = default_widen_factor)
     widen = axis[:widen]
-    widen isa Bool   && return widen ? default_widen_factor : nothing
+    widen isa Bool   && return widen ? factor : nothing
     widen isa Number && return widen
     widen == :auto   || @warn "Invalid value specified for `widen`: $widen"
 
@@ -604,7 +604,7 @@ function widen_factor(axis::Axis)
     for sp in axis.sps
         for series in series_list(sp)
             if series.plotattributes[:seriestype] in _widen_seriestypes
-                return default_widen_factor
+                return factor
             end
         end
     end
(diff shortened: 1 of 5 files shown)


src/axes.jl
```diff
@@ -560,21 +589,26 @@ const _widen_seriestypes = (
     :scatter3d,
 )
 
-function default_should_widen(axis::Axis)
-    if axis[:widen] isa Bool
-        return axis[:widen]
-    end
+const default_widen_factor = 1.06
```
BeastyBlacksmith, line 592: Can we make this a default keyword argument to `widen_factor` ?

yha: I'm not sure how that could work, since `widen_factor` takes the axis as argument and chooses whether to use the default based on its `:widen` and `:lims` attributes.

yha: Never mind, I think I understand now what you mean - `widen_factor(axis::Axis; default_widen_factor = 1.06)`?
The problem is you might want to do `scale_lims(..., default_widen_factor)` as is done now in the `bar` recipe. So the default factor would need be specified again there or in the methods of `scale_lims`.

BeastyBlacksmith: I see. I missed that. Still it would be good to have
```julia
widen_factor(axis::Axis; factor = default_widen_factor)
```
then, IMO.

yha: done

t-bltg:
Can this be merged ?

BeastyBlacksmith (approved):
There are conflicts to be fixed, but otherwise can be merged.

t-bltg:
@yha, this PR broke the docs:
```julia
 ┌ Error: failed to run `@example` block in src/tutorial.md:296-301
│ ```@example tutorial
│ #Pkg.add("RDatasets")
│ using RDatasets, StatsPlots
│ iris = dataset("datasets", "iris")
│ @df iris marginalhist(:PetalLength, :PetalWidth)
│ ```
│   value =
│    MethodError: no method matching widen(::Float64, ::Float64, ::Symbol)
│    Closest candidates are:
│      widen(::T) where T at operators.jl:895
└ @ Documenter.Expanders ~/.julia/packages/Documenter/yf96B/src/Utilities/Utilities.jl:32
```

Can you have a look to fix this ?

Issue #4366, opened 2022-09-21
[BUG] Crash on annotate with Unitful


ivan-boikov:
## Details

Using `annotate!` with Unitful coordinates fails on GR, InspectDR, PGFPlotsX and UnicodePlots:
```julia
Plots.plot([0,1]u"s", [0,2]u"m")
annotate!([0.5]u"s", [1]u"m", text("annotation"))
```
on PlotlyJS no crash, but annotation ends up on top left corner of plot. Without units no crashes on all backends, but it is assumed that units are the same as in `plot()` call, i.e. here annotation will be at (0.5s,1m)
```julia
Plots.plot([0,1]u"s", [0,2]u"m")
annotate!([0.5], [1], text("annotation"))
```
Not sure if it belongs here, Unitful or UnitfulRecipes though. Or if I should split the report per backend.

GR stacktrace:
```julia
ERROR: MethodError: no method matching wctondc(::Quantity{Float64, 𝐓, Unitful.FreeUnits{(s,), 𝐓, nothing}}, ::Quantity{Float64, 𝐋, Unitful.FreeUnits{(m,), 𝐋, nothing}})
Stacktrace:
[1] gr_display(sp::Plots.Subplot{Plots.GRBackend}, w::Measures.AbsoluteLength, h::Measures.AbsoluteLength, viewport_canvas::Vector{Float64}) 
    @ Plots ~/.local/share/julia/packages/Plots/FCUr0/src/backends/gr.jl:1050
[2] gr_display(plt::Plots.Plot{Plots.GRBackend}, fmt::String)
    @ Plots ~/.local/share/julia/packages/Plots/FCUr0/src/backends/gr.jl:699
[3] gr_display
    @ ~/.local/share/julia/packages/Plots/FCUr0/src/backends/gr.jl:665 [inlined]
[4] _display(plt::Plots.Plot{Plots.GRBackend}) 
    @ Plots ~/.local/share/julia/packages/Plots/FCUr0/src/backends/gr.jl:2250
[5] display(#unused#::Plots.PlotsDisplay, plt::Plots.Plot{Plots.GRBackend})
    @ Plots ~/.local/share/julia/packages/Plots/FCUr0/src/output.jl:164
[6] display(x::Any)
    @ Base.Multimedia ./multimedia.jl:328
[7] #invokelatest#2
    @ ./essentials.jl:729 [inlined]
[8] invokelatest
    @ ./essentials.jl:726 [inlined]
[...]

Issue #4114, opened 2022-02-22
[FR] UnitfulRecipes.jl as dependency?


juliohm:
Could this package be a dependency or maybe be ported fully to Plots.jl? I don't think users would complain about the additional Unitful.jl dependency in Plots.jl?

Can we get a all-batteries-included experience in terms of units without forcing users to load UnitfulRecipes.jl manually?

BeastyBlacksmith:
Personally I am in favor of this. It should be noted that this adds ~0.2 seconds of using time making it ~3s in total for Plots on my machine.
@briochemc what is your opinion on this?

briochemc:
Sorry for the delay, for some reason I did not get notified of this... 

I'm also in favor of this but I would like to note that there are a few rough edges that have not been ironed out (yet?) in UnitfulRecipes.

Pinging @jw3126 and @gustaphe too :)

jw3126:
Thanks, @juliohm for bringing this up! To me the question really is how often do people want unitful plotting. Personally, I use it quite a bit, but it might be that 99.9% of julia users don't. I really have no idea. Can one get insight into this? Is there telemetry info about how often Plots.jl + Unitful.jl is simultaneously installed or something like that?

juliohm:
Unitful.jl is a super lightweight dependency and I wonder if it wouldn't be
much easier for *end users* to just have these recipes added here.
Alternatively if they are just recipes, they could go into Unitful.jl
directly?

Em seg., 7 de mar. de 2022 04:43, Jan Weidner ***@***.***>
escreveu:

jw3126:
> Unitful.jl is a super lightweight dependency

It does add 0.2s load time, I would not call this super lightweight.

> I wonder if it wouldn't be much easier for *end users* to just have these recipes added here.

I fully agree it would be very nice for end users if unitful plots worked out of the box. Probably many users of Plots + Unitful are not even aware that they can work together nicely.

jw3126:
> Alternatively if they are just recipes, they could go into Unitful.jl
directly?

This is a very good suggestion. I imagine that most Plots.jl users don't use Unitful.jl, but that many (most?) Unitful.jl users also use Plots.jl .

mkborregaard:
Yes that seems like the right design

gustaphe:
[...]

PR #4370, opened 2022-09-22
conditionally load `UnitfulRecipes`

t-bltg:
Fix https://github.com/JuliaPlots/Plots.jl/issues/4366.
Fix https://github.com/JuliaPlots/Plots.jl/issues/4114.


t-bltg:
@BeastyBlacksmith, ok ?

t-bltg:
On second thought, I don't think `@eval Main` is a good idea.

PR #4371, opened 2022-09-24
Integrate `UnitfulRecipes` into `Plots`

t-bltg:
Fix https://github.com/jw3126/UnitfulRecipes.jl/issues/79.
Fix https://github.com/JuliaPlots/Plots.jl/issues/4366.
Fix https://github.com/JuliaPlots/Plots.jl/issues/4114.

Conditionally import `UnitfulRecipes` when `Unitful` is imported.

Per https://github.com/JuliaPlots/Plots.jl/issues/4366#issuecomment-1256919034, copy `src` and `test` code from `UnitfulRecipes` into `Plots`, with minor `Plots` adjustements (related to `@require`, `@reexport`).

Following this PR [`UnitfulRecipes`](https://github.com/jw3126/UnitfulRecipes.jl) should eventually be deprecated / archived.

Cc significant `UnitfulRecipes` contributors: @briochemc @gustaphe @jw3126.

Rework some `Plots` tests (causing import issues), which has the sided effect to fix long standing `Plots` tests warnings.


gustaphe:
Sounds good to me in principle, I'm happy to give any of my contributions to Plots.jl. 

I find it hard to say anything about the implementation, there seems to be quite some unrelated edits etc. But if Plots.jl maintainers approve I don't object.

src/unitful.jl
```diff
@@ -0,0 +1,277 @@
+# previously https://github.com/jw3126/UnitfulRecipes.jl
+# authors: Benoit Pasquier (@briochemc) - Jan Weidner (@jw3126)
```
jw3126, line 2:
```suggestion
# authors: Benoit Pasquier (@briochemc) - David Gustavsson (@gustaphe) - Jan Weidner (@jw3126)
```

src/unitful.jl
```diff
@@ -0,0 +1,277 @@
...
+
+module UnitfulRecipes
```
jw3126, line 4: I assume this is a 1:1 copy of UnitfulRecipes?

src/unitful.jl
t-bltg: Sure, I've copied the contents of https://github.com/jw3126/UnitfulRecipes.jl/blob/master/Project.toml, which weren't up-to-date I guess.

src/unitful.jl
t-bltg: I've made some slight syntax changes, fixed https://github.com/jw3126/UnitfulRecipes.jl/issues/79, changed some imports, but this should be really close to `UnitfulRecipes@master`. Anyway the tests were also copied and not modified (apart from some renames for consistency with others `Plots` tests, and some additions for https://github.com/jw3126/UnitfulRecipes.jl/issues/79). So we should be covered by CI which runs all the `UnitfulRecipes` tests ...

briochemc:
I'm all for integrating UnitfulRecipes.jl into Plots.jl as well! Just make sure this does not blow up TTFP! (Does it even affect it?)

src/unitful.jl
```diff
@@ -0,0 +1,293 @@
...
+"""
+    P_str(s)
+
+Creates a string that will be Protected from recipe passes.
+
+Example:
+```julia
+julia> plot(rand(10)*u"m", xlabel=P"This label is protected")
+
+julia> plot(rand(10)*u"m", xlabel=P"This label is not")
```
BeastyBlacksmith, line 218: I don't get this. In both cases `P` is being used. Also what is the effect of a label being protected ?
I guess it won't show units by default then?

BeastyBlacksmith (commented):
If there is any documentation we should also move that into main documentation.

@briochemc, @gustaphe, @jw3126 if any of you needs/wants push rights to keep working on this just react with a 👍🏻.

src/unitful.jl
t-bltg: Fixed.

t-bltg:
> Just make sure this does not blow up TTFP! (Does it even affect it?)

Regarding TTFP:

**master**
```julia
using Unitful, UnitfulRecipes, Plots

const a = 1u"m/s^2"
v(t) = a * t
x(t) = a / 2 * t^2

main() = begin
  t = (0:.01:100) * u"s"
  plot(x.(t), v.(t), xlabel="position", ylabel="speed") |> display
  return
end

@time main()
# 9.626337 seconds (5.55 M allocations: 289.536 MiB, 2.14% gc time, 96.40% compilation time: 15% of which was recompilation)
```

**PR**
```julia
using Unitful, Plots

const a = 1u"m/s^2"
v(t) = a * t
x(t) = a / 2 * t^2

main() = begin
  t = (0:.01:100) * u"s"
  plot(x.(t), v.(t), xlabel="position", ylabel="speed") |> display
  return
end

@time main()
# 9.159979 seconds (5.53 M allocations: 288.254 MiB, 1.45% gc time, 96.22% compilation time: 17% of which was recompilation)
```

t-bltg:
> If there is any documentation we should also move that into main documentation.

See https://github.com/JuliaPlots/PlotDocs.jl/pull/302 (might need some rework).

gustaphe:
Concerning the push right question, by "this" do you mean this PR or (this part of) (now) Plots.jl? In the former case I don't think there is any need, the transfer seems quite straightforward. In the latter case, I will probably keep contributing, and it's up to you if you want that to happen by PR or if I should merge my own changes (in conjunction with the others) like before. I'm not intimately familiar with the granularity of GitHub permissions.

BeastyBlacksmith:
> Concerning the push right question, by "this" do you mean this PR or (this part of) (now) Plots.jl? In the former case I don't think there is any need, the transfer seems quite straightforward. In the latter case, I will probably keep contributing, and it's up to you if you want that to happen by PR or if I should merge my own changes (in conjunction with the others) like before. I'm not intimately familiar with the granularity of GitHub permissions.

I meant the latter. I think, you know the code best and therefore I would prefer having you being able to make calls on your own, therefore I will send you an invite.

jw3126:
> If there is any documentation we should also move that into main documentation.
> 
> Also the issues need to be moved.
> 
> @briochemc, @gustaphe, @jw3126 if any of you needs/wants push rights to keep working on this just react with a 👍🏻.

Thanks! At the moment I don't use Unitful much, so I probably do not need push access. Should that change, I will come back to you.

t-bltg:
The docs build are successful locally and look decent in a browser, but https://github.com/JuliaPlots/PlotDocs.jl/pull/302 needs a new version of Plots with this PR merged.

Note that I've disabled the notebook example in the docs for now, but this can be reworked later.

So I will probably merge this PR when CI is green, unless comments / objections. Following that, the `UnitfulRecipes` docs should be available in https://docs.juliaplots.org/dev/.

briochemc:
> > Just make sure this does not blow up TTFP! (Does it even affect it?)
> 
> Regarding TTFP

That's nice! But I was thinking more about TTFP for non-Unitful users.

t-bltg:
There is no impact since loading of the added `UnitfulRecipes` code in this PR is conditioned to loading `Unitful` first. But I should only trust what I see / measure so here are the timings:

**master**
```julia
@time begin
  using Plots
  plot(1:2) |> display
end
 13.199953 seconds (9.41 M allocations: 612.501 MiB, 4.65% gc time, 56.57% compilation time: 12% of which was recompilation)
```

**pr**
```julia
@time begin
  using Plots
  plot(1:2) |> display
end
 13.254682 seconds (9.41 M allocations: 612.282 MiB, 4.77% gc time, 55.89% compilation time: 12% of which was recompilation)
```

t-bltg:
@jw3126, now that this is released starting from `Plots@1.34.2`, some todos:

1. [archive](https://docs.github.com/en/repositories/archiving-a-github-repository/archiving-repositories) https://github.com/jw3126/UnitfulRecipes.jl;
2. (maybe add `julia` upper bound as suggested in https://discourse.julialang.org/t/how-do-i-deprecate-a-package/14129/2);
3. add deprecation note in https://github.com/jw3126/UnitfulRecipes.jl/blob/master/README.md;
4. transfer opened issues to Plots;
5. (don't know if possible: transfer opened PRs to Plots).

Thanks all for the comments and suggestions.

The new docs are available at https://docs.juliaplots.org/stable/unitfulrecipes/unitfulrecipes/.

PR #4376, opened 2022-09-26
temp disable `savefig` during precomp on `windows`

t-bltg:


t-bltg:
Looks like yanking latest `Plots` release is more appropriate, until we know further.

Issue #4375, opened 2022-09-26
Plots v1.34.1 - Cannot Precompile in new project environment


paschermayr:
Hi there, 

1. I am unable to precompile Plots.jl on Windows-Julia 1.7.3 in a new project environment with only Plots.jl installed.

```
      Status `XXXX\projects\TestProject\Project.toml`
  [91a5bcdd] Plots v1.34.1
```

Error Message when trying to use Plots.jl:

```
GKS: cairoplugin.dll: can't load library, error 126 (0x7e)
ERROR: LoadError: SystemError: opening file "XXXX\\AppData\\Local\\Temp\\jl_g5Ue4As5MP.png": No such file or directory
Stacktrace:
  [1] systemerror(p::String, errno::Int32; extrainfo::Nothing)
```

I tried to solve this via `Pkg.build("GR")` from previous suggestions, but it did not help either.

2. Actually, I wanted to submit another issue for Plots `v1.34.0`, where the ylabel could no longer be chosen as a symbol, but the package was updated so fast I did not manage to submit. Would this work in Plots `v1.34.1`?:

```
using Plots
plot(1,2; ylabel = :test)
```

Best regards, 
Patrick

t-bltg:
@jheinen, even though `GR@master` suite is green for `windows`, this does sound related to the GR changes we did in `0.68 - 0.69`.

Since I'm not a windows user, do you have a hint ?

paschermayr:
> No, do you know which Plots version supported this, if any ?

Yes, 1.33.0 and below should support this. I do not know when this was introduced, but I used symbols as labels in my code for at least a year now.

> @paschermayr, does pinning Plots@1.34.0 work ?

Yes, then I can use Plots again. However, using a symbol for labels does not work here anymore as before (version 1.33.0 and below). In a fresh environment:

```
Pkg.pin(name="Plots", version="1.34.0")
using Plots

plot(1,2) #working
plot(1,2; ylabel = :testing) #not working

MethodError: no method matching iterate(::Symbol)
Closest candidates are:
  iterate(!Matched::Union{LinRange, StepRangeLen}) at C:\Users\Pasch\AppData\Local\Programs\Julia-1.7.3\share\julia\base\range.jl:826
  iterate(!Matched::Union{LinRange, StepRangeLen}, !Matched::Integer) at C:\Users\Pasch\AppData\Local\Programs\Julia-1.7.3\share\julia\base\range.jl:826
[...]

PR #4377, opened 2022-09-26
revert `GR` bump until `4375` is fixed

t-bltg:
https://github.com/JuliaPlots/Plots.jl/issues/4375.


t-bltg:
@jheinen, https://github.com/JuliaPlots/Plots.jl/issues/4375 is definitively linked to `GR@0.69` (since CI is green for this PR with `GR@0.68`) so I'm reverting this change until we fixed the aforementioned issue.

I should have been more careful before tagging `1.34.1` in https://github.com/JuliaPlots/Plots.jl/commit/4a60bea9e81129637e8e6d39788b9d3d091dc969.

If I may ask, could you open up a PR when increasing the GR version next time instead of pushing directly to `Plots@master` ?
This will prevent breaking `master` by running CI first, and also and mistakes like tagging too quickly.

jheinen:
Sorry for the confusion. Will open a pull request for this purpose in the future. Since in GR all tests (and also our CI system) had reported no more errors, I had assumed that everything is fine.

I can not explain the error. I will try tomorrow to reproduce the error somehow on a Windows system using plain GR.

Unfortunately, I will not be able to support from Wednesday (for 2 weeks) as I am on vacation and have no internet access.

t-bltg:
Thanks, this is not urgent as long as we keep GR to `0.68` for now.

I'm also limited in my ability to fix this since I'm on `linux`.

The issue is apparently triggered when saving the figure to a file.

Have some nice offline holidays ;)

PR #4402, opened 2022-09-30
use `withenv` instead of persistent `ENV` variables

t-bltg:
Leaving persistent `ENV` variables is a bad habbit.
If we really need those set because of the way `GR` works by design, we should clear them after usage.


t-bltg:
Windows `1.8` failure is unrelated.

PR #4411, opened 2022-10-01
Refactor to reduce allocations in `slice_arg!`

BioTurboNick:
It's a minor improvement. Refactored to avoid allocations due to type instability in `slice_arg!`

Test example:
```
const xxxx = rand(10, 10000) # lots of small series
plot(xxxx);
```

Before: `866.029 ms (4472221 allocations: 233.98 MiB)`

After: `832.345 ms (4470177 allocations: 232.73 MiB)`

(Numbers both include the effect of an unmerged RecipesPipeline PR, but the impact is the same)

### DIFF
--- a/src/args.jl
+++ b/src/args.jl
@@ -1753,11 +1753,13 @@ end
 # InputWrapper just gives the contents
 # anything else is returned as-is
 function slice_arg(v::AMat, idx::Int)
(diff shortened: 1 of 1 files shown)


t-bltg (approved):
lgtm

PR #4412, opened 2022-10-01
Faster extrema calculations for large series

BioTurboNick:
Due to type instability, the `expand_extrema` methods were very slow. Adding a type assert fixes that and greatly improves the performance of plotting large series.

Test example:
```
const x6 = rand(1000000, 10) # several large series
plot(x6);
```

Before: `1.004 s (19999416 allocations: 457.95 MiB)`

After: `209.896 ms (4526 allocations: 152.85 MiB)`

(Numbers both include the effect of an unmerged RecipesPipeline PR, but the impact is the same)


t-bltg (approved):
lgtm

PR #4414, opened 2022-10-02
Unitful ticks and lims

gustaphe:
Solves an issue from UnitfulRecipes: https://github.com/jw3126/UnitfulRecipes.jl/issues/82#issuecomment-1264380067

```julia
plot(randn(5)u"m")
plot!(; ylims=(0u"cm", 5u"cm"), yticks=[1u"cm", 2u"cm", 3u"cm"])
```

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,7 +1,7 @@
 name = "Plots"
 uuid = "91a5bcdd-55d7-5caf-9e0b-520d859cae80"
 author = ["Tom Breloff (@tbreloff)"]
-version = "1.35.0"
+version = "1.35.1"
 
 [deps]
 Base64 = "2a0f44e3-6c83-55bd-87e4-b1978d98bd5f"
--- a/Project.toml
+++ b/Project.toml
@@ -1,7 +1,7 @@
 name = "Plots"
 uuid = "91a5bcdd-55d7-5caf-9e0b-520d859cae80"
 author = ["Tom Breloff (@tbreloff)"]
-version = "1.35.1"
+version = "1.35.0"
 
 [deps]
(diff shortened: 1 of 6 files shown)


t-bltg (approved):
LGTM, pending CI.
You have merge rights now - if I'm not mistaken - so you can merge at will.

t-bltg:
We have a formatter in `Plots`, you need to run:
```julia
using JuliaFormatter
format(["src", "test"])
```
for the format check to pass.

Project.toml
```diff
@@ -1,7 +1,7 @@
 name = "Plots"
 uuid = "91a5bcdd-55d7-5caf-9e0b-520d859cae80"
 author = ["Tom Breloff (@tbreloff)"]
-version = "1.35.0"
+version = "1.35.1"
```
t-bltg, line 4: We usually prefer to push a dedicated commit with the ver num in the commit text for version bump (this help visually inspecting commits).

gustaphe: oh, okay. I'll do some force push magic once CI is happy.

t-bltg: If you do it now, you'll spare waiting for CI twice ;)

t-bltg: The policy here in `Plots` is to squash `PRs` before merging, again for clarity.
So what I meant was remove the version bump in this `PR`, merge when `CI` is green, and then edit `Project.toml` in the github interface to only increase the version, with the version in the title commit as e.g. https://github.com/JuliaPlots/Plots.jl/commit/87d2470df3203e378f43c6c2a877789c2d0e2a24.
I wasn't clear enough on that one (you'll know for the next `PR`, no need to change it now).

PR #4421, opened 2022-10-04
Don't assume aspect ratio if none given

### LINKED ISSUE #4416: [BUG] `aspect_ratio` not preserved after `plot!` call for Unitful data
Normally if you set the `aspect_ratio` in one `plot` call then it is preserved in subsequent `plot!` calls.  However, this preservation does not work currently for Unitful data.  MWE:

## Details

```julia
using Unitful: mm
using Plots: plot, plot!

u = 1
p1 = plot((1:3)u, (2:4)u, aspect_ratio=:equal)
plot!((2:3)u, (2:3)u) # aspect ratio remains :equal as expected

u = 1mm
p2 = plot((1:3)u, (2:4)u, aspect_ratio=:equal) # correct ratio here
plot!((2:3)u, (2:3)u) # aspect ratio no longer :equal now
#plot!(aspect_ratio = :equal) # restores :equal (so this is my current work around)

plot(p1,p2)
```

### Backends

I have tested this only with GR.

### Versions

Plots.jl version: v1.35.0
Backend  version (`]st -m <backend(s)>`): GR#master
Output of `versioninfo()`:

Julia Version 1.8.2
Commit 36034abf260 (2022-09-29 15:21 UTC)
Platform Info:
  OS: macOS (x86_64-apple-darwin21.4.0)
  CPU: 8 × Intel(R) Core(TM) i7-8559U CPU @ 2.70GHz
  WORD_SIZE: 64
  LIBM: libopenlibm
  LLVM: libLLVM-13.0.1 (ORCJIT, skylake)
  Threads: 8 on 8 virtual cores
Environment:
  JULIA_NUM_THREADS = 8

@t-bltg I am posting this here in Plots rather than over at UnitfulRecipes.  I hope that's the right process now.  This is not urgent because the work-around is an easy fix.

gustaphe:
Close #4416

### DIFF
--- a/src/unitful.jl
+++ b/src/unitful.jl
@@ -123,7 +123,7 @@ Attribute fixing
 ===============#
 # Aspect ratio
 function fixaspectratio!(attr, u, axisletter)
-    aspect_ratio = get!(attr, :aspect_ratio, :auto)
+    aspect_ratio = get(attr, :aspect_ratio, :auto)
     if aspect_ratio in (:auto, :none)
         # Keep the default behavior (let Plots figure it out)
         return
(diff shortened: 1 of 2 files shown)


t-bltg (approved):
LGTM.

PR #4444, opened 2022-10-07
Adds extra_kwarg legend_wfactor to control the width of legend with gr.

Gregstrq:
This PR adds an extra_kwarg `legend_wfactor` for GR backend.
This extra_kwarg is multiplied with `legend_width_factor` and effectively controls the width of the legend.

The reason for this PR is the following. I have been trying to create a common legend for a grid of subplots, which consists of several columns. For that, I created empty subplots with only the legend displayed.
The problem was that the default legend width was so small, that it was hard to discern the linestyles.

### DIFF
--- a/src/backends/gr.jl
+++ b/src/backends/gr.jl
@@ -1324,6 +1324,7 @@ function gr_get_legend_geometry(viewport_plotarea, sp)
     end
 
     legend_width_factor = (viewport_plotarea[2] - viewport_plotarea[1]) / 45 # Determines the width of legend box
+    legend_width_factor *= get(sp[:extra_kwargs], :legend_wfactor, 1)
     legend_textw = legendw
     legend_rightw = legend_width_factor
     legend_leftw = legend_width_factor * 4


BeastyBlacksmith:
Note, that, if this gets accepted, needs to get documented [here](https://docs.juliaplots.org/stable/backends/#Supported-:subplot-:extra_kwargs)

t-bltg:
Please also add a before / after image to this PR, for discussion.

Gregstrq:
Here is the result with `legend_wfactor=1` (default):
narrow_legend

And here is the result with `legend_wfactor=2`:
wide_legend

Funny thing. If I use `legend_wfactor=2`, the image looks correct in Pluto, but if I save it directly to `pdf`, the width of the legend is the same as in the first example. The image that I attach here was obtained by saving to `svg` and applying transforming with `svg2pdf`.

t-bltg:
Thanks for the images, looks nice.

As @BeastyBlacksmith, you can also make a `PR` in [`PlotDocs`](https://github.com/JuliaPlots/PlotDocs.jl) to modify [this file](https://github.com/JuliaPlots/PlotDocs.jl/blob/master/docs/src/backends.md#fine-tuning) for the new `legend_wfactor`.

> Funny thing. If I use legend_wfactor=2, the image looks correct in Pluto, but if I save it directly to pdf, the width of the legend is the same as in the first example. The image that I attach here was obtained by saving to svg and transforming with svg2pdf.

This is worth investigating.

t-bltg:
> Not sure if github notifies you when other PR or issue mentions the current PR.

Yes, I've seen those.

> I am ready to open an issue once we deal we this PR.

It would be better to try to fix this now, although it might be a lower-level issue with the `GR` backend.

So I'll defer to @BeastyBlacksmith's judgment to see if this can go in now or wait for the `pdf` issue to be fixed.

BeastyBlacksmith:
> If I use `legend_wfactor=2`, the image looks correct in Pluto, but if I save it directly to `pdf`, the width of the legend is the same as in the first example. The image that I attach here was obtained by saving to `svg` and transforming with `svg2pdf`.

The problem description reads like an implementation issue rather than a backend issue to me (though I haven't checked). So we should resolve this before merging. We'd either need to check whether we can reproduce the issue on plain GR or check whether we have different code paths for svg and pdf and currently miss one.

Gregstrq:
Here is an MWE to play around:
```
using Plots, LaTeXStrings

l = grid(1,3)
pl1 = plot((1:2)', label = [L"N=18" L"N=16"], linestyle=[:solid :dashdot], frame=:none, size = (0.3,0.15).*390,
        extra_kwargs=Dict(:subplot=>Dict(:legend_wfactor=>2.0)))
pl2 = plot((1:2)', label = [L"N=14" L"N=12"], linestyle=[:dashdotdot :dash], color=[3 4], frame=:none, size = (0.3,0.15).*390,
        extra_kwargs=Dict(:subplot=>Dict(:legend_wfactor=>2.0)))
pl3 = plot(1, label = L"N=10", linestyle=:dot, color=5, frame=:none, size = (0.3,0.15).*390,
        extra_kwargs=Dict(:subplot=>Dict(:legend_wfactor=>2.0)))
pl = plot(pl1, pl2, pl3, layout = l, legendfontsize=8, size=(1.0,0.15).*390)
```

Gregstrq:
How does one create a grid on plain GR?

t-bltg:
I you mean a grid of `subplot`s, this is not possible in `GR`.

t-bltg:
Superseded by https://github.com/JuliaPlots/Plots.jl/pull/4496.

PR #4445, opened 2022-10-07
DUMMY: investigating Windows precompilation failure in 1.8.2

BioTurboNick:
I was curious about why Plots had an issue precompiling on Windows in https://github.com/JuliaLang/julia/issues/46989, and I haven't been able to reproduce the issue on my system. So I'm trying to revert the workaround to see if the problem remains in the CI.


t-bltg:
As I recall, it appears to produce the same error.

BioTurboNick:
Nearest I can figure is that GR is failing to create the initial temporary file in `_show` (or at least, not in the expected location), which Plots then tries to read from to copy over to the intended permanent file.

Why it only seems to happen in CI so far, unclear. I'll keep poking.

BioTurboNick:
Irony: Using tmate to debug it on the CI machine, but it doesn't support scrolling, and the stack trace is too damn long 😆. And sadly resistant to my AbbreviatedStackTraces.

BioTurboNick:
Either GR is failing to create the file, perhaps via some change in libuv?, or it's being deleted before control returns to Plots. And unclear how precompilation environment differs from the normal REPL execution, where things work fine.

Since LibUV is tightly bound to the Julia version, I don't think it'll be possible to debug that here. So next is to look at GR.

Inquiring if there's an easy way to get debug information out of GR: https://github.com/sciapp/gr/issues/161

Next obvious question, perhaps, is why this error doesn't show up in GR.jl's CI?

Other question is how far back in Plots does this problem happen? The failure first appears in Plots v.1.34.0 - when this whole method of precompilation was introduced.

BioTurboNick:
GR.savefig works fine in the same environment.

t-bltg:
I agree that reproducing the issue in plain `GR` might help narrowing the issue.

BioTurboNick:
I think I've nailed it.

The width and height of the viewport are too small. 1024x768 leads to failure to render and save the file.

If I force `display_width_px` and `display_height_px` to be the values I observed from precompilation on the CI machine, I can reproduce the failure locally.

https://github.com/JuliaPlots/Plots.jl/blob/e9fb3b35eadb951965da34abcd1553482b7d547c/src/backends/gr.jl#L671-L672

Two things to act on:
1) is to make this code robust to small display port sizes
2) GR should presumably produce an error when it is unable to write to the file?

t-bltg:
> But it does suggest that GR is silently failing internally.

The does show the limit of having a `c` backend, an not a pure `julia` one.
Until we have something approaching that (or have `sciapp/gr` low level debugging tools), we are forced to cope with nasty bugs like this one.

mkitti:
Did this ever come to any conclusion? I've been thinking of taking a look.

BioTurboNick:
> Did this ever come to any conclusion? I've been thinking of taking a look.

I don't believe so. I didn't want to mess with GR so I stopped there.

t-bltg:
Superseded by https://github.com/JuliaPlots/Plots.jl/pull/4617.

PR #4334, opened 2022-09-04
rework precompilation

t-bltg:
1. rework examples structure (mostly syntax, split imports);
2. use `SnoopPrecompile`.

PR #4451, opened 2022-10-09
Adjust workaround for Windows precompilation on Julia 1.8.2+

BioTurboNick:
My goal here is to still precompile on normal computers, but only skip it if there's a problem in CI

Providing a warning will also allow the issue to be monitored, and if someone ever encounters it outside of the CI system, maybe we can learn more about the issue.

Adjustment to #4406, based on work in #4445

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -55,6 +55,8 @@ jobs:
       # surface errors during precompilation
       # auto-precompile in build step suppresses errors currently.
       - name: Precompile
+        env:
+          JULIA_DEBUG: Plots
         shell: julia --project=@. --color=yes {0}
         run: |
           using Pkg; Pkg.precompile()
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -49,7 +49,8 @@ jobs:
       - uses: julia-actions/cache@v1
 
       - uses: julia-actions/julia-buildpkg@latest
-        env: JULIA_PKG_PRECOMPILE_AUTO: 0
+        env:
+          JULIA_PKG_PRECOMPILE_AUTO: 0
       
       # surface errors during precompilation
       # auto-precompile in build step suppresses errors currently.
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -47,7 +47,16 @@ jobs:
         with:
           version: ${{ matrix.version }}
       - uses: julia-actions/cache@v1
+
       - uses: julia-actions/julia-buildpkg@latest
+        env: JULIA_PKG_PRECOMPILE_AUTO: 0
+      
+      # surface errors during precompilation
+      # auto-precompile in build step suppresses errors currently.
+      - name: Precompile
+        shell: julia --project=@. --color=yes {0}
+        run: |
+          using Pkg; Pkg.precompile()
 
       - name: Test upstream RecipesBase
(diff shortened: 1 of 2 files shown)


BioTurboNick:
This works to avoid the CI error but the warning doesn't show up...

t-bltg:
I'm a bit reluctant to use `try catch`, but if there is not other solution :shrug: 

Can you check that this has no impact on `TTFP` by running [this snippet](https://github.com/JuliaPlots/Plots.jl/pull/4334#issuecomment-1250745773) against this PR and master ?

For this `PR` to go in, it's mandatory that the warning shows up somehow ...

BioTurboNick:
Will do, though my expectation is that try/catch has 0 cost *unless* an exception occurs.

Also will see if there's a way to get the warning to show up...

BioTurboNick:
Hmm. Tried a few things and no way so far to get precompilation to print anything unless there's an error thrown?

Inquiring here: https://github.com/JuliaLang/Pkg.jl/issues/3224

src/precompilation.jl
```diff
@@ -5,25 +5,35 @@ if get(ENV, "PLOTS_PRECOMPILE", "true") == "true"
...
-                $func()
-            end)
+                    try
+                        # During precompilation on the Windows GitHub CI in Julia 1.8.2+,
+                        # can fail here due to GR potentially failing to write out the
+                        # temporary file. Pending a fix, swallow the error and continue.
+                        $func()
+                    catch
+                        @debug "Plots: Failed a trial save during precompilation"
+                    end
```
t-bltg, line 34:
```suggestion
                        $func()
                    catch e
                        if e isa SystemError && Sys.iswindows() && get(ENV, "CI", "false") == "true"
                            # During precompilation on the Windows GitHub CI in Julia 1.8.2+,
                            # can fail here due to GR potentially failing to write out the
                            # temporary file. Pending a fix, swallow the error and continue.
                        else
                            throw(e)
                        end
                        @debug "Plots: Failed a trial save during precompilation"
                    end
```

If the warning doesn't work as expected, then maybe silently ignore precompilation `SystemEror` on `windows` `ci`, else hard error.

.github/workflows/ci.yml
```diff
@@ -47,7 +47,19 @@ jobs:
...
+          JULIA_PKG_PRECOMPILE_AUTO: 0
+      
+      # surface errors during precompilation
+      # auto-precompile in build step suppresses errors currently.
+      - name: Precompile
+        env:
+          JULIA_DEBUG: Plots
+        shell: julia --project=@. --color=yes {0}
+        run: |
+          using Pkg; Pkg.precompile()
```
t-bltg, line 62:
```suggestion
```

That does not seem to work as intended right ? No warning appear on `windows 1.8.2`.

BioTurboNick: Yes, at the moment it does not. Was hoping for another workaround from the Pkg issue.

src/precompilation.jl
BioTurboNick: That's fair. We probably want `rethrow()` though?

t-bltg: Ha, yes I took `rethrow(e)` note in https://docs.julialang.org/en/v1/base/base/#Base.rethrow too seriously.

t-bltg:
Superseded by https://github.com/JuliaPlots/Plots.jl/pull/4617.

PR #4453, opened 2022-10-12
Apply recipes to keyword arguments

gustaphe:
There are some keyword arguments which it makes sense to interpret as "on a data axis" - meaning that it would be useful to query the recipe pipeline for how to interpret user provided types. 

Some such arguments that I have identified are
- [x] lims
- [x] ticks
- [ ] annotation coordinates
- [ ] ribbon/error/fillto
- [ ] `marker_z` and friends
- [ ] colorbar lims and ticks
- [ ] `marker_size` (this one is questionable)

So far, a number of these have been hardcoded for `Unitful` and `Dates`, which are two of the most obvious use cases, but if a user wanted to be able to set limits in their own data type they would have to either override unexported Plots functions or handle the conversion to `Tuple{Float, Float}` themselves. With this PR, a type recipe is enough.

### DIFF
--- a/src/axes.jl
+++ b/src/axes.jl
@@ -81,6 +81,7 @@ end
 function attr!(axis::Axis, args...; kw...)
     # first process args
     plotattributes = axis.plotattributes
+    dummy_attributes = Dict{Symbol, Any}(:plot_object=>first(axis.sps).plt)
     for arg in args
         process_axis_arg!(plotattributes, arg)
     end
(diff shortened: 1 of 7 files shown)


src/axes.jl
```diff
@@ -81,6 +81,7 @@ end
 function attr!(axis::Axis, args...; kw...)
     # first process args
     plotattributes = axis.plotattributes
+    dummy_attributes = Dict{Symbol, Any}(:plot_object=>first(axis.sps).plt)
```
BeastyBlacksmith: What are those needed for?

gustaphe: Good question. `RecipesPipeline._apply_type_recipe` has
https://github.com/JuliaPlots/Plots.jl/blob/399ee6a415dde2b3339f44cbc09e1eab616d5dd4/RecipesPipeline/src/type_recipe.jl#L18-L25
(does Github show a summary?)
where it asks the first argument for `:plot_object`, which it then uses for pre- and postprocessing. Perhaps the function should instead be extended to accept an `Axis` object instead or something, this just seemed like the minimum intrusion operation to me.

Ickaser:
Has Plots moved far enough towards 2.0 that it would be difficult to revive this PR? Time is limited as always but this seems like a functionality related to eventually solving #4741, which would be useful to me, so I would be willing to make some efforts here

BeastyBlacksmith:
I don't think any of the relevant portions of the code have changed, so it shouldn't be hard to port the PR.
It's not yet completely clear to me, what the implications of doing this are though.

PR #4456, opened 2022-10-14
add quality test using `Aqua`

t-bltg:
- ~~drop `Contour`, `LaTeXStrings` and `Latexify` hard dependencies, since they are only used by `pgfplotsx` backend;~~
- resolve methods ambiguities detected by [`Aqua`](https://github.com/JuliaTesting/Aqua.jl);
- fix `pie` aspect_ratio bug (leading to a `Gaston` warning);
- syntax simplifications;
- cleanup dead code;
- improve coverage.


t-bltg:
> What is the benefit of that

Well, `pgfplotsx` is an optional backend, so should be `Contour` and `Latexify`.
Compatibility is still checked in [here](https://github.com/JuliaPlots/Plots.jl/pull/4456/files#diff-f6be56dc5fd8749a6cfca149f4d324d86aee499f760c2c2f95c4672598acca0fR1194-R1196).

This ambiguity was spotted by `Aqua` tests as stale dependencies (since only triggered by `@require`).

BeastyBlacksmith:
> Well, `pgfplotsx` is an optional backend, so should `Contour` and `Latexify` be optional dependencies for `Plots`. Maybe we should create a glue package for `PGFPlotsX` ?

These are all packages where its not worth the hassle to put that much work into it over just shipping them. In principle I agree it would be nice to have a better mechanism for opional dependencies, especially because of compat. But that would need to land in Pkg first.

For reference:
```julia
julia> @time_imports using Contour
      4.6 ms  Contour
julia> @time_imports using Latexify
      0.6 ms  Requires
      0.9 ms  LaTeXStrings
     27.5 ms  MacroTools
      0.4 ms  Formatting
      4.6 ms  OrderedCollections
     32.9 ms  Latexify 41.86% compilation time (7% recompilation)
```

t-bltg:
Thanks, I've removed the breaking `PGFPlotsX` changed, which should maybe be included in a future `Plots 2.0`.

BeastyBlacksmith:
> In principle I agree it would be nice to have a better mechanism for optional dependencies, especially because of compat. But that would need to land in Pkg first.

Relevant PR to track: https://github.com/JuliaLang/Pkg.jl/pull/3216

PR #4469, opened 2022-10-21
bump cov - more `PyPlot` exs

t-bltg:
- nearly all `PyPlot` disabled examples are ok now;
- bump coverage (:tired_face:, never enough), add more tests.

### DIFF
--- a/src/backends/gr.jl
+++ b/src/backends/gr.jl
@@ -2087,19 +2087,15 @@ function gr_draw_surface(series, x, y, z, clims)
             # Only triangles - connections have to be 0-based (indexing)
             ci, cj, ck = series[:connections]
             if !(length(ci) == length(cj) == length(ck))
-                throw(
-                    ArgumentError(
-                        "Argument connections must consist of equally sized arrays.",
-                    ),
-                )
+                "Argument connections must consist of equally sized arrays." |>
+                ArgumentError |>
+                throw
             end
             cns = map(i -> ([3, ci[i] + 1, cj[i] + 1, ck[i] + 1]), eachindex(ci))
         else
-            throw(
-                ArgumentError(
-                    "Unsupported `:connections` type $(typeof(series[:connections])) for seriestype=$st",
-                ),
-            )
+            "Unsupported `:connections` type $(typeof(series[:connections])) for seriestype=$st" |>
+            ArgumentError |>
+            throw
         end
         facecolor = if series[:fillcolor] isa AbstractArray
             series[:fillcolor]
--- a/src/backends/gr.jl
+++ b/src/backends/gr.jl
@@ -542,7 +542,7 @@ end
 
 function _cbar_unique(values, propname)
     out = last(values)
-    if any(x != out for x in values)
+    if any(map(x -> x != out, values))
         @warn "Multiple series with different $propname share a colorbar. " *
               "Colorbar may not reflect all series correctly."
     end
@@ -696,9 +696,7 @@ function gr_display(plt::Plot, dpi_factor = 1)
     gr_fill_viewport(viewport_canvas, plt[:background_color_outside])
 
     # subplots:
-    for sp in plt.subplots
-        gr_display(sp, w * px, h * px, viewport_canvas)
-    end
+    foreach(sp -> gr_display(sp, w * px, h * px, viewport_canvas), plt.subplots)
 
     GR.updatews()
 end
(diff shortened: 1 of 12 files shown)


src/backends/gr.jl
```diff
@@ -542,7 +542,7 @@ end
 
 function _cbar_unique(values, propname)
     out = last(values)
-    if any(x != out for x in values)
+    if any(map(x -> x != out, values))
```
BeastyBlacksmith, line 545: that is turning something that does not allocate into something that does
```julia
julia> f(val) = any(x != 1 for x in val)
f (generic function with 1 method)

julia> g(val) = any(map(x -> x != 1, val))
g (generic function with 1 method)

julia> @btime f($(collect(1:100)))
  1.936 ns (0 allocations: 0 bytes)
true

julia> @btime g($(collect(1:100)))
  28.060 ns (1 allocation: 160 bytes)
```

src/backends/gr.jl
```diff
@@ -1493,18 +1487,14 @@ function gr_draw_axes(sp, viewport_plotarea)
         x_bg, y_bg = RecipesPipeline.unzip(GR.wc3towc.(area_x, area_y, area_z))
         GR.fillarea(x_bg, y_bg)
 
-        for letter in (:x, :y, :z)
-            gr_draw_axis_3d(sp, letter, viewport_plotarea)
-        end
+        foreach(letter -> gr_draw_axis_3d(sp, letter, viewport_plotarea), (:x, :y, :z))
```
BeastyBlacksmith, line 1490: I'm not very comfortable with the amount of closures you are introducing lately. I hope, you tested, that you don't run into https://github.com/JuliaLang/julia/issues/15276
it would be more robust to wrap thos in `let` block for safety.

src/backends/gr.jl
t-bltg: I missed that it was a generator, will fix later.

src/backends/gr.jl
t-bltg: I'll think about it. To me atm, being concise is more important. Also, this is not a critical function called a thousand times per plot. But I get your point.

src/backends/gr.jl
t-bltg: https://github.com/JuliaPlots/Plots.jl/pull/4470

src/backends/gr.jl
BeastyBlacksmith: Its not this specific place. You do that all over the codebase, which is fine if you made sure this does not impact performance, otherwise it will just make work to revert it where it did.

> To me atm, being concise is more important.

Users usually disagree about that ^^

t-bltg: **master**
```julia
julia> using Plots; @time plot(1:2) |> display
  6.766723 seconds (713.40 k allocations: 35.765 MiB, 0.65% gc time, 95.10% compilation time: 15% of which was recompilation)
```

**1.35.0**
```julia
julia> using Plots; @time plot(1:2) |> display
  6.649556 seconds (714.62 k allocations: 35.905 MiB, 0.67% gc time, 95.05% compilation time: 15% of which was recompilation)
```

BeastyBlacksmith: That case is covered by the benchmarks ci job. But there is is enough code that was changed that is not covered by this.
I'm just saying you should test _the specific function you are changing_ before merge, whether stylistic changes impact performance.

t-bltg: I hear that, but there is theory and practice.
I'm also saying that having good development practices involves good coverage, which I'm trying hard to improve here. I'm not even talking about the backend bugs, ...
Performance discussions should always happen last.

BeastyBlacksmith: > I'm also saying that having good development practices involves good coverage, which I'm trying hard to improve here.

I see that. And in general, I appreciate that. But potentially making existing code worse just to have a better score on some debatable development metric is the wrong way to go.

t-bltg: IMO, there is nothing to be discussed when dealing with widely used codebases such as `Plots`.
Robustness **should** prevail over performance.
There will be a time where we will focus on reducing allocations, which is `julia`'s weak point these days.
But first robustness through coverage (which doesn't mean I will write sloppy code). Non tested code is bound to fail.

BeastyBlacksmith: Its not increasing robustness if you make functional equivalent changes to reduce line count.
Either that code paths gets tested or not. It's gaming numbers.
I mean sometimes it is beneficial to use `map` instead of a comprehension and sometimes its not.

PR #4479, opened 2022-10-30
eachindex call incorrectly assumes that the returned index is IndexLinear

bkamins:
I am not sure if other calls to `eachindex` in Plots.jl have the same problem. The fix should resolve https://discourse.julialang.org/t/plots-borking-on-sentinelarrays-produced-by-csv-read/89505, but maybe there are other cases in the code when the same issue would be raised (I have checked the `eachindex` calls, but often I was not sure if the type on which it is called returns `IndexLinear` by default or not).

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -97,6 +97,7 @@ PlotlyKaleido = "f2990250-8cf9-495f-b13a-cce12b45703c"
 PyCall = "438e738f-606a-5dbb-bf0a-cddfbfd45ab0"
 PyPlot = "d330b81b-6aea-500a-939a-2ce795aea3ee"
 RDatasets = "ce6b1742-4840-55fa-b093-852dadbb1d8b"
+SentinelArrays = "91c51154-3ec4-41a3-a24f-3f23e20d615c"
 StableRNGs = "860ef19b-820b-49d6-a774-d7a799459cd3"
 StaticArrays = "90137ffa-7385-5640-81b9-e52037218182"
 StatsPlots = "f3b207a7-027a-5e70-b257-86293d7955fd"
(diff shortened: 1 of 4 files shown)


bkamins:
This fix assumes that `x` in the code is guaranteed to be `AbstractArray`. I am not sure if it is guaranteed (I would assume it is, but I am not sure as from the signature it is not clear), so someone knowing Plots.jl should confirm this. Thank you!

bkamins:
Also, more generally - it would be good to confirm that `_cycle` works correctly with OffsetArrays.jl (I am not using it, but this might be another issue of a similar nature).

bkamins:
x-ref https://github.com/JuliaLang/julia/pull/47389

t-bltg:
Thanks, git blame show that this is not due to a recent change.

Some tests (and maybe `OffsetArrays` ones) should be added to https://github.com/JuliaPlots/Plots.jl/blob/master/test/test_utils.jl, so that we don't run into this again.

bkamins:
>  you don't mind me pushing in your patch-1 branch ?

sure - it will be much faster for someone knowing the internals to make these things correct than me. Thank you!

t-bltg:
Thanks, I'm trying to understand the failing example to cook up a test.
Do you know which package wraps data into `SentinelArrays` ? Is it `DataFrames` ?

bkamins:
No, CSV.jl creates `SentinelArrays`. I recommend you use SentinelArrays.jl directly for tests.

A reproducer is:
```
julia> using Plots, SentinelArrays

julia> scatter(ChainedVector([[1, 2], [3, 4]]), 1:4)
ERROR: MethodError: no method matching _cycle(::ChainedVector{Float64, Vector{Float64}}, ::SentinelArrays.ChainedVectorIndex{Vector{Float64}})
```

Another reproducer is:
```
julia> using OffsetArrays

julia> scatter(OffsetArray(1:4, -2), 1:4)
```
which produces the following plot (clearly incorrect due to index alignment and `_cycle` design):
image

bkamins:
Yes - that is why this PR only fixes SentinelArrays.jl issue. For OffsetArrays.jl a decision would need to be made what is the desired behavior.

Issue #4481, opened 2022-10-30
[FR] Add support for skipmissing


bkamins:
Currently the following fails:
```
plot(skipmissing(1:5))
```
It would be convenient that Plots.jl materialized `SkipMissing` objects as vectors automatically, so that users do not have to do:
```
plot(skipmissing(1:5) |> collect)
```
(this is what they have to do now.)

CC @t-bltg @nalimilan

t-bltg:
I think this is reasonable to ask.

PR #4482, opened 2022-10-30
collect `skipmissing`

t-bltg:
Fix https://github.com/JuliaPlots/Plots.jl/issues/4481.


bkamins (approved):
It is nice that there is an abstraction layer, so that the fix is indeed easy. Thank you!

Issue #2206, opened 2019-10-02
Horizontal legends


fbruetting:
Would be cool if we could have legend also in the style of horizontal rows instead of a vertical lists. 😎️

mkborregaard:
Yes very cool

szcf-weiya:
It would be easy in PGFPlotsX with the help of `extra_kwargs`, since `legend style={legend columns=-1}` in pgfplots draws the legend horizontally. For example,
```julia
plot(rand(3,2), extra_kwargs=:subplot, legend=:outertop, legend_columns=-1)
```
![image](https://user-images.githubusercontent.com/13688320/103476971-2f404700-4df5-11eb-9e88-dbc70c6f321a.png)

songhanzhang:
Has this feature been added in the GR backend?

BeastyBlacksmith:
No, currently only implemented for the pgfplotsx-backend

t-bltg:
> Has this feature been added in the GR backend ?

As of `Plots@1.36.0`.

Issue #4491, opened 2022-11-03
[BUG] Legend at `:top` and `:bottom` dislocated to the left


luboshanus:
## Details

Legend with position at `legend=:top` and `:bottom` is dislocated to the left edge and almost out of frame.

<img width="760" alt="Screenshot 2022-11-03 at 11 37 38" src="https://user-images.githubusercontent.com/2473206/199700420-52672475-f976-470b-bd39-93d5e5bb17c7.png">

Similar appears for `legend=:bottom`.

### Backends

This bug occurs on ( insert `x` below )

Backend      | yes | no  | untested
-------------|-----|-----|---------
gr (default) |  x   |     |
pyplot       |     |     | x
plotlyjs     |     |     | x
pgfplotsx    |     |     | x
unicodeplots |     |     | x
inspectdr    |     |     | x
gaston       |     |     | x

### Versions

[91a5bcdd] Plots v1.35.7

```
Julia Version 1.8.2
Commit 36034abf260 (2022-09-29 15:21 UTC)
Platform Info:
  OS: macOS (x86_64-apple-darwin21.4.0)
  CPU: 20 × Intel(R) Xeon(R) W-2150B CPU @ 3.00GHz
  WORD_SIZE: 64
  LIBM: libopenlibm
  LLVM: libLLVM-13.0.1 (ORCJIT, skylake-avx512)
  Threads: 1 on 20 virtual cores
```

t-bltg:
Sorry about this. Simmilar to https://github.com/JuliaPlots/Plots.jl/issues/4485. Will fix.

Issue #4485, opened 2022-11-01
subplots + legends + layout broken in 1.35.6


wmkoolen:
Dear all,

After updating to 1.35.6 the legends in subplots are not placed correctly anymore. This is fine with 1.35.5. Here's a minimal example:

```julia
using Plots

a = plot(1:5,1:5,label="bla", legend=:left)
b = plot(1:5,1:5,label="flu")
plot(a,b, layout=@layout [a;b])

```
![example](https://user-images.githubusercontent.com/29552933/199191794-3658424a-ebf1-46a2-a5e3-1f75eaf16ca1.svg)

Notice how the "a" plot's legend is gone. A plot(a) reveals that it is still there in the original. Also if the layout argument is omitted, the plots look fine. So the combination of subplots with layout messes things up.

t-bltg:
Thanks for reporting, I'll have a look asap.

t-bltg:
Found it, just an embarassing typo.

Issue #2423, opened 2020-02-29
Legends metaissue


mkborregaard:
An overhaul of Plots' legend capabilities would be really useful. Here are some relevant open issues

-[] `contour` legend does not scale with `clims` #533
-[] Legend position with PlotlyJS #574 
-[] shared legend for subplots on some backends #673 
-[] Documentation didn't allow me to figure out how to change the names in legends #995
-[] Legend doesn't appear on quiver plot #1589
-[] `aspect_ratio = :equal` with `legend = false` not correct for `plotlyjs()` #1603
-[] legendfont does not affect legendtitle #1856
-[] markerstrokewidth ignored for legend with PyPlot backend #2033
-[] Legend placement with gr appears buggy #2202
-[] Horizontal legends #2206
-[] legend = :outerright is missing from the documentation #2207
-[] Legend positioned far from borders overlaping data - GR #2330
-[] control the size of colorbars? #2345
-[] Errorbar is missing in the legend #2360
-[] legend position with outer legends and vertical subplots #2369
-[] Twinx doesn't combine legends #2417 

As well as a desire to share color bars and legends among subplots, allow multiple columns in legends and improvement of the alignment of axes (both with and without `link`) in layouts with legends.

isentropic:
https://github.com/JuliaPlots/Plots.jl/issues/2033 should be fixed

isentropic:
#1856 closed

mkborregaard:
Yay @isentropic you're on a roll!

isentropic:
Perhaps there could be another explicit legend object, that is actually an empty axis with empty plot on it. Similar to what you showed here https://github.com/JuliaPlots/Plots.jl/issues/673 but it would read as,
```
plot(
           scatter(rand(100,3), legend = false),
               scatter(rand(100,3), legend = false),
                   legend((1:3)),  # actually it is scatter((1:3)', xlim = (4,5), legend = true, framestyle = :none) under the hood
                   layout = @layout([a b c{0.1w}]))

```

isentropic:
That is we could perhaps add a new series type called `legend` that would actually be an invisible shrunk axis so that all backends can support it. This way it won't feel as a hack and can be presented in documentation.
`legend` would be able to have few recipes supporting only shapes and words given to it

isentropic:
This way we can solve,
#2345
#673
#995
#2369

infixing (inplotting) legend series on top of other axes would allow users to have an explicit control over legends if they wish

[...]

PR #4496, opened 2022-11-05
`gr`: support horizontal legends

t-bltg:
Fix https://github.com/JuliaPlots/Plots.jl/issues/2206.
Fix https://github.com/JuliaPlots/Plots.jl/issues/4497 (and add regression tests).

It's a tough one since the `gr` code is a mess, non generic, easily breakable (some legend code paths are not tested in ref imgs), etc...

Add the ability to draw legend bounding boxes (for debugging, when using `Plots.debugplots(true)`).
Fix centering of legend boxes (`:top` and `:bottom`), and position of legend title.

These examples should cover most of the legend boxes possibilities:
```julia
julia> using Plots; Plots.debugplots(true);
julia> kw = (legend_column=-1, legend_title="title");
julia> plot([1:2 reverse(1:2)]; marker=:circle, kw...)
julia> plot(1:2, fillrange=reverse(1:2); kw...)
julia> plot(rand(100); reg = true, fill = (0, :green), kw...)
julia> hspan([1, 2, 3, 4]; kw...)
```

Supersedes https://github.com/JuliaPlots/Plots.jl/pull/4444.

Xref https://github.com/JuliaPlots/Plots.jl/issues/2423.

Added new examples to cover the issues encountered in https://github.com/JuliaPlots/Plots.jl/issues/4485 and https://github.com/JuliaPlots/Plots.jl/issues/4491:
**1.35.8 - inner - ex64**
64_old

**pr - inner - ex64**
ref64

**1.35.8 - outer - ex65**
65_old

**pr - outer - ex65**
ref65

### DIFF
--- a/src/backends/gr.jl
+++ b/src/backends/gr.jl
@@ -1126,59 +1126,71 @@ end
 
 function gr_get_legend_geometry(vp, sp)
     legendn = 0
-    legendw = dy = 0.0
+    legendw = dy = 0.0  # vertical
+    legendh = dx = 0.0  # horizontal
+
+    vertical = (legend_column = sp[:legend_column]) == 1
     if sp[:legend_position] !== :none
         GR.savestate()
         GR.selntran(0)
         GR.setcharup(0, 1)
         GR.setscale(0)
-        if sp[:legend_title] !== nothing
+        if (ttl = sp[:legend_title]) !== nothing
             gr_set_font(legendtitlefont(sp), sp)
-            tbx, tby = gr_inqtext(0, 0, string(sp[:legend_title]))
-            l, r = extrema(tbx)
-            b, t = extrema(tby)
-            dy = t - b
-            legendw = r - l
+            (l, r), (b, t) = extrema.(gr_inqtext(0, 0, string(ttl)))
+            legendh = dy = t - b
+            legendw = dx = r - l
             legendn += 1
         end
         gr_set_font(legendfont(sp), sp)
         for series in series_list(sp)
             should_add_to_legend(series) || continue
-            tbx, tby = gr_inqtext(0, 0, string(series[:label]))
-            l, r = extrema(tbx)
-            b, t = extrema(tby)
+            (l, r), (b, t) = extrema.(gr_inqtext(0, 0, string(series[:label])))
             dy = max(dy, t - b)
+            dx = max(dx, r - l)
             legendw = max(legendw, r - l) # Holds text width right now
+            legendh = max(legendh, t - b)
             legendn += 1
         end
-
         GR.setscale(GR.OPTION_X_LOG)
         GR.selntran(1)
         GR.restorestate()
     end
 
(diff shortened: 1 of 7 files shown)


src/backends/gr.jl
```diff
@@ -1188,49 +1217,44 @@ function gr_update_viewport_legend!(vp, sp, leg)
...
-            vp.xmin += leg.w
+        if leg.vertical
+            if yaxis[:mirror]
+                vp.xmin += leg.w
+            else
+                vp.xmax -= leg.w
+            end
         else
-            vp.xmax -= leg.w
+            # FIXME: update these when vertical
```
BeastyBlacksmith, line 1257: what should happen here? (and you probably meant horizontal)

BeastyBlacksmith (commented):
Am I right, that this only covers `legend_columns = -1`? What will happen if someone passes `legend_columns = 2`?
Maybe its a good idea to catch that and print a warning.

src/backends/gr.jl
t-bltg: Thanks, do you know what `:inline` does ?

t-bltg:
> Am I right, that this only covers legend_columns = -1? 

Yes, other values are not taken into account (`vertical` when `legend_columns = 1`).

src/backends/gr.jl
BeastyBlacksmith: I don't really know, but I would expect it to show the labels at the end of the lines.

BeastyBlacksmith: probably this: https://github.com/JuliaPlots/Plots.jl/pull/2521

BeastyBlacksmith: I'd say you can leave it as it was before

t-bltg: Done, I've checked with `plot(rand(10, 4), legend = :inline, xflip = true)`, this was not concerned by the layout of the legend box.

BeastyBlacksmith: What about `plot(rand(10, 4), legend = :inline, ymirror = true, legend_columns = -1)` ?

t-bltg: inline

Left padding is wrong, I will fix this.

t-bltg:
The following shows the correctness of the bboxes.

```julia
julia> using Plots
julia> Plots.debugplots(true);
julia> plot([1:2, reverse(1:2)]; marker=:circle, leg_title="ttl", legend_column=1)
julia> plot([1:2, reverse(1:2)]; marker=:circle, leg_title="ttl", legend_column=2)
```

t-bltg:
@BeastyBlacksmith, I've update all the images in this PR, and I think we have the correct behavior now.

Do you have any more comments, before this goes in ?

I am going to make it `Plots@1.36.0` after merge, since there are some slight box positioning changes / fixes in vertical mode.

BeastyBlacksmith (approved):
Thank you for working on this :)

Issue #4502, opened 2022-11-08
[BUG] Barplot in logscale has bars starting at different levels


giordano:
## Details

```julia
bar([1 2 3], [10  20 50]; yscale=:log10)
```
produces

![image](https://user-images.githubusercontent.com/765740/200643464-e3768cd7-d998-4a6f-8f63-1b98ecc494f3.png)

@t-bltg pointed out that one can use

```julia
bar([1 2 3], [10  20 50], yscale=:log10, fillto=1)
```
to force the bar to start from a given value, but it'd be nice to have a nicer default behaviour :slightly_smiling_face: 

### Backends

This bug occurs on ( insert `x` below )

Backend      | yes | no  | untested
-------------|-----|-----|---------
gr (default) |  x   |     |
pyplot       |     |     |
plotlyjs     |     |     |
pgfplotsx    |     |     |
unicodeplots |     |     |
inspectdr    |     |     |
gaston       |     |     |

### Versions

Plots.jl version: v1.35.0
Backend  version (`]st -m <backend(s)>`): v0.69.2
Output of `versioninfo()`:

```
julia> versioninfo()
Julia Version 1.8.2
Commit 36034abf260 (2022-09-29 15:21 UTC)
Platform Info:
  OS: Linux (x86_64-linux-gnu)
  CPU: 8 × Intel(R) Core(TM) i7-4870HQ CPU @ 2.50GHz
  WORD_SIZE: 64
  LIBM: libopenlibm
  LLVM: libLLVM-13.0.1 (ORCJIT, haswell)
  Threads: 1 on 8 virtual cores
```

PR #4503, opened 2022-11-08
baseline fix for bar plots using log scales

t-bltg:
Fix https://github.com/JuliaPlots/Plots.jl/issues/4502.

At least values within the same decade are unified to the same value (decade start).
Would that be better @giordano ?

```julia
bar([1 2 3], [11 25 85]; yscale=:log10)
```

**pr**
pr

**master**
master

The problem lies in the way we process series.
Can we preprocess a common minimal value before mapping to [bar series](https://github.com/JuliaPlots/Plots.jl/blob/master/src/recipes.jl#L408-L504) ?

- [x] check performance

### DIFF
--- a/src/args.jl
+++ b/src/args.jl
@@ -531,6 +531,9 @@ const _suppress_warnings = Set{Symbol}([
     :relative_bbox,
     :layout_insets,
     :force_minpad,
+    :x_extrema,
+    :y_extrema,
+    :z_extrema,
 ])
 
 is_subplot_attr(k) = k in _all_subplot_args
--- a/src/pipeline.jl
+++ b/src/pipeline.jl
@@ -152,7 +152,7 @@ end
 
 function RecipesPipeline.process_sliced_series_attributes!(plt::Plots.Plot, kw_list)
     # determine global extrema
-    xe = xe = ze = NaN, NaN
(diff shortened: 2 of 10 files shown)


BeastyBlacksmith:
Can't we just set `fillrange --> 1` in the bar recipe if the yscale is a logscale?

t-bltg:
Could be a cheap solution too.

giordano:
For what is worth, the original data in my example span multiple decades (10 and 100 mainly), reason why I was trying to use logscale

t-bltg:
We could compute global series extrema somewhere here https://github.com/JuliaPlots/Plots.jl/blob/master/RecipesPipeline/src/series_recipe.jl#L10-L29 ...

If we choose an arbitrary default value of `1`, we'll get the same issue for e.g. `bar([1 2 3], [0.01 10 100]; yscale=:log10)`.

t-bltg:
Computing extrema along all series now allow the following:

```julia
bar([1 2 3], [0.02 125 10_000]; yscale=:log10)
```

**pr**
pr

**master**
master

giordano:
Looks great! And I can still tweak it with fillto if I need, right?

t-bltg:
Yes, as long as positive. Can also be an vector of positive values.

PR #4507, opened 2022-11-11
Bump GR version

jheinen:
set compat entry git GR to 0.70


BeastyBlacksmith:
Is there a reason why we should not allow GR 0.69.5?

jheinen:
> Is there a reason why we should not allow GR 0.69.5?

No, as long as the correct `GR_jll` is installed.

t-bltg:
Can anyone confirm this has no issue for  `vscode` on `windows` ?

jheinen:
> Can anyone confirm this has no issue for `vscode` on `windows` ?

On Windows 10 with Julia 1.8.2 it works fine:

Screenshot 2022-11-11 at 12 13 36

jheinen:
@t-bltg : I have no idea how to fix the benchmark test. You probably want to make more tests. From my side, the PR can otherwise be merged.

Project.toml
```diff
@@ -44,7 +44,7 @@ Conda = "1"
 Contour = "0.5 - 0.6"
 FFMPEG = "0.2 - 0.4"
 FixedPointNumbers = "0.6 - 0.8"
-GR = "0.69.5"
+GR = "0.70"
```
t-bltg, line 47:
```suggestion
GR = "0.69.5 - 0.70"
```

As @BeastyBlacksmith suggested.

jheinen: Should be ok.

jheinen:
It seems (see Windows test results), that we have to force the compat entry to be 0.70.

t-bltg:
I've asked on slack for the weird `libdSFMT` error in `Random` on `windows` and `julia 1.8`.

t-bltg:
On `windows` and `julia 1.6` we have a similar `The parameter is incorrect.` but affecting `libgit2`: this suggests bad interaction when loading low level libraries ? :thinking:

t-bltg:
Something is definitively related to `GR` and needs investigation, since a PR opened and tested now passes all `windows` tests (https://github.com/JuliaPlots/Plots.jl/pull/4510).

jheinen:
> Something is definitively related to `GR` and needs investigation, since a PR opened and tested now passes all `windows` tests (#4510).

I don't think so. I only see low-level I/O errors in `libuv`. Why should they happen only in Windows???

t-bltg:
> And, why did it work 2 hours ago?

I don't know, let's re-run the jobs again.

t-bltg:
Fails again, whilst jobs are fine on master (https://github.com/JuliaPlots/Plots.jl/actions/runs/3447742701/jobs/5754071011).
I'm not saying this is induced by `GR` itself, but `ci` failure is somehow induced by this `PR`, period.

jheinen:
The CI system seems to be less reproducible at the moment than the software under test. No idea how to proceed, as all GR tests work fine.

t-bltg:
> The CI system seems to be less reproducible at the moment than the software under test.

This is a totally unjustified affirmation. A computer does what it's told to do. If ci is failing now there is a logical explanation. Did you inspect the logs ?

Working job:, uses `GR@0.70.0`: https://github.com/JuliaPlots/Plots.jl/actions/runs/3442266041/jobs/5742671840#step:9:67
Failing job, uses `GR@0.70.1`: https://github.com/JuliaPlots/Plots.jl/actions/runs/3446919468/jobs/5754149813#step:9:181.

So `GR@0.70.1` broke it.

mkitti:
Here is an explanation of the Windows issue:
https://github.com/jheinen/GR.jl/pull/500#issuecomment-1312586737

PR #4511, opened 2022-11-12
Add compat with GR =0.70.0

mkitti:


t-bltg:
I'd rather have first a testing mechanism such as https://github.com/jheinen/GR.jl/pull/498 (to be discussed), and `GR#master` fixed first.

mkitti:
Yes, but there is also an independent issue here with PkgBenchmark. That is not a GR issue.

Thr fix in GR is probably just to rely on GRDIR as was done before 0.69. Changing the DLL search order appears to mess up the global space even more than setting PATH.

t-bltg:
`BenchmarkCI` erroring has been seen before, and we probably hit https://github.com/tkf/BenchmarkCI.jl/issues/131.

t-bltg:
Superseded by https://github.com/JuliaPlots/Plots.jl/pull/4544.

mkitti:
Noting for posterity that GR.jl v0.70.0 works fine with Plots.jl. GR.jl v0.70.1 does not work with Plots.jl but was yanked. GR.jl v0.71.0 formally reintroduces the state in GR.jl v0.70.0 via https://github.com/jheinen/GR.jl/pull/500 by withdrawing use of `SetDefaultDllDirectories` on Windows. This was problematic since using `SetDefaultDllDirectories` with `LOAD_LIBRARY_SEARCH_DEFAULT_DIRS` caused Julia's use of `LoadLibraryEx` with `LOAD_WITH_ALTERED_SEARCH_PATH` to fail. Rather than using `SetDefaultDllDirectories` with `LOAD_LIBRARY_SEARCH_DEFAULT_DIRS`, we instead use `LOAD_LIBRARY_SEARCH_DEFAULT_DIRS` as a fallback with `LoadLibraryEx` which does not interfere with Julia's loading mechanism.

In GR.jl v0.70.0, the use of `SetDefaultDllDirectories` was bugged. The symbol `kernel32` was not defined:
https://github.com/jheinen/GR.jl/blob/b31a11ba6e24e6e72563864407f0374ff76ce7ce/src/funcptrs.jl#L46

In GR.jl v0.70.1, the use of `SetDefaultDllDirectories` was enabled by replacing the symbol with a string literal:
https://github.com/jheinen/GR.jl/blob/4db03d7af4ad03f8359daac814eadc09197fd7dd/src/funcptrs.jl#L58

In GR.jl v0.71.0, `SetDefaultDllDirectories` was removed. However, the use of `AddDllDirectories` remains:
https://github.com/jheinen/GR.jl/blob/db3e5f53738be892b23317d673179a32b0e50910/src/funcptrs.jl#LL50C53-L50C53

gr framework v0.71.0 (to which GR.jl v0.71 is pinned) implements a fallback loading mechanism with `LoadLibraryEx` and `LOAD_LIBRARY_SEARCH_DEFAULT_DIRS` that can take advantage of the multiple directory information provided via `AddDllDirectories`:
https://github.com/sciapp/gr/blob/129ec060419ef2a911e6c87c32e0406da5526a18/lib/gks/plugin.c#L63

Overall the use of the [Win32 Loader API](https://learn.microsoft.com/en-us/windows/win32/api/libloaderapi/), should provide a robust way to communicate multiple directories to GR plugins on Windows going forward. In particular, this eliminates the dependence on the environment variable `PATH` to communicate such information.

PR #4515, opened 2022-11-13
simplify condition

t-bltg:
Enhance clarity.

### DIFF
--- a/src/backends/gr.jl
+++ b/src/backends/gr.jl
@@ -244,10 +244,11 @@ function gr_polyline3d(x, y, z, func = GR.polyline3d)
     end
 end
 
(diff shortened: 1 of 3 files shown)


src/backends/gr.jl
```diff
@@ -244,19 +244,19 @@ function gr_polyline3d(x, y, z, func = GR.polyline3d)
     end
 end
 
+has_latex_inline_math(s::AbstractString) = occursin(r".*\$[^\$]+?\$.*", s) ≢ nothing
```
bkamins, line 247: maybe just `!==` as  `≢ ` does not display well in many fonts?

t-bltg: You're right, habits :shrug:, fixed.

t-bltg:
Well, the `10^{` is arbitrary, so I changed to `^{` and this breaks `unitful` tests.
This is britle and needs rework.

Issue #4513, opened 2022-11-12
[FR] Use Preferences.jl to configure default loaded backend


mkitti:
Loading GR by default can cause issues since it loads so many other packages, notably JLLs.

For example, loading QT via GR can result in the PyPlot backend not working:
https://discourse.julialang.org/t/trouble-using-plots-on-linux/90061/32?u=mkitti

To simplify this, let's provide a Preferences.jl mechanism so that users can load their preferred backend by default. This will also help to minimize issues and accelerate plots by not loading unnecessary packages that we are not using.

mkitti:
The Discourse post got moved. See https://discourse.julialang.org/t/trouble-using-plots-on-macos/90182/19?u=mkitti

t-bltg:
This will also also speedup for the chosen default backend.
If a user only uses a declared backend, we should [precompile examples](https://github.com/JuliaPlots/Plots.jl/blob/394034718da68af036b863c5b24ece5e7cce83f2/src/precompile.jl) for that backend.

t-bltg:
This is going to be a tough one, because of the constraints of module initialization in `__init__`, precompilation using `SnoopPrecompile` and the way we load backends (currently using `@eval` and `@require`) in `Plots` ...

I tried some things, but going nowhere atm.

PR #4517, opened 2022-11-14
replace PLOTS_DEFAULT_BACKEND by Preferences

BeastyBlacksmith:
I think this is a necessary step for #4513 with this you can do the following
```julia
julia> ENV["PLOTS_DEFAULT_BACKEND"] = "PyPlot"
"PyPlot"

julia> ENV["PLOTS_PRECOMPILE"] = "false"
"false"

julia> using Plots; backend()
Plots.PyPlotBackend()

julia> gr(); plot(1:5)  # still works
```


t-bltg:
I remember having tried that in another `PR` but was failing `ci`, don't remember why.

src/init.jl
```diff
@@ -64,7 +64,7 @@ function __init__()
             )
         end,
     )
-    if _default_backend() != :gr
+    if _default_backend() === :gr
```
BeastyBlacksmith, line 67: That wasn't a typo since, we need the requires only if GR is not the default backend

BeastyBlacksmith:
I went with replacing the environment variable with Preferences, since it also works on CI.
This still does not adress installing Plots without GR, but for that we need optional dependencies.

src/init.jl
t-bltg: Sorry, I got no notifications, because the PR was on draft.

src/Plots.jl
```diff
@@ -35,6 +35,7 @@ using Base.Meta
 using Requires
 using Reexport
 using Unzip
+@reexport using Preferences
```
t-bltg, line 38:
```suggestion
using Preferences
```

I'd prefer not exporting this.

src/backends.jl
```diff
@@ -289,7 +286,7 @@ end
 # ------------------------------------------------------------------------------
 # gr
 
-_initialize_backend(pkg::GRBackend) = nothing  # COV_EXCL_LINE
+_initialize_backend(pkg::GRBackend) = @eval Plots import GR  # COV_EXCL_LINE
```
t-bltg, line 289:
```suggestion
_initialize_backend(pkg::GRBackend) = nothing  # COV_EXCL_LINE
```

Since `GR` is still imported in `backends/gr.jl`, I don't think this is needed.

src/Plots.jl
BeastyBlacksmith: So you'd recommend users to call `Plots.Preferences.set_preference!(Plots, "default_backend" => "plotly")`?
I thought that is a bit long

t-bltg:
> I went with replacing the environment variable with Preferences

Hum, won't this removal be breaking ?
We have to handle this with `Preferences`.
My idea was to write a small `PlotsPreferences` module and on `__init__` of that module, fetch the default backend from `Preferences`, with `PLOTS_DEFAULT_BACKEND` preempting the `Preferences` defined value.

Something similar to what we wrote for GR:
https://github.com/jheinen/GR.jl/blob/eb612a063d6ad092bc93d0161a84c067cde7f654/src/preferences.jl#L51-L53

src/Plots.jl
t-bltg: I would prefer writing our own `PlotsPreferences` module, define our backend selection helpers / functions (with error checks), and export only what's necessary from that module (leaking into `Plots`).

See for example the `GRPreferences` module:
https://github.com/jheinen/GR.jl/blob/master/src/preferences.jl#L97-L118.

BeastyBlacksmith:
> Hum, won't this removal be breaking ?

As far as I can tell, that environment variable is not documented and thus not part of the API and I couldn't make it work on CI. 
I would also wait to document usage of the Preferences until it has some battle testing.

t-bltg:
I think an environment variable pre-empting the `Preferences` based mechanism will still be useful in the context of debugging for un-experienced users.

An env var can always be set, while locating and / or editing persistent files written by `Preferences` on the disk is more cumbersome (especially for julia beginners).

t-bltg:
Superseded by https://github.com/JuliaPlots/Plots.jl/pull/4566.

Issue #4508, opened 2022-11-11
`minorgrid=true` interacts with `yminorticks` inconsistently


ryofurue:
When `minorgrid=true` is in force, `yminorticks=false` stops working. See the code below.

This is inconvenient because you cannot control the minor ticks with `yminorticks` alone.  I wish that the "meaning" of `minorgrid=true` were "Plot minor grid lines at minor ticks" and that `minorgrid` itself didn't affect the placement of minor ticks at all.  The present behavior can be understood if `minorgrid=true` means "Plot minor grid lines. Period."

Julia 1.8.2 (through juliaup) on macOS 12.6

````julia
using Plots

function myplot(ys; yminorticks)
  plot(ys; minorgrid=true,
       ylims=(-0.2,1.2),
       yticks = 0:0.2:1, yminorticks)
end

ys = rand(Float64, 10)
myplot(ys; yminorticks=2) # gets 1 minor tick
myplot(ys; yminorticks=1) # gets 4 minor ticks!
myplot(ys; yminorticks=false) # gets 4 minor ticks!
````

**Comment by t-bltg** [comment]
I believe that `myplot(ys; yminorticks=false) # gets 4 minor ticks!` is to be fixed.

But otherwise, the feature is documented to be the divider of the default `yminorticks`:
```julia
julia> plotattr(:Axis, :minorticks)
`minorticks` is of type Integer.

Intervals to divide the gap between major ticks into.

Axis attribute,  defaults to `false`.
```

`myplot(ys; yminorticks=1)  # 5 ÷ 1 = 5 minor ticks intervals (4 ticks)`
`myplot(ys; yminorticks=2)  # 5 ÷ 2 = 2 minor ticks intervals (1 tick)`

PR #4528, opened 2022-11-19
fix `minorticks`

t-bltg:
Fix https://github.com/JuliaPlots/Plots.jl/issues/4508.

~~`minorticks` is very confusing since the name suggests the number of ticks, not the number of intervals which is `n° ticks + 1`).~~
==> refused as of "breaking".

~~As @ryofurue mentioned, the current behavior is very much inconsistent.~~
==> apparently, being inconsistent is acceptable.

This PR fixes these problems:
```julia
using Plots

myplot(ys; kw...) =
  plot(ys; ylims=(-0.2,1.2), yticks = 0:0.2:1, kw...)

ys = rand(Float64, 10)
myplot(ys; minorgrid=true, yminorticks=:auto)    # default number of minor intervals (5) - ticks (4)
myplot(ys; minorgrid=true, yminorticks=true)     # default number of minor intervals (5) - ticks (4)
myplot(ys; minorgrid=true, yminorticks=:none)    # no minor ticks
myplot(ys; minorgrid=true, yminorticks=nothing)  # no minor ticks
myplot(ys; minorgrid=true, yminorticks=false)    # no minor ticks
myplot(ys; minorgrid=true, yminorticks=0)        # 0 intervals - 0 minor ticks (makes no sense though)
myplot(ys; minorgrid=true, yminorticks=1)        # 1 interval  - 0 minor ticks
myplot(ys; minorgrid=true, yminorticks=2)        # 2 intervals - 1 minor tick
myplot(ys; minorgrid=true, yminorticks=3)        # 3 intervals - 2 minor ticks
myplot(ys; minorgrid=true, yminorticks=4)        # 4 intervals - 3 minor ticks
...

myplot(ys; minorgrid=false, yminorticks=:auto)   # no minor ticks with default `:auto`
myplot(ys; minorgrid=false, [...])  # same as above
```

I've modified the `lens` example to take 1 minor tick (and hence regress on test examples), since `minorticks` wasn't tested before with an integer argument.

~~`PyPlot` has been adjusted since `AutoMinorLocator` uses the number of intervals.~~


t-bltg:
@BeastyBlacksmith, ok ?

BeastyBlacksmith (commented):
As much as I agree, but we can't change semantics of an attribute in a minor release

BeastyBlacksmith:
> * the keyword `minorticks` is a currently a lie, and is in no way related `intervals` as it claims;

Your example in https://github.com/JuliaPlots/Plots.jl/issues/4508#issuecomment-1320838148 showed that it does what it says apart from edge cases which could very well be handled differently (which would indeed be a bugfix)

> * it wasn't tested anywhere in our tests before this `PR` so technically it is not breaking anything (the lens ex was modified but on purpose for this PR);

Thats not how SemVer works. That we don't have tests for something does not mean you can't break it. On the contrary, it increases the probability that you break something accidentally.
Its not about breaking tests, but about breaking someone elses code.

> * the implementation was absolute non sense and totally inconsistent: I think the current is unambiguous (and I'd like someone to prove me wrong or give counter arguments - unless this happens, nothing should prevent this);

That might be and bringing the implementation to match the documentation would be the way to go here.

> * I consider the `2.0` argument as invalid: we are not following a schedule (timeline) here with deadlines or due dates. as such, this is equivalent to saying that this will never happen.

That we don't have a timeline doesn't mean it won't ever happen. I'm open to talk about making a plan, but we also need to collect things that should go in a breaking release.

t-bltg:
This makes absolutely no sense.
I understand the users being lost to these kind of arguments, and going elsewhere (I'm absolutely not thinking about `Makie` there).

I've reverted `minorticks` to be the number of intervals, even if giving `minorticks = 0` makes absolutely no sense mathematically.

I've updated the julia code in https://github.com/JuliaPlots/Plots.jl/pull/4528#issue-1456340615.
So I'm going to merge this pending `CI` since the description change of `minorticks` has been removed.

Issue #3003, opened 2020-09-24
[BUG] Inconsitent linewidth in legend


ValentinKaisermayer:
The line width in the legend is not consistent with different plotting options.

```julia
using Plots
plot(dpi=200)
plot!(rand(10), lw=5, label="a")
plot!(rand(10), lw=5, fill=(0, .3), label="b")
```
![test](https://user-images.githubusercontent.com/50108075/94159782-3d8cd780-fe84-11ea-8adb-fd4668d77d95.png)

### Backends

This bug occurs on ( insert `x` below )

Backend      | yes | no  | untested
-------------|-----|-----|---------
gr (default) |  x   |     |
pyplot       |     |     | x
plotly       |     |     | x
plotlyjs     |     |     | x
pgfplotsx    |     |     | x
inspectdr    |     |     | x

### Versions

Plots.jl version:
Backend  version (`]st -m`):
```
GR v0.52.0
Plots v1.6.6
```
Output of `versioninfo()`:
```
Julia Version 1.5.0
Commit 96786e22cc (2020-08-01 23:44 UTC)       
Platform Info:
  OS: Windows (x86_64-w64-mingw32)
  CPU: Intel(R) Core(TM) i5-6200U CPU @ 2.30GHz
  WORD_SIZE: 64    
  LIBM: libopenlibm
  LLVM: libLLVM-9.0.1 (ORCJIT, skylake)
Environment:
  JULIA_EDITOR = "C:\Program Files\Microsoft VS Code\Code.exe"
  JULIA_NUM_THREADS =
```

PR #4530, opened 2022-11-20
multiply `legend_font_pointsize` by `line_width`

t-bltg:
Fix https://github.com/JuliaPlots/Plots.jl/issues/3003.

The inconsistency should be fixed.

Needs new reference images. Maybe we need some non-linearity or flag here to revert to non-scaling of legend line width. TBD.

PR
ref1

master
https://docs.juliaplots.org/dev/gallery/gr/generated/gr-ref1/#gr_demo_1

### DIFF
--- a/src/args.jl
+++ b/src/args.jl
@@ -2053,12 +2053,14 @@ ensure_gradient!(plotattributes::AKW, csym::Symbol, asym::Symbol) =
             cgrad(alpha = plotattributes[asym])
     end
 
-_replace_linewidth(plotattributes::AKW) =
+const DEFAULT_LINEWIDTH = Ref(1)
+
 # get a good default linewidth... 0 for surface and heatmaps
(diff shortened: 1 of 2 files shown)


BeastyBlacksmith (approved):
I mean.. I don't like that scaling at all, but being consistent is better than not

t-bltg:
Maybe multiply by `min(lw, 5)`, so that it doesn't get too large in the legend ?
:thinking:

t-bltg:
I've arbitrarily clamped the legend linewidth in the rand `[0.1 - 10]` (default is `lw=1`) so that legend boxes with `lw` outside this range still look acceptable.

BeastyBlacksmith:
Looks reasonable 👍🏻

syou83syou83:
Can anybody share an example code how this problem has been solved? I don't quite get it. Thank you very much. @t-bltg @BeastyBlacksmith

Issue #3399, opened 2021-03-29
Change default legend position to :outertopright


niclasmattsson:
This is halfway between a bug report and a feature request, so I'm not tagging it with either.

Please change the default legend position to :outertopright. Here's my one and only argument: a visualization package should not default to hiding part of the data you're trying to visualize.

isentropic:
> a visualization package should not default to hiding part of the data you're trying to visualize.
well the arguably biggest plotting package matplotlib does

niclasmattsson:
Really? I haven't used Python or matplotlib in a long time, but I thought it used a heuristic to place the legend where it doesn't overlap the data. So if Plots.jl had a similar heuristic then an inside legend would be OK, but otherwise it really should place the legend outside.

daschw:
I agree, that it would be nicer. Maybe we could change it with the next major release. For now, you can make it your default by adding `const PLOTS_DEFAULTS = Dict{Symbol, Any}(:legend=>:outertopright)` to your `~/.julia/config/startup.jl`. You can add any plot attribute to this dict.

isentropic:
I still think that the current implementation is good already.

It's actually `legend=:best`
'`Best`' for pyplot is actually a smart legend placement so that it does not overlap with data. 

Backends are supposed to find the best location that does not overlap with data but still manages to be within the plot. This is the best of both worlds, it is compact and does not intersect.

Some backends are unable to find the best location for the legend, hence they just fallback to `topright`

All in all, I think the current default is really good.

isentropic:
What could be changed is the inclusion of the legend by default for single series data. If only a single series is added perhaps the default `y1` should not be placed at all

lmiq:
I have written a simple function set the legend position in a tentative minimal-overlapping position (see https://discourse.julialang.org/t/why-plots-jl-produce-figures-looks-not-as-good-as-other-packages-even-with-pyplot/74946/30?u=leandromartinez98)

What if Plots, with a function like that, defined the 'best' position and send it to the backend, instead of expecting the backend to support that feature? Apparently pyplot is the only one having that option.

t-bltg:
I agree that having a `:best` placement for `GR` would be an improvement, and it could be used by other backends as well.

@lmiq, can you create a `PR` ?

We'd need to `@btime ...` the placement algorithm obviously.

lmiq:
I can try, the function which is there is already fast (on the tenths of micro seconds), and can probably be optimized still.

[...]

PR #4536, opened 2022-11-21
implements best legend place holder

lmiq:
Implemented the best legend position placeholder: It will choose between bottomleft/right or topleft/right, depending on a sample of the data, trying to maximize the difference to the closest point. The function is fast (5-10 microseconds), and should be good enough for a guess of the best position. 

Tries to address: https://github.com/JuliaPlots/Plots.jl/issues/3399

Test it with:
```julia
using Plots
scatter(randn(100), randn(100))
```

### DIFF
--- a/src/backends/gr.jl
+++ b/src/backends/gr.jl
@@ -1103,16 +1103,7 @@ function gr_legend_pos(sp::Subplot, leg, vp)
         return gr_legend_pos(lp, vp)
     end
     if (leg_str = string(lp)) == "best"
-        lp = _guess_best_legend_position(:best,sp) 
-        if lp == :topleft
-            leg_str = "topleft"
-        elseif lp == :bottomleft
-            leg_str = "bottomleft"
-        elseif lp == :bottomright
-            leg_str = "bottomright"
-        else lp == :topright
-            leg_str = "topright"
-        end
+    leg_str = string(_guess_best_legend_position(lp, sp))
     end
     xpos = if occursin("left", leg_str)
         vp.xmin + if occursin("outer", leg_str)
--- a/src/backends/gr.jl
+++ b/src/backends/gr.jl
@@ -1102,9 +1102,9 @@ function gr_legend_pos(sp::Subplot, leg, vp)
     elseif !(lp isa Symbol)
         return gr_legend_pos(lp, vp)
     end
-    if (leg_str = string(lp)) == "best"
-        leg_str = string(_guess_best_legend_position(lp, sp))
-    end
+
+    leg_str = string(_guess_best_legend_position(lp, sp))
+
     xpos = if occursin("left", leg_str)
         vp.xmin + if occursin("outer", leg_str)
             -leg.pad - leg.w - leg.xoffset - !ymirror * gr_axis_width(sp, yaxis)
--- a/src/backends/gr.jl
+++ b/src/backends/gr.jl
@@ -1103,7 +1103,16 @@ function gr_legend_pos(sp::Subplot, leg, vp)
         return gr_legend_pos(lp, vp)
     end
     if (leg_str = string(lp)) == "best"
-        leg_str = _find_best_legend_position(sp) 
+        lp = _guess_best_legend_position(:best,sp) 
+        if lp == :topleft
+            leg_str = "topleft"
+        elseif lp == :bottomleft
+            leg_str = "bottomleft"
+        elseif lp == :bottomright
+            leg_str = "bottomright"
+        else lp == :topright
+            leg_str = "topright"
+        end
     end
     xpos = if occursin("left", leg_str)
         vp.xmin + if occursin("outer", leg_str)
--- a/src/backends/gr.jl
+++ b/src/backends/gr.jl
@@ -1103,7 +1103,7 @@ function gr_legend_pos(sp::Subplot, leg, vp)
         return gr_legend_pos(lp, vp)
     end
     if (leg_str = string(lp)) == "best"
-        leg_str = "topright"  # NOTE: can we do better auto-positioning, as `PyPlot` does ?
+        leg_str = _find_best_legend_position(sp) 
     end
     xpos = if occursin("left", leg_str)
         vp.xmin + if occursin("outer", leg_str)
--- a/src/backends/gr.jl
+++ b/src/backends/gr.jl
@@ -1103,7 +1103,7 @@ function gr_legend_pos(sp::Subplot, leg, vp)
         return gr_legend_pos(lp, vp)
     end
     if (leg_str = string(lp)) == "best"
-    leg_str = string(_guess_best_legend_position(lp, sp))
+        leg_str = string(_guess_best_legend_position(lp, sp))
     end
     xpos = if occursin("left", leg_str)
         vp.xmin + if occursin("outer", leg_str)
--- a/src/utils.jl
+++ b/src/utils.jl
@@ -1151,12 +1151,11 @@ end
 # closest point is chosen as the best position.
 #
 function _find_best_legend_position(plt;nsamples=50)
-    xlims = xlims(plt)
-    ylims = ylims(plt)
-    dmin_max = zero(promote_type(eltype(xlims),eltype(ylims)))
+    xl = xlims(plt)
+    yl = ylims(plt)
+    dmin_max = zero(promote_type(eltype(xl),eltype(yl)))
     ibest = 0
-    for (i, lim) in enumerate(Iterators.product(xlims,ylims))
-
+    for (i, lim) in enumerate(Iterators.product(xl,yl))
         for series in plt.series_list
             x = series[:x]
             y = series[:y]
--- a/src/utils.jl
+++ b/src/utils.jl
@@ -1145,7 +1145,7 @@ function _dmin_point(dmin, x, y, lim, scale)
 end
 function _dmin_series(lim, scale, x, y, nsamples)
     dmin = +Inf
-    step = max(1, div(length(x), nsamples))
+    step = 1
     lim = lim ./ scale
     n = max(1, div(min(nsamples, length(x)), 2))
     # Run from the extremes of the dataset inwards
@@ -1182,7 +1182,7 @@ function _guess_best_legend_position(xl, yl, plt, nsamples)
 end
 
 # Main function
-function _guess_best_legend_position(lp::Symbol, plt; nsamples = 100)
+function _guess_best_legend_position(lp::Symbol, plt; nsamples = 50)
     lp === :best || return lp
     xl = xlims(plt)
     yl = ylims(plt)
--- a/src/utils.jl
+++ b/src/utils.jl
@@ -1177,7 +1177,7 @@ function _guess_best_legend_position(xl, yl, plt, nsamples)
         end
     end
     # this inversion favors :topright in case of draws, without cost
-    ibest = findmax(dist_to_lims[i] for i in 4:-1:1)[2]
+    ibest = findmax(@view(dist_to_lims[4:-1:1]))[2]
     return (:topright, :topleft, :bottomright, :bottomleft)[ibest]
 end
 
--- a/src/utils.jl
+++ b/src/utils.jl
@@ -1168,8 +1168,9 @@ function _guess_best_legend_position(xl, yl, plt, nsamples)
             dist_to_lims[i] = min(dist_to_lims[i], _dmin_series(lim, scale, x, y, nsamples))
         end
     end
-    ibest = findmax(dist_to_lims)[2]
-    return (:bottomleft, :bottomright, :topleft, :topright)[ibest]
+    # this inversion favors :topright in case of draws, without cost
+    ibest = findmax(dist_to_lims[i] for i in 4:-1:1)[2]
+    return (:topright, :topleft, :bottomright, :bottomleft)[ibest]
 end
 
 # Main function
--- a/src/utils.jl
+++ b/src/utils.jl
@@ -1130,11 +1130,15 @@ _document_argument(s::Symbol) =
     _fmt_paragraph(_argument_description(s), leadingspaces = 6 + length(string(s)))
 
 #
+#
+# The following functions implement the guess of the optimal legend position,
+# from the data series.
+#
 # Computes the distance of the plot limit to a sample of points
 # equally spaced in the set and keeps the minimum distance found.
 #
 function _dmin_series(lim, scale, x, y, nsamples)
-    dmin = typemax(eltype(lim))
+    dmin = +Inf
     step = max(1, div(length(x), nsamples))
     lim = lim ./ scale
     for i in 1:min(nsamples, length(x))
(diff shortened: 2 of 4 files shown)


t-bltg:
Nice, I'm going to try it now.

Don't forget to run `using JuliaFormatter; format(["src", "test"])` for the check to pass.

src/utils.jl
```diff
@@ -1128,3 +1128,49 @@ _argument_description(s::Symbol) =
...
+
+#
+# This function guesses which is the best legend position by computing
+# the distance of uniform set of points of the data set to each of the 
+# edges of the plot. The edge that is found to more distant to the 
+# closest point is chosen as the best position.
+#
+function _find_best_legend_position(plt;nsamples=50)
+    xlims = Plots.xlims(plt)
+    ylims = Plots.ylims(plt)
```
t-bltg, line 1155:
```suggestion
    xlims = xlims(plt)
    ylims = ylims(plt)
```

src/utils.jl
```diff
@@ -1128,3 +1128,49 @@ _argument_description(s::Symbol) =
...
+#
+# Computes the distance of the plot limit to a sample of points
+# equally spaced in the set and keeps the minimum distance found.
+#
+function _dmin_series(lim,x,y,nsamples)
+    dmin = typemax(eltype(lim))
+    step = max(1,div(length(x),nsamples))
+    for i in 1:min(nsamples,length(x))
+        isample = firstindex(x) + (i-1)*step
+        d = sum(abs2, lim .- (x[isample],y[isample]))
```
t-bltg, line 1141: I think we can unroll by hand and avoid the allocations here ?

src/utils.jl
```diff
@@ -1128,3 +1128,49 @@ _argument_description(s::Symbol) =
...
+# closest point is chosen as the best position.
+#
+function _find_best_legend_position(plt;nsamples=50)
+    xlims = Plots.xlims(plt)
+    ylims = Plots.ylims(plt)
+    dmin_max = zero(promote_type(eltype(xlims),eltype(ylims)))
+    ibest = 0
+    i = 0
+    for lim in Iterators.product(xlims,ylims)
+        i += 1
```
t-bltg, line 1160:
```suggestion
    for (i, lim) in enumerate(Iterators.product(xlims,ylims))

```

src/utils.jl
lmiq: I don't think the allocations are coming from there, `lim` is a tuple, known at compile time. I've tested that. I think the allocations are coming from the fact that `xlims` are returned as `Tuple{Any,Any}`, which is beyond what I can deal with in this functions. I tried adding a function barrier but that removed 2 allocations only and was not worth the complication (the effect on time was irrelevant).

t-bltg:
Here are the changed reference images.
`ref4`
ref4
`ref7`
ref7
`ref8`
ref8
`ref12`
ref12
`ref13`
ref13
`ref15`
ref15
`ref17`
ref17
`ref18`
ref18
`ref19`
ref19
`ref23`
ref23
`ref25`
ref25
`ref26`
ref26
`ref29`
ref29
`ref30`
ref30
`ref36`
ref36
`ref37`
ref37
`ref44`
ref44
`ref46`
ref46
`ref48`
ref48
`ref51`
ref51
`ref53`
ref53
`ref56`
ref56
`ref58`
ref58
`ref61`
ref61
`ref62`
![ref62](https://user-images.githubusercontent.com/13423344/203096381-1e89230c-5add-43ca-

src/backends/gr.jl
```diff
@@ -1103,7 +1103,16 @@ function gr_legend_pos(sp::Subplot, leg, vp)
...
+        lp = _guess_best_legend_position(:best,sp) 
+        if lp == :topleft
+            leg_str = "topleft"
+        elseif lp == :bottomleft
+            leg_str = "bottomleft"
+        elseif lp == :bottomright
+            leg_str = "bottomright"
+        else lp == :topright
+            leg_str = "topright"
+        end
```
t-bltg, line 1115: What I meant was:

```suggestion
    leg_str = string(_guess_best_legend_position(lp, sp))
```

and remove `    if (leg_str = string(lp)) == "best"`.

src/utils.jl
```diff
@@ -1128,3 +1128,50 @@ _argument_description(s::Symbol) =
...
+#
+# This function guesses which is the best legend position by computing
+# the distance of uniform set of points of the data set to each of the 
+# edges of the plot. The edge that is found to more distant to the 
+# closest point is chosen as the best position.
+#
+function _guess_best_legend_position(lp::Symbol,plt;nsamples=50)
+    if lp != :best
+        return lp
+    end
```
t-bltg, line 1156:
```suggestion
    lp === :best || return lp
```

Nitpick on concision.

lmiq:
> Here are the changed reference images (25 out of 65).

Actually most of them look weird to me, it is not what I was expecting. But now that I know those tests exist, I can check each of them before committing changes. 

(The histogram is case is a case apart, as the plot does not show the data series, and what I do know is based on the data series).

t-bltg:
> But now that I know those tests exist, I can check each of them before committing changes.

The refs files are checked when running `test Plots`.

You can set `julia> ENV["VISUAL_REGRESSION_TESTS_AUTO"] = true` before `pkg> test Plots` (see https://github.com/JuliaPlots/Plots.jl/blob/e6a068934f3f0a4a3b6e0a8dc6ca81744be81717/test/runtests.jl#L53) in order to update them automatically, and partially run the test-suite (faster).

lmiq:
I'll close the PR for now to work on it further without disturbing you.

t-bltg:
You could also have marked it as a draft (it disables notifications ;)).

lmiq:
I think now it is doing mostly a good job. It could be slightly better if I was able to get the `:size` attribute of the plot there, but I'm getting an error trying to access `plt.attr.explicit[:size]` from within `_guess_best_legend_position`. 

Nevertheless, from my tests, the default positions now in the tests look mostly good, and better than before.

t-bltg:
> but I'm getting an error trying to access plt.attr.explicit[:size] from within _guess_best_legend_position

I think `plt[:size]` would work.

t-bltg:
Really nice work. I only think `ref19` would be better with a top-right margin, but this is marginal.

A test is also broken:
```julia
Examples: Error During Test at [...]/Plots.jl/test/test_backends.jl:211
  Got exception outside of a @test
  MethodError: no method matching typemax(::Type{Real})
  Closest candidates are:
    typemax(::Union{DateTime, Type{DateTime}}) at [...]/share/julia/stdlib/v1.8/Dates/src/types.jl:453
    typemax(::Union{Date, Type{Date}}) at [...]/share/julia/stdlib/v1.8/Dates/src/types.jl:455
    typemax(::Union{Time, Type{Time}}) at [...]/share/julia/stdlib/v1.8/Dates/src/types.jl:457
    ...
  Stacktrace:
    [1] (::Plots.var"#82#84"{DataType})(#unused#::Int64)
      @ Plots [...]/Plots.jl/src/utils.jl:1161
    [2] ntuple(f::Plots.var"#82#84"{DataType}, n::Int64)
      @ Base ./ntuple.jl:19
    [3] _guess_best_legend_position(lp::Symbol, plt::Plots.Subplot{Plots.GRBackend}; nsamples::Int64)
      @ Plots [...]/Plots.jl/src/utils.jl:1161
    [4] _guess_best_legend_position
      @ [...]/Plots.jl/src/utils.jl:1155 [inlined]
    [5] gr_legend_pos(sp::Plots.Subplot{Plots.GRBackend}, leg::NamedTuple{(:yoffset, :xoffset, :base_markersize, :base_factor, :has_title, :vertical, :entries, :space, :texth, :textw, :span, :pad, :dy, :dx, :w, :h), Tuple{Float64, Float64, Float64, Float64, Bool, Bool, Int64, Float64, Float64, Float64, Float64, Float64, Float64, Float64, Float64, Float64}}, vp::Plots.GRViewport{Float64})
      @ Plots [...]/Plots.jl/src/backends/gr.jl:1120
    [6] gr_add_legend(sp::Plots.Subplot{Plots.GRBackend}, leg::NamedTuple{(:yoffset, :xoffset, :base_markersize, :base_factor, :has_title, :vertical, :entries, :space, :texth, :textw, :span, :pad, :dy, :dx, :w, :h), Tuple{Float64, Float64, Float64, Float64, Bool, Bool, Int64, Float64, Float64, Float64, Float64, Float64, Float64, Float64, Float64, Float64}}, viewport_area::Plots.GRViewport{Float64})
      @ Plots [...]/Plots.jl/src/backends/gr.jl:991
```

lmiq:
Let me fix all those things and see if I can use the plot size to improve corner case (elongated plots). When I'm ready I remove the Draft flag (sorry not making it Draft before, I couldn't find where to do it until you commented it).

lmiq:
@t-bltg 

Concerning example 19: it is possible to improve the position of the guess by increasing the `nsamples` parameter. But that comes with a cost. 

Maybe we could have two options: `:auto`, with the current behavior, and `:best`, in which the complete data is sampled to place the legend (slower for large data sets).  What you think?

t-bltg:
I think the current is still acceptable for ref19.

Can you post the results of a simple benchmark: `@btime show(devnull, plot(1:2))` with an without this PR, to see what the induced time increase looks like ?
If it is marginal, we can maybe increase the default `nsamples`.

I'd be in favor to `:best` as default, but maybe @BeastyBlacksmith has another opinion.
PyPlot's default is `:best` I think.

src/backends/gr.jl
```diff
@@ -1103,7 +1103,7 @@ function gr_legend_pos(sp::Subplot, leg, vp)
         return gr_legend_pos(lp, vp)
     end
     if (leg_str = string(lp)) == "best"
-        leg_str = "topright"  # NOTE: can we do better auto-positioning, as `PyPlot` does ?
+        leg_str = string(_guess_best_legend_position(lp, sp))
     end
```
t-bltg, line 1107: This can be simplified to 
```julia
leg_str = string(_guess_best_legend_position(lp, sp))
```

src/utils.jl
```diff
@@ -1128,3 +1128,63 @@ _argument_description(s::Symbol) =
...
+    return dmin
+end
+
+#
+# This function guesses which is the best legend position by computing
+# the distance of uniform set of points of the data set to each of the 
+# edges of the plot. The edge that is found to more distant to the 
+# closest point is chosen as the best position.
+#
+# Function barrier because lims are type-unstable
```
t-bltg, line 1167: Would it make sense to fix the type instability of `xlims`, `ylims` ?

t-bltg:
Nice, I really like where this is going.

I don't have anything to add (PR is tested, benchmarked, carefully implemented ...).

When you think that this `PR` is ready, you can run `Plots` test suite locally, update the reference images, and submit a `PR` to https://github.com/JuliaPlots/PlotReferenceImages.jl.

If we are satisfied with the new refs images (changes can be inspected in the github interface), we merge it, and we run the `Plots` test suite here.

When `ci` is green and `PR` approved here -> merge.

test/test_utils.jl
```diff
@@ -189,3 +189,67 @@ end
     # `CSV` produces `SentinelArrays` data
     @test scatter(ChainedVector([[1, 2], [3, 4]]), 1:4) isa Plot
 end
+
+@testset "Best legend position" begin
+    x = 0:0.01:2
+    plt = plot(x, x, label = "linear")
+    plt = plot!(x, x .^ 2, label = "quadratic")
+    plt = plot!(x, x .^ 3, label = "cubic")
+    @test Plots._guess_best_legend_position(:best, plt) == :topleft
```
t-bltg, line 198:
```suggestion
    @test Plots._guess_best_legend_position(:best, plt) === :topleft
```

And :arrow_down: as well.

lmiq:
Corresponding image PR: https://github.com/JuliaPlots/PlotReferenceImages.jl/pull/142

t-bltg (approved):
I'd like @BeastyBlacksmith to comment before merging the reference images.

To me, there is no ambiguity on the automatic placement (at least on the ref images).

lmiq:
I just updated a comment. Nothing will change.

BeastyBlacksmith (approved):
I'd be more happy if it kept `:topright` in the cases where it is impossible not to overlap with data (or maybe even choose `:outertopright` then).
But I like it either way. Good work!

src/utils.jl
```diff
@@ -1128,3 +1128,56 @@ _argument_description(s::Symbol) =
...
+    for isample in lastindex(x):(-step):(lastindex(x) - n)
+        dmin = _dmin_point(dmin, x[isample], y[isample], lim, scale)
+    end
+    return dmin
+end
+#
+# Function barrier because lims are type-unstable
+#
+function _guess_best_legend_position(xl, yl, plt, nsamples)
+    scale = (maximum(xl) - minimum(xl), maximum(yl) - minimum(yl))
```
t-bltg, line 1164:
```suggestion
    scale = (maximum(xl) - minimum(xl), maximum(yl) - minimum(yl))
```

Isn't it faster to use `extrema` here (one single pass) ?

src/utils.jl
```diff
@@ -1128,3 +1128,56 @@ _argument_description(s::Symbol) =
...
+    # this inversion favors :topright in case of draws, without cost
+    ibest = findmax(@view(dist_to_lims[4:-1:1]))[2]
+    return (:topright, :topleft, :bottomright, :bottomleft)[ibest]
+end
+# Main function
+function _guess_best_legend_position(lp::Symbol, plt; nsamples = 50)
+    lp === :best || return lp
+    xl = xlims(plt)
+    yl = ylims(plt)
+    return _guess_best_legend_position(xl, yl, plt, nsamples)
```
t-bltg, line 1182:
```suggestion
    return _guess_best_legend_position(xlims(plt), ylims(plt), plt, nsamples)
```

src/utils.jl
lmiq: Doesn't make any difference, and makes the code uglier, IMO (have do add two lines to first compute the extrema and then use it)

lmiq:
> where it is impossible not to overlap

If we could obtain the coordinates of the legend box we could try that. But I didn't find how to do it, so we are just examining the plot edges relationship with the data. We can try some different heuristics, such as summing some distances, but we would need a case where the current strategy is clearly bad. Also biased examples against any other strategy should be tested.

src/utils.jl
```diff
@@ -1167,3 +1167,54 @@ _argument_description(s::Symbol) =
...
+    dist_to_lims = fill(+Inf, 4) # faster than tuple
+    for series in plt.series_list
+        x = series[:x]
+        y = series[:y]
+        for (i, lim) in enumerate(Iterators.product(xl, yl))
+            dist_to_lims[i] = min(dist_to_lims[i], _dmin_series(lim, scale, x, y, nsamples))
+        end
+    end
+    # this inversion favors :topright in case of draws, without cost
+    ibest = findmax(@view(dist_to_lims[4:-1:1]))[2]
```
t-bltg, line 1213: Couldn't we detect overlap here by adding a distance threshold ?
Such that we use `findfirst` instead and on failure (`nothing`) return `:topright` as fallback.

lmiq: To actually detect overlap we need the coordinates of the legend box. With that information we could compute at roughly the same cost how many atoms fall into it. 

Using any other heuristic without that  may be of course fine in particular cases, but I'm not sure if it is worth the trouble.

t-bltg: I dislike introducing heuristics in a deterministic algorithm too.
I'm happy with the current, but I'm also anticipating users complaining about the new behavior.
Let me think about that while ci is running.

lmiq: I didn't mean something with aleatory behavior, just a different strategy.

I'm pretty sure someone will complain, but that will happen with any strategy :-D

What I can try rather easily is to use a sum of inverse distances, or similar, and see if you like the results better.

t-bltg: Regarding the `coordinates of the legend box`, the legend geometry is computed here https://github.com/JuliaPlots/Plots.jl/blob/816d0f4c5cda98f13300cdf265338995d58723e5/src/backends/gr.jl#L1248-L1265

And you namedtuple `leg` is accessible in the parent calling function https://github.com/JuliaPlots/Plots.jl/blob/816d0f4c5cda98f13300cdf265338995d58723e5/src/backends/gr.jl#L1105, so you can maybe pass this nt to `_guess_best_legend_position` (and make it conditional, defaulting to `nothing` for other backends that might not compute legend geometry).

My little drawing in https://github.com/JuliaPlots/Plots.jl/blob/816d0f4c5cda98f13300cdf265338995d58723e5/src/backends/gr.jl#L1219-L1224 might be useful here.

BeastyBlacksmith: I'm against calling backend functions in frontend code. We'd need a utility function in the frontend then, that backends can overload.

lmiq:
I'll post the results of this different placement strategy, which does better if there is a bunch of points in one corner, but the closest overlap is in the other (should I have created a new branch? I'm still confused with this)

lmiq:
Here they are: https://github.com/JuliaPlots/PlotReferenceImages.jl/pull/143/files

It actually looks good, IMO. Maybe better, particularly in the scatter plots.

t-bltg:
Thanks, merged, let's try that now.

t-bltg:
I've have cleaned up reference images between versions `1.36.3` and `1.36.4`.

t-bltg:
Should we merge ?

t-bltg:
Thanks @lmiq, this is a really nice contribution :tada: .

Issue #2410, opened 2020-02-24
OffsetArray and Plots.heatmap


goretkin:
This doesn't seem to work neither with the GR backend (output below) or the PyPlot backend (stacktrace below).

```julia
import LinearAlgebra: I
import OffsetArrays: OffsetArray
import Plots
Plots.pyplot()
#Plots.gr()
bmp = OffsetArray(collect(I(11)), -5:5, -5:5)
Plots.heatmap(bmp)
```

PyPlot:
```
julia> Plots.heatmap(bmp)
Error showing value of type Plots.Plot{Plots.PyPlotBackend}:
ERROR: PyError ($(Expr(:escape, :(ccall(#= /Users/goretkin/.julia/packages/PyCall/kAhnQ/src/pyfncall.jl:43 =# @pysym(:PyObject_Call), PyPtr, (PyPtr, PyPtr, PyPtr), o, pyargsptr, kw))))) <class 'TypeError'>
TypeError('Dimensions of C (11, 11) are incompatible with X (6) and/or Y (6); see help(pcolormesh)')
  File "/Users/goretkin/.julia/conda/3/lib/python3.7/site-packages/matplotlib/__init__.py", line 1805, in inner
    return func(ax, *args, **kwargs)
  File "/Users/goretkin/.julia/conda/3/lib/python3.7/site-packages/matplotlib/axes/_axes.py", line 5971, in pcolormesh
    X, Y, C = self._pcolorargs('pcolormesh', *args, allmatch=allmatch)
  File "/Users/goretkin/.julia/conda/3/lib/python3.7/site-packages/matplotlib/axes/_axes.py", line 5564, in _pcolorargs
    C.shape, Nx, Ny, funcname))

Stacktrace:
 [1] pyerr_check at /Users/goretkin/.julia/packages/PyCall/kAhnQ/src/exception.jl:60 [inlined]
 [2] pyerr_check at /Users/goretkin/.julia/packages/PyCall/kAhnQ/src/exception.jl:64 [inlined]
 [3] _handle_error(::String) at /Users/goretkin/.julia/packages/PyCall/kAhnQ/src/exception.jl:81
 [4] macro expansion at /Users/goretkin/.julia/packages/PyCall/kAhnQ/src/exception.jl:95 [inlined]
 [5] #110 at /Users/goretkin/.julia/packages/PyCall/kAhnQ/src/pyfncall.jl:43 [inlined]
[...]

Issue #4520, opened 2022-11-15
[FR] Improve performance of plotting heatmaps


BioTurboNick:
Reported on Slack by Jesse Chan, repeated calls to `plot!` to produce overlying heatmaps was very slow.

```
const mat = randn(512, 512)
function f(x)
    plot(); for _ in 1:10000
        heatmap!(x)
    end
end
```

Allocation profiler shows the large majority of allocations occurring in `heatmap_edges`:
![image](https://user-images.githubusercontent.com/1438610/202042099-e8caf599-bec0-49f3-826f-5f6fb16daecb.png)

And the large majority of computation time in `expand_extrema!` (which calls `heapmap_edges`) and `update_clims` (my old nemesis):
![image](https://user-images.githubusercontent.com/1438610/202042301-80dcc98e-05d0-4114-8b10-ee8c5a845617.png)

t-bltg:
@jlchan, can you check if `Plots#master` improves the current or if it is still unacceptably slow ?

jlchan:
Thanks @t-bltg. Using @BioTurboNick's snippet above, on Plots v1.36.1, I get
```julia
julia> @btime f($mat)
  511.680 ms (290431 allocations: 215.73 MiB)
```
On `Plots#master`
```julia
julia> @btime f($mat)
  509.745 ms (81638 allocations: 207.71 MiB)
```

t-bltg:
Thanks, it's disappointing.
I'll give it another try later, memory allocation is just insane on my end.

BioTurboNick:
Thanks for tackling this! I was going to poke around this week/weekend. Can still if you get stumped.

jlchan:
I appreciate it! FYI I've found a workaround by adding another custom recipe. However, improving multiple `plot!` times would certainly help when I teach Julia to students who are coming from Matlab.

t-bltg:
> Can still if you get stumped.

There is still room !
I'm filling up 64Gb RAM here, and starting to swap with that example.
Don't know why gc doesn't jump in though :thinking:.

BioTurboNick:
Haha sorry. My system has 48 GB of RAM and caps out at 21 GB consumed by Julia. But you could easily cut the number down, Or cut the matrix size by half, too. I just chose a reasonably large one to show the effect, and I didn't pay attention to RAM usage haha.

Manually running GC doesn't reduce the number here.

Just holding on to all the plot data looks like 19.5 GB of RAM.

[...]

PR #4549, opened 2022-11-24
Performance improvements to heatmap path

BioTurboNick:
Partially addresses #4520 

Type inferrence is often a big issue here:
```
const mat = randn(512, 512);
function f(x, n)
    plot()
    for _ in 1:n
        heatmap!(x)
    end
end
@btime f(mat, 100)

Master: 1.151 s (26455855 allocations: 614.24 MiB)
Adding one type annotation to `expand_extrema!(a::Axis, Surf::Surface)`: 546.681 ms (241455 allocations: 214.24 MiB)
```

### DIFF
--- a/src/components.jl
+++ b/src/components.jl
@@ -737,7 +737,7 @@ zvalues(
 # -----------------------------------------------------------------------
 
 function expand_extrema!(a::Axis, surf::Surface)
-    ex = a[:extrema]
+    ex = a[:extrema]::Extrema
     for vi in surf.surf
         expand_extrema!(ex, vi)
     end


BioTurboNick:
Any idea why colorbars use a different extrema logic than the spatial axis limits? @t-bltg 

Colorbars ignore nans, axis limits ignore infinites

t-bltg:
> Any idea why colorbars use a different extrema logic than the spatial axis limits ?

Clueless :)

If the current test suite passes with your modifications, I'd say go for it.
Cutting down run time by 50% is a nice improvement.

t-bltg:
It would be nice to fix https://github.com/JuliaPlots/Plots.jl/issues/2410 along the way ;)

BioTurboNick:
I've tried immutable Extrema objects and other organizations; I often end up losing some optimization the compiler is doing in the current state of this PR.

I've poked around at this quite a bit right now, and I think we're up against the minimum possible time if we hold these invariants:
- ignore nans, infinites
- extrema guaranteed to exactly fit all values

If feels ridiculous that the vast majority of computation time, for large numbers of large series, is spent finding extrema.

Idea - Provide an option to control extrema-finding:
- Exact (current)
- Sampling (strided subset?)
- Disabled (not performed, depends entirely on provided limits)

Sampling would maybe become the default, taking in all data up to a threshold and then sampling the rest. Strided would likely be more performant and deterministic than random, though slight danger of a patterned input missing important values.

Thoughts?

t-bltg:
> If feels ridiculous that the vast majority of computation time, for large numbers of large series, is spent finding extrema.

Agreed.
Traversing large series can be expensive, but it is sometimes needed.
I've recently added such a traversal (https://github.com/JuliaPlots/Plots.jl/blob/master/src/pipeline.jl#L156-L159) in this bug fix PR https://github.com/JuliaPlots/Plots.jl/pull/4503.

Providing an option is indeed a good idea, since OPs problem is a bit of a niche one.
IIUC, his data mostly consists of `NaN`s, with small non overlapping patches. Overlaying these patches creates the final heatmap. This is not how we regularly proceed. We usually create the final heatmap before plotting it (single series), by assembling patches.
I would remain as is by default and could be toggled upon performance issues. Considering this problem is a niche one, I'd like the "performance" option to be generic for other series too, and not only dedicated to `heatmap` plotting problems (maybe relevant for `mesh3d` patches, or large `scatter`s).

OP provided another case, which would be worth investigating (are the slow paths the same as for `heatmap` ?):
```julia
x = randn(100, 1_000)
y = randn(100, 1_000)
z = randn(100, 1_000)
scatter(x, y, zcolor=z)  # slowish
scatter(vec(x), vec(y), zcolor=vec(z))  # fast ?
```

You can maybe find inspiration from https://github.com/JuliaPlots/Plots.jl/pull/4536, which uses a deterministic "distance to series" algorithm for computing the legend position.

BioTurboNick:
Yes, it's almost entirely in paths that affect all plots - updating colorbar limits and axis extents.

If I just simply say "take the first 1024 points, then sample ~1024 more in stride" for both I drastically cut the computation time.

```
65.056 ms (28453 allocations: 202.16 MiB)
```

BioTurboNick:
I noticed that during the (I think last stage?) of series recipes, the data is copied out and also infinites and missings are replaced with NaNs at the same time.

So a couple notes:
1. Could we consolidate these passes, checking and storing the extrema at the same time as copying the data?
2. If we're guaranteed to not have infinites after that first pass, we can skip checking for them later, which would also save some time. Checking for nans is still a bit slow though.

t-bltg:
> Could we consolidate these passes, checking and storing the extrema at the same time as copying the data

Of course.

> If we're guaranteed to not have infinites after that first pass, we can skip checking for them later, which would also save some time.

I agree that this is the way to go: store some minimal statistics about a series in a struct or a nt e.g. extrema, has/has not nans, has/has not inf, ...

PR #4559, opened 2022-11-28
[RecipesBase] Use interpolation instead of qualification to drop dependance on the namespace of RecipesBase users

### LINKED ISSUE #4557: [BUG] `using RecipesBase: @recipe` is broken
The output of `@recipe` assumes that `RecipesBase` is loaded. This makes `using RecipesBase: @recipe` not work. Ideally, it should work, failing that, documenting the limitation would be nice.

LilithHafner:
We can't assume that folks will have RecipesBase in their namespace.

Fixes #4557

### DIFF
--- a/RecipesBase/src/RecipesBase.jl
+++ b/RecipesBase/src/RecipesBase.jl
@@ -123,7 +123,7 @@ function create_kw_body(func_signature::Expr)
             push!(
                 cleanup_body.args,
                 :(
-                    $is_key_supported($(QuoteNode(k))) ||
+                    $RecipesBase.is_key_supported($(QuoteNode(k))) ||
                     delete!(plotattributes, $(QuoteNode(k)))
                 ),
             )
--- a/RecipesBase/src/RecipesBase.jl
+++ b/RecipesBase/src/RecipesBase.jl
@@ -181,16 +181,16 @@ function process_recipe_body!(expr::Expr)
                     :(plotattributes[$k] = $v)
                 else
                     # if the user has set this keyword, use theirs
-                    :($is_explicit(plotattributes, $k) || (plotattributes[$k] = $v))
+                    :($RecipesBase.is_explicit(plotattributes, $k) || (plotattributes[$k] = $v))
                 end
 
                 expr.args[i] = if quiet
                     # quietly ignore keywords which are not supported
-                    :($is_key_supported($k) ? $set_expr : nothing)
+                    :($RecipesBase.is_key_supported($k) ? $set_expr : nothing)
                 elseif require
                     # error when not supported by the backend
                     :(
-                        $is_key_supported($k) ? $set_expr :
+                        $RecipesBase.is_key_supported($k) ? $set_expr :
                         error(
                             "In recipe: required keyword ",
                             $k,
@@ -296,10 +296,10 @@ macro recipe(funcexpr::Expr)
             @nospecialize
             $kw_body
             $cleanup_body
-            series_list = $RecipeData[]
+            series_list = $RecipesBase.RecipeData[]
             func_return = $func_body
             func_return === nothing ||
-                push!(series_list, $RecipeData(plotattributes, $wrap_tuple(func_return)))
+                push!(series_list, $RecipeData(plotattributes, $RecipesBase.wrap_tuple(func_return)))
             series_list
         end |> esc,
     )
@@ -333,7 +333,7 @@ macro series(expr::Expr)
     esc(quote
         let plotattributes = copy(plotattributes)
             args = $expr
-            push!(series_list, $RecipeData(plotattributes, $wrap_tuple(args)))
+            push!(series_list, $RecipesBase.RecipeData(plotattributes, $RecipesBase.wrap_tuple(args)))
             nothing
         end
     end)
@@ -370,10 +370,10 @@ function _userplot(expr::Expr)
         quote
             $expr
             export $funcname, $funcname2
-            Core.@__doc__ $funcname(args...; kw...) = $plot($typename(args); kw...)
-            Core.@__doc__ $funcname2(args...; kw...) = $plot!($typename(args); kw...)
-            Core.@__doc__ $funcname2(plt::$AbstractPlot, args...; kw...) =
-                $plot!(plt, $typename(args); kw...)
+            Core.@__doc__ $funcname(args...; kw...) = $RecipesBase.plot($typename(args); kw...)
+            Core.@__doc__ $funcname2(args...; kw...) = $RecipesBase.plot!($typename(args); kw...)
+            Core.@__doc__ $funcname2(plt::$RecipesBase.AbstractPlot, args...; kw...) =
+                $RecipesBase.plot!(plt, $typename(args); kw...)
         end,
     )
 end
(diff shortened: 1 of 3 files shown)


LilithHafner:
CI is already failing after just a few minutes <3

BeastyBlacksmith:
We can't expecpt them to have all those internal functions in the namespace either, I'd say.
The proper fix would be to document that users should at least use
```julia
using ReciperBase: RecipesBase, @recipe
```

t-bltg:
As Simon suggested, improving the docs is probably the easiest.

With a [minor nitpick](https://github.com/JuliaLang/julia/issues/39235):
```julia
import RecipesBase: RecipesBase, @recipe
```

RecipesBase/test/runtests.jl
```diff
@@ -1,47 +1,47 @@
-# Run tests with `import RecipesBase` instead of `using RecipesBase` to test
-# that objects like `AbstractPlot` are properly prefixed with `RecipesBase.` in
-# the macros.
-import RecipesBase
+# Run tests with `import RecipesBase as RB` instead of `using RecipesBase`
+# or `import RecipesBase` to test that macros do not depend on the
+# namespace of the emeclosing scope.
```
BeastyBlacksmith, line 3: Isn't the issue that `import RecipesBase` still imports the `RecipesBase` namespace while `using RecipesBase: @recipe` does not?

LilithHafner:
The original approach (fb77991) should work if https://github.com/JuliaLang/julia/issues/47726 is fixed upstream.

LilithHafner:
Failing that, a slightly more verbose approach should work following this pattern
```julia
julia> module NanoRecipesBase
           function apply_recipe end
           macro recipe(x)
               quote
                   let M = $NanoRecipesBase
                       M.apply_recipe() = $x
                   end
               end
           end
       end
Main.NanoRecipesBase

julia> module User
           import Main.NanoRecipesBase: @recipe
           try
               println(NanoRecipesBase)
           catch UndefVarError
               println("Good")
           end
           @recipe 1729
       end
Good
Main.User

julia> NanoRecipesBase.apply_recipe()
1729
```

RecipesBase/test/runtests.jl
LilithHafner: We could use `import RecipesBase: @recipe` in testing, but that would add `@recipe` to the namespace. This adds only `RB` which is an abbreviation that is not even used in the source code (afaik)

LilithHafner:
> The proper fix would be to document that users should at least use

> As Simon suggested, improving the docs is probably the easiest.

We should only document an unexpected limitation/bug if we can't fix it. This PR should fix it :)

LilithHafner:
The format check gave a lot of diffs. Is there a way to avoid transcribing all those non-functional-changes manually?

t-bltg:
https://docs.juliaplots.org/latest/contributing/#Write-code,-and-format

t-bltg (approved):
Nice.

RecipesBase/src/RecipesBase.jl
```diff
@@ -332,18 +330,13 @@ end
...
-            end
-        end,
-    )
+    esc(quote
+        let plotattributes = copy(plotattributes)
+            args = $expr
+            push!(series_list, $RecipeData(plotattributes, $wrap_tuple(args)))
+            nothing
+        end
+    end)
```
t-bltg, line 339:
```suggestion
    quote
        let plotattributes = copy(plotattributes)
            args = $expr
            push!(series_list, $RecipeData(plotattributes, $wrap_tuple(args)))
            nothing
        end
    end |> esc
```

Minor stylistic enhancement.

t-bltg:
Seems like using `GlobalRef` triggers a segfault apocalypse here :laughing:

LilithHafner:
Strange. I guess I'll just revert to 5440348 "fix formatting".

LilithHafner:
I don't understand the `GlobalRef` segfaults so I asked on Discourse: https://discourse.julialang.org/t/what-is-a-globalref-and-how-i-use-it/90971

LilithHafner:
Thanks [@ffevotte](https://discourse.julialang.org/u/ffevotte) from discourse for the latest and best workaround for julia#47726

t-bltg:
Looks good, thanks for getting to the bottom of this with a cleaner solution. Still intrigued by the `GlobalRef` segfault though :thinking:

RecipesBase/test/runtests.jl
BeastyBlacksmith, line 3:
```suggestion
# namespace of the enclosing scope.
```

BeastyBlacksmith (approved):

RecipesBase/src/RecipesBase.jl
```diff
@@ -123,7 +123,7 @@ function create_kw_body(func_signature::Expr)
             push!(
                 cleanup_body.args,
                 :(
-                    RecipesBase.is_key_supported($(QuoteNode(k))) ||
+                    $is_key_supported($(QuoteNode(k))) ||
```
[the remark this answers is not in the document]
LilithHafner: Because there is no need to interpolate the module at method _calls_. interpolating the function directly is more straightforward and clear than interpolating the module and then looking up the function within the module.

LilithHafner:
I'm happy with [de5a423](https://github.com/JuliaPlots/Plots.jl/pull/4559/commits/de5a423e7a5911fdca2fdae117c4f2db80539b29), but I believe [e9e12d6](https://github.com/JuliaPlots/Plots.jl/pull/4559/commits/e9e12d6a5a69c55a8c4eee486d7b0980ea7a4b38) contains unnecessarily indirect interpolation.

RecipesBase/src/RecipesBase.jl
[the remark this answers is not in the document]
t-bltg: This is inconsistent with https://github.com/LilithHafner/Plots.jl/blob/patch-1/RecipesBase/src/RecipesBase.jl#L90 then.

RecipesBase/src/RecipesBase.jl
```diff
@@ -87,7 +87,7 @@ function get_function_def(func_signature::Expr, args::Vector)
     elseif func_signature.head ≡ :call
         func = Expr(
             :call,
-            :(RecipesBase.apply_recipe),
+            :($RecipesBase.apply_recipe),
```
t-bltg, line 90:
```suggestion
            :($apply_recipe),
```

We tries that, but it failed, right ?

LilithHafner: Yes. This is a definition, so we need indirect interpolation because of JuliaLang/julia#47726.
In all other cases the more direct approach works (note that all tests pass on [de5a423](https://github.com/JuliaPlots/Plots.jl/pull/4559/commits/de5a423e7a5911fdca2fdae117c4f2db80539b29))

t-bltg: In this case, and because this is a low-level julia detail (maybe bug, I'm not competent to say it is), I would prefer to be explicit and consistent, instead of being implicit and kindof obscure.
IMO, it is not immediately obvious to the reader of `RecipesBase` code why you can use `:($is_key_supported)` in some place and not `:($apply_recipe)`.

LilithHafner:
> Do you implicit that those added interpolations have a runtime performance penalty ?

No. There should not be a runtime penalty.

LilithHafner:
Ultimately, it's just a style/formatting choice. I prefer direct interpolation over module interpolation + qualified reference because it is shorter, but either should work, and y'all are more likely to be maintaining this than me so feel free to pick whatever style you prefer.

t-bltg:
We disagree on this one, and I explained my position in https://github.com/JuliaPlots/Plots.jl/pull/4559#discussion_r1035672905.

Thanks for the contribution and the detailed analysis here !

RecipesBase/src/RecipesBase.jl
t-bltg: Side note: [Explicit is better than implicit](https://peps.python.org/pep-0020/#the-zen-of-python), taking the good bits from `Python` ...

Issue #4561, opened 2022-11-28
[BUG] New crash with patch release


juliohm:
Some recent *patch* releases introduced new crashes as usual... I can't build the documentation of packages that were building just fine.

## Details

```julia
ERROR: LoadError: BoundsError: attempt to access 20-element Vector{Float64} at index [100]
Stacktrace:
  [1] getindex
    @ ./array.jl:924 [inlined]
  [2] _dinv_series(lim::Tuple{Float64, Int64}, scale::Tuple{Float64, Float64}, x::StepRangeLen{Float64, Base.TwicePrecision{Float64}, Base.TwicePrecision{Float64}, Int64}, y::Vector{Float64}, nsamples::Int64, weight::Float64)
    @ Plots ~/.julia/packages/Plots/820OU/src/utils.jl:1186
  [3] _dinv_series(lim::Tuple{Float64, Int64}, scale::Tuple{Float64, Float64}, x::StepRangeLen{Float64, Base.TwicePrecision{Float64}, Base.TwicePrecision{Float64}, Int64}, y::Vector{Float64}, nsamples::Int64)
    @ Plots ~/.julia/packages/Plots/820OU/src/utils.jl:1180
  [4] _guess_best_legend_position(xl::Tuple{Float64, Float64}, yl::Tuple{Int64, Float64}, plt::Plots.Subplot{Plots.GRBackend}, nsamples::Int64)
    @ Plots ~/.julia/packages/Plots/820OU/src/utils.jl:1203
  [5] _guess_best_legend_position(lp::Symbol, plt::Plots.Subplot{Plots.GRBackend}, nsamples::Int64)
    @ Plots ~/.julia/packages/Plots/820OU/src/utils.jl:1217
  [6] _guess_best_legend_position
    @ ~/.julia/packages/Plots/820OU/src/utils.jl:1216 [inlined]
  [7] gr_legend_pos(sp::Plots.Subplot{Plots.GRBackend}, leg::NamedTuple{(:yoffset, :xoffset, :base_markersize, :base_factor, :has_title, :vertical, :entries, :space, :texth, :textw, :span, :pad, :dy, :dx, :w, :h), Tuple{Float64, Float64, Float64, Float64, Bool, Bool, Int64, Float64, Float64, Float64, Float64, Float64, Float64, Float64, Float64, Float64}}, vp::Plots.GRViewport{Float64})
[...]

PR #4562, opened 2022-11-28
support a y indexing different from x

lmiq:
This change supports, in the `_guess_best_legend_position` function, different offsets for y and x. Aiming to address https://github.com/JuliaPlots/Plots.jl/issues/4561.

However, it seems that GR does not support those different offsets, thus the bug must be another.

### DIFF
--- a/src/utils.jl
+++ b/src/utils.jl
@@ -1197,7 +1197,7 @@ function _dinv_series(lim, scale, x, y, nsamples, weight = 100.0)
     yoffset = firstindex(y) - firstindex(x)
     for i in firstindex(x):max(1, div(min(nsamples, length(x)), 2))
         # This bound checking is probably redundant
-        (_checkbounds(x, i, j) && _checkbounds(y, i+yoffset, j+yoffset)) || continue
+        (_checkbounds(x, i, j) && _checkbounds(y, i + yoffset, j + yoffset)) || continue
         dinv += (
             inv(1 + weight * d_point(x[i], y[i + yoffset], lim, scale)) +
             inv(1 + weight * d_point(x[j], y[j + yoffset], lim, scale))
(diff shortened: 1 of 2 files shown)


t-bltg:
@lmiq, I'm back so we try to fix this.

What would be conservative atm, and until we know more of the bug is to force [boundschecking](https://docs.julialang.org/en/v1/devdocs/boundscheck/#The-bounds-checking-call-hierarchy) for a (hopefully small) performance penalty, and release a patch version quickly.

We can always remove / rework that later.

lmiq:
It seems that he has a case where `lastindex(x)` points to an invalid index, which might point to a bug in some array interface definition somewhere else.

t-bltg:
Wait, isn't that and effect of https://github.com/JuliaArrays/OffsetArrays.jl/issues/77 ?

We index series using a linear index, but they can be n-dimensional :thinking:

lmiq:
These plots fail, but not at the guessing of the legend position:

```
julia> x = OffsetArray(rand(3,3), -2, -2);

julia> heatmap(x)
ERROR: BoundsError: attempt to access 3-element OffsetArrays.IdOffsetRange{Int64, Base.OneTo{Int64}} with indices -1:1 at index [2]
Stacktrace:
  [1] throw_boundserror(A::OffsetArrays.IdOffsetRange{Int64, Base.OneTo{Int64}}, I::Tuple{Int64})
    @ Base ./abstractarray.jl:703
  [2] checkbounds
    @ ./abstractarray.jl:668 [inlined]
  [3] getindex
    @ ~/.julia/packages/OffsetArrays/WvkHl/src/axes.jl:224 [inlined]
```

t-bltg (approved):

t-bltg:
Also, we didn't check `legnth(y)` and assumed `length(x) == length(y)`, is that always valid ?

t-bltg:
@lmiq, remove draft & merge now ?

t-bltg:
Yeah, `plot(1:100, collect(Float64, 1:20))` errors, but not in `_guess_best_legend_position`.
Weird case.

PR #4563, opened 2022-11-28
support cyclic indexing 

lmiq:
Add support for cyclic indexing on y in the best legend placeholder.

### DIFF
--- a/src/utils.jl
+++ b/src/utils.jl
@@ -1187,24 +1187,20 @@ function d_point(x, y, lim, scale)
     isnan(d) && return 0.0
     d
 end
+# Cycle y indices while taking care of possible offsets
 function _yindex(i, x, y)
     iy = i - firstindex(x) + 1
-    if length(y) < length(x)
-        iy = iy % length(y)
-        if iy == 0
-            iy = length(y)
-        end
-    end
+    iy = mod1(iy, length(y))
     iy + firstindex(y) - 1
 end
 function _dinv_series(lim, scale, x, y, nsamples, weight = 100.0)
-    length(x) > 0 || return +Inf
-    length(y) > 0 || return +Inf
+    (nx = length(x)) > 0 || return +Inf
+    (ny = length(y)) > 0 || return +Inf
     lim = lim ./ scale
     dinv = 0.0
     # Run from the extremes of the dataset inwards
-    j = lastindex(x)
-    for i in firstindex(x):max(1, div(min(nsamples, length(x)), 2))
+    i, j = firstindex(x), lastindex(x)
+    for i in i:max(1, div(min(nsamples, nx), 2))
         dinv += (
             inv(1 + weight * d_point(x[i], y[_yindex(i, x, y)], lim, scale)) +
             inv(1 + weight * d_point(x[j], y[_yindex(j, x, y)], lim, scale))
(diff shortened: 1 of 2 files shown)


lmiq:
I'll probably need to check how this combines with offset indexes first

t-bltg:
I would implement this like so:

```julia
diff --git a/src/utils.jl b/src/utils.jl
index 35a2cee9..9413e780 100644
--- a/src/utils.jl
+++ b/src/utils.jl
@@ -1187,20 +1187,22 @@ function d_point(x, y, lim, scale)
     isnan(d) && return 0.0
     d
 end
-_checkbounds(x, i, j) = checkbounds(Bool, x, i) && checkbounds(Bool, x, j)
+
+# length(series[:x]) can be != length(series[:y]), must cycle
+_cyclic(i, nx, ny, yoffset) = (ny < nx ? mod1(i, ny) : i) + yoffset
+
 function _dinv_series(lim, scale, x, y, nsamples, weight = 100.0)
-    length(x) > 0 || return +Inf
+    (nx = length(x)) > 0 || return +Inf
+    (ny = length(y)) > 0 || return +Inf
     lim = lim ./ scale
     dinv = 0.0
     # Run from the extremes of the dataset inwards
-    j = lastindex(x)
-    yoffset = firstindex(y) - firstindex(x)
-    for i in firstindex(x):max(1, div(min(nsamples, length(x)), 2))
-        # This bound checking is probably redundant
-        (_checkbounds(x, i, j) && _checkbounds(y, i + yoffset, j + yoffset)) || continue
+    i, j = firstindex(x), lastindex(x)
+    yoffset = firstindex(y) - i
+    for i in i:max(1, div(min(nsamples, nx), 2))
         dinv += (
-            inv(1 + weight * d_point(x[i], y[i + yoffset], lim, scale)) +
-            inv(1 + weight * d_point(x[j], y[j + yoffset], lim, scale))
+            inv(1 + weight * d_point(x[i], y[_cyclic(i, nx, ny, yoffset)], lim, scale)) +
+            inv(1 + weight * d_point(x[j], y[_cyclic(j, nx, ny, yoffset)], lim, scale))
         )
         j -= 1
     end
```

lmiq:
> I would implement this like so:

Oh, I just see that now, and implemented something similar, added some tests. We can use those tests with your implementation. I'm out for a moment now.

t-bltg:
@lmiq, maybe you can use one of the `_cycle` methods we define in https://github.com/JuliaPlots/Plots.jl/blob/2f493a5a8edff26dac2d6ef4462153a1db720ff6/src/utils.jl#L186-L199 ?

lmiq:
Tests passed locally, no reference image changed.

src/utils.jl
```diff
@@ -1187,20 +1187,18 @@ function d_point(x, y, lim, scale)
     isnan(d) && return 0.0
     d
 end
-_checkbounds(x, i, j) = checkbounds(Bool, x, i) && checkbounds(Bool, x, j)
 function _dinv_series(lim, scale, x, y, nsamples, weight = 100.0)
-    length(x) > 0 || return +Inf
+    (nx = length(x)) > 0 || return +Inf
+    (ny = length(y)) > 0 || return +Inf
```
t-bltg, line 1192:
```suggestion
    length(y) > 0 || return +Inf
```

t-bltg:
> Tests passed locally, no reference image changed.

Awesome, simple and efficient.

t-bltg:
Could you add a note in the code above `_cycle` to explain why it is needed (maybe point to a github issue).

t-bltg (approved):

t-bltg:
Thanks for fixing !

lmiq:
Thank you for the nice tips, I've learned a few things in the process.

PR #4566, opened 2022-11-28
select `Plots` backend via `Preferences` (or env variable)

t-bltg:
Fix https://github.com/JuliaPlots/Plots.jl/issues/4513.
Supersedes https://github.com/JuliaPlots/Plots.jl/pull/4517 (borrows the added test).

This PR:
- allows to select the default backend using an environment variable e.g. `PLOTS_DEFAULT_BACKEND='unicodeplots'` or make it persistent via [`Preferences.jl`](https://github.com/JuliaPackaging/Preferences.jl), e.g. `Plots.set_default_backend!(:unicodeplots)`, all case-insensitive;
- allows precompilation for the selected backend, thus improving `ttfp` for non `GR` backends;
- avoid potential `GR` runtime problems (low-level `c`, since `GR` is conditionally loaded);
- still allows to swap backends at runtime with `Requires` (making this PR non-breaking);
- add `Plots.diagnostics` function.

We cannot remove the hard dependency on `GR` in `Project.toml` since it would be breaking, but it would be trivial to move `GR` from `[deps]` to `[extras]` in `Plots 2.0`.
One would then have to type `Pkg.add(["Plots", "<mandatory chosen backend>"])` to use `Plots`, which I think is acceptable to ask and has the benefit of reducing `Plots` dependencies imprint (you won't have to download `GR` (and thus artifacts) if you chose to stick to another backend).

Ping @BeastyBlacksmith and @mkitti for discussion / review of the implementation.


mkitti (commented):
In the special case of GR, should we do `import GR` within the top-level of Plots if it is the default backend?

Perhaps we should consider a more radical redesign in the near future?

1. Each of the pieces of backend glue code could be independent subpackages. GRPlotsBackend, PyPlotPlotsBackend, etc.
2. Most of the code currently in Plots.jl gets moved to a package named AbstractPlots.jl. All the backend packages depend on AbstractPlots.jl.
3. Plots.jl becomes a small package which just depends on GRPlotsBackend.jl by default. Via Preferences.jl it can optionally load any of the other backend packages and act as a proxy for that package.

The advantage of this is that each backend now can be precompiled and has its own compiled cache. Also the project dependency tree becomes a bit clearer. One could get all the features of Plots by just doing `using SomePlotsBackend.jl` directly.

For loading with Requires.jl can then just do `@eval using GRPlotsBackend.jl` similar to what @timholy recommended here:
https://discourse.julialang.org/t/categoricalarrays-forces-openspecfun-jll-recompilation/87759/18?u=mkitti

The impact here for this PR is still acknowledging that the GR backend is a special case for the current architecture. That is Plots.jl has declared a dependency on GR.jl. One consequence of that special case is that we load GR.jl within Plots.jl rather than Main. Eventually we move that responsibility to GRPlotsBackend.jl.

t-bltg:
> In the special case of GR, should we do import GR within the top-level of Plots if it is the default backend?

This PR removes that, in order for users not wanting to use the default backend to load `GR`.
I must still investigate why the downstream test are failing, but Plots test suite is green locally.

Thanks for the valuable comments here.

> The advantage of this is that each backend now can be precompiled and has its own compiled cache.

I haven't though of that, and it a clear advantage.

> For loading with Requires.jl can then just do @eval using GRPlotsBackend.jl

Wouldn't the load time be increased, since you parse to evaluate the common Plots code in each backend ?

> Most of the code currently in Plots.jl gets moved to a package named AbstractPlots.jl

This is a good idea for the long term, we'd have to export a lot of symbols from this module.

Overall I think this is the plan for the long term (let's open an issue about this specific re-design ?), but for the moment I'd like to be the least disruptive and try this out. I'm sure we're going to hit bugs and minimizing the changes minimizes the chances of hitting some.

src/backends.jl
```diff
@@ -131,21 +132,41 @@ CurrentBackend(sym::Symbol) = CurrentBackend(sym, _backend_instance(sym))
...
-        end
+function load_default_backend()
+    CURRENT_BACKEND.sym =
+        @load_preference("backend", get(ENV, "PLOTS_DEFAULT_BACKEND", "gr")) |>
+        lowercase |>
+        Symbol
+    backend(CURRENT_BACKEND.sym)
+end
+
+function set_backend!(backend::Union{Nothing,AbstractString,Symbol} = nothing; kw...)
```
BeastyBlacksmith, line 143:
```suggestion
function set_default_backend!(backend::Union{Nothing,AbstractString,Symbol} = nothing; kw...)
```

src/backends.jl
```diff
@@ -131,21 +132,41 @@ CurrentBackend(sym::Symbol) = CurrentBackend(sym, _backend_instance(sym))
...
-            backend(sym)
-        else
-            @warn """You have set PLOTS_DEFAULT_BACKEND=$env_default, but it is not a valid backend package.
-            Choose from: \n\t$(join(sort(_backends), "\n\t"))
-            """
-            _fallback_default_backend()
-        end
+function load_default_backend()
+    CURRENT_BACKEND.sym =
+        @load_preference("backend", get(ENV, "PLOTS_DEFAULT_BACKEND", "gr")) |>
```
BeastyBlacksmith, line 137:
```suggestion
        @load_preference("default_backend", get(ENV, "PLOTS_DEFAULT_BACKEND", "gr")) |>
```

src/backends.jl
```diff
@@ -131,21 +132,41 @@ CurrentBackend(sym::Symbol) = CurrentBackend(sym, _backend_instance(sym))
...
+        lowercase |>
+        Symbol
+    backend(CURRENT_BACKEND.sym)
+end
+
+function set_backend!(backend::Union{Nothing,AbstractString,Symbol} = nothing; kw...)
+    value = backend === nothing ? nothing : lowercase(string(backend))
+    set_preferences!(Plots, "backend" => value; kw...)
```
BeastyBlacksmith, line 145:
```suggestion
    set_preferences!(Plots, "default_backend" => value; kw...)
```

src/backends.jl
```diff
@@ -131,21 +132,41 @@ CurrentBackend(sym::Symbol) = CurrentBackend(sym, _backend_instance(sym))
...
+    rm.(Base.find_all_in_cache_path(Base.module_keys[Plots]))
+    nothing
+end
+
+function diagnostics(io::IO = stdout)
+    from = if has_preference(Plots, "backend")
```
BeastyBlacksmith, line 151:
```suggestion
    from = if has_preference(Plots, "default_backend")
```

src/backends.jl
```diff
@@ -416,6 +442,10 @@ function _initialize_backend(pkg::PlotlyBackend)
         end
         _check_compat(PlotlyBase)
         _check_compat(PlotlyKaleido)
+        @eval begin
+            const PlotlyBase = Main.PlotlyBase
+            const PlotlyKaleido = Main.PlotlyKaleido
+        end
```
BeastyBlacksmith, line 448: what is that for?

test/runtests.jl
```diff
@@ -63,3 +63,5 @@ for name in (
         include("test_$name.jl")
     end
 end
+
+Plots.set_backend!(; force = true)  # clear out `Preferences` file
```
BeastyBlacksmith, line 67: Does that mean that everyone who runs the tests will have his preferences wiped?

mkitti: It depends how they run the tests. If they use `Pkg.test`, then no. That creates a new temporary environment where the LocalPreferences.jl file will be stored.

t-bltg: So currently, a package is imported in `Main` (hack for not adding it as a hard dependency in `Project.toml`). However, for some packages, you need to access the module itself (e.g. https://github.com/JuliaPlots/Plots.jl/blob/4fff117f5587abe30361ec367346752f10c5cbe6/src/backends/pyplot.jl#L12). We import `GR` directly in plots, but other backends need this. And for consistency, I've added this for all backends (here plotly).

src/backends.jl
t-bltg: So currently, a package is imported in `Main` (hack for not adding it as a hard dependency in `Project.toml`). However, for some packages, you need to access the module itself (e.g. https://github.com/JuliaPlots/Plots.jl/blob/4fff117f5587abe30361ec367346752f10c5cbe6/src/backends/pyplot.jl#L12). We import `GR` directly in plots, but other backends need this. And for consistency, I've added this for all backends (here plotly).

mkitti:
Is [a687eba](https://github.com/JuliaPlots/Plots.jl/pull/4566/commits/a687ebad068e1d20cd8ab831ab53d45ec7a8760f) necessary? Julia should see that the LocalPreferences.jl has changed and invalidate the cache automatically.

From the [Preferences.jl README](https://github.com/JuliaPackaging/Preferences.jl):
> Preferences that are accessed during compilation are automatically marked as compile-time preferences, and any change recorded to these preferences will cause the Julia compiler to recompile any cached precompilation .ji files for that module.

src/backends.jl
BeastyBlacksmith: I wonder how well this works in Pluto (which does not evaluate in Main). But if I understood you right, if it worked before, it should still work, right?

mkitti:
> From my own experiments in [#4566 (comment)](https://github.com/JuliaPlots/Plots.jl/pull/4566#issuecomment-1330465229), this doesn't work if I don't clear the package cache (can someone check ?).
> 
> I'd rather have a better solution for triggering precompilation, but there seems to be none: https://discourse.julialang.org/t/invalidate-precompilation-cache/90992.

Something is wrong then.

src/backends.jl
t-bltg: It should. Also, we have to distinguish the precompilation phase (which occurs before `__init__` where `@require` quicks in).

mkitti:
Do we always load the preference during precompilation? Perhaps load preference should be called outside of the ENV dependent conditional.

https://github.com/JuliaLang/julia/blob/edc8241acf912fc009419ccc37ff8e6b3d7807c6/base/loading.jl#L2161-L2163

test/runtests.jl
t-bltg: Should be fixed by https://github.com/JuliaPlots/Plots.jl/pull/4566/commits/c920c49662a17d3a046aaf111019f304f2b6cdd0.

t-bltg:
This is already was is done in https://github.com/t-bltg/Plots.jl/blob/c920c49662a17d3a046aaf111019f304f2b6cdd0/src/init.jl#L173.

But maybe I should enclose this in a `@static if`, let me check.

src/init.jl
```diff
@@ -163,3 +165,41 @@ function __init__()
         @reexport using .UnitfulRecipes
     end
 end
+
+##################################################################
+# COV_EXCL_START
+if bool_env("PLOTS_PRECOMPILE", "true") && bool_env("JULIA_PKG_PRECOMPILE_AUTO", "true")
+    @precompile_setup begin
+        backend()  # get from `Preferences` or env, and initialize backend
```
mkitti, line 173: Do we initialize the backend if we don't do this precompilation?

t-bltg: No, then you'll have to explicitly call e.g. `gr()`.

mkitti, line 173:
```suggestion

# Get from `Preferences` or env, and initialize backend
# This must always be done during precompilation.
# Otherwise, the cache will not invalidate when preferences change.
const default_backend = backend()
##################################################################
# COV_EXCL_START
if bool_env("PLOTS_PRECOMPILE", "true") && bool_env("JULIA_PKG_PRECOMPILE_AUTO", "true")
    @precompile_setup begin
```

mkitti: We at least need to load the preference unconditionally for cache invalidation to work properly.

t-bltg: That seems to work, thanks. I guess it's because of the assignment to a `const` variable ?

src/backends.jl
```diff
@@ -131,21 +132,44 @@ CurrentBackend(sym::Symbol) = CurrentBackend(sym, _backend_instance(sym))
...
-            backend(sym)
-        else
-            @warn """You have set PLOTS_DEFAULT_BACKEND=$env_default, but it is not a valid backend package.
-            Choose from: \n\t$(join(sort(_backends), "\n\t"))
-            """
-            _fallback_default_backend()
-        end
+function load_default_backend()
+    CURRENT_BACKEND.sym =
+        load_preferences(Plots, "default_backend", get(ENV, "PLOTS_DEFAULT_BACKEND", "gr")) |>
```
mkitti, line 137: This logic seems backwards. The environment variable should take precedence over Preferences.jl.

src/init.jl
t-bltg: Wait no, I take that back.

```julia
$ JULIA_PKG_PRECOMPILE_AUTO=0 julia -e 'import Plots; Plots.set_default_backend!(; force=true)'
$ julia -e 'using Plots'  # precompiles
$ JULIA_PKG_PRECOMPILE_AUTO=0 julia -e 'import Plots; Plots.set_default_backend!(:gaston; force=true)'
$ julia -e 'using Plots'  # no precompilation

File: src/backends.jl (hunk 8 above)
Comment (t-bltg) [reply src/backends.jl:137]: Right.

Comment (mkitti) [inline src/backends.jl:137]:
```suggestion
const default_backend_string = load_preferences(Plots, "default_backend", "gr")
function load_default_backend()
    CURRENT_BACKEND.sym =
        get(ENV, "PLOTS_DEFAULT_BACKEND", default_backend_string) |>
```

t-bltg: Mmh, we probably don't need this to be dynamic.

t-bltg: This is the correct fix.

t-bltg:
The only remaining problem here, is that setting the backend via an environment variable does not re-trigger precompilation.

mkitti (commented):
It is currently possible to end up in a situation where we cannot load Plots.jl if the backend dependency is not present.

We should check if the backend dependencies are present before switching the preference.

We can check `Pkg.dependencies()` to see if the needed package is present and a direct dependency.

src/backends.jl
```diff
@@ -16,6 +16,7 @@ backend_name() = CURRENT_BACKEND.sym
 _backend_instance(sym::Symbol)::AbstractBackend =
     haskey(_backendType, sym) ? _backendType[sym]() : error("Unsupported backend $sym")
 
+backend_package_name() = _backend_packages[backend_name()]
 backend_package_name(sym::Symbol) = _backend_packages[sym]
```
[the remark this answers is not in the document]
t-bltg: Now that ci is green, time to polish.

src/backends.jl
```diff
@@ -130,22 +131,48 @@ end
...
 
 # ---------------------------------------------------------
+# this must always be done during precompilation, otherwise
+# the cache will not invalidate when preferences change
+const DEFAULT_BACKEND_STRING = load_preference(Plots, "default_backend", "gr")
+
+function load_default_backend()
+    # environment variable preempts the `Preferences` based mechanism
+    CURRENT_BACKEND.sym =
+        something(get(ENV, "PLOTS_DEFAULT_BACKEND", nothing), DEFAULT_BACKEND_STRING) |>
```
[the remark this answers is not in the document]
t-bltg: No, this is a simplification indeed.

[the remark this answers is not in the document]
t-bltg: Fixed.

src/backends.jl
```diff
@@ -130,22 +131,48 @@ end
...
-            Choose from: \n\t$(join(sort(_backends), "\n\t"))
-            """
-            _fallback_default_backend()
-        end
+function set_default_backend!(
+    backend::Union{Nothing,AbstractString,Symbol} = nothing;
+    kw...,
+)
+    value = backend === nothing ? nothing : lowercase(string(backend))
+    set_preferences!(Plots, "default_backend" => value; kw...)
```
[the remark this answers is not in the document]
t-bltg: There is a warning thrown [here](https://github.com/JuliaPlots/Plots.jl/blob/master/src/backends.jl#L141-L144), but tha can probably we made safer for `Preferences`.

t-bltg:
> We can check Pkg.dependencies() to see if the needed package is present and a direct dependency.

`Pkg.dependencies` is marked [experimental](https://pkgdocs.julialang.org/v1/api/#Pkg.dependencies), so I don't know if that should be used.

src/backends.jl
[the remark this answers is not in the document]
t-bltg: We already have such a runtime check in https://github.com/JuliaPlots/Plots.jl/blob/master/src/Plots.jl#L16-L30, let's reuse it.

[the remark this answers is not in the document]
t-bltg: Fixed.

mkitti:
> > We can check Pkg.dependencies() to see if the needed package is present and a direct dependency.
> 
> `Pkg.dependencies` is marked [experimental](https://pkgdocs.julialang.org/v1/api/#Pkg.dependencies), so I don't know if that should be used.

Using `Pkg.Types.read_project` sounds worse. It's not documented.

What if we just try to import the package before we set the preference? Put a try catch around it and give a friendly error if we cannot import the package.

Another alternative is we change the API so you can only set the default to the currently used backend.

t-bltg:
> Using Pkg.Types.read_project sounds worse. It's not documented.

Agreed, but in the context of adding something through a PR, we should either consider it or fix `Pkg.Types.read_project`.

> What if we just try to import the package before we set the preference? Put a try catch around it and give a friendly error if we cannot import the package.

That could work, although I'm always reluctant to use `try ... catch`, to me this is either an indicating of a poor algorithm and / or laziness to write something clear. Let me think more about this.

> Another alternative is we change the API so you can only set the default to the currently used backend.

I don't see a clear reason for restricting this.

mkitti:
> I don't see a clear reason for restricting this.

It forces the user to demonstrate that they can load the backend successfully before changing it to be the default.

One example is trying to use Gaston.jl without having gnuplot installed. When I tried this, I could not get Plots.jl beyond precompilation.

t-bltg:
> One example is trying to use Gaston.jl without having gnuplot installed. When I tried this, I could not get Plots.jl beyond precompilation.

We have no way of controlling this, and this is beyond `Plots`.
`Gaston` can we loaded, but if `gnuplot` is not installed on the system, it won't work.

Project.toml
```diff
@@ -46,6 +47,8 @@ FFMPEG = "0.2 - 0.4"
 FixedPointNumbers = "0.6 - 0.8"
 GR = "0.69.5 - 0.71"
 Gaston = "1"
+HDF5 = "0.16"
+InspectDR = "0.4"
```
t-bltg, line 51: Those weren't set before, and the enforced checks requested by @mkitti have the benefit of spotting this mistake.

mkitti: Cool

mkitti:
Overall this looks good to me. I have had not had time to test the final form personally yet

mkitti (approved):

t-bltg:
Thanks for reviewing.
I'll probably let it rest another day for testing and merge it tomorrow.

mkitti:
Just a thought.  Those `@require` blocks look quite standardized. Would it make sense to make a new macro? Also could you b pull the UUIDs from Project.toml?

t-bltg:
Yeah, I've already reworked some stuff here, thanks.
But there is always this annoying `PlotyBase` / `Plotly` inconsistencies :cry:.
I'm trying to test precompilation for all backends at the moment, but there are still a few bug arising.

test/test_backends.jl
```diff
@@ -138,6 +140,63 @@ with(:pgfplotsx) do
...
+            end
+            exit(res.n_passed == 2 ? 0 : 1)
+            """,
+        )
+        @test success(run(```$(Base.julia_cmd()) $script```))
+    end
+
+    is_pkgeval() || for be in ALL_BACKENDS
+        @test_logs Plots.set_default_backend!(be)  # test the absence of warnings
+        @test Base.compilecache(Base.module_keys[Plots]) isa String  # test default precompilation
```
t-bltg, line 194: This tests `SnoopPrecompile` precompilation for `Preferences` set backends, making the approach more robust.

mkitti:
Here's one way to get access to the UUIDs in Project.toml so you don't have to repeat them in the code.

```julia
julia> begin
           using TOML
           PlotsTOML = TOML.parsefile(joinpath(dirname(dirname(pathof(Plots))), "Project.toml"))
           depdict = merge(PlotsTOML["deps"], PlotsTOML["extras"])
       end
Dict{String, Any} with 65 entries:
  "Pkg"                   => "44cfe95a-1eb2-52ea-b672-e2afdf69b78f"
  "OffsetArrays"          => "6fe1bfb0-de20-5000-8ca7-80f57d26f881"
  "PlotlyBase"            => "a03496cd-edff-5a9b-9e67-9cda94a718b5"
  "RelocatableFolders"    => "05181044-ff0b-4ac5-8273-598c1e38db00"
  "Base64"                => "2a0f44e3-6c83-55bd-87e4-b1978d98bd5f"
  "ImageMagick"           => "6218d12a-5da1-5696-b52f-db25d2ecc6d1"
  "Latexify"              => "23fbe1c1-3f47-55db-b15f-69d7ec21a316"
  "Measures"              => "442fdcdd-2543-5da2-b0f3-8c86c306513e"
  "RecipesBase"           => "3cdcf5f2-1ef4-517c-9805-6587b60abb01"
  "Requires"              => "ae029012-a4dd-5104-9daa-d747884805df"
  "PlotlyJS"              => "f0f68f2c-4968-5e81-91da-67840de0976a"
  "Showoff"               => "992d4aef-0814-514b-bc4d-f2e9a6c4116f"
  "GR"                    => "28b8d3ca-fb5f-59d9-8090-bfdbd6d07a71"
  "Distributions"         => "31c24e10-a181-5473-b8eb-7969acd0382f"
  "UUIDs"                 => "cf7118a7-6976-5b1a-9a39-7adc72f591a4"
  "FixedPointNumbers"     => "53c48c17-4a7d-5ca2-90c5-79b7896eea93"
  "JLFzf"                 => "1019f520-868f-41f5-a6de-eb00f4b6a39c"
  "LinearAlgebra"         => "37e2e46d-f89d-539d-b4ee-838fcccc9c8e"
  "VisualRegressionTests" => "34922c18-7c2a-561c-bac1-01e79b2c4c92"
  "HDF5"                  => "f67ccb44-e63f-5c2f-98bd-6dc0ccc4ba2f"
  "LibGit2"               => "76f85450-5226-5b5a-8eaa-529ad045b433"
  "FFMPEG"                => "c87230d0-a227-11e9-1b43-d7ebe4e7570a"
  "GeometryBasics"        => "5c1252a2-5f33-56bf-86c9-59e7332b4326"
  "Dates"                 => "ade2ca70-3891-5945-98fb-dc099432e06a"
  "UnicodeFun"            => "1cfade01-22cf-5700-b092-accc4b62d6e1"
  "FileIO"                => "5789e2e9-d7fb-5bc7-8068-2c6fae9b9549"
  "PlotThemes"            => "ccf2f8ad-2431-5c83-bf29-c5338b663b6a"
  "Contour"               => "d38c429a-6771-53c6-b99e-75d170b6e991"
  "Printf"                => "de0858da-6303-5e67-8744-51eddeeeb8d7"
  "Test"                  => "8dfed614-e22c-5e08-85e1-65c5234f0b40"
  "Random"                => "9a3f8284-a2c9-5f02-9a11-845980a1fd5c"
  "Gaston"                => "4b11ee91-296f-5714-9832-002c20994614"
  "StatsPlots"            => "f3b207a7-027a-5e70-b257-86293d7955fd"
  "Conda"                 => "8f4d0f93-b110-5947-807f-2305c1781a2d"
  "Unzip"                 => "41fe7b60-77ed-43a1-b4f0-825fd5a5650d"
  "REPL"                  => "3fa0cd96-eef1-5676-8a61-b3b8758bbffb"
  "RecipesPipeline"       => "01d81517-befc-4cb6-b9ec-a95719d0359c"
  "PyCall"                => "438e738f-606a-5dbb-bf0a-cddfbfd45ab0"
  "PlotUtils"             => "995b91a9-d308-5afd-9ec6-746e21dbc043"
  "JSON"                  => "682c06a0-de6a-54ab-a142-c8b1cf79cde6"
  "StatsBase"             => "2913bbd2-ae8a-5f71-8c99-4fb6c76f3a91"
  "Images"                => "916415d5-f1e6-5110-898d-aaa5f9f070e0"
  "PyPlot"                => "d330b81b-6aea-500a-939a-2ce795aea3ee"
  "Aqua"                  => "4c88cf16-eb10-579e-8560-4a9242c79595"
  "TestImages"            => "5e47fb64-e119-507b-a336-dd2b206d9990"
  "Unitful"               => "1986cc42-f94f-5a68-af5c-568840ba703d"
  "StaticArrays"          => "90137ffa-7385-5640-81b9-e52037218182"
  "UnicodePlots"          => "b8865327-cd53-5732-bb35-84acbb429228"
  "Reexport"              => "189a3867-3050-52da-a836-e630ba90ab69"
  "InspectDR"             => "d0351b0e-4b05-5898-87b3-e2a8edfddd1d"
  "NaNMath"               => "77ba4419-2d1f-58cd-9bb1-8ffee604a2e3"
  "Downloads"             => "f43a241f-c20a-4ad4-852c-f6b1247861c6"
  "Statistics"            => "10745b16-79ce-11e8-11f9-7d13ad32a3b2"
  "SentinelArrays"        => "91c51154-3ec4-41a3-a24f-3f23e20d615c"
  "PlotlyKaleido"         => "f2990250-8cf9-495f-b13a-cce12b45703c"
  "PGFPlotsX"             => "8314cec4-20b6-5062-9cdb-752b83310925"
  "StableRNGs"            => "860ef19b-820b-49d6-a774-d7a799459cd3"
  "LaTeXStrings"          => "b964fa9f-0449-5b57-a5c2-d3ea65f4040f"
  "SnoopPrecompile"       => "66db9d55-30c0-4569-8b51-7e840670fc0c"
  ⋮                       => ⋮

julia> depdict["GR"]
"28b8d3ca-fb5f-59d9-8090-bfdbd6d07a71"

julia> depdict["HDF5"]
"f67ccb44-e63f-5c2f-98bd-6dc0ccc4ba2f"

julia> depdict["PlotlyBase"]
"a03496cd-edff-5a9b-9e67-9cda94a718b5"
```

mkitti:
You could integrate it by doing something like this:

```julia
using TOML
const PlotsTOML = TOML.parsefile(joinpath(dirname(dirname(pathof(Plots))), "Project.toml"))
const depdict = merge(PlotsTOML["deps"], PlotsTOML["extras"])

macro load(backend, pkg)
    uuid = depdict[pkg]
    quote
        backend_name() === $backend || @require $pkg = $uuid begin
            include(_path($backend))
        end
    end |> esc
end
```

t-bltg:
Some timings:

# UnicodePlots

Non-default:
```julia
julia> @time using Plots
  7.260609 seconds (9.33 M allocations: 639.850 MiB, 9.49% gc time, 17.18% compilation time: 51% of which was recompilation)
julia> @time unicodeplots()
  6.657163 seconds (8.79 M allocations: 578.477 MiB, 3.78% gc time, 27.97% compilation time)
julia> @time plot(1:2) |> display
 14.595231 seconds (25.33 M allocations: 1.344 GiB, 4.43% gc time, 99.84% compilation time: 16% of which was recompilation)
```

Persistent default `Plots.set_default_backend!(:unicodeplots)`:
```julia
julia> @time using Plots
 15.047585 seconds (21.17 M allocations: 1.417 GiB, 7.59% gc time, 15.70% compilation time: 25% of which was recompilation)
julia> backend()
Plots.UnicodePlotsBackend()
julia> @time plot(1:2) |> display
  7.226316 seconds (1.76 M allocations: 91.479 MiB, 0.67% gc time, 99.69% compilation time: 16% of which was recompilation)
```

For `unicodeplots`, TTFP is reduced from `~28s` downto `~22s`.

# PythonPlot

Non-default:
```julia
julia> @time using Plots
  7.260609 seconds (9.33 M allocations: 639.850 MiB, 9.49% gc time, 17.18% compilation time: 51% of which was recompilation)
julia> @time pythonplot()
 10.171920 seconds (9.58 M allocations: 515.003 MiB, 2.48% gc time, 75.43% compilation time: 37% of which was recompilation)
julia> @time plot(1:2) |> display
 14.901739 seconds (17.68 M allocations: 932.919 MiB, 4.05% gc time, 94.80% compilation time: 9% of which was recompilation)
```

Persistent default `Plots.set_default_backend!(:pythonplot)`:
```julia
julia> @time using Plots
 16.153475 seconds (18.67 M allocations: 1.111 GiB, 5.79% gc time, 52.54% compilation time: 43% of which was recompilation)
julia> backend()
Plots.PythonPlotBackend()
julia> @time plot(1:2) |> display
 10.825664 seconds (4.06 M allocations: 215.154 MiB, 1.29% gc time, 93.19% compilation time: 27% of which was recompilation)
```

For `pythonplot`, TTFP is reduced from `~32s` downto `~27s`.

# Gaston

Non-default:
```julia
julia> @time using Plots
  7.260609 seconds (9.33 M allocations: 639.850 MiB, 9.49% gc time, 17.18% compilation time: 51% of which was recompilation)
julia> @time gaston()
  1.190615 seconds (1.11 M allocations: 59.159 MiB, 3.78% gc time, 75.65% compilation time)
julia> @time plot(1:2) |> display
  8.914270 seconds (10.12 M allocations: 534.611 MiB, 2.63% gc time, 95.48% compilation time: 11% of which was recompilation)
```

Persistent default `Plots.set_default_backend!(:gaston)`:
```julia
julia> @time using Plots
  6.950285 seconds (8.95 M allocations: 612.747 MiB, 9.02% gc time, 17.96% compilation time: 48% of which was recompilation)
julia> backend()
Plots.GastonBackend()
julia> @time plot(1:2) |> display
  6.308577 seconds (1.21 M allocations: 61.623 MiB, 0.54% gc time, 95.13% compilation time: 16% of which was recompilation)
```

For `gaston`, TTFP is reduced from `~17s` downto `~13s`.

t-bltg:
I think I'll just merge this as is, as the numbers do show some improvements, and non-regression for `GR`, and I don't want to fight `Plots` design choices no more.

I think it's time to run `JET` over `plot(1:2)` and work on the thousands of runtime dispatches we hit.

t-bltg:
Thanks for the comments and the help here.

mkitti:
My pleasure. So are we dripping pyplot support?

t-bltg:
No, still supported for plots 1.x, but no more development and bug fixes. Seems acceptable no ? Maintaining both is a lot of work ...

mkitti:
Apparently, this line is not effective for some reason? https://github.com/JuliaPlots/Plots.jl/blob/62c14074cc90063a4375266c00c14b1219ccb3f8/src/backends.jl#L1370

BeastyBlacksmith:
I can't reproduce that. Which version is that?

KristofferC:
The following is the repro:

```
using Plots
using Plots.PlotMeasures
using LaTeXStrings
using PGFPlotsX
using DelimitedFiles

pgfplotsx(size=(900,900), legend=true,grid = false,framestyle =:box,xlim = (1.1,5.3),ylim = (-0.1,3.0),xminorticks=2,yminorticks=2)
```

Plots 1.38.3

BeastyBlacksmith:
Hmm.. so the issue seems to be loading `PGFPlotsX` before calling `pgfplotsx` (which is not neccessary, but should work anyways)

mkitti:
Removing `using PGFPlotsX` succeeds for me.

```julia
using Plots
using Plots.PlotMeasures
using LaTeXStrings
using DelimitedFiles

pgfplotsx(size=(900,900), legend=true,grid = false,framestyle =:box,xlim = (1.1,5.3),ylim = (-0.1,3.0),xminorticks=2,yminorticks=2)
```

PR #4582, opened 2022-12-06
Fix colorbar positioning when multiple subplots are present

### LINKED ISSUE #2691: Colorbar issue for  side by side contours plot with Plotly.
Hi!

When using Plotly as backend and try to plot two contour plots side by side, each with its own colorbar, the two colorbars seem to appear on top of each other. 

The same behaviour/bug can even be seen on the [Plotly example page](http://docs.juliaplots.org/latest/generated/plotly/#plotly-ref22-1).

Can I do something to fix this or will it be fixed in upcoming updates of the library?

stevenaleung:
Closes #2691

Contour plot example:
```julia
data1 = rand(3,3);
data2 = rand(3,3) .+ 10;
p1 = contour(data1, colorbar=:right)
p2 = contour(data2, colorbar=:right)
plot(p1, p2)
```

Left: before
Right: after
  

Heatmap example:
```julia
data1 = rand(3,3);
data2 = rand(3,3) .+ 10;
p1 = heatmap(data1, colorbar=:right)
p2 = heatmap(data2, colorbar=:right)
plot(p1, p2)
```

Left: before
Right: after

### DIFF
--- a/src/backends/plotly.jl
+++ b/src/backends/plotly.jl
@@ -81,6 +81,12 @@ function plotly_domain(sp::Subplot)
     pcts = plotly_apply_aspect_ratio(sp, sp.plotarea, pcts)
     x_domain = [pcts[1], pcts[1] + pcts[3]]
     y_domain = [pcts[2], pcts[2] + pcts[4]]
+    if hascolorbar(sp)
+        subplot_width = pcts[3]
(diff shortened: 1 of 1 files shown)


BeastyBlacksmith:
wonderful, thank you!

Issue #4583, opened 2022-12-07
[BUG] Plots don't render using Plotly in v1.37 ("_display is not defined for this backend")


niclasmattsson:
Using Plotly in version 1.37 returns "_display is not defined for this backend" and no plot is rendered. The same error is reported in old issues like #2407 and #2698, but in those issues the problem was calling `plotly()` inside a function. That is not the case here. MWE below, which works just fine in Plots 1.36 but not in 1.37.

```
julia> using Plots

julia> plotly()
┌ Warning: For saving to png with the `Plotly` backend `PlotlyBase` and `PlotlyKaleido` need to be installed.
│   err =
│    ArgumentError: Package PlotlyBase not found in current path.
│    - Run `import Pkg; Pkg.add("PlotlyBase")` to install the PlotlyBase package.
└ @ Plots C:\Users\niclas\.julia\packages\Plots\m8l21\src\backends.jl:508
Plots.PlotlyBackend()

julia> plot(rand(10))  # No plot rendered.
┌ Warning: _display is not defined for this backend.
└ @ Plots C:\Users\niclas\.julia\packages\Plots\m8l21\src\output.jl:216
```
### HOWEVER, there's a weird workaround...

Installing and using PlotlyJS but still running the `plotly()` backend works in 1.37.

```
julia> using Plots, PlotlyJS

julia> plotly()
┌ Warning: For saving to png with the `Plotly` backend `PlotlyBase` and `PlotlyKaleido` need to be installed.
│   err =
│    ArgumentError: Package PlotlyBase not found in current path, maybe you meant `import/using .PlotlyBase`.
│    - Otherwise, run `import Pkg; Pkg.add("PlotlyBase")` to install the PlotlyBase package.
└ @ Plots C:\Users\niclas\.julia\packages\Plots\m8l21\src\backends.jl:508
Plots.PlotlyBackend()

julia> Plots.plot(rand(10))   # Works, renders in default web browser.
```

### Backends

This bug occurs on ( insert `x` below )

**(FYI there's no plotly line in the default issue template.)**

Backend      | yes | no  | untested
-------------|-----|-----|---------
gr (default) |     |   x  | 
plotly     |   x  |     |      
pyplot       |     |     | x
plotlyjs     |     |   x  | 
pgfplotsx    |     |     | x
unicodeplots |     |     | x
inspectdr    |     |     | x
gaston       |     |     | x

### Versions

Plots.jl version: 1.37
[...]

PR #4584, opened 2022-12-07
fix plotly incongruity

t-bltg:
Maybe fix https://github.com/JuliaPlots/Plots.jl/issues/4583.

I don't like doing this (exceptions to the already complicated load / requires mechanism).

Can you test that it works @BeastyBlacksmith ?

### DIFF
--- a/src/Plots.jl
+++ b/src/Plots.jl
@@ -8,7 +8,7 @@ if isdefined(Base, :Experimental) && isdefined(Base.Experimental, Symbol("@max_m
 end
 
 using Pkg, Dates, Printf, Statistics, Base64, LinearAlgebra, SparseArrays, Random
-using SnoopPrecompile, Preferences, Requires, Reexport
+using SnoopPrecompile, Preferences, Requires, Reexport, RelocatableFolders
 using Base.Meta
 using Unzip
 @reexport using RecipesBase
--- a/src/Plots.jl
+++ b/src/Plots.jl
@@ -168,11 +168,11 @@ include("examples.jl")
 include("plotattr.jl")
 include("backends.jl")
 const CURRENT_BACKEND = CurrentBackend(:none)
-include("backends/web.jl")
 include("output.jl")
 include("ijulia.jl")
 include("fileio.jl")
 include("shorthands.jl")
+include("backends/web.jl")
 include("init.jl")
 
 end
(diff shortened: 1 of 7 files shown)


BeastyBlacksmith:
```julia
julia> using Plots; plotly(); plot(rand(10))
[ Info: Precompiling Plots [91a5bcdd-55d7-5caf-9e0b-520d859cae80]
┌ Warning: backend `GR` is not installed.
└ @ Plots ~/.julia/dev/Plots/src/backends.jl:33
[ Info: GR
[ Info: Precompiling PlotlyKaleido [f2990250-8cf9-495f-b13a-cce12b45703c]
```
This works. I'm a bit irritated by that warnong though

BeastyBlacksmith:
we might want to add a simple test with the ploty backend, that creates a html-file

BeastyBlacksmith:
I mean
```┌ Warning: backend `GR` is not installed.
└ @ Plots ~/.julia/dev/Plots/src/backends.jl:33
```

t-bltg:
> we might want to add a simple test with the ploty backend, that creates a html-file

As I said, I'm not `plotly` competent at all, so if you have a test idea ...

t-bltg:
> Warning: backend `GR` is not installed.

How can that happen ? `GR` is a hard depedency.

BeastyBlacksmith:
> > Warning: backend `GR` is not installed.
> 
> How can that happen ? `GR` is a hard depedency of `Plots`, maybe run `instantiate` ?

Hmm.. I had this, but it doesn't show up on CI. Must be my system then.

BeastyBlacksmith:
> I mean
> 
> ```
> ┌ Warning: backend `GR` is not installed.
> └ @ Plots ~/.julia/dev/Plots/src/backends.jl:33
> ```

FWIW, I've seen that warning popping up in multiple occasions now, also on other peoples machines. It's working afterwards nevertheless, but it is irritating.

Secondly, I happened to notice that you can't use the plotlybase functionality if you haven't PlotlyKaleido installed, which ideally shouldn't be.

t-bltg:
> FWIW, I've seen that warning popping up in multiple occasions now, also on other peoples machines. It's working afterwards nevertheless, but it is irritating.

Maybe an environment stack issues. Could help if we had a reproducer.

PR #4598, opened 2022-12-15
process magic marker attribute consistently

BeastyBlacksmith:
I noticed that the example in https://docs.juliaplots.org/stable/pipeline/#An-example-command didn't work on some backends, e.g. `pgfplotsx`.

I think proccessing of the markershapes should not depend on whether a backend supports a certain markershape or not.


src/components.jl
```diff
@@ -294,7 +294,8 @@ function font(args...; kw...)
         elseif sym === :rotation
             rotation = kw[sym]
         elseif sym === :color
-            color = parse(Colorant, kw[sym])
+            col = kw[sym]
+            color = col isa Colorant ? col : parse(Colorant, col)
```
t-bltg, line 298: Ha, sorry, that got fixed on master.

PR #4604, opened 2022-12-15
Attribute description enhancement - proof of demo

Ininterrompue:
Enhances descriptions for the following histogram-related attributes as a proof of demonstration:
* `bins`
* `show_empty_bins`
* `weights`

Two issues:
1. The links are definitely not ideal - help wanted to make this clean. Goal: series attributes should link to relevant series types docstrings on the API page.
2. There's a bunch of newlines `\n` as-is - help wanted to hide these

### DIFF
--- a/src/arg_desc.jl
+++ b/src/arg_desc.jl
@@ -2,6 +2,9 @@ const AStr = AbstractString
 const ColorType = Union{Symbol,Colorant,PlotUtils.ColorSchemes.ColorScheme,Integer}
 const TicksType = Union{AVec{Real},Tuple{AVec{Real},AVec{AStr}},Symbol}
 
+link_histogram   = "[`histogram`](https://docs.juliaplots.org/stable/api/#Plots.histogram-Tuple)"
+link_histogram2d = "[`histogram2d`](https://docs.juliaplots.org/stable/api/#Plots.histogram2d-Tuple)"
+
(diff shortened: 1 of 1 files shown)


src/arg_desc.jl
```diff
@@ -2,6 +2,9 @@ const AStr = AbstractString
 const ColorType = Union{Symbol,Colorant,PlotUtils.ColorSchemes.ColorScheme,Integer}
 const TicksType = Union{AVec{Real},Tuple{AVec{Real},AVec{AStr}},Symbol}
 
+link_histogram   = "[`histogram`](https://docs.juliaplots.org/stable/api/#Plots.histogram-Tuple)"
```
t-bltg, line 5: I think this is a good idea, although this is not really aimed toward being maintainable.

t-bltg:
I don't want to block any good effort towards having cleaner and enhanced docs.

So @Ininterrompue tell us when you want this merged.

@BeastyBlacksmith, does this looks good to you ?

BeastyBlacksmith:
In its current state it doesn't produce usable docs. Though its heading in a good direction.

src/arg_desc.jl
BeastyBlacksmith: an option would be to insert some sort of tag like e.g. `[:histogram:]` and then replace that with the link when building the ducumentation.

Ideally that would then end up in the docstring as well, but I don't see a way to make that happen

PR #4615, opened 2023-01-02
support plotting offset images

simeonschaub:
`plot(::OffsetMatrix{<:Colorant})` currently fails, because the recipe
assumes all axes are one-based. I wasn't sure how to add a test for
this, so I'd appreciate any pointers.

### DIFF
--- a/src/recipes.jl
+++ b/src/recipes.jl
@@ -1309,7 +1309,7 @@ function clamp_greys!(mat::AMat{<:Gray})
 end
 
 @recipe function f(mat::AMat{<:Gray})  # COV_EXCL_LINE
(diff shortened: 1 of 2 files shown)


BeastyBlacksmith:
I think adding a test that passes an OffsetArray to `heatmap` would catch this, no?
Would be enough to check that the recipe does not throw, no need for display.

simeonschaub:
`heatmap` for `OffsetArray` is actually broken in other fun ways, but I added two tests for my initial use case.

simeonschaub:
Any idea what's going on with the Conda errors?

BeastyBlacksmith:
Not yet :/

PR #4617, opened 2023-01-03
enable more precompilation on Windows

KristofferC:
This works around JuliaLang/julia/issues/46989 and give good speedups on 1.9.0:

Before:

```julia
julia> @time (p = plot(rand(2,2)); display(p))
  3.102817 seconds (9.33 M allocations: 580.555 MiB, 3.48% gc time, 141.40% compilation time: <1% of which was recompilation)
```

After:

```julia
julia> @time (p = plot(rand(2,2)); display(p))
  0.190534 seconds (75.80 k allocations: 4.387 MiB, 93.92% compilation time: 7% of which was recompilation)
```


Sacha0 (approved):
🥲

KristofferC:
I can haz release? :P

PR #4622, opened 2023-01-06
use === instead of == to compare Symbols

ranocha:
I found invalidations from loading both RecipesBase.jl and SciMLBase.jl using the beta of Julia 1.9:
```julia
julia> using Pkg; Pkg.activate(temp=true); Pkg.add("SciMLBase"); Pkg.develop("RecipesBase")

julia> using SnoopCompileCore; invalidations = @snoopr(using RecipesBase, SciMLBase); using SnoopCompile

julia> trees = invalidation_trees(invalidations)
[...]
 inserting ==(retcode::SciMLBase.ReturnCode.T, s::Symbol) @ SciMLBase ~/.julia/packages/SciMLBase/QqtZA/src/retcodes.jl:348 invalidated:
   backedges: 1: superseding ==(x, y) @ Base Base.jl:127 with MethodInstance for ==(::Any, ::Symbol) (12 children)
```

I fixed invalidations by using `===` instead of `==` when comparing to a `Symbol`, which is also recommended in the docstring of `Symbol`:
```julia
help?> Symbol
search: Symbol
[...]
  Symbols are immutable and should be compared using ===. The implementation re-uses the same object for all
  Symbols with the same name, so comparison tends to be efficient (it can just compare pointers).
[...]
```

### DIFF
--- a/RecipesBase/src/RecipesBase.jl
+++ b/RecipesBase/src/RecipesBase.jl
@@ -73,10 +73,10 @@ _is_arrow_tuple(expr::Expr) =
     expr.head ≡ :tuple &&
     !isempty(expr.args) &&
     isa(expr.args[1], Expr) &&
-    expr.args[1].head == :(-->)
+    expr.args[1].head === :(-->)
 
-_equals_symbol(x::Symbol, sym::Symbol) = x == sym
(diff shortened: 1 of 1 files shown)


ranocha:
CI fails with a lot of `Path to conda environment is not valid` errors that do not seem to be related to this PR?

PR #4624, opened 2023-01-07
work around CI caching issues

simeonschaub:


simeonschaub:
Alternatively, we might want to try @giordano's suggestion in https://github.com/julia-actions/cache/issues/42#issuecomment-1374495622 and just downgrade instead of removing caching alltogether

t-bltg:
@simeonschaub, do you want a wipeout of caches in https://github.com/JuliaPlots/Plots.jl/actions/caches ?

simeonschaub:
Yes, let's try that

t-bltg:
> Yes, let's try that

Done.

simeonschaub:
@t-bltg could you rerun CI once the first run finishes? We should make sure that caching works properly now

t-bltg:
I've opened a dummy PR at https://github.com/JuliaPlots/Plots.jl/pull/4625.

giordano:
I'm not sure cache is hit on different branches/PR until it's in the default branch.

t-bltg:
Re-running failing ci in https://github.com/JuliaPlots/Plots.jl/pull/4615 now.

simeonschaub:
Sorry, I meant rerunning CI just on this branch, but as @giordano says, this might not actually do anything. I propose that we just merge this if CI passes and then verify by creating a PR on top of that new master branch

t-bltg:
Deleting the cache seems to have worked:

https://github.com/JuliaPlots/Plots.jl/actions/runs/3859523069/jobs/6579133488 (failing)
vs
https://github.com/JuliaPlots/Plots.jl/actions/runs/3859523069/jobs/6584314006 (running)

simeonschaub:
Yes, but I believe that's just due to it now not using any cache at all. Once another cache is created with julia-actions/cache v1.2 on master again, I'd expect those jobs to fail as before, so we should merge this before any other commits are merged into the main branch.

simeonschaub:
Looks like this worked. Should be good to merge, but immediately before merging we should delete the caches again, because it looks like those other retriggered jobs created new caches.

t-bltg:
In the end I don't think this PR is necessary.
I've re-run some failing jobs on previous commits in master using new cache files, and we'll see if ci fails again or not.

simeonschaub:
If you look in the CI logs, those are all not hitting any existing caches though. So just rerunning a bunch of failed jobs without a plan is a bit counter-productive right now, because it doesn't trigger the original issue and just creates more potentially corrupt cache files for later jobs

t-bltg:
> If you look in the CI logs, those are all not hitting any existing caches though

https://github.com/JuliaPlots/Plots.jl/pull/4626 hits caches correctly.

PR #4639, opened 2023-01-19
Add :unitformat to _base_supported_args

xwjqv:
This prevents `:unitformat` from getting into `extra_kwargs`
and crashing the PGFPlotsX backend.

### DIFF
--- a/src/backends.jl
+++ b/src/backends.jl
@@ -315,6 +315,7 @@ const _base_supported_args = [
     :show_empty_bins,
     :z_order,
     :permute,
+    :unitformat,
 ]
 
 function merge_with_base_supported(v::AVec)


BeastyBlacksmith:
What is `:unitformat` and what does it do?

xwjqv:
`unitformat` is an attribute for setting the format of units from Unitful.jl in the axis labels. The code for that is in [src/unitful.jl](https://github.com/JuliaPlots/Plots.jl/blob/master/src/unitful.jl)
[usage examples in documentation](https://docs.juliaplots.org/stable/generated/unitfulrecipes_examples/#Unit-formatting)

Before this change the PGFPlotsX backend crashed, when setting `unitformat`, because the value can't be converted to TeX or ends up in the .tex file and latex does not know what to do with it.

BeastyBlacksmith:
Thanks for the clarification. LGTM

Issue #4631, opened 2023-01-12
[FR] remove explicit GR dependency from `Project.toml`


t-bltg:
Users would have to install an explicit backend (`] add Plot GR`) to use `Plots`.
Following https://github.com/JuliaPlots/Plots.jl/pull/4566#issue-1467108667, mentioned in https://github.com/JuliaPlots/Plots.jl/issues/4565#issuecomment-1371805519.
For Plots 2.0, since breaking.

PR #4641, opened 2023-01-26
fix warning

t-bltg:


BeastyBlacksmith (approved):
To test this I did the following:
```
JULIA_DEPOT_PATH=$(mktemp -d) julia
```
```julia
pkg> add Plots
# Ctrl-C when precompilation started
julia> using Plots
```

BeastyBlacksmith:
Need to fix CI though ^^

t-bltg:
The warning also appeared in `PlotUtils` (https://github.com/JuliaPlots/PlotUtils.jl/actions/runs/4015693175/jobs/6897730588#step:6:387), so I could find a fix.

CI failure is odd, will have a look into that.

t-bltg:
That's all related to `GR` being a hard `Plots` dependency.

This warning would not have occurred if https://github.com/JuliaPlots/Plots.jl/issues/4631 was fixed.

PR #4642, opened 2023-01-26
simplify `open`

t-bltg:
While trying to fix `nightly`.
Seems like there is a missing `close` somewhere...

### DIFF
--- a/src/backends/web.jl
+++ b/src/backends/web.jl
@@ -25,22 +25,19 @@ embeddable_html(plt::AbstractPlot) = html_head(plt) * html_body(plt)
 function open_browser_window(filename::AbstractString)
     @static if Sys.isapple()
         return run(`open $(filename)`)
-    end
-    @static if Sys.islinux() || Sys.isbsd()    # Sys.isbsd() addition is as yet untested, but based on suggestion in https://github.com/JuliaPlots/Plots.jl/issues/681
+    elseif Sys.islinux() || Sys.isbsd()    # Sys.isbsd() addition is as yet untested, but based on suggestion in https://github.com/JuliaPlots/Plots.jl/issues/681
         return run(`xdg-open $(filename)`)
-    end
-    @static if Sys.iswindows()
+    elseif Sys.iswindows()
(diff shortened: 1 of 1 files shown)


src/backends/web.jl
```diff
@@ -25,22 +25,19 @@ embeddable_html(plt::AbstractPlot) = html_head(plt) * html_body(plt)
...
-    @warn "Unknown OS... cannot open browser window."
 end
 
 function write_temp_html(plt::AbstractPlot)
     html = standalone_html(plt; title = plt.attr[:window_title])
     filename = string(tempname(), ".html")
-    output = open(filename, "w")
-    write(output, html)
-    close(output)
+    write(filename, html)
```
BeastyBlacksmith, line 40: I am surprised this works without opening a file first

t-bltg: See https://docs.julialang.org/en/v1/base/io-network/#Base.write

And https://github.com/JuliaLang/julia/blob/0231c224a1ff25aba975a182b465e06de10f72e5/base/io.jl#L459

BeastyBlacksmith: Thats handy

t-bltg: Yup, as convenient as `read(filename, String)`.

BeastyBlacksmith:
> Seems like there is a missing `close` somewhere...

Are you sure? It looks kore like a bug in Pkg to me.

t-bltg:
I mean by that on `nightly` we hang waiting for some `i/o` operations, I vaguely recall seeing something similar in some discussion in the julia repo (but don't recall which one ...).
```
[pid 8003] waiting for IO to finish:
 TYPE[FD/PID]       @UV_HANDLE_T->DATA
 pipe[18]           @0x29f8660->0x7f1664428880
 pipe[12]           @0x307d3b0->0x7f1664428920
 process[8015]      @0x2be3af0->0x7f1664c696d0
```

Probably minor (likely to be fixed by future commits on `julia@master`), but let's keep an eye on this.

PR #4647, opened 2023-02-01
Fix code loading for pgfplotsx

BeastyBlacksmith:
Fix https://github.com/KristofferC/PGFPlotsX.jl/issues/308

Ref: https://github.com/JuliaPlots/Plots.jl/pull/4566#issuecomment-1410529861


mkitti:
Is this unique to this backend?

BeastyBlacksmith:
I think so, yes

Issue #4589, opened 2022-12-07
[FR] use package extensions


t-bltg:
X ref  https://github.com/JuliaLang/julia/pull/47695

mkitti:
Documentation here:
https://pkgdocs.julialang.org/dev/creating-packages/#Weak-dependencies

PR #4649, opened 2023-02-08
package extensions - weak deps

t-bltg:
Fix https://github.com/JuliaPlots/Plots.jl/issues/4589.

Note that we don't replace the `@require` statements for backends loading since their treatment is somewhat "special": https://github.com/JuliaPlots/Plots.jl/blob/ef504453651743928ac09314ec28fb7b4b4804c5/src/backends.jl#L100

I'm unsure this logic can be moved to package extensions (neither if it's worth it).

NOTE: the formatter check failure is investigated in https://github.com/tkf/JuliaProjectFormatter.jl/pull/7.


t-bltg:
I fail to see:
1. why this PR increases (slightly) invalidations - not blocking;
2. most importantly, why [CI segfaults](https://github.com/JuliaPlots/Plots.jl/actions/runs/4133983381/jobs/7144558678#step:10:2300) on `PythonPlot` related stuff on `linux - ubuntu` after testing ends, while a dummy PR opened at the same moment passes CI (https://github.com/JuliaPlots/Plots.jl/pull/4650) - this is blocking.

I cleaned up the actions caches, but it didn't seem to help :shrug: .

I guess another look at this (maybe @mkitti, or @BeastyBlacksmith) is necessary.

t-bltg:
Setting the `ENV` variable `MPLBACKEND: "agg"` seems to solve the segfaults.

ext/IJuliaExt.jl
```diff
@@ -1,10 +1,15 @@
-const _use_local_dependencies = Ref(false)
-const _use_local_plotlyjs = Ref(false)
+module IJuliaExt
```
BeastyBlacksmith, line 1: Did you test, that this works in jupyter notebooks?

ext/UnitfulExt.jl
```diff
@@ -200,27 +197,14 @@ Base.ncodeunits(n::S) = ncodeunits(n.content)
...
-```julia
-julia> plot([0,1]u"m", [1,2]u"m/s^2", xlabel=P"This label will NOT display units")
-
-julia> plot([0,1]u"m", [1,2]u"m/s^2", xlabel="This label will display units")
-```
-"""
-macro P_str(s)
-    return ProtectedString(s)
-end
+Plots.protectedstring(s) = ProtectedString(s)
```
BeastyBlacksmith, line 201: why does the macro needs to be moved?

ext/IJuliaExt.jl
t-bltg: Good point, I haven't used the Ijulia kernel in notebooks for a long time but this must be checked. Will do.

ext/UnitfulExt.jl
t-bltg: It doesn't seem possible to export symbols from the weak extensions modules.

So the "fix" here I found is to define a `protectedstring` helper in Plots (generic function with no method) and define a method `Plots.protectedstring` in the weak ext module.

ext/IJuliaExt.jl
t-bltg: IJulia

Looks like it's working. Should I check something else (I'm not familiar with IJulia) ?

BeastyBlacksmith: thats fine

ext/UnitfulExt.jl
BeastyBlacksmith: would it work if we only have `macro P_str end` in Plots and then the implementation in the extension?

Probably macros don't hade a methods table, right?

t-bltg: You cannot do something like this:
```julia
macro Plots.P_str(...)
    # ...
end
```
And using the following doesn't work either (breaks `UnitfulExt` precompilation):
```julia
@eval Plots macro P_str(...)
    # ...
end
```

BeastyBlacksmith: > Finally, this doesn't work (no `P_str` symbol exported):
> 
> ```julia
> export P_str
> macro P_str(...)
>     # ...
> end
> ```

I would expect that this exports to `Plots` rather than Main

t-bltg:
I'll probably merge this in the next few days, unless new comments.

PR #4671, opened 2023-02-17
gr: add support for user fonts

jheinen:


t-bltg:
Does GR have a font cache mechanism ?
`gr_set_font` is called a lot and `GR.loadfont` might be costly (I haven't checked) ?

jheinen:
Yes, there is a font cache mechanism in GR. This MR only adds support for additional/other fonts in the user's or the system-wide font directories - `GR.loadfont()` is only called once in this case and the font dictionary is extended.

t-bltg (approved):

BeastyBlacksmith:
Do we need to bump compat or would this also work with GR v0.69.5?

jheinen:
> Do we need to bump compat or would this also work with GR v0.69.5?

No, the `GR.loadfont` function was already introduced in version 0.64.3.

PR #4672, opened 2023-02-17
Silence precompilation process

giordano:
Precompiling `Plots` causes an annoying message being printed in julia v1.9+:
```
(tmp) pkg> precompile
Precompiling environment...
  47 dependencies successfully precompiled in 141 seconds. 86 already precompiled.
  1 dependency had warnings during precompilation:
┌ Plots [91a5bcdd-55d7-5caf-9e0b-520d859cae80]
│  [ Info: GR
└  
```
due to https://github.com/JuliaPlots/Plots.jl/blob/65cf8a1bcffd79e359f4d1e4ec615e8f57c81976/src/init.jl#L88 I don't know why there's a call to `@info` there, that hardly looks like a bottleneck, but if you really want to keep it you should at least silence it.  This PR uses a `NullLogger` to achieve that.  Adding the `Logging` standard library doesn't add anything to load time, since it's already an indirect dependency through many other dependencies.

Alternatively, `@info` can simply be removed, and keep only the call to `backend_package_name()`, if that's what you want to precompile.


t-bltg:
This `@info` line was added for debugging purposes intended for beginners using Plots and hitting precompilation issues as we use to see on discourse. It can be tricky to help someone remotely especially on errors related to precompilation. I have nothing against silencing it by default but I want a simple equivalent (again aimed at beginners) to see which default backend is being precompiled.

t-bltg:
Using `@debug` instead of `@info` is fine with me.

t-bltg (approved):

PR #4645, opened 2023-01-31
Added arbitrary legend_columns for gr

briederer:
Since I modified the code for myself I thought it would be nice to also upload so others can use it.
Especially since it is asked quite often e.g. on [Discourse](https://discourse.julialang.org/t/plot-legend-with-multiple-columns/29195) and #2423.
For now it is just a quick hacky solution but it works with `legend_title` and without. Also it differs between `legend_column=-1` as a full horizontal version and `legend_column=nseries` where only the labels are printed horizontally but the `legend_title` is printed center-aligned above the series.

So maybe even if it is not merged it may help some people or be useful when switching to `Plots 2.0`

The following code produces the output and the figure below
```julia 
data = rand(10, 3)
leg_cols = -1:4

leg_plots(; kw...) = begin
  map((leg_col -> begin
      plot(data; marker=:circle, ticks=:none, plot_title=string("legend_columns = ", leg_col), legend_columns=leg_col, kw...)
    end), leg_cols)
end
(w, h) = Plots._plot_defaults[:size]
with(scalefonts=1, size=(w, h)) do
  plot(leg_plots()..., leg_plots(legend_title="Test")...; layout=(4, 3))
end
```

``` julia
┌ Warning: n° of legend_column=0. Assuming vertical layout.
└ @ Plots ~/.julia/dev/Plots/src/backends/gr.jl:1223
┌ Warning: n° of legend_column=4 is larger than n° of series=3
└ @ Plots ~/.julia/dev/Plots/src/backends/gr.jl:1220
┌ Warning: n° of legend_column=0. Assuming vertical layout.
└ @ Plots ~/.julia/dev/Plots/src/backends/gr.jl:1223
┌ Warning: n° of legend_column=4 is larger than n° of series=3
└ @ Plots ~/.julia/dev/Plots/src/backends/gr.jl:1220
```
multi_colums_gr

PR #4678, opened 2023-03-03
Adding arbitrary legend_columns for python-backends

briederer:
As a follow up to #4645 I am trying to implement the feature also for other backends.

As asked in https://github.com/JuliaPlots/Plots.jl/pull/4645#issuecomment-1453323122 by @lmanzanillas I have started with the `PythonPlot` backend, which has been rather trivial in principle. However, I could not reproduce the behaviour of a fully horizontal legend for `legend_columns=-1` like in `gr()`. I.e. the `legend_title` being on the same line as the series labels. 

I'll have a look at the other backends too and then also create Test Images for the docs.

# Samples

```julia
data = rand(10, 4)
leg_cols = -2:5

leg_plots(; kw...) = begin
  map((leg_col -> begin
      plot(data; marker=:circle, ticks=:none, plot_title=string("legend_columns = ", leg_col), legend_columns=leg_col, kw...)
    end), leg_cols)
end
(w, h) = (600,600)
with(scalefonts=0.5, size=(w, h)) do
  plot(leg_plots()..., leg_plots(legend_title="Test")...; layout=(4, 4))
end
```

## GR
sample_gr

```julia
┌ Warning: n° of legend_column=0 has undefined behaviour. Assuming vertical layout.
└ @ Plots ~/.julia/dev/Plots/src/backends/gr.jl:1226
┌ Warning: n° of legend_column=4 is larger than n° of series=3
└ @ Plots ~/.julia/dev/Plots/src/backends/gr.jl:1223
┌ Warning: n° of legend_column=0 has undefined behaviour. Assuming vertical layout.
└ @ Plots ~/.julia/dev/Plots/src/backends/gr.jl:1226
┌ Warning: n° of legend_column=4 is larger than n° of series=3
└ @ Plots ~/.julia/dev/Plots/src/backends/gr.jl:1223
```

## PythonPlot
sample_pythonplot

```julia
┌ Warning: n° of legend_column=0 has undefined behaviour. Assuming vertical layout.
└ @ Plots ~/.julia/dev/Plots/src/backends/pythonplot.jl:1438
┌ Warning: n° of legend_column=4 is larger than n° of series=3
└ @ Plots ~/.julia/dev/Plots/src/backends/pythonplot.jl:1435
┌ Warning: n° of legend_column=0 has undefined behaviour. Assuming vertical layout.
└ @ Plots ~/.julia/dev/Plots/src/backends/pythonplot.jl:1438
┌ Warning: n° of legend_column=4 is larger than n° of series=3
└ @ Plots ~/.julia/dev/Plots/src/backends/pythonplot.jl:1435
```

## Other Backends

To be done.


briederer:
Also interesting to note:
`PythonPlot` orders the series differently in the multi-column layout (i.e. column-major). Since this is also the way how julia treats matrices it is probably the preferred way. So I would suggest tho change `GR` to this behaviour too.

briederer:
Apparently for the some backends the `legend_column=-1` possibility is not even implemented so I won't touch it.
The remaining backends are therefore:
- [ ] `Gaston`
- [ ] `PGSPlotsX`

BeastyBlacksmith:
> Also interesting to note: `PythonPlot` orders the series differently in the multi-column layout (i.e. column-major). Since this is also the way how julia treats matrices it is probably the preferred way. So I would suggest tho change `GR` to this behaviour too.

FWIW pgfplots also sorts these row-major

briederer:
Hmm 🤔 
I mean zo be fair probably is row-major a bit more intuitive in a horizontal legend since (at least in the english language) we are reading fromm left to right and then jump in the next line.
I don't have any strong opinions, except for wanting a common behavior for the different backends.

BeastyBlacksmith:
It would be possible to let the sign carry the majorness.
Not sure, if I'd like that though ^^

briederer:
That's a good point.
I'll change it tomorrow. Since it's already there but only commented it should not take too much time.

However, I also want to give the pgfplotsx backend another look.
Since it is implemented apparently in the backend it should be doable.

BeastyBlacksmith:
> However, I also want to give the pgfplotsx backend another look.
> Since it is implemented apparently in the backend it should be doable.

It is already working for that. Consider:
```julia
julia> using Plots; pgfplotsx()
Plots.PGFPlotsXBackend()

julia> plot(rand(4,4), legend_columns = 2)

```

Its just that your example errors because pgfplots errors on too small labels for some reason...

BeastyBlacksmith:
@briederer is this good to go?

briederer:
Oh sorry, totally forgot about this.
I am not sure if GR and python are consistent at the moment and it still lacks proper docs and/or TestImages.
I am currently very limited in my time, but I'll have another look next week (unless someone else has time and wants to jump in 😅).

BeastyBlacksmith:
bump

BeastyBlacksmith:
kindly reminder

PR #4679, opened 2023-03-04
Add AbstractVector `Shape(x, y)` constructor

nick4f42:
Added a constructor for `Shape(x, y)` where `x` and `y` are abstract vectors. Before, `Shape(x, y)` would only accept `Vector`s.


t-bltg (approved):

t-bltg:
Thanks !

PR #4680, opened 2023-03-04
Wrapping of power in PGFPlotsX labels

### LINKED ISSUE #4676: [BUG] Log scale axis labels not brace-wrapped in PGFPlotsX
## Details
Log scale axis labels with a power of two symbols or more is broken
```julia
using Plots
pgfplotsx()
PGFPlotsX.latexengine!(PGFPlotsX.PDFLATEX) # doubt it's relevant, but just in case - I use it
plot(
    plot([1e-5, 1e10]; title="Automatic labels",
        yticks=[1e-5, 1e0, 1e10]),
    plot([1e-5, 1e10]; title="Manual labels",
        yticks=([1e-5, 1e0, 1e10], ["\$10^{-5}\$", "\$10^{0}\$", "\$10^{10}\$"])),
    legend=:none, yscale=:log10
)
```
image

Looking at the generated .tex
```julia
savefig("plot.tex")
open("plot.tex") do f
    for line in eachline(f)
        i0 = findfirst("yticklabels=", line)
        i1 = findfirst("ytick=", line)
        if !isnothing(i0)
            println(line[i0[1]:i1[1]-3])
        end
    end
end
```
```
yticklabels={{$10^-5$,$10^0$,$10^10$}}
yticklabels={{$10^{-5}$,$10^{0}$,$10^{10}$}}
```
we see that the power is not wrapped automatically. 
The behavior changed between [v1.38.1](https://github.com/JuliaPlots/Plots.jl/releases/tag/v1.38.1) and [v1.38.2](https://github.com/JuliaPlots/Plots.jl/releases/tag/v1.38.2), so I suspect #4618 was almost there - the tests only covered single-digit powers.

Looks like it can be fixed here
https://github.com/JuliaPlots/Plots.jl/blob/5f29256c5e08eca051b59c61ada27b3cce129439/src/axes.jl#L115
by changing to 
```julia
Dict{Symbol,Function}(:log10 => x -> "10^{$x}", :log2 => x -> "2^{$x}", :ln => x -> "e^{$x}")
```
though it's a low-level function and I am not confident enough to open a PR.

### Backends

gr (default) |     |  x   |
pythonplot   |     |     | x
plotlyjs     |     |     |  x
pgfplotsx    |  x   |     |
unicodeplots |     |  x   | x
inspectdr    |     |     | x
gaston       |     |     | x

### Versions

Plots.jl version: ```v1.38.6```
```
[28b8d3ca] GR v0.71.7
[8314cec4] PGFPlotsX v1.5.3
[b8865327] UnicodePlots v3.4.0
```
Output of `versioninfo()`:
```
Julia Version 1.8.1
Commit afb6c60d69a (2022-09-06 15:09 UTC)
Platform Info:
  OS: Linux (x86_64-linux-gnu)
  CPU: 12 × AMD Ryzen 5 3600 6-Core Processor
  WORD_SIZE: 64
  LIBM: libopenlibm
  LLVM: libLLVM-13.0.1 (ORCJIT, znver2)
  Threads: 1 on 12 virtual cores
```

ivan-boikov:
Fixes #4676

I'd like to bring to attention that the tests for axes are not backend-agnostic - current tests are passed because GR overrided ```labelfunc```, but PGFPlotsX did not and failed.
I added a new test, but it changes the backend during a run, which I think is not good.

### DIFF
--- a/src/backends/pgfplotsx.jl
+++ b/src/backends/pgfplotsx.jl
@@ -81,6 +81,8 @@ curly(obj) = "{$(string(obj))}"
 latex_formatter(formatter::Symbol) = formatter in (:plain, :latex) ? formatter : :latex
 latex_formatter(formatter::Function) = formatter
 
+labelfunc(scale::Symbol, backend::PGFPlotsXBackend) = labelfunc_tex(scale)
+
 pgfx_halign(k) = (left = "left", hcenter = "center", center = "center", right = "right")[k]
 
 function (pgfx_plot::PGFPlotsXPlot)(plt::Plot{PGFPlotsXBackend})
--- a/test/test_axes.jl
+++ b/test/test_axes.jl
@@ -27,6 +27,10 @@
     @test Plots.get_labels(:engineering, float.(500:500:1500), :identity) ==
           ["500.×10^{0}", "1.00×10^{3}", "1.50×10^{3}"]
     @test Plots.get_labels(:latex, 1:3, :identity) == ["\$1\$", "\$2\$", "\$3\$"]
+    # GR is used during tests and it correctly overrides labelfunc(), but PGFPlotsX does not and will fail
+    pgfplotsx()
+    @test Plots.get_labels(:auto, 1:3, :log10) == ["10^{1}", "10^{2}", "10^{3}"]
+    gr()
     @test Plots.get_labels(:auto, 1:3, :log10) == ["10^{1}", "10^{2}", "10^{3}"]
     @test Plots.get_labels(:auto, 1:3, :log2) == ["2^{1}", "2^{2}", "2^{3}"]
     @test Plots.get_labels(:auto, 1:3, :ln) == ["e^{1}", "e^{2}", "e^{3}"]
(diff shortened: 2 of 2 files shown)


t-bltg:
> I added a new test, but it changes the backend during a run, which I think is not good.

To temporarily swap the backend, use the following:
```julia
with(:pgfplotsx) do
   plot(..)
end
# backend restored to the previously selected one
```

ivan-boikov:
Huh, interesting, thanks! Will update in a bit.

t-bltg:
Shouldn't we cover multiple digits cases in the tests ?

ivan-boikov:
> Shouldn't we cover multiple digits cases in the tests ?

I don't think that's necessary due to how we add the braces. If they are there, they _will_ cover the entire number, it's actually harder to add them incorrectly. https://github.com/JuliaPlots/Plots.jl/blob/555d278fe1b2b861a96855fc64aeeb08c8894f54/src/axes.jl#L118-L122 I think that's why in other tests only single digit powers are tested.
So yeah, I made a wrong accusation in the bug report, sorry about that.

t-bltg:
I see, thanks for clarifying.

t-bltg (approved):

PR #4681, opened 2023-03-04
Support for disabling axis ticks in UnicodePlots

### LINKED ISSUE #4657: [FR] Support `xticks`, `yticks` with UnicodePlots
UnicodePlots supports `xticks` and `yticks` (at least as `Bool` arguments) since JuliaPlots/UnicodePlots.jl#234, but passing `xticks=false` or `yticks=false` using Plots.jl does nothing. It’s also not listed as supported in the [documentation](https://docs.juliaplots.org/stable/gallery/unicodeplots/generated/unicodeplots/). But since it’s there in UnicodePlots, it would be nice to have access to it from Plots.jl too.

ivan-boikov:
Fixes #4657


t-bltg:
In the UnicodePlots context, `get_ticks` can be expensive (it actually computes ticks which are unused for this backend).

I would instead add a `has_ticks` helper in `axes.jl` loosely defined as:
```julia
_has_ticks(::Any) = true  # catchall
_has_ticks(s::Symbol) = s !== :none
_has_ticks(::Nothing) = false
_has_ticks(b::Bool) = b

has_ticks(axis) = get(axis, :ticks, nothing) |> _has_ticks
```

, and use the following instead:
```julia
xticks = has_ticks(xaxis),
yticks = has_ticks(yaxis),
```

ivan-boikov:
> I would instead add a `has_ticks` helper in `axes.jl`

Good call. I will also add overloads for ```Tuple``` and ```AbstractVector``` in a few moments if that's alright.

src/axes.jl
```diff
@@ -321,6 +321,16 @@ get_ticks(ticks::Bool, args...) =
...
 
+# do not specify array item type to also catch e.g. "xlabel=[]" and "xlabel=([],[])"
+_has_ticks(v::AVec) = !isempty(v)
+_has_ticks(t::Tuple{AVec,AVec}) = !isempty(t[1])
+_has_ticks(s::Symbol) = s !== :none
+_has_ticks(b::Bool) = b
+_has_ticks(::Nothing) = false
+_has_ticks(::Any) = true
+
+has_ticks(axis::Plots.Axis) = get(axis, :ticks, nothing) |> _has_ticks
```
t-bltg, line 332:
```suggestion
has_ticks(axis::Axis) = get(axis, :ticks, nothing) |> _has_ticks
```

t-bltg (approved):

PR #4682, opened 2023-03-05
Add svg+xml generation to precompile setup

oschulz:
I think this speeds up first plot in VScode and Jupyter a bit more, subjectively.


oschulz:
What does that format check not like here?

t-bltg (approved):

t-bltg:
Thanks !

oschulz:
Thanks!

PR #4684, opened 2023-03-07
Do not include plotly.jl twice

antholzer:
At the moment when setting `:plotly` as the default backend (using Preferences) one gets quite a few warning of the type
```
WARNING: Method definition is_subplot_supported(Plots.PlotlyBackend) in module Plots at /home/stephan/.julia/dev/Plots/src/backends/plotly.jl:4 overwritten on the same line (check for duplicate calls to `include`).
  ** incremental compilation may be fatally broken for this module **
```
They are the result of `src/backend/plotly.jl` being included twice. The following commit is a workaround that seems to fix this.

### DIFF
--- a/src/backends.jl
+++ b/src/backends.jl
@@ -534,7 +534,7 @@ _post_imports(::PlotlyBackend) = @eval begin
     const PlotlyKaleido = Main.PlotlyKaleido
     # FIXME: in Plots `2.0`, `plotly` backend should be re-named to `plotlybase`
     # so that we can trigger include on `@require` instead of this
-    include(_path(:plotly))
+    CURRENT_BACKEND.sym === :plotly || include(_path(:plotly))
     include(_path(:plotlybase))
 end
 function _initialize_backend(pkg::PlotlyBackend)
@@ -553,7 +553,9 @@ function _initialize_backend(pkg::PlotlyBackend)
         # NOTE: `plotly` is special in the way that it does not require dependencies for displaying a plot
         # as a result, we cannot rely on the `@require` mechanism for loading glue code
         # this is why it must be done here.
-        @eval include(_path(:plotly))
+        if CURRENT_BACKEND.sym !== :plotly
+            @eval include(_path(:plotly))
+        end
     end
 end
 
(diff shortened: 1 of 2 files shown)


t-bltg:
No: if this is backend specific, then it should be fixed in `src/backends.jl`, somewhere along

https://github.com/JuliaPlots/Plots.jl/blob/20726c880f06cfcee1428205f72d70d07e7e53b5/src/backends.jl#L537-L538

t-bltg:
`is_subplot_supported` and `is_string_supported` seem to have been removed in https://github.com/JuliaPlots/Plots.jl/commit/03a9fc341920634b00531b06a21761442c59faf8, can you remove them in this PR ?

We could also use `Plot.PLOTS_DEFAULT_BACKEND == "plotly" || ...` as an alternative (since it is loaded from `Preferences`), I'm wondering if this is more robust than using `CURRENT_BACKEND.sym` ...

src/backends.jl
```diff
@@ -534,7 +534,7 @@ _post_imports(::PlotlyBackend) = @eval begin
     const PlotlyKaleido = Main.PlotlyKaleido
     # FIXME: in Plots `2.0`, `plotly` backend should be re-named to `plotlybase`
     # so that we can trigger include on `@require` instead of this
-    include(_path(:plotly))
+    CURRENT_BACKEND.sym === :plotly || include(_path(:plotly))
```
t-bltg, line 537:
```suggestion
    PLOTS_DEFAULT_BACKEND == "plotly" || include(_path(:plotly))
```

src/backends.jl
```diff
@@ -553,7 +553,9 @@ function _initialize_backend(pkg::PlotlyBackend)
         # NOTE: `plotly` is special in the way that it does not require dependencies for displaying a plot
         # as a result, we cannot rely on the `@require` mechanism for loading glue code
         # this is why it must be done here.
-        @eval include(_path(:plotly))
+        if CURRENT_BACKEND.sym !== :plotly
+            @eval include(_path(:plotly))
+        end
```
t-bltg, line 558:
```suggestion
        PLOTS_DEFAULT_BACKEND == "plotly" || @eval include(_path(:plotly))
```

t-bltg (changes requested):
Let's use `PLOTS_DEFAULT_BACKEND` instead since the issue is related to `Preferences` in the first place.

t-bltg:
I'm also curious to know if only removing the function definition lines actually suffice to solve your issue.

antholzer:
Sorry I did not include the full output of the warnings. All function definitions have one not just the ones I removed above.

Here is the output when I remove these two functions (actually only one since the other one was commented) but do not include the fix: https://gist.github.com/antholzer/64b71a4ea7ac32461b603f4e7de3fdc6)

t-bltg (approved):

PR #4690, opened 2023-03-13
Add supported arguments to backend.jl

nilsLue:
I have added arguments that are supported by a specific backend but not yet mentioned on the website.
I also corrected mistakes I made.


src/backends.jl
```diff
@@ -975,7 +1032,116 @@ const _pythonplot_seriestype = _pyplot_seriestype
 const _pythonplot_marker     = _pyplot_marker
 const _pythonplot_style      = _pyplot_style
 const _pythonplot_scale      = _pyplot_scale
-const _pythonplot_attr       = _pyplot_attr
+
+const _pythonplot_attr = merge_with_base_supported([
```
nilsLue, line 1036: The argument plot_title only works for subplots with the pyplot backend and not with pythonplot. 
Therefore the two backends do not support the same attributes. So I can  add supported arguments only to the pyplot backend I copied them before and added them to the pythonplot backend.

PR #4707, opened 2023-03-26
Fix logarithmic units

### LINKED ISSUE #4700: UnitfulExt not working for logarithmic units
UnitfulExt currently does not work for logarithmic units.
Example:
```Julia
using Plots
using Unitful:m, dB, dBm
plot( (1:10)m) # works
plot( (1:10)dB ) # does not work, produces error
plot( (1:10)dBm ) # does not work, produces error
```
The error reads
```
ERROR: MethodError: no method matching isinf(::Gain{Unitful.LogInfo{:Decibel, 10, 10}, :?, Int64})
```

gustaphe:
## Description
Closes #4700 

Partially based off of work by @ykonter (who is not in zenodo, as far as I can tell). Creates underscored extensions of Unitful's `unit` and `ustrip` to deal with logarithmic units.

This is not perfect, note how the y axis in the example doesn't rescale on the `plot!` call. But there is more than one interpretation of what that call should do anyway, so perhaps good enough should be left good enough here.

## Example
```julia
julia> plot((1:10)u"dBm", (1:10)u"dB"; seriestype=:scatter)

julia> plot!((1:10)u"mW", (1:10); seriestype=:scatter)
```
logunits

## Attribution

### DIFF
--- a/ext/UnitfulExt.jl
+++ b/ext/UnitfulExt.jl
@@ -5,9 +5,9 @@ module UnitfulExt
 
 import Plots: Plots, @ext_imp_use, @recipe, PlotText, Subplot, AVec, AMat, Axis
 import RecipesBase
-@ext_imp_use :import Unitful Quantity unit ustrip Unitful dimension Units NoUnits
+@ext_imp_use :import Unitful Quantity unit ustrip Unitful dimension Units NoUnits LogScaled logunit MixedUnits Level Gain uconvert
 
-const MissingOrQuantity = Union{Missing,<:Quantity}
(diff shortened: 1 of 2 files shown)


t-bltg (approved):
Xref https://github.com/jw3126/UnitfulRecipes.jl/pull/72 and https://github.com/jw3126/UnitfulRecipes.jl/pull/73 (to my current knowledge).
I think a partial fix is better than nothing.

gustaphe:
No documentation changes. The only thing I think is not in line with previous documentation is the conversion weirdness around dimensionless units vs logarithmic ones, and I'm not sure Plots.jl is the place you should be reading about that.

PR #4735, opened 2023-04-28
Fixed logic of warning on dimension mismatch

yha:
## Description

Fixes spurious warnings when `missing`/`NaN` is at the start or end of input vector, as in
```
plot([missing; 1:4], color=1:5)
```
Adds stricter checks with warnings in other cases. Tests pass even when these warnings are changed to errors, but I kept them as warnings to be less breaking.

`warn_on_attr_dim_mismatch` really shouldn't have anything to do with segments, so maybe it should be called from elsewhere in the pipeline, but I'm not sure where.

## Attribution

### DIFF
--- a/test/test_animations.jl
+++ b/test/test_animations.jl
@@ -43,7 +43,7 @@ end
         circleplot(x, y, i, line_z = 1:n, cbar = false, framestyle = :zerolines)
     end when i % 5 == 0 fps = 10
 
-    @test_throws LoadError macroexpand(
+    @test_throws ErrorException macroexpand(
         @__MODULE__,
         quote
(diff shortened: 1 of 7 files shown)


test/test_animations.jl
```diff
@@ -43,7 +43,7 @@ end
         circleplot(x, y, i, line_z = 1:n, cbar = false, framestyle = :zerolines)
     end when i % 5 == 0 fps = 10
 
-    @test_throws LoadError macroexpand(
```
BeastyBlacksmith, line 46: We can't change this as long as we test on 1.6 also

Issue #4730, opened 2023-04-26
[BUG][PGFPlotsX][Unitful] setting axis labels duplicates units


xwjqv:
When setting axis labels while using units and the PGFPlotsX backend,
the units appear twice in the label.

This probably happens because of the changes in #4723.
The units are also incorrectly italicized (maybe use [UnitfulLatexify.jl](https://github.com/gustaphe/UnitfulLatexify.jl) for it)
### Example 
```julia
using Plots, Unitful
pgfplotsx()
plot([1u"s",2u"s"],[1u"m",2u"m"],xlabel="t",ylabel="r")
````
![svg-image](https://user-images.githubusercontent.com/35635774/234439558-9ad5a851-0160-474f-8893-59016acbdca1.png)
#### crashing example
This has probably the same cause as #4722
```julia
using Plots, Unitful
pgfplotsx()
plot([1u"s",2u"s"],[1u"m/s^2",2u"m/s^2"],xlabel="t", ylabel="a")
```
### Backends

This bug occurs on ( insert `x` below )

Backend      | yes | no  | untested
-------------|-----|-----|---------
gr (default) |     | x    |
pythonplot   |     |     |x
plotlyjs     |     |    x |
pgfplotsx    | x    |     |
unicodeplots |     |   x  |
inspectdr    |     |  x   |
gaston       |     |  x   |

### Versions

Plots.jl version: master
Backend  version (`]st -m <backend(s)>`): PGFPlotsX v1.5.3
Output of `versioninfo()`:
```
Julia Version 1.8.5
Commit 17cfb8e65ea (2023-01-08 06:45 UTC)
Platform Info:
  OS: Linux (x86_64-linux-gnu)
  CPU: 16 × AMD Ryzen 7 1700 Eight-Core Processor
  WORD_SIZE: 64
  LIBM: libopenlibm
  LLVM: libLLVM-13.0.1 (ORCJIT, znver1)
  Threads: 8 on 16 virtual cores
Environment:
  JULIA_REVISE_WORKER_ONLY = 1
```

BeastyBlacksmith:
CC: @t-bltg

t-bltg:
Hum sorry for this.

Unitful was written for GR and we never tested it on PGFPlotsX. So everything is left to be to be correctly implemented here.

There are choice to be made. I'd say that the units should be in math mode for handling braces/parens (hence rendered slightly italicized). But it's to be discussed.

Contributions welcome 🙂.

[...]

PR #4747, opened 2023-05-13
Fix UnitfulExt for PGFPlotsX take 2

BeastyBlacksmith:
## Description

Fix https://github.com/JuliaPlots/Plots.jl/issues/4730
## Attribution

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -92,7 +92,7 @@ UnicodeFun = "0.4"
 UnicodePlots = "3.4"
 UnitfulLatexify = "1"
 Unzip = "0.1 - 0.2"
-julia = "1.6"
+julia = "1.9"
 
 [extras]
 Aqua = "4c88cf16-eb10-579e-8560-4a9242c79595"
--- a/Project.toml
+++ b/Project.toml
@@ -92,7 +92,7 @@ UnicodeFun = "0.4"
 UnicodePlots = "3.4"
 UnitfulLatexify = "1"
 Unzip = "0.1 - 0.2"
-julia = "1.9"
+julia = "1.6"
 
 [extras]
 Aqua = "4c88cf16-eb10-579e-8560-4a9242c79595"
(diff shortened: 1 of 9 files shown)


gustaphe:
LGTM

test/test_unitful.jl
```diff
@@ -9,7 +12,7 @@ zseries(pl, idx = length(pl.series_list)) = pl.series_list[idx].plotattributes[:
 testfile = tempname() * ".png"
 
 macro isplot(ex) # @isplot macro to streamline tests
-    :(@test $(esc(ex)) isa Plot)
+    :(@test $(esc(ex)) isa Plots.Plots.Plot)
```
t-bltg, line 15: `Plots.Plot` ?

Project.toml
```diff
@@ -71,19 +87,12 @@ RelocatableFolders = "0.3, 1"
...
 Unzip = "0.1 - 0.2"
-julia = "1.6"
-
-[extensions]
-FileIOExt = "FileIO"
-GeometryBasicsExt = "GeometryBasics"
-IJuliaExt = "IJulia"
-ImageInTerminalExt = "ImageInTerminal"
-UnitfulExt = "Unitful"
+julia = "1.9"
```
t-bltg, line 95: Why are we dropping support for LTS ?

BeastyBlacksmith: I don't know exactly how to make tests pass on earlier versions. The failures can be seen in https://github.com/JuliaPlots/Plots.jl/actions/runs/4993428661

t-bltg: Since `LaTeXStrings` is a hard  `Plots` dependency, you shouldn't use `@ext_imp_use` but regular imports or `Plots.LaTeXStrings...`.

Also, I'd sugest to add the `UnitfulLatexify` dependency to the `[weakdeps]` and `[extensions]` sections since it is only used in a extension.

t-bltg: I thought we were aiming at reducing `Plots` dependencies.
Since `Unitful` is a weak dependency, I think `UnitfulLatexify` should also be an extension.
I'm sure you can make it work through extensions with a bit of effort.

BeastyBlacksmith: https://julialang.slack.com/archives/C67EFTEF3/p1684332705496899

Issue #4719, opened 2023-04-10
[BUG] IJuliaExt causes problems with IJulia in default environment


oschulz:
## Details

IJulia is a bit peculiar in that you can really only have one version of it at the same time, otherwise one get's these "IJulia version doesn't match ..." warnings. This is because the (default) Jupyter kernelspec is bound to a specific IJulia version (since "kernel.json" points to `".julia/packages/IJulia/SOME_VERSION/src/kernel.jl"`). So one (at least me) typically keeps IJulia in the default (e.g. "v1.9") environment and out of specific project environment like the the "Project/Manifest.toml" kept next to a notebook.

This leads to the following problem with the current Plots version on Julia v1.9, though:

```
using Plots

[ Info: Precompiling IJuliaExt [2f4121a4-3b3a-5ce6-9c5e-1f2673ce168a]
ERROR: LoadError: ArgumentError: Package IJulia [7073ff75-c697-5162-941a-fcdaad2a7d2a] is required but does not seem to be installed:
 - Run `Pkg.instantiate()` to install all recorded dependencies.
```

This is very inconvenient, the only solution seems to be having IJulia in the notebook environment, but this would require having a separate Jupyter kernel for each notebook environment as well (or at lease having quite a few kernels). It's also really confusion to (esp. newer) users, because they can run `using IJulia` in the same notebook before `using Plots` cell without error, so IJulia is "there" (just not in an extension-compatible way).

IJulia is special here because it's automatically loaded without being part of the notebook, and because the "kernel.json" issue, Plot's other Pkg-extensions don't have to be this problem.

I would suggest the we handle IJulia via Requires again. Most (all) `..._ijulia_...` function in "IJuliaExt.jl" could be defined in Plots itself, since they don't depend on IJulia, so Requires would only need to load a few lines IJulia-specific code.

### Backends

Not backend-specific.

### Versions

Plots.jl version: v1.38.9
Backend  version (`]st -m <backend(s)>`): Not backend-specific
Output of `versioninfo()`:

```
Julia Version 1.9.0-rc2
Commit 72aec423c2a (2023-04-01 10:41 UTC)
Platform Info:
  OS: Linux (x86_64-linux-gnu)
  CPU: 16 × Intel(R) Core(TM) i9-9880H CPU @ 2.30GHz
  WORD_SIZE: 64
  LIBM: libopenlibm
  LLVM: libLLVM-14.0.6 (ORCJIT, skylake)
```

[...]

PR #4749, opened 2023-05-19
Use Base.require workaround in IJuliaExt

oschulz:
Addresses issue #4719.


oschulz:
Don't merge yet, still adding an improvement.

t-bltg (commented):
I'm a bit reluctant in making Plots messier than it already is.

That being said, we can have a work around if it is the only solution left, since it affects multiple end users.

At the very least, we should address 
https://github.com/JuliaPlots/Plots.jl/issues/4719#issuecomment-1554061512 so that we can link to an upstream issue in this PR.

oschulz:
Ok, should be good now.

oschulz:
TTFP doesn't seem to be increased noticeably, compared to "extension-mode".

PR #4752, opened 2023-05-20
Singleargustrip

gustaphe:
## Description

There was a bug hiding behind a test that didn't actually render the part that was supposed to be tested. Annotations with unitful coordinates used a non-working version of `_ustrip`. This fixes that.

## Attribution

### DIFF
--- a/ext/UnitfulExt.jl
+++ b/ext/UnitfulExt.jl
@@ -287,21 +287,32 @@ getaxisunit(a::Axis) = getaxisunit(a[:guide])
 #==============
 Fix annotations
 ===============#
-Plots.locate_annotation(
-    sp::Subplot,
-    x::MissingOrQuantity,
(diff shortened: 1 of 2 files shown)


gustaphe:
I can't fix the visual regressions because my system never shows a confirmation box, and I would be really surprised if they are caused by this PR.

BeastyBlacksmith:
I can reproduce the failures on master as well

gustaphe:
By the way, I can't run the test suite with the ENV setting either, because i get a segfault partway through. It might just be my computer being bad, or something about my python install. If there's no visual regression I've just used CI for tests in the past, but that doesn't work when I need to update the references.

(Perhaps a dream could be to get visual regression prompts as a comment on the PR, that would also automatically let the community see what your change does if it's visual)

Issue #4774, opened 2023-06-09
[BUG] Histogram log scale is broken


Moelf:
```julia
data = randn(10^6);

histogram(data; yscale=:log10)


DomainError with -4.773573484531328:

log will only return a complex result if called with a complex argument. Try log(Complex(x)).

    throw_complex_domainerror(::Symbol, ::Float64)@math.jl:33
    _log(::Float64, ::Val{:ℯ}, ::Symbol)@log.jl:301
    log@log.jl:267[inlined]
    log@math.jl:343[inlined]
    round_base@utils.jl:84[inlined]
    floor_base(::Float64, ::Float64)@utils.jl:79
    macro expansion@recipes.jl:440[inlined]
    apply_recipe(::AbstractDict{Symbol, Any}, ::Type{Val{:bar}}, ::Any, ::Any, ::Any)@RecipesBase.jl:300
    _process_seriesrecipe(::Any, ::Any)@series_recipe.jl:50
    _process_seriesrecipe(::Any, ::Any)@series_recipe.jl:62
    _process_seriesrecipes!(::Any, ::Any)@series_recipe.jl:27
    recipe_pipeline!(::Any, ::Any, ::Any)@RecipesPipeline.jl:99
    _plot!(::Plots.Plot, ::Any, ::Any)@plot.jl:223
    #plot#188@plot.jl:102[inlined]
    plot@plot.jl:93[inlined]
    #histogram#369@RecipesBase.jl:427[inlined]
    histogram@RecipesBase.jl:427[inlined]
    top-level scope@[Local: 1](http://localhost:1234/edit?id=12711bf8-0717-11ee-1283-dbe7b91dd08a#)[inlined]
```

hu-jing-ya:
Same problem

PR #4782, opened 2023-06-23
reset fillto for logscale barplots

BeastyBlacksmith:
## Description
Fix https://github.com/JuliaPlots/Plots.jl/issues/4774

@Moelf, @t-bltg this solution is a bit ugly, because it will cause bar plots with different magnitudes on different series to have different baselines, but since you don't have any information on the other series in the recipe, I don't see a clever way to do that also.

I.e.
```julia
bar([1 2 3], [0.02 125 10_000]; yscale = :log10)
```
gives
image

## Attribution


Moelf:
I feel like at some point we touched some code related to this, but this looks reasonable in my opinion

PR #4812, opened 2023-08-26
test upcoming julia `1.10`

t-bltg:
## Description

Currently at `beta2` [stage](https://github.com/JuliaLang/julia/releases/tag/v1.10.0-beta2).

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -41,9 +41,9 @@ jobs:
           - os: ubuntu-latest
             prefix: xvfb-run
             version:  '1.8' # only test intermediate release on `ubuntu` to spare resources
-          # - os: ubuntu-latest
-          #   prefix: xvfb-run
-          #   version: '~1.9.0-0'  # upcoming julia version, next `rc`
+          - os: ubuntu-latest
+            prefix: xvfb-run
+            version: '~1.10.0-0'  # upcoming julia version, next `rc`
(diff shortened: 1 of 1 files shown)


BeastyBlacksmith:
I do wonder if the "changing your default backend" stuff is holdable in 1.10

BeastyBlacksmith:
Well, I started doing that in https://github.com/BeastyBlacksmith/Plots.jl/pull/12, but its also unlikely I will be able to finish that before julia 1.10 gets released.
On the other hand it seems a bit odd to me, having to make breaking changes just to keep things working for a minor release, but I guess that ship has sailed.

axsk:
> I do wonder if the "changing your default backend" stuff is holdable in 1.10

For me `using Plots` or even precompiling it fails with setting the default backend in my startup.jl via `ENV["PLOTS_DEFAULT_BACKEND"] = "Plotly"`.

```julia
julia> using Plots
Precompiling Plots
        Info Given Plots was explicitly requested, output will be shown live 
WARNING: Method definition _plotly_framestyle(Symbol) in module Plots at /home/htc/bzfsikor/.julia/packages/Plots/sxUvK/src/backends/plotly.jl:3 overwritten on the same line (check for duplicate calls to `include`).
ERROR: LoadError: Method overwriting is not permitted during Module precompile.
```

However, it seems to work after precompiling once without the startup file...

PR #4815, opened 2023-08-29
place an upper bound fo PythonPlot

BeastyBlacksmith:
## Description

This should be a remedy to avoid segfaults and other issues when choosing pythonplot as the default backend.

xref: https://github.com/JuliaPy/PythonPlot.jl/pull/26#issuecomment-1556799332

## Attribution

## Things to consider


t-bltg:
This is not a fix but merely a workaround.

PR #4830, opened 2023-10-04
Add missing Animation constructor

agdestein:
## Description

This adds a constructor `Animation(dir)` in addition to the two existing ones `Animation()` and `Animation(dir, frames)`.
The docstring already mentions `Animation(dir)` as an option (see #4568).

## Attribution

## Things to consider

### DIFF
--- a/src/animation.jl
+++ b/src/animation.jl
@@ -13,7 +13,7 @@ struct Animation
     frames::Vector{String}
 end
 
-Animation() = Animation(convert(String, mktempdir()), String[])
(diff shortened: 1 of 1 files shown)


BeastyBlacksmith:
Would you mind adding a test?

agdestein:
If `dir` does not exist, adding a frame throws a `SystemError` (missing directory). We could alternatively create this directory in the constructor, e.g. `ispath(dir) || mkpath(dir)`. This is also done for the empty constructor with `mktempdir`.

BeastyBlacksmith:
I think the current behaviour is reasonable.

Issue #2218, opened 2019-10-11
Feature request: plot axes and ticks on all four sides


mihrits:
I would like to be able to make a plot that has (duplicated) ticks on all four sides.

Simple matplotlib example (although matplotlib in python already has axes on all four sides as default) from python3:

```python
import matplotlib.pyplot as plt

plt.bar(range(10),range(10))
plt.tick_params(top=True, right=True)
plt.show()
```
It's a requisite for plots, when submitting an article to a scientific journal in my field.

BeastyBlacksmith:
That should be `plot(1:10, 1:10, framestyle = :box)`.

mihrits:
With Julia v1.4.1 and Plots v1.0.12 `plot(1:10, 1:10, framestyle = :box)` still doesn't work for me. It does have borders on all four sides, but not the ticks.

BeastyBlacksmith:
which backend?

mihrits:
GR

BeastyBlacksmith:
I see, looking at the documentation, the only backend that gives you the ticks at the moment is `pgfplotsx`.

mihrits:
I'm sorry, I don't know how to use that. I couldn't find it in the docs either.

Haoran-Feng:
It seems [GRUtils](https://heliosdrm.github.io/GRUtils.jl/stable/) has the ticks on all four sides by default.

isentropic:
I think pyplot does this with box frame style

mihrits:
GRUtils solution seems good to me. Thanks!

BeastyBlacksmith:
Keeping open unti its fixed

orlox:
Not a proper fix, but here is a workaround I found. I just create a twin axis in each direction, match the x and y boundaries and remove labels using a copy_ticks function.
```julia
function twiny(sp::Plots.Subplot)
    sp[:top_margin] = max(sp[:top_margin], 30Plots.px)
    plot!(sp.plt, inset = (sp[:subplot_index], bbox(0,0,1,1)))
    twinsp = sp.plt.subplots[end]
    twinsp[:xaxis][:mirror] = true
    twinsp[:background_color_inside] = RGBA{Float64}(0,0,0,0)
    Plots.link_axes!(sp[:yaxis], twinsp[:yaxis])
    twinsp
end
twiny(plt::Plots.Plot = current()) = twiny(plt[1])

function copy_ticks(sp::Plots.Subplot)
[...]

PR #4832, opened 2023-10-10
Fix border ticks of framestyle :box

lch:
## Description

As mentioned in Issue #2218 , when using the GR backend with attribute `framestyle = :box`, there are no ticks presents on the opposite side of border (border depends on attribute `mirror`). And PR #4488 has not solved the problem and closed the previous issue. 

Here I'm providing another small PR trying to fix the problem, which just add missing segements when drawing the plot with attribute `framestyle = :box`.

Also mentioned by @BeastyBlacksmith in the issue, the behaviour of `:box` should be like that. However I'm not sure whether it is better to provides some attribute switches to control whether those ticks should be placed on the opposite side.

## Attribution

## Things to consider

### DIFF
--- a/src/axes.jl
+++ b/src/axes.jl
@@ -886,6 +886,21 @@ function axis_drawing_info(sp, letter)
                 grid_factor_2d[] / ax_length,
                 ax[:tick_direction] !== :none,
             )
+            if sp[:framestyle] === :box
+                add_major_or_minor_segments_2d(
+                    sp,
+                    ax,
+                    oax,
+                    reverse(oas),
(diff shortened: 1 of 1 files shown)


BeastyBlacksmith:
can you run `using JuliaFormatter; format(["src", "test"])` to get the formatting right?

oidw:
Hi, is it possible to also have the original behavior (no ticks on the opposite side of border) when setting `framestyle = :box`? 
Or, is there any other method to achieve that (figure with box frame but no ticks on top and right axes)?

PR #4849, opened 2023-12-01
Fix warning for failed integration with PlotlyBase & PlotlyKaleide

blegat:
When one of these two packages is not installed, an `ArgumentError` is thrown. As the code was doing `err isa ArgumentError || @warn ...`, nothing is warned and the user does not see any error or warning.
With the current PR, when one of the packages is not in the environment or is not installed, the correct warning is displayed.
Moreover, other errors are rethrown instead of ignored


BeastyBlacksmith:
Thanks a lot !

Issue #4855, opened 2023-12-24
[BUG]: linetype=:steppre broken with OffsetArrays


adityam:
## Details
`linetype=:steppre` does not work with OffsetArrays as shown by the following example:

```
data = OffsetArray(rand(11),-5:5)
plot(data)  # works
plot(data,linetype=:steppre) # fails
```

The error message is:

```
BoundsError: attempt to access 11-element OffsetArrays.IdOffsetRange{Int64, Base.OneTo{Int64}} with indices -5:5 at index [6]

Stacktrace:
  [1] throw_boundserror(A::OffsetArrays.IdOffsetRange{Int64, Base.OneTo{Int64}}, I::Tuple{Int64})
    @ Base ./abstractarray.jl:744
  [2] checkbounds
    @ ./abstractarray.jl:709 [inlined]
  [3] getindex
    @ ~/.julia/packages/OffsetArrays/0MOrf/src/axes.jl:224 [inlined]
  [4] make_steps(x::OffsetArrays.IdOffsetRange{Int64, Base.OneTo{Int64}}, st::Symbol, even::Bool)
    @ Plots ~/.julia/packages/Plots/sxUvK/src/recipes.jl:203
  [5] macro expansion
    @ ~/.julia/packages/Plots/sxUvK/src/recipes.jl:216 [inlined]
  [6] apply_recipe(plotattributes::AbstractDict{Symbol, Any}, #unused#::Type{Val{:steppre}}, x::Any, y::Any, z::Any)
    @ Plots ~/.julia/packages/RecipesBase/BRe07/src/RecipesBase.jl:300
  [7] _process_seriesrecipe(plt::Any, plotattributes::Any)
    @ RecipesPipeline ~/.julia/packages/RecipesPipeline/BGM3l/src/series_recipe.jl:50
  [8] _process_seriesrecipes!(plt::Any, kw_list::Any)
    @ RecipesPipeline ~/.julia/packages/RecipesPipeline/BGM3l/src/series_recipe.jl:27
  [9] recipe_pipeline!(plt::Any, plotattributes::Any, args::Any)
    @ RecipesPipeline ~/.julia/packages/RecipesPipeline/BGM3l/src/RecipesPipeline.jl:99
 [10] _plot!(plt::Plots.Plot, plotattributes::Any, args::Any)
    @ Plots ~/.julia/packages/Plots/sxUvK/src/plot.jl:223
[...]

PR #4860, opened 2024-01-04
Fix offset array hist

isentropic:
## Description
Fixing [this](https://github.com/JuliaPlots/Plots.jl/issues/4855) probably wouldn't work with like star wars arrays or some weird thing but at least it should work with linear arrays
## Attribution

## Things to consider


isentropic:
@t-bltg @BeastyBlacksmith this is probably safe to merge, but I've been so disconnected from Plots for a while, maybe something changed in the workflow

isentropic:
probably don't squash as I tried to fix python tests here too

isentropic:
well pythonplot doesn't bug out now, but something else is wrong i'm not sure what

PR #4861, opened 2024-01-04
Warn on extrakws

isentropic:
## Description
Consider a case from https://docs.juliaplots.org/stable/backends/#Fine-tuning
```
x = range(-3, 3, length=30)
surface(
  x, x, (x, y)->exp(-x^2 - y^2), c=:viridis, legend=:none,
  nx=50, ny=50, display_option=Plots.GR.OPTION_SHADED_MESH,  # <-- series[:extra_kwargs]
)
```
Upon mispelling of any kws it is interpretted as an extra_kw which leads to silent ignoring of kws:
```
# before
plot(rand(10), newweirdarg=true)
```
just plots.
Now it outputs:
```
 Warning: Extra keyword argument newweirdarg not explicitly supported with Plots.GRBackend().  Choose from: legend_hfactor, legend_wfactor, nx, ny, display_option
└ @ Plots ~/Projects/Plots.jl/src/args.jl:1564
```

Granted this is a crutch as I just look over a list of predefined kws, but in principle this mechanism could be more sophisiticated per backend. What you think?


BeastyBlacksmith:
In principle I am on board with the idea, but e.g. the PGFPlotsX backend makes deliberate use of this as in it passes all extra keywords to the backend and they usually just work. 
See #2200 for reference. So to get it right, we would need to check against every valid keyword of pgfplots (and keep it up to date).

isentropic:
We could just ignore this check for PGF plots, it could just support all of them

BeastyBlacksmith:
Yeah, if we make the check use a predicate function, that could just return true in that case

PR #4862, opened 2024-01-05
Get the plotly.min.js file from PlotlyJS.jl

fonsp:
There is code in Plots.jl to download the `plotly.min.js` file from the plotly CDN and store it in a scratch space, but the dependency PlotlyJS.jl already contains the `plotly.min.js` as one of its artifacts! This PR uses this file for the plotly backend.

This artifacts approach is very reliable (also on limited network connections, since artifacts are served by Julia's pkg server), so this PR reuses the file from PlotlyJS.jl in Plots.jl :)

Depends on https://github.com/JuliaPlots/PlotlyJS.jl/pull/480


isentropic:
Can you please add the other PR here too, and let's see the combined PR runs. Plotly needs some fixing and maybe this would fix the failing tests on windows

fonsp:
Hey! The easiest would be to wait for https://github.com/JuliaPlots/PlotlyJS.jl/pull/480 to be released, I can combine the PRs afterwards

src/init.jl
```diff
@@ -1,10 +1,10 @@
-using Scratch
 using REPL
+import PlotlyJS
```
fonsp, line 2: @isentropic It looks like this caused a test failure because PlotlyJS is not a dependency. How does that work? I thought PlotlyJS is always loaded by default?

isentropic:
probably plots gets loaded somewer here https://github.com/JuliaPlots/Plots.jl/blob/fcf2e74258e7eb7dbba36c91182d3e8aa3d7c712/src/backends.jl#L561

isentropic:
not sure I never understood why plotly has more than 1 package

src/init.jl
BeastyBlacksmith: > When I run `import Plots; plotly(); plot(1:10)` in Pluto, does that load the PlotlyJS package?

No, only `import Plots; plotlyjs(); plot(1:10)` will load PlotlyJS, `plotly()` is special in that it is the only backend that is usable without any 3rd party (julia) package, but it can only display to a web browser, while `plotlyjs` has a standalone display via Blink.jl

fonsp:
Aha! Then I think this PR only makes sense if PlotlyJS was also added as a regular dep of Plots.jl, to ensure that it is always loaded and available to have access to the artifact.

But maybe https://github.com/JuliaPlots/Plots.jl/pull/4884 is a better alternative?

PR #4863, opened 2024-01-05
Use data URL for local plotly asset

### LINKED ISSUE #4854: [BUG] offline Plot in Pluto results in empty Plot
## Details

In offline mode, the `plotly.min.js` file is being injected with a reference to the local filesystem. Since Pluto is served over `http://localhost`, the `file:///` paths can't be loaded. (Note that Pluto static exports _do_ work with the offline mode, since both the HTML file and the Plotly file are served over `file:///` and there is no protocol mismatch).

A fix could be to follow what PlutoPlotly does and inject the library files into the Pluto frontend distribution files, which are being served over http when Pluto is running:

https://github.com/JuliaPluto/PlutoPlotly.jl/blob/8c46e64be948899059722b74b101478c83326534/src/local_plotly_library.jl#L62-L69

and

https://github.com/JuliaPluto/PlutoPlotly.jl/blob/8c46e64be948899059722b74b101478c83326534/src/show.jl#L24-L32 

### Backends
Plotly

gr (default) |     |     | x 
pythonplot   |     |     | x 
plotlyjs     |  x  |     | 
pgfplotsx    |     |     | x
unicodeplots |     |     | x
inspectdr    |     |     | x 
gaston       |     |     | x 

### Versions

Plots.jl version:  Plots v1.39.0
Backend  version (`]st -m <backend(s)>`): - 
Output of `versioninfo()`:

```
julia> versioninfo()
Julia Version 1.9.4
Commit 8e5136fa29 (2023-11-14 08:46 UTC)
Build Info:
  Official https://julialang.org/ release
Platform Info:
  CPU: 8 × 11th Gen Intel(R) Core(TM) i7-1165G7 @ 2.80GHz
  WORD_SIZE: 64
  LIBM: libopenlibm
  LLVM: libLLVM-14.0.6 (ORCJIT, tigerlake)
  Threads: 1 on 8 virtual cores
Environment:
  JULIA_DEPOT_PATH = D:\.julia
```

fonsp:
Currently, when including the `plotly.min.js` locally in HTML, this is done using a `file://` URL. Importing scripts using a file URL is only allowed in browsers if the page origin is also a file URL, i.e. if you open `file://...something.html`.

This PR uses Base64 to include the script using a data URL, which is supported in all environments, including `http` and `https` origins. This means that the plotly backend can be used with `_use_local_dependencies[]` in web-based IDEs like Pluto and other web sites.

Fix #4854, depends on #4862

### DIFF
--- a/src/backends/plotly.jl
+++ b/src/backends/plotly.jl
@@ -1055,7 +1055,7 @@ html_body(plt::Plot{PlotlyBackend}) = plotly_html_body(plt)
 
 plotly_url() =
     if _use_local_dependencies[]
-        "file:///" * _plotly_local_file_path[]
+        _plotly_data_url()
     else
         "https://cdn.plot.ly/$_plotly_min_js_filename"
     end
--- a/src/init.jl
+++ b/src/init.jl
@@ -4,7 +4,7 @@ import Base64
 
 const _plotly_local_file_path = Ref{Union{Nothing,String}}(nothing)
 const _plotly_data_url_cached = Ref{Union{Nothing,String}}(nothing)
-_plotly_data_url() = 
+_plotly_data_url() =
     if _plotly_data_url_cached[] === nothing
         _plotly_data_url_cached[] = "data:text/javascript;base64,$(Base64.base64encode(read(_plotly_local_file_path)))"
     else
--- a/src/init.jl
+++ b/src/init.jl
@@ -4,7 +4,8 @@ import Base64
 
 const _plotly_local_file_path = Ref{Union{Nothing,String}}(nothing)
 const _plotly_data_url_cached = Ref{Union{Nothing,String}}(nothing)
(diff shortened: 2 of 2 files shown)


pankgeorg:
This PR doesn't depend on https://github.com/JuliaPlots/Plots.jl/pull/4862

I see https://github.com/JuliaPlots/Plots.jl/pull/4884 as a better alternative to #4862 (which would defeat the purpose of having the `plotly()` backend, but the artifact there is 110MB, which is.. not ideal for a package that already takes a long time to run.

So my suggestion is, can we move forward with this PR (which unblocks the issue with `file:///` URLS over insecure contexts, until we figure out a nice way to depend on the artifacts? (I also don't want to plagiarize this code in my other PR 😅)

@isentropic, @fonsp proactively thank you both 🙏🏽

BeastyBlacksmith:
Looks good to me and thanks for taking care 🙂

PR #4884, opened 2024-02-02
feat: use artifacts instead of PlotlyJS

pankgeorg:
## Description
Alternative to #4862, includes #4863

Replaces the download from CDN/scratchspace with an artifact


src/init.jl
```diff
@@ -1,13 +1,25 @@
-using Scratch
 using REPL
+import Base64
```
sjkelly, line 2:
```suggestion
import Base64
using Artifacts
```

sjkelly: This should fix CI: https://github.com/JuliaPlots/Plots.jl/actions/runs/7759723372/job/21164417261?pr=4884#step:10:740

pankgeorg:
> The advantage of this PR over #4862 is that this PR will always have access to the `plotly.min.js` asset, while #4862 only works when PlotlyJS is loaded by the user, right?

Yes, exactly. We also directly link to the Plotly release assets, which is also the disadvantage of this PR (the artifact being 110MB)

disberd:
@pankgeorg I created a repository last week to experiment with javascript and bundling that might be relevant for this.
The repo is at https://github.com/disberd/PlotlyArtifactsESM and it automatically bundles and release in an artifact ready format the plotly library.
If you are still pursuing this you might have a look at it and let me know if it can be used or if you see issues with its implementation

BeastyBlacksmith:
now that PlotlyKaleido hosts these as Artifacts anyway (https://github.com/JuliaPlots/PlotlyKaleido.jl/pull/27) we might be able to use these?

PR #4902, opened 2024-03-08
Added error messages in grid function for invalid values for heights and widths

### LINKED ISSUE #4212: [FR] grid function: Throw errors on bad settings
Hi! It would be great if the `grid` function errors when the heights and widths are not correctly set — so the users can understand the incorrect behavior through the error messages. In the following example, we should throw an error if one of the values is zero or if the sum is greater than one. Also, the argument should allow `PlotMeasures` units as they create error messages that are difficult to understand for newcomers. Cheers

```julia
# ╔═╡ dd200ca8-dc0a-11ec-09bc-6b12e6752417
using Plots

# ╔═╡ bb729008-f813-422c-b976-64f1a8f51195
using PlutoUI

# ╔═╡ 0022b3c5-1883-434d-9d03-31be2f5ebb0f
@bind fraction Slider(0.0:0.1:1.0, default=0.4, show_value=true)

# ╔═╡ c93e3243-b265-4c4b-8b98-4ecf011d6ba6
plot(
	plot(log), plot(exp), plot(sqrt), plot(x -> x^2),
	legend = false,
	layout = grid(2, 2, 
		heights = [fraction, 1.0 - fraction],
		widths = [fraction, 1.0 - fraction])
)
```
image

image

Gythre-9:
Fix #4212 
The `grid` function (`GridLayout`) now throws error messages when heights and widths of the grid are not correctly set. This ensures that users receive informative feedback when specifying incorrect values, making it easier to correct issues related to grid dimensions.

I added error messages for the following conditions:
- The sum of heights is not equal to 1.
- The sum of widths is not equal to 1.
- Widths values are not in the range (0, 1).
- Heights values are not in the range (0, 1).

I have tested these changes by providing incorrect values for heights and widths, ensuring that the error messages are triggered as expected. Additionally, I have confirmed that the function continues to work correctly with valid input and with no input at all (standard values).

### DIFF
--- a/src/layouts.jl
+++ b/src/layouts.jl
@@ -215,24 +215,24 @@ function GridLayout(
     heights = nothing,
     kw...,
 )
-# Check the values for heights and widths if values are provided
-if heights !== nothing && widths !== nothing
-    if sum(heights) != 1 
-    error("The sum of heights must be 1!")
-end
-if sum(widths) != 1
-    error("The sum of widths must be 1!")
-end
-if all(x -> 0 < x < 1, widths) == false 
-    error("Values for widths must be in the range (0, 1)!")
-end
-if all(x -> 0 < x < 1, heights) == false
-    error("Values for heights must be in the range (0, 1)!")
-end
-else
-    heights = zeros(dims[1])
-    widths = zeros(dims[2])
-end
+    # Check the values for heights and widths if values are provided
+    if heights !== nothing && widths !== nothing
+        if sum(heights) != 1
+            error("The sum of heights must be 1!")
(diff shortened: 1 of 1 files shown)


BeastyBlacksmith:
Don't worry the CI failures are mostly unrelated to these changes. However, please format your code with
```julia
using JuliaFormatter
format(["src", "test", "ext"], verbose=true)
```

Gythre-9:
Ok, I have formatted the code using JuliaFormatter (or at least, I think i did). Let me know if any further edits are needed!

Issue #4565, opened 2022-11-28
[FR] Plots `2.0`


t-bltg:
To many issues need stuff discussed but postponed to `2.0` because rightfully considered breaking.
Latest example at https://github.com/JuliaPlots/Plots.jl/issues/4561.

Should we branch somehow and start working on `2.0` ?

I've already closed `~80` stale issues this we, we need to go through all the remaining opened issues, close the stale ones (or won't fix because of an upstream problem, or dependencies), apply labels to start focusing on important things (like scaling, fonts, backend selection, precompilation, consistency, ...).

t-bltg:
Thanks for the comments, don't hesitate to label stuff, and adjust the list here.

BeastyBlacksmith:
I wanted to write this issue soonish anyway, so thanks for opening.

I think an important part is to define a common set of goals and non-goals we want to accomplish with this breaking release and only break things with a clear motivation.

For starters I am going to list my personal goals:

## Towards a get-API and improving attribute management in general

I started this in #https://github.com/JuliaPlots/Plots.jl/pull/3069 #2854 and then in #4489.
Apart from a few functions like `Plots.xlims` users don't have reliable way to access plot attributes, especially in recipes.
This is a missing abstraction that people circumvent by reaching into internals which makes changing this breaking.
My plan was to add a `struct` like entrypoint which will create attributes, aliases, documentation and getter-functions from one place in the source code.
Thanks to aliases that could have been non-breaking if people would newer directly read from the `plotattributes` Dictionairy in recipies ...
I also propose to have a dedicated syntax in `RecipesBase` for that (like e.g. `<--`).

## Don't export symbols that are not part of the API

I don't think that needs explanaition

## Change defaults

I was for a long time gatekeeper of defaults considering this a (visually) breaking change. This would be the time to have that never ending discussion about what the best default color palatte would be (but in a separate issue, please).

## Explicit cycling

While cycling is useful in many cases, especially in combination with layers of recipes it causes some really hard to debug issues (thinking of https://github.com/JuliaPlots/GraphRecipes.jl/issues/163 and others). I want that to be opt-in in Plots 2.0 rather than the default.

---

Thats probably not all, but the first things that came to my head.

EldarAgalarov:
Need to properly re-implement plot_title, because it completely broken: https://github.com/JuliaPlots/Plots.jl/issues/4612

isentropic:
[...]

Issue #4596, opened 2022-12-14
[FR] Harmonize how series data are used


BioTurboNick:
For 2.0, we could harmonize all series-data fields to behave the same. `x`, `y`, `z`, `marker_z`, `fill_z`, etc. should all just be treated like different dimensions and all processed the same way.

Related: #4077

Issue #4567, opened 2022-11-29
[FR] redesign backend dependencies


t-bltg:
Some nice design ideas by @mkitti in https://github.com/JuliaPlots/Plots.jl/pull/4566#pullrequestreview-1196734092:

> Perhaps we should consider a more radical redesign in the near future?

> 1. Each of the pieces of backend glue code could be independent subpackages. GRPlotsBackend, PyPlotPlotsBackend, etc.
> 2. Most of the code currently in Plots.jl gets moved to a package named AbstractPlots.jl. All the backend packages depend on AbstractPlots.jl.
> 3. Plots.jl becomes a small package which just depends on GRPlotsBackend.jl by default. Via Preferences.jl it can optionally load any of the other backend packages and act as a proxy for that package.

> The advantage of this is that each backend now can be precompiled and has its own compiled cache. Also the project dependency tree becomes a bit clearer. One could get all the features of Plots by just doing using SomePlotsBackend.jl directly.

> For loading with Requires.jl can then just do @eval using GRPlotsBackend.jl similar to what @timholy recommended here:
https://discourse.julialang.org/t/categoricalarrays-forces-openspecfun-jll-recompilation/87759/18?u=mkitti

> The impact here for this PR is still acknowledging that the GR backend is a special case for the current architecture. That is Plots.jl has declared a dependency on GR.jl. One consequence of that special case is that we load GR.jl within Plots.jl rather than Main. Eventually we move that responsibility to GRPlotsBackend.jl.

JuliaLang/julia/pull/47040 or JuliaLang/julia/pull/47695 could help too.

BeastyBlacksmith:
> 1. Each of the pieces of backend glue code could be independent subpackages. GRPlotsBackend, PyPlotPlotsBackend, etc.

Back in the days (6 years ago) that was how things were done, e.g. https://github.com/JuliaPlots/PlotsGR.jl

t-bltg:
Implemented in https://github.com/JuliaPlots/Plots.jl/pull/4904.

PR #4904, opened 2024-03-12
Plots `v2` a.k.a `2.0`

BeastyBlacksmith:
This is the development branch for Plots 2.0.

## What changed
- No default backend in `PlotsBase`, `Plots` becomes a wrapper for `PlotsBase` + `GR` (per https://github.com/JuliaPlots/Plots.jl/issues/4565#issuecomment-2004679230), making it non-breaking (should we discuss a default backend change ?);
- Backends are purely extensions with no `@require`;
- The internals are more organized into modules and moved around;
- Backend init and backend code is only found in its extensions, this should help with backend interference and annoyances about initialization and precompilation of plots. This should make backends modular and implemented much easier and without hacks into Plots.jl codebase;
- Restored `Preferences` chosen backend;

Example:
```julia
import UnicodePlots
using Plots 
unicodeplots()
plot(1:10)
```
## What works
- `PlotsBase` and `Plots` tests;
- `Preferences` based backend selection.
- Precompilation statements for Plots, using an alternate backend

## What does not work

# Help needed:
- [x] write backend agnostic precompile statements for `PlotsBase`
- [x] backport changes from v1 since last rebase
- [ ] report breakages
- [x] move documentation back into this repo
- [x] start updating the docs
- [ ] update the news.md
- [ ] register `GraphRecipes` and `StatsPlots` from this monorepo to the [`General`](https://github.com/JuliaRegistries/General) registry.

# Decisions to be made
- [x] deprecate inspectDR ?
- [ ] move to column major order ?
- [ ] change the defaults
- [ ] change the default backend ?

Issues needing attention please add more or checkoff resolved:
https://github.com/JuliaPlots/Plots.jl/issues/4565#issue-1466819814
Copy of that:
Labeled issues:
- [x] https://github.com/JuliaPlots/Plots.jl/issues/4532
- [ ] https://github.com/JuliaPlots/Plots.jl/issues/4028
- [ ] https://github.com/JuliaPlots/Plots.jl/issues/3001
- [ ] https://github.com/JuliaPlots/Plots.jl/issues/2980
- [ ] https://github.com/JuliaPlots/Plots.jl/issues/2129
- [ ] https://github.com/JuliaPlots/Plots.jl/issues/2047
- [x] https://github.com/JuliaPlots/Plots.jl/issues/4181
- [x] https://github.com/JuliaPlots/Plots.jl/issues/4513
- [x] https://github.com/JuliaPlots/Plots.jl/issues/4631
- [ ] https://github.com/JuliaPlots/Plots.jl/issues/4350
- [x] https://github.com/JuliaPlots/Plots.jl/issues/4567
- [ ] https://github.com/JuliaPlots/Plots.jl/issues/4596

Labeled PRs:
- [x] https://github.com/JuliaPlots/Plots.jl/pull/4326
- [ ] https://github.com/JuliaPlots/Plots.jl/pull/3942
- [x] https://github.com/JuliaPlots/Plots.jl/pull/3744
- [x] https://github.com/JuliaPlots/Plots.jl/pull/3732
- [ ] https://github.com/JuliaPlots/Plots.jl/pull/3473
- [ ] https://github.com/JuliaPlots/Plots.jl/pull/3357
- [x] https://github.com/JuliaPlots/Plots.jl/pull/3351
- [x] Rework adding legends #3732
---
Closes #4326


BeastyBlacksmith:
@t-bltg do you intend to release this as a 1.x release?

t-bltg:
If we can make it non breaking, it would make things easier and end endless discussions.

BeastyBlacksmith:
Well, there would be a few changes that we'd need to roll back for this, like removal of pgfplots and I removed some functions also.

But I also don't quite see how to make switching of backends non-beraking since, what was 
```julia
using Plots
pythonplot()
```
is now
```julia
using Plots
import PythonPlot
pythonplot()
```

isentropic:
I still don't get why can't 2.0 be breaking and why having a helpful error message isn't enough as a solution as per my comment in the issue #4565. Having GR as the default backend feels bad to me. I understand that nobody changes the backend but this way backend malfunctioning could be diagnosed easier.

 Then if the backend doesn't work people can learn about other backends. I really don't buy the keep gr argument by searching GitHub code. Plots 2.0 isn't just for the users but also for debugging and long term stability. 

Keeping gr as the default is not good imo. Flux got rid of it's CUDA for this reason and it works fine, new users can adapt we just need to leave good info/warn messages

isentropic:
I'm sorry that I'm late into the discussion and if some undoing needs to be done but we should prioritise long term good with 2.0. GR being the default is also what defers people from using plots, if users can learn about the flexibility that's great too. Pyplot imo works more stably with various fonts and scalings.

The purpose of package extensions is exactly this. Splitting into plotsbase and just plots to keep the standard interface means that we never need to go to 2.0 then. 

2.0 by definition is breaking and that's ok. Plots is not a low level package and hardly any packages take dep on Plots. Interactive usage is easier to adapt to 2.0 anyways in case the help messages are well stated.

We also need to think about column major order to make Plots play nicely with others and streamline the column major way. These two changes are by far the most important, both being breaking and that's ok cause it's 2.0.

isentropic:
I get why people might object with needing to import GR explicitly, but you know adding such import help long term in that people would realize that there are more backends than just GR, and that the errors are coming because of 'import GR' not 'using Plots'. **Im not kidding half of our issues are because GR didn't build or something like that.** In the current scenario users try to import Plots it errors and then there's frustration. With gr free plots GR will throw it's warnings right awa and we can throw warnings and stuff like your GR loading failed, try these other backends.

t-bltg:
> But I also don't quite see how to make switching of backends non-beraking since, what was

I've omitted this, so I've restored the 2.0 label.

> We also need to think about column major order to make Plots play nicely with others and streamline the column major way.

I get this, but if no one implements it, this shouldn't block a release.

> Having GR as the default backend feels bad to me.

I would be in favor of dropping GR as the default backend. But on the other adding `PlotsBase` without any backend dependency seems a good compromise.

In https://github.com/JuliaPlots/Plots.jl/pull/4914, I've finalized the way the modules and dependency should be written with a lot of cleanups, and restoration of the precompile statements, but unfortunately, it broke the reference tests in subtle ways.

I like the new splitting of `PlotsBase` into smaller modules initiated in this branch, but it needs more work and cleanup, without breaking the tests.

t-bltg:
I've made progress and fixed the tests in https://github.com/JuliaPlots/Plots.jl/pull/4914, everything is passing locally, and partially in CI.

Currently, there is no solution to generate precompile statements for a different backend than GR in `Plots` since we cannot `@eval Main import UnicodePlots` for example.

So we need at least to generate backend independent precompile statement for `PlotsBase`.

t-bltg:
@BeastyBlacksmith, how can we move forward now ?

Embedded modules need more cleanup, but at least the port to extensions is now complete, with 100% previous tests passing.

t-bltg:
I'd say we merge this now in `master`, create a `v1` branch for backports and move on.

We can probably ask users to try an alpha version of `Plots 2`.

BeastyBlacksmith:
> @BeastyBlacksmith, how can we move forward now ?
> 
> Embedded modules need more cleanup, but at least the port to extensions is now complete, with 100% previous tests passing.

Thanks a lot! Next steps would involve porting all the 2.0 labeled PRs and merge them into this branch. I won't merge this branch prior to the meeting at JuliaCon and even after that, there is no need to rush this. We should make sure we get all the breaking changes in that have accumulated over the years.

There will be a good amount of time before we get that oppurtunity again.

t-bltg:
Thanks, please keep in mind that I probably won't work on this anymore since my time is very limited, so make sure to backport the changes made in `master` into this branch, so that it doesn't go out of sync and stalls.

t-bltg:
Don't you think that making v2 the current master/main would force developers to work towards the next version ?

I mean we can always merge PRs into a v1 branch and tag backports releases.

Else I'm afraid given the amount of active contributors that this would be a nightmare to maintain in sync between branches ...

BeastyBlacksmith:
What I probably will do is sync them and then make `v2` the new default branch, possibly renaming `master` to `v1`. I don't think we need to merge these at all.

t-bltg:
`nightly` and `1.11.0-beta1` are passing :tada:, it has been a while ...

t-bltg:
`GraphRecipes` and `StatsPlots` are now integrated into `Plots` repo, so no more commits should be merged in https://github.com/JuliaPlots/GraphRecipes.jl or https://github.com/JuliaPlots/StatsPlots.jl.

t-bltg:
@BeastyBlacksmith, can you fix the [`DOCUMENTER_KEY`](https://github.com/JuliaPlots/Plots.jl/actions/runs/11325262164/job/31491790630?pr=4904#step:5:7972) issue so that we can have dev docs for v2 deployed ?

t-bltg:
I've change the deploydocs url to https://github.com/JuliaPlots/Plots.jl.git in https://github.com/JuliaPlots/Plots.jl/commit/a6562c14dc4e9f23b58958807f1afd19073dc8dd specifically because I was hitting a deployment failure from Documenter.

BeastyBlacksmith:
> I thought we were dropping the PlotDocs repo ?

Not entirely, it should still host the artifacts to not blow the repo size of Plots, but the generating code should be here.

t-bltg:
I see.

But still, we will hit this non-deployment criterion because of the url: https://github.com/JuliaPlots/Plots.jl/actions/runs/11317817828/job/31471690944#step:5:7953

After changing it, we hit another error, related to `DOCUMENTER_KEY`: https://github.com/JuliaPlots/Plots.jl/actions/runs/11318156624/job/31484813837#step:5:7945

t-bltg:
I think we need to follow https://documenter.juliadocs.org/stable/man/hosting/#Out-of-repo-deployment.

BeastyBlacksmith:
That key is already present. We used to trigger the build from here in v1 also

t-bltg:
4. done in https://github.com/JuliaPlots/Plots.jl/pull/4904/commits/bb14fd5d5278001e63d035c91951a6c27b1a2cf0, let's see how it goes ...

t-bltg:
As I suspected, there is still an issue with permissions, hence I think `DOCUMENTER_KEY` is the culprit here, but I have no further means to investigate:
https://github.com/JuliaPlots/Plots.jl/actions/runs/11501098562/job/32012907078?pr=4904#step:5:8035.

BeastyBlacksmith:
Works now

t-bltg:
Awesome thanks, https://docs.juliaplots.org/dev looks functional !

t-bltg:
I think the docs are now fully functional on `v2`, e.g. https://docs.juliaplots.org/dev/gallery/gr/generated/gr-ref002/#gr_ref002.

t-bltg:
@BeastyBlacksmith, would you be in favor of integration [PlotThemes](https://github.com/JuliaPlots/PlotThemes.jl) into this monorepo ?

BeastyBlacksmith:
> The docs are now fully functional on `v2`,

Thanks a lot !

> would you be in favor of integration [PlotThemes](https://github.com/JuliaPlots/PlotThemes.jl) into this monorepo ?

Yeah, thats probably a good idea

t-bltg:
CI green, finally 💚🐍🍃🌱🐸♻️☘️🟢🟩🍏🍵!

PlotsBase/src/shorthands.jl
```diff
@@ -210,15 +211,27 @@ julia> hline([-1,0,2])
...
+# a lone position is the common case, and without these a real would fall through to the
+# generic single-argument recipes: a float has none, and an integer means "n empty series"
+for func in (:hline, :vline)
+    bang = Symbol(func, :!)
+    @eval begin
+        $func(v::Real, args...; kw...) = $func([v], args...; kw...)
+        $bang(v::Real, args...; kw...) = $bang([v], args...; kw...)
+        $bang(plt::Plot, v::Real, args...; kw...) = $bang(plt, [v], args...; kw...)
+    end
+end
```
BeastyBlacksmith, line 233: note that this does not work if a recipe returns a vline, because that will hit `plot(..., seriestype = :vline)` and not these methods

t-bltg: @AJ0070

AJ0070: Confirmed. Integers are the worse half: `plot(0.73; seriestype = :vline)` errors, `plot(3; seriestype = :vline)` quietly gives three empty series. The #5788 methods sit on the `vline(...)` entry point, so anything that reaches `plot` with the seriestype already set walks straight past them.

`_series_data_vector` already gets `plotattributes`, so it knows the seriestype. A trait next to `is3d`:

```julia
takes_positions(st) = false
for st in (:hline, :vline, :hspan, :vspan)
    @eval takes_positions(::Type{Val{Symbol($(string(st)))}}) = true
end
takes_positions(st::Symbol) = takes_positions(Val{st})
takes_positions(plotattributes::AbstractDict) =
    takes_positions(get(plotattributes, :seriestype, :path))
```

then wrap in the two scalar methods:

```julia
_series_data_vector(x::Real, plotattributes) =
    takes_positions(plotattributes) ? [_prepare_series_data([x])] : [_prepare_series_data(x)]
_series_data_vector(n::Integer, plotattributes) =
    takes_positions(plotattributes) ? [_prepare_series_data([n])] : [zeros(0) for i in 1:n]
```

`plot(3)` still gives three empty series since its seriestype is `:path`, and `plot(0.73)` still errors. It covers `hspan(0.73)` too, which #5788 left out, and it makes those six shorthand methods redundant: I deleted them locally and shorthands still passes 67/67, RecipesPipeline tests too.

Downside is that it lands in RecipesPipeline rather than PlotsBase, and `plot(3; seriestype = :vline)` becomes one line instead of three empty series, which is what `vline(3)` already does. Shall I put it up?

BeastyBlacksmith: Please set it up and add tests accordingly

AJ0070: Done, #5791. Trait plus the two `_series_data_vector` methods, and the six shorthand methods from #5788 come back out since they are redundant once the pipeline handles it. Tests cover the recipe case both ways round, a float and an integer, plus the direct `plot(x; seriestype = ...)` call and that `:path` still means empty series.

PR #4922, opened 2024-04-26
 Fix downloads count badge

pricklypointer:
Fixed the download count badge using https://juliapkgstats.com/. The current badge is broken because the old API and PkgVisBoard is not maintained.

This PR:
- Replaces the broken badge with a new badge that uses the new API

Please review and provide any feedback. Thanks for considering this PR!

## Attribution

## Things to consider


t-bltg (approved):

t-bltg:
Awesome @pricklypointer thanks for your contribution!

Issue #4927, opened 2024-05-01
[FR] Reproducible output with PGFPlotsX backend


ajdunlap:
When using the PGFPlotsX backend, the output is different every time you run it even with the same code because UUIDs are used for series IDs. It would be nice if these UUIDs were generated in a non-random way so that the same input would give you the same output. (The reason I want this is that the externalize library of tikz recompiles all of my figures every time I regenerate them, even if the code used to generate them hasn't actually changed.)

BeastyBlacksmith:
Fixed in [#4932](https://github.com/JuliaPlots/Plots.jl/pull/4932)

PR #4932, opened 2024-05-06
Making PGFPlotsX use deterministic series IDs

ajdunlap:
## Description

This is a first attempt at addressing #4927 . One downside I can see is that this may impact thread-safety. It also may be a bit slower if there are many many series, but it seems like there would have to be really a lot of series for the maximum call to take a significant amount of time.

## Attribution

## Things to consider

### DIFF
--- a/src/backends/pgfplotsx.jl
+++ b/src/backends/pgfplotsx.jl
@@ -267,7 +267,7 @@ function (pgfx_plot::PGFPlotsXPlot)(plt::Plot{PGFPlotsXBackend})
             end
             for (series_index, series) in enumerate(series_list(sp))
                 # give each series a uuid for fillbetween
-                series_id = uuid4()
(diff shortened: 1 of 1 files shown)


BeastyBlacksmith:
The problem I see here is, that when you have multiple such outputs in a larger tex-file you really want the ids to be unique to the whole document and not just to a single plot.

ajdunlap:
I just tried a test file with two identical PGFPlots with identical "name path" values and it compiled correctly and didn't complain about the repeated "name path" values. Of course it also doesn't complain when there are repeated name path values in the *same* plot, but then of course the output is not correct. But if it is essential to have unique name paths even through multiple plots in the same document, then indeed something more complicated is required here.

BeastyBlacksmith (approved):
I'm okay with this then, I think this should be the only use of UUIDs in the codebase and thus we can remove this dependency then.
Can you also open a PR with this against the `v2` branch?

Issue #4202, opened 2022-05-15
[BUG] Grid covering axes


putianyi889:
## Details
```julia
plot(1:5,1:5,1:5, xlims=(1,5), ylims=(1,5), zlims=(1,5), framestyle=:box,foreground_color_grid=:red, gridalpha=1, thickness_scaling=5, color=:transparent)
```
on pgfplotsx():

![download](https://user-images.githubusercontent.com/44583944/168468230-b0ada899-3a43-4c15-953f-63eb066c43f4.png)

### Backends

This bug occurs on ( insert `x` below )

Backend      | yes | no  | untested
-------------|-----|-----|---------
gr (default) |  x |     |
pyplot       |     |     |x
plotlyjs     |     |     |x
pgfplotsx    |   x |     |
unicodeplots |     |     |x
inspectdr    |     |     |x
gaston       |     |     |x

### Versions

Plots.jl version: 1.29.0
Backend  version (`]st -m <backend(s)>`): GR 0.64.2, PGFPlotsX 1.5.0
Output of `versioninfo()`:
```
Julia Version 1.7.2
Commit bf53498635 (2022-02-06 15:21 UTC)
Platform Info:
  OS: Windows (x86_64-w64-mingw32)
  CPU: AMD Ryzen 7 4800H with Radeon Graphics
  WORD_SIZE: 64
  LIBM: libopenlibm
  LLVM: libLLVM-12.0.1 (ORCJIT, znver2)
Environment:
  JULIA_EDITOR = code
  JULIA_NUM_THREADS = 16
```

PR #4945, opened 2024-06-18
GR: draw gridlines below all axes

dd0:
## Description

GR currently plots all components of the x-axis first, and then the y-axis (and possibly z). Since horizontal gridlines belong to the y-axis, if there's a tick at the low end of `xrange`, the corresponding gridline will cover the x-axis (see issue https://github.com/JuliaPlots/Plots.jl/issues/4202).

This fixes the issue for GR by splitting `gr_draw_axis` into three stages, so that we first draw all minor gridlines, then all major gridlines, and the rest of the axis on top.

With the change, a few image tests would need to be updated, in which the x-axis colour is now no longer affected by the (transparent) overlapping gridline -- as far as I can see, no visible changes. I assume that this is done separately in https://github.com/JuliaPlots/PlotReferenceImages.jl/?

## Attribution

## Things to consider

### DIFF
--- a/src/backends/gr.jl
+++ b/src/backends/gr.jl
@@ -1407,6 +1407,8 @@ function gr_draw_axes(sp, vp)
         x_bg, y_bg = RecipesPipeline.unzip(GR.wc3towc.(area_x, area_y, area_z))
         GR.fillarea(x_bg, y_bg)
 
+        foreach(letter -> gr_draw_axis_minorgrid_3d(sp, letter, vp), (:x, :y, :z))
+        foreach(letter -> gr_draw_axis_grid_3d(sp, letter, vp), (:x, :y, :z))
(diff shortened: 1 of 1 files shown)


BeastyBlacksmith:
> I assume that this is done separately in [JuliaPlots/PlotReferenceImages.jl](https://github.com/JuliaPlots/PlotReferenceImages.jl/)?

Yes, that is correct

BeastyBlacksmith:
I set that up for you in https://github.com/JuliaPlots/PlotReferenceImages.jl/pull/151
Can you open a second PR to the v2 branch with the same changes?

dd0:
Sorry about taking so long to respond -- I now made a second PR with essentially the same changes for v2.

BeastyBlacksmith:
Thanks! Consider adding your name to our `.zenodo.json` file (see #3503 for reference)

PR #4955, opened 2024-07-03
lazily load Pkg

KristofferC:
This is a bit of a hack to work around the fact that depending on Pkg is quite expensive in 1.11. Here, we only load it when checking the compatibility / install status of a backend and also skipping that when the backend is GR.jl (since we know that is compatible due to resolving together with Plots).

I'll admit this is kind of ugly and a bit of "TTFP hacking". However, I don't think Plots.jl should depend on a package manager at all so 🤷.

TTFP change 2.1s to 1.7s


KristofferC:
PlotsBase does the same: https://github.com/JuliaPlots/Plots.jl/blob/f32247a819bc938a133bb3403770814194821408/PlotsBase/src/backends.jl#L45

isentropic:
I dont think this function is needed at all now https://github.com/JuliaPlots/Plots.jl/blob/f32247a819bc938a133bb3403770814194821408/PlotsBase/src/backends.jl#L27
but lets wait for @BeastyBlacksmith to confirm

BeastyBlacksmith:
Thats true, that code can go in `v2`. However for the 1.x series we would need to up the julia compat, since `pkgversion` requires julia 1.9+

KristofferC:
> since pkgversion requires julia 1.9+

It is easy to read the project file of the package and extract the version. I just didn't remember when `pkgversion` was introduced.

KristofferC:
(Looking forward to v2 which seems to have a great design avoiding these "hacks")

PR #4963, opened 2024-07-13
Add docs folder

BeastyBlacksmith:
## Description

This adds the docs folder formerly on PlotDocs.jl here, so we have the generating code more easily in sync with the codebase.

The documentation still should be deployed to PlotDocs.jl, se we don't increase the repository size to an unreasonable amount.

TODO:

- [ ] update source to be compatible with Documenter 1.x 

## Attribution
- [x] am listed in [.zenodo.json](https://github.com/JuliaPlots/Plots.jl/blob/2463eb9f8065c52ed8314f6e541664c5b9db88d2/.zenodo.json) (see https://github.com/JuliaPlots/Plots.jl/issues/3503)


isentropic:
Could you clarify a bit more why is this needed? I do not understand that much

BeastyBlacksmith:
It makes contributing and updating the documentation easier when you can do it in the same PR

isentropic:
Oh just for 2.0 right?

BeastyBlacksmith:
thats right

Issue #4961, opened 2024-07-13
[FR] Show legend by default only if labels have been set


mgkuhn:
At the moment the default is to always show a legend, i.e.`legend=true`, even if no label was provided by the user. The default label "y1" is not really useful, especially not if there is just a single series. As a result, users have to add `legend=false` very frequently.

For v2 it would be nice if the legend is only shown by default if `label` values have been provided, especially if there is only a single series.

Current result:
`plot(sin)` looks like `plot(sin; legend=true, label=("y1"))`

Expected result:
`plot(sin)` should look like `plot(sin; legend=false)`
`plot(sin; label=("sin"))` should look like `plot(sin; legend=true, label=("sin"))`

PR #4964, opened 2024-07-18
do not plot legend upon single series

isentropic:
Stolen from https://github.com/JuliaPlots/Plots.jl/pull/3732/files

Basically do not insert label in case there is only single series being plotted. For me it feels like a good default. Legend would show in case you provide a custom label for a single series it of course. Having `y1` to be displayed is surely weird if it is by its own


isentropic:
I'll fix the tests, once we agree if it goes into 2.0 or not

BeastyBlacksmith:
I'm in favor

isentropic:
there is a small problem here, `plot(rand(10), label="y1")` would still suppress the label. Basically the current logic just checks if there is single series && label="y1" which is okay practically, but maybe someone really wants to display "y1" as a single series.

I propose adding a new keyword `smart_label=true` to have this behavior active. Otherwise I can't see how we can unhack this or return to previous behavior for compatibility reasons

BeastyBlacksmith:
> if there is single series && label="y1"

Yeah, I would prefer to set the label to `""` if there is only a single series or to set `legend_position = :none`.

isentropic:
https://github.com/JuliaPlots/Plots.jl/issues/4961

isentropic:
I dived into the processing pipeline and how things are handled, but there doesn't seem to be a way to introduce this functionality without adding a new keyword without hacks. The problem this in this statement
> Yeah, I would prefer to set the label to "" if there is only a single series or to set legend_position = :none.
The pipeline updates the legend_position before processing series. So we can't set `legend_position= :none` because series haven't been processed, and there is no way to know how many series are present.

Another way is to do this after series are processed. Now that all series are processed, the labels are already set to `"y1","y2"` and legend_position is set to `:best`. At this point we can't know if plots set the labels to `y1,y2` or the user set it like in the first place.

Another way might be while series are processed. But there is problem in that it is not easy to know how many more series are to come, that is it is kinda impossible to tell if there will be only one series in the subplot.

A dirty solution is to make a new arg `auto_labelling=false, true, :smart` true would correspond to the current behavior, false would completely disable y1,y2,y3... smart will correspond to autolabelling in case there is more than one series. Having this arg, we still can't tell if the user put that `y1` label there or not, but the user can enable `y1` label by setting `auto_labelling=true`

BeastyBlacksmith:
> Another solution is just to disable this auto labelling into `y1,y2,y3` I'm not sure how is that even useful in the first place.

Thats fine with me too

isentropic:
Well, I simplified the change to `label=:none` by default. So there is no longer weird numbering like `y1,y2...` but users can reenable it by `label=:auto` to restore the past behavior. Now, there is simply no label by default

isentropic:
Gotta say now that there is no default label, all images look much cleaner. Remind me what is the policy about making reference images. I'm on a mac, and my ref images look always different even for the ones that don't change. How should I handle this

isentropic:
I have a feeling that we should disable reference image checking in macos too. And maybe for the purposed of a cleaner transition maybe I should embed PlotsRefImages repo inside here the same way you wanted to that with PlotsDocs? @BeastyBlacksmith

BeastyBlacksmith:
> PlotsRefImages repo

Thats pretty big. But it wouldn't be impossible to automate the creation of the ref images and opening of the pull request via CI, that would also ensure that its built on the same architecture

Issue #3001, opened 2020-09-22
[FR] Safe figure saving


isentropic:
Perhaps `savefig` should be safe by default and error upon already existing file, and have a `unsafe tag`.
This has hurt me multiple with multiple libraries. In scientific context it is better not to override by default I think. This is breaking, but it should be better long term.

t-bltg:
I'm skeptical about this. I believe this should be the other way around (if ever implemented), `savefig` with an optional `safe=false` keyword, defaulting to overwriting the files.

isentropic:
Maybe we can reopen this @t-bltg ?
Unsafe saving has burned all of us at some point. Maybe we discuss this with the community>

oameye:
I am with @isentropic on this one.

PR #4965, opened 2024-07-18
Safe figure saving by default, + new arg

isentropic:
Like I suggested before in https://github.com/JuliaPlots/Plots.jl/issues/3001


isentropic:
```julia
julia> p = plot(rand(10), safe_saving=true);

julia> savefig(p, "this.png")
"/Users/z/Projects/Plots.jl/this.png"

julia> savefig(p, "this.png")
┌ Warning: Filename /Users/z/Projects/Plots.jl/this.png already exists, defaulting to prevent overriding. To disable this behavior, provide `:safe_saving=false` kwarg, i.e. `plot(rand(10), safe_saving=false)`
└ @ PlotsBase ~/Projects/Plots.jl/PlotsBase/src/output.jl:142

julia> p = plot(rand(10), safe_saving=false);

julia> savefig(p, "this.png")
"/Users/z/Projects/Plots.jl/this.png"
```

isentropic:
I'll fix the tests, once we agree if it goes into 2.0 or not

BeastyBlacksmith:
> I'll fix the tests, once we agree if it goes into 2.0 or not. I was also thinking instead of making this a property of a plot, maybe just make it a flag in `savefig` feature, but then maybe someone would want this unsafe behavior by default.

I would be in favor of making it a part of `savefig`, same goes for `tex_output_standalone`. If changing the default is the main concern we could make it a [`ScopedValue`](https://github.com/vchuravy/ScopedValues.jl)

isentropic:
I can't understand how scopedvalues can help here. Can you give an example please?

BeastyBlacksmith:
```julia
julia> using ScopedValues

julia> const DEFAULT_OVERWRITE = ScopedValue(false)
ScopedValue{Bool}(0)

julia> savefig(;overwrite = DEFAULT_OVERWRITE) = @show overwrite[]
savefig (generic function with 1 method)

julia> savefig()
overwrite[] = false
false

julia> with(DEFAULT_OVERWRITE => true) do
         savefig()
       end
overwrite[] = true
true
```

isentropic:
I see, I meant about script usage. Probably some people have their research pipelines such that there is some saving done inside. I was thinking that a user would exclude the flag in Prefernces.jl or startup.jl
Now that I'm looking into these, where are the docs about setting the defaults? I can't seem to find the docs about default setting anymore

BeastyBlacksmith:
I think its the tip on the bottom of this page: https://docs.juliaplots.org/stable/install/

isentropic:
So you still think ScopedValue is the right solution here?

BeastyBlacksmith:
I do. I mean, you can still do something like
```julia
const DEFAULT_OVERWRITE = ScopedValue(get(PLOTS_DEFAULTS, :overwrite, false))
```
To get change it based on some value in `startup.jl`.

Issue #4962, opened 2024-07-13
[FR] Sticks with arrow heads


mgkuhn:
I currently use the following workaround to add arrow heads to sticks plots (e.g. to signify Dirac deltas in signal-processing figures):
```
uparrow   = Shape([(-1.5,-3), (0,0), (1.5,-3), (0,0), (-1.5,-3)])
downarrow = Shape(uparrow.x, -uparrow.y)

data = [2, 1, 3, -1, 0, -0.0]

sticks(data;
       marker=[isless(x,0.0) ? downarrow : uparrow for x in data],
       markerstrokecolor=:auto)
```

I wonder if it would be possible to
- [ ] add `:uparrow` as a predefined marker shape
- [ ] cause `sticks` to flip the marker shape vertically if the data value is less than `0.0`, in order to cause shapes like uparrow to point downwards (where `-0.0` should also flip downwards)
- [ ] cause `markercolor=nothing` to cause the marker not to be filled

It seems `Shape` was mainly intended for closed polygons (documented?). At least in GR the polygon is closed to be filled, even if the stroke around it is not closed. Since it doesn't currently seem possible to disable filling of the marker, I had to close the arrowhead stroke by going backwards on the same path. Ideally, it should be possible to use an unfilled open shape `uparrow = Shape([(-1.5,-3), (0,0), (1.5,-3)])` instead.

mgkuhn:
`markercolor=nothing` still does not yet cause markers to not be filled (without using transparency).

In pgfplots, a hack is to use `extra_kwargs=Dict(:series => Dict("mark" => "o"))` to switch to unfilled markers,

PR #4977, opened 2024-09-04
Adding uparrow as a predefined marker and creating unfilled markershapes

leckerbeon:
## Description
Adds `:uparrow` as a predefined markershape and causes `markercolor=nothing` to create an unfilled open shape as requested in #4962. 
The changes also make it possible to create unfilled markers of every predefined markershape and also add an `:downarrow` shape. 
Unfortunately I couldn´t quite figure out how to make sticks flip the marker shape vertically if the data value is less or equal to `0.0.`
## Attribution
- [X] I am listed in [.zenodo.json](https://github.com/JuliaPlots/Plots.jl/blob/2463eb9f8065c52ed8314f6e541664c5b9db88d2/.zenodo.json) (see https://github.com/JuliaPlots/Plots.jl/issues/3503)

## Things to consider

### DIFF
--- a/PlotsBase/src/utils.jl
+++ b/PlotsBase/src/utils.jl
@@ -103,19 +103,14 @@ function _update_series_attributes!(plotattributes::AKW, plt::Plot, sp::Subplot)
     end
 
     # update markerstrokecolor
-    if plotattributes[:markerstrokecolor] ≡ :match
-        stroke_color = plotattributes[:seriescolor]
+    plotattributes[:markerstrokecolor] = if plotattributes[:markerstrokecolor] ≡ :match
+        get_series_color(plotattributes[:seriescolor], sp, plotIndex, stype)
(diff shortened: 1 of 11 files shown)


PlotsBase/src/utils.jl
```diff
@@ -104,13 +104,13 @@ function _update_series_attributes!(plotattributes::AKW, plt::Plot, sp::Subplot)
 
     # update markerstrokecolor
     plotattributes[:markerstrokecolor] = if plotattributes[:markerstrokecolor] ≡ :match
-        plot_color(sp[:foreground_color_subplot])
+        get_series_color(plotattributes[:seriescolor], sp, plotIndex, stype)
```
BeastyBlacksmith, line 107: This causes the markerstroke to always be the same as the markercolor. I don't think we want this. E.g.
```julia
using Plots
scatter(rand(10))
```
now looks like this
image

Issue #5009, opened 2024-10-31
Incorrect Behavior of Bats under (:permute)


Jonas-a-Zimmermann:
## Details
Using `permute` on the `Bar` recipe leads to incorrectly set `xlims`. This is due to the recipe changing the limits of the axis before the axis are permuted (hence, e.g. for a histogram of data in the range(-10,10); the lower limit is set to -10; Further downstream the limits of the axis are only further expanded, leading to (-10; max(bins)), where max(bins) is the maximum height of the bins.

See also #5007 
### Backends

This bug occurs on ( insert `x` below )

Backend      | yes | no  | untested
-------------|-----|-----|---------
gr (default) |   x |     |
pythonplot   |     |     |x
plotlyjs     |     |     |x
pgfplotsx    |     |     |x
unicodeplots |     |     |x
inspectdr    |     |     |x
gaston       |     |     |x

### Versions

Plots.jl version:
Backend  version (`]st -m <backend(s)>`):
Output of `versioninfo()`:

BeastyBlacksmith:
Related to #4949

Issue #4949, opened 2024-06-20
[BUG] When using permute = (:x, :y) on a bar plot the bars are not automatically centered


georgemetzger:
## Details

When plotting a horizontal bar plot with permute the plot is not centered; the bars run to the edges and bar_edges does not seem to have any effect. Plots can be adjusted with ylims! but this would become tedious when refreshing data or making new plots.

![plot](https://github.com/JuliaPlots/Plots.jl/assets/18706316/491fa6e8-8c3b-4ee9-a049-d4d4d352ba7e)

### Backends

This bug occurs on ( insert `x` below )

Backend      | yes | no  | untested
-------------|-----|-----|---------
gr (default) |   x  |     |
pythonplot   |     |     | x
plotlyjs     |  x   |     | 
pgfplotsx    |     |     | x
unicodeplots |     |     | x
inspectdr    |     |     | x
gaston       |     |     | x

### Versions

Plots.jl version: v1.40.4
Backend  version (`]st -m <backend(s)>`):
Output of `versioninfo()`:
Julia Version 1.10.3
Commit 0b4590a550 (2024-04-30 10:59 UTC)
Build Info:
  Official https://julialang.org/ release
Platform Info:
  OS: Windows (x86_64-w64-mingw32)
  CPU: 20 × 13th Gen Intel(R) Core(TM) i7-1370P
  WORD_SIZE: 64
  LIBM: libopenlibm
  LLVM: libLLVM-15.0.7 (ORCJIT, goldmont)
Threads: 1 default, 0 interactive, 1 GC (on 20 virtual cores)
Environment:
  JULIA_EDITOR = code
  JULIA_NUM_THREADS =

PR #5010, opened 2024-10-31
Incorrect Behavior of Bars under (:permute) -> remove the expand_extrema call

Jonas-a-Zimmermann:
## Description
Removed the `expand_extrema!` call in the `bar` recipe, as this leads to an issue with the permute option; 
see https://github.com/JuliaPlots/Plots.jl/issues/5009

## Attribution
- [x] I am listed in [.zenodo.json](https://github.com/JuliaPlots/Plots.jl/blob/2463eb9f8065c52ed8314f6e541664c5b9db88d2/.zenodo.json) (see https://github.com/JuliaPlots/Plots.jl/issues/3503) (Part of this PR)

### DIFF
--- a/.zenodo.json
+++ b/.zenodo.json
@@ -779,6 +779,11 @@
         {
             "name": "Patrick Jaap",
             "type": "Other"
+        },
+        {   
+            "affiliation": "Ludwig-Maximilians-Universiaet Muenchen",
+            "name": "Jonas Zimmermann",
+            "type": "Other"
         }
     ],
     "upload_type": "software"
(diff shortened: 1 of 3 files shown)


Jonas-a-Zimmermann:
I suppose you could move the code from https://github.com/JuliaPlots/Plots.jl/pull/4768/commits/e786ff5de674de40d5235de0009b839d73fc8b0a into the bar recipe itself, however i do net see the benefit of that.
(well i suppose it could solve the widening isssue #4949 aswell)

However, (I think) the usage of `expand_extrema` is in the best case redundant (it is and has to be called downstream, to ensure linking axis works), and in the worst case errors like these happen, thus just not using it within recipes seems like the proper way to me.
However, do note that i am not **that** familiar with the codebase.

PR #5024, opened 2024-11-12
check version in ci - fix precompilation

t-bltg:
- fix alternative backend precompilation statements.;
- reduce number of `using`;
- move dependencies to package extensions (e.g. `PGFPlotsX`).


timholy:
I suspect that you'll want to change the order of these statements:
https://github.com/JuliaPlots/Plots.jl/blob/55e84dad62a2622962e975fbe1f562b27bdc4a2d/PlotsBase/src/init.jl#L178-L179

The problem is that the thunk will be compiled before PrecompileTools can turn on tracing. See https://github.com/JuliaLang/PrecompileTools.jl/pull/53 and https://github.com/JuliaLang/PrecompileTools.jl/issues/58

PR #5027, opened 2024-11-16
set offsets to 0

BeastyBlacksmith:
## Description

Closes #3942 

Ports the changes from #3942, didn't check yet if we need further modifications

## Attribution

## Things to consider

### DIFF
--- a/PlotsBase/ext/GRExt.jl
+++ b/PlotsBase/ext/GRExt.jl
@@ -979,10 +979,10 @@ function PlotsBase._update_min_padding!(sp::Subplot{GRBackend})
 
     # Add margin given by the user
     padding = (
-        left = Ref(2mm + sp[:left_margin]),
-        top = Ref(2mm + sp[:top_margin]),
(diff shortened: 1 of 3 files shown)


PlotsBase/ext/GRExt.jl
```diff
@@ -894,7 +894,10 @@ function gr_set_tickfont(sp, letter::Symbol; kw...)
...
 end
 
-# size of the text with no rotation
+"""
+Size of the text with no rotation.
+Returns `(width, height)`.
+"""
 function gr_text_size(str)
     GR.savestate()
     GR.selntran(0)
```
BeastyBlacksmith, line 903: @jheinen can you give what the numbers returned by `GR.inqtext` mean in terms of the size of the displayed area?
Do they have a unit?
Further down I try to add padding for the labels and somehow I need all kinds of correction factors to make it look good, but I am not sure that it scales well when I change the size of the plot.

PR #5058, opened 2025-03-17
add df macro

### LINKED ISSUE #3336: Move df macro here?

BeastyBlacksmith:
## Description

fix #3336

Closes #3351.

## Attribution

## Things to consider


mkborregaard:
yay! 🎉

Issue #5062, opened 2025-03-19
[BUG] `zscale` is not working for surface plots


JakubVanek:
## Details

The GR integration code currently crashes when attempting to use a logarithmic Z axis scale on a surface plot.

MWE:
```julia
using Plots
Qvel = 10 .^ range(start=-7.0, stop=+3.0, step=0.1)
Qacc = 10 .^ range(start=-7.0, stop=+3.0, step=0.1)
pos_mse = ones(length(Qvel), length(Qacc))
plot(Qvel, Qacc, pos_mse, st=:surface, xscale=:log10, yscale=:log10, zscale=:log10)
```

<details>
<summary>Julia error and stack trace</summary>

```
Error showing value of type Plots.Plot{Plots.GRBackend}:
ERROR: MethodError: no method matching getindex(::@NamedTuple{ln::Int64, log2::Int64, log10::Int64}, ::Bool)
The function `getindex` exists, but no method is defined for this combination of argument types.

Closest candidates are:
  getindex(::NamedTuple, ::AbstractVector{Symbol})
   @ Base namedtuple.jl:171
  getindex(::NamedTuple, ::Tuple{Vararg{Symbol}})
   @ Base namedtuple.jl:170
  getindex(::NamedTuple, ::Colon)
   @ Base namedtuple.jl:169
  ...

Stacktrace:
  [1] gr_set_window(sp::Plots.Subplot{Plots.GRBackend}, vp::Plots.GRViewport{Float64})
    @ Plots ~/.julia/packages/Plots/Ec1L1/src/backends/gr.jl:1374
  [2] gr_display(sp::Plots.Subplot{Plots.GRBackend}, w::Measures.AbsoluteLength, h::Measures.AbsoluteLength, vp_canvas::Plots.GRViewport{Float64})
    @ Plots ~/.julia/packages/Plots/Ec1L1/src/backends/gr.jl:948
  [3] (::Plots.var"#508#509"{Int64, Int64, Plots.GRViewport{Float64}})(sp::Plots.Subplot{Plots.GRBackend})
    @ Plots ~/.julia/packages/Plots/Ec1L1/src/backends/gr.jl:688
  [4] foreach(f::Plots.var"#508#509"{Int64, Int64, Plots.GRViewport{Float64}}, itr::Vector{Plots.Subplot})
    @ Base ./abstractarray.jl:3187
[...]

PR #5063, opened 2025-03-19
GR: Fix zscale support in surface plots

JakubVanek:
## Description

Previously, `zscale` was being set to a Bool value (the result of the set membership check). This caused a crash when querying the `gr_z_log_scales` tuple on the next line, because that array is indexed by symbols like :log10.

Given that the lines above assign the actual symbols to `xscale` and `yscale`, it seems likely that `zscale` should behave the same.

Fixes: https://github.com/JuliaPlots/Plots.jl/issues/5062

## Attribution

## Things to consider


BeastyBlacksmith:
Nice catch! Thank you

PR #5066, opened 2025-03-26
add macros that turn unit from cm/mm/in to px

kaikuhn:
## Description
It can be useful to be able to define the size of plots in other units.
Therefor these macros convert the tuple of size in px before the size tuple is evaluated by the function.

## Attribution

## Things to consider
It shouldn't affect anything since the macros are executed before the tuple is transferred to the plot function.

## Considerations
I have seen that other values like left_margin are implemented by the Measure package.
To apply this method to the plot size the changes would have more impact, because the pixels unit needs to be recalculated for the selected DPI value after receipt.

### DIFF
--- a/PlotsBase/src/Commons/Commons.jl
+++ b/PlotsBase/src/Commons/Commons.jl
@@ -44,6 +44,7 @@ export width, height, leftpad, toppad, bottompad, rightpad
 export origin, left, right, bottom, top, bbox, bbox!
 export DEFAULT_BBOX, DEFAULT_MINPAD, DEFAULT_LINEWIDTH
 export MM_PER_PX, MM_PER_INCH, DPI, PX_PER_INCH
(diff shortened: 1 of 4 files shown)


BeastyBlacksmith:
Why do these need to be macros and what benefit do they have over the converter functions we already have further down in the code?

kaimacuser:
> Why do these need to be macros and what benefit do they have over the converter functions we already have further down in the code?

Hello [BeastyBlacksmith](https://github.com/BeastyBlacksmith),

Thank you for your message.
It needn't to be macros. It was just a suggestion for an easy way of setting the figure size without changing anything else in the code.
As I mentioned it would be nice to have a similar syntax for setting absolute length as it is handled for e.g left_margin using the Measure package (It is the same syntax it is in pyplot).
Is there anything planned in this direction?

PR #5068, opened 2025-04-03
Fix plot append!

### LINKED ISSUE #4825: [FR] `append!` cannot accept collections (as would be expected from the name)
The signature of the `append!` method through which all others pass accepts only `Real...`, meaning it cannot actually add multiple datapoints to a series. It is my understanding that in Julia, `append!` is generally meant to be the bulk equivalent of `push!`, but because of this signature, `append!` for plots is just a less flexible version of `push!`. 

The problematic method definition: https://github.com/JuliaPlots/Plots.jl/blob/7cd27c470e10d8364ab935935eb58a97eae55ce3/src/utils.jl#L810C1-L810C90
image

If I understand correctly, just making the type of `args` less restrictive, to include vectors, would fix this.
The signature of the previous method should probably also be changed for consistency but that would be a breaking change.

Ouroboros42:
Fixes #4825

## Description
Allow appending vectors of data to plot, using a syntax consistent with `push!`.

### DIFF
--- a/PlotsBase/src/Plots.jl
+++ b/PlotsBase/src/Plots.jl
@@ -86,6 +86,8 @@ Base.iterate(plt::Plot) = iterate(plt.subplots)
 
 Base.push!(plt::Plot, args::Real...) = push!(plt, 1, args...)
 Base.push!(plt::Plot, i::Integer, args::Real...) = push!(plt.series_list[i], args...)
+Base.append!(plt::Plot, args::Real...) = append!(plt, 1, args...)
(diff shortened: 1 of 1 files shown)


AJ0070:
This looks superseded by #5796, which landed the same `AbstractVector...` methods along with the `extend_series_data!` fix for range-backed series and a guard for the ambiguity Aqua flagged. @Ouroboros42 got to the method signatures first here.

Ouroboros42:
Awesome!

PR #5072, opened 2025-04-15
interpolate scratch_dir to avoid precompilation failure in 1.12

lmiq:
## Description

Interpolating the `scratch_dir` variable in the precompilation directive solves the precompilation failure observed in v1.12-beta1. 

## Attribution

## Things to consider


lmiq:
I think it would be nice, if this is a good solution, to make a 1.40.13 release with this as soon as possible, because otherwise CI runs will fail for all packages in Julia 1.12 pre-releases that depend on Plots.jl (on extensions, for example).

BeastyBlacksmith:
Thanks! Would be nice, if you also could port that to the v2 branch.

PR #5091, opened 2025-05-11
rework `Project.toml` using `[sources]` - add `test/Project.toml`

t-bltg:
## Description
Rework `Project.toml` using `[sources]` and split test dependencies to an added new file `test/Project.toml`.

This has the benefit of getting rid of the infamous `[targets]` section in `Project.toml`.


BeastyBlacksmith:
Does that break usage on julia 1.10 (which doesn't have `[sources]` AFAIK)?
If so, you would want to update the julia compat entry.

t-bltg:
It does not since CI is passing and we test using Julia lts.

Issue #5093, opened 2025-05-27
[BUG] UnitfulExt problems on v2 branch


Ickaser:
## Details

1. On the current v2 branch, to load the Unitful extension (and therefore recipes for Unitful types), UnitfulLatexify has to be manually imported: 
 https://github.com/JuliaPlots/Plots.jl/blob/4d289ba9c9dc1cd09a0abd942e2b201802fc2d93/PlotsBase/Project.toml#L75
(By importing UnitfulLatexify, Latexify and LaTeXStrings are also loaded and the extension is activated.)
It looks like UnitfulLatexify is used to make the extension work with PGFPlotsX.
I see why these dependencies were added to the extension, and it is likely that many users would want UnitfulLatexify anyway. I got caught off guard, though, when wanted to start a pull request related to the Unitful extension and had to look at the source to figure out I needed UnitfulLatexify. If this is deliberately a breaking change for 2.0 and that will be documented, I think I can understand some of the motivation, but wanted to be sure this isn't an accident.

2. The extension checks if the current backend is PGFPlotsX, and if PGFPlotsX isn't loaded (or something, I don't have it installed) these lines error out:
https://github.com/JuliaPlots/Plots.jl/blob/4d289ba9c9dc1cd09a0abd942e2b201802fc2d93/PlotsBase/ext/UnitfulExt.jl#L236
https://github.com/JuliaPlots/Plots.jl/blob/4d289ba9c9dc1cd09a0abd942e2b201802fc2d93/PlotsBase/ext/UnitfulExt.jl#L244

### Backends

This bug occurs on ( insert `x` below )

Backend      | yes | no  | untested
-------------|-----|-----|---------
gr (default) |     |     |
pythonplot   |     |     |
plotlyjs     |     |     |
pgfplotsx    |     |     |
unicodeplots |     |     |
inspectdr    |     |     |
gaston       |     |     |

### Versions

Plots.jl version: current v2 branch, commit 4d289ba9c9dc1cd09a0abd942e2b201802fc2d93
Backend  version (`]st -m <backend(s)>`):
Output of `versioninfo()`:

t-bltg:
Yeah, reworking `UnitfulExt` for clarity and consistency with the available backends is more than welcome (I'm too short on time).

Ickaser:
This isn't fully resolved yet

PR #5094, opened 2025-05-27
UnitfulExt: Check if PGFPlotsX is available, then check if it's the current backend

Ickaser:
## Description

Fix part of #5093.

## Attribution
(as of now)

## Things to consider

### DIFF
--- a/.zenodo.json
+++ b/.zenodo.json
@@ -797,6 +797,12 @@
         {
             "name": "Finn Eisenach",
             "type": "Other"
+        },
+        {
+            "affiliation": "Purdue University",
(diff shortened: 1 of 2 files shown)


Ickaser:
> Does `backend_name() === :pgfplotsx` work?

That seems to work, and looks a little cleaner to me--going with that for now.

Ickaser:
The CI failures here seem unrelated to me--GR reference image 62 is failing on everything but Windows, see https://github.com/JuliaPlots/PlotReferenceImages.jl/pull/154/files.

t-bltg:
Thanks !

Issue #4822, opened 2023-09-12
[BUG][UnitfulExt] Setting label on axis with unitful limits errors


skleinbo:
# MWE

```julia
using Plots, Unitful

julia> y(t) = 1/2*t^2;

julia> plot(y, xlims=(0u"s", 5u"s"))

julia> xlabel!("Time")
Error showing value of type Plots.Plot{Plots.GRBackend}:
ERROR: DimensionError:  and s are not dimensionally compatible.
```

<details><summary>Full stacktrace</summary>

```julia
Error showing value of type Plots.Plot{Plots.GRBackend}:
ERROR: DimensionError:  and s are not dimensionally compatible.
Stacktrace:
  [1] #s103#141
    @ ~/.julia/packages/Unitful/J4AJj/src/conversion.jl:7 [inlined]
  [2] var"#s103#141"(::Any, s::Any, t::Any)
    @ Unitful ./none:0
  [3] (::Core.GeneratedFunctionStub)(::Any, ::Vararg{Any})
    @ Core ./boot.jl:602
  [4] uconvert(a::Unitful.FreeUnits{(), NoDims, nothing}, x::Quantity{Int64, 𝐓, Unitful.FreeUnits{(s,), 𝐓, nothing}})
    @ Unitful ~/.julia/packages/Unitful/J4AJj/src/conversion.jl:72
  [5] ustrip
    @ ~/.julia/packages/Unitful/J4AJj/src/utils.jl:35 [inlined]
  [6] _ustrip(u::Unitful.FreeUnits{(), NoDims, nothing}, x::Quantity{Int64, 𝐓, Unitful.FreeUnits{(s,), 𝐓, nothing}})
    @ UnitfulExt ~/.julia/packages/Plots/sxUvK/ext/UnitfulExt.jl:333
  [7] _broadcast_getindex_evalf
    @ ./broadcast.jl:683 [inlined]
  [8] _broadcast_getindex
    @ ./broadcast.jl:656 [inlined]
  [9] #31
    @ ./broadcast.jl:1088 [inlined]
 [10] ntuple
    @ ./ntuple.jl:49 [inlined]
[...]

PR #5095, opened 2025-05-27
Add `unit` axis attribute (for v2)

### LINKED ISSUE #4750: [BUG] DimensionError (units do not match) generated when using twinx() to plot data with Unitful units
## Details
When trying to plot two data series with different Unitful units on the same plot but on seperate vertical axes, using `twinx()`, an error is returned informing the user that the units are not compatible with each other. Please see MWE 1 for an example.

My suspicion is that the check applied to ensure that the units are compatible when plotting Unitful data happens before a check to see if the to-be-plotted series is on a new, or different axis. The 'are the units compatible' check still needs to occur, though, on the common axis (in this case the horizontal x-axis).

By modifying the units of the second data series to be 'compatible' with the first data series, as shown in MWE 2, then the  bug report #4741 is reproduced, where the plot is created successfully but the secondary y-axis label is missing.

I had a look into the stack trace from the error, and it seemed to my uninformed eyes like here was the first possible candidate:
```
 [16] macro expansion
    @ ~/.julia/packages/RecipesPipeline/BGM3l/src/user_recipe.jl:131 [inlined]
```
which links to this line of code:
[131] `did_replace |= y !== (newy = _apply_type_recipe(plotattributes, y, :y))`

Thank you - please let me know if I can help more diagnosing the problem here, I'm not very skilled or experienced, but I'm enthusastic to try to be helpful.

Joe
_____

### MWE 1 - Dimension error
```
using Plots
using Unitful

gr()

x = [t for t in 0:1:10].*u"s"
y1 = [v for v in 10:-1:0].*u"V"
y2 = [i for i in 0:1:10].*u"A"

plot(x, y1; label = "Left axis data - units = $(unit(eltype(y1)))")
twinx()
plot!(x, y2; label = "Right axis data - units = $(unit(eltype(y2)))") # DimensionError generated when calling this line
```

### MWE 2 - Dimension error gone, bug report #4741 reproduced
```
using Plots
using Unitful

gr()

x = [t for t in 0:1:10].*u"s"
y1 = [v for v in 10:-1:0].*u"V"
y2 = [i for i in 0:1:10].*u"mV".*1000

plot(x, y1; label = "Left axis data - units = $(unit(eltype(y1)))")
twinx()
plot!(x, y2; label = "Right axis data - units = $(unit(eltype(y2)))") # y-axis unit label now missing
```

### Backends

gr (default) |  x   |     |
pythonplot   |  x   |     |
plotlyjs     |   x  |     |
pgfplotsx    |     |     | x
unicodeplots |  x   |     |
inspectdr    |  x   |     |
gaston       |     |     | x

### Versions

Plots.jl version:  [91a5bcdd] Plots v1.38.12
Unitful.jl version: [1986cc42] Unitful v1.14.0
Backend  versions (`]st -m <backend(s)>`): 
    - GR:  [28b8d3ca] GR v0.72.5
    - pythonplot: [274fc56d] PythonPlot v1.0.2
    - plotlyjs: [f0f68f2c] PlotlyJS v0.18.10
    - unicodeplots: [b8865327] UnicodePlots v3.5.3
    - inspectdr: [d0351b0e] InspectDR v0.4.3

Output of `versioninfo()`:
```
julia> versioninfo()
Julia Version 1.9.0
Commit 8e630552924 (2023-05-07 11:25 UTC)
Platform Info:
  OS: macOS (x86_64-apple-darwin22.4.0)
  CPU: 4 × Intel(R) Core(TM) i5-5250U CPU @ 1.60GHz
  WORD_SIZE: 64
  LIBM: libopenlibm
  LLVM: libLLVM-14.0.6 (ORCJIT, broadwell)
  Threads: 1 on 4 virtual cores
Environment:
  JULIA_EDITOR = code
  JULIA_NUM_THREADS = 
  JULIA_IMAGE_THREADS = 1
```

Ickaser:
## Description

In an attempt to fix #4750 and #4822 , following the suggestion in https://github.com/JuliaPlots/Plots.jl/issues/4822#issuecomment-1732512054 to move units from the label (so a string) to an actual axis attribute.

This currently generates correct plots and solves the above issues, but the axis label behavior is probably still a little wonky--e.g., units may get added to an axis label after the user sets the label not to have them. That would be an important part of making this work properly, I think, so I'll mark this "draft" for now, but I would welcome suggestions about how that should behave: for example, should `unitformat` always be used, and users have to set `unitformat` to `:nothing` if they don't want the units printed? Or perhaps respect the `ProtectedString` API? In the past I found the `ProtectedString` to be a little annoying but haven't thought yet about how it will interact.

The current code is almost certainly not as elegant as it should be, and once I have an idea of desired label behavior there will probably be some pieces leftover that can be deleted.

## MWE for testing

```julia
using Unitful, UnitfulLatexify # UnitfulLatexify currently needed to trigger extension, see #5093 
using Plots

x = range(1u"s", 10u"s", length = 100)
y = @. x * 5u"m/s" 

plot(u"hr", u"m") 
plot!(x, y)
plot!(xlabel="time")
plot!(x, -y)
pl2 = twinx()
plot!(pl2, x, 1 ./y)
plot!(pl2, ylabel="1/distance")
```

## Attribution
 (assuming #5094  merges)

## Things to consider


Ickaser:
@gustaphe should probably weigh in on this, I suspect

Ickaser:
> I am generally in favor, have not checked the details yet. I can't even run the code, something about `key :pgfplotsx not found` on line 240 of UnitfulExt, but that may be down to some misconfiguration on this end. 

That's a two-line fix in #5094 😅, I left that as a separate PR because it seemed like it would need less discussion.

> The _elegant_ way to do this would be to have axis properties `unit`, `guide` and `unitformat` which are stored separately and spit out a string at the point of display. No need for a `ProtectedString` if you can just set your `unitformatter` to `(a,b) -> string(a)` or something. The somewhat less elegant way is to have `unit`, `guide_string` and `unitformat`, and then updating `guide` every time one of them changes. I guess the current status is this second option. Workable.

The _current_ current depends whether a plot call happened last, so I need to change at least a bit. Possibly I could split into two PRs, one with a less-elegant fix for 1.x and a better one for 2.0, depending how easy it is for me to figure this out.

Ickaser:
I believe all the tests are passing now. To make that so, I had to do a couple things that may merit discussion.
- Move `ProtectedString` into `PlotsBase`, rather than the extension. 2.0 might be the most natural place to drop this API, so if desired, that could be implemented now or moved to a second PR so this can be backported to 1.x. 
- If `""` is passed as a guide label, old API does not add units. Seems a reasonable expectation, so if `""` is passed to a guide, set `axis[:unitformat]=:none` as a side effect, achieving the desired result. This could potentially be annoying if someone passes a guide string later on and expects the unit to come back. An alternative might be to wrap the empty label in a ProtectedString, if we keep that API.
- The PGFPlotsX unit handling requires Latexify, so I refer to the import in the PGFPlotsX extension.

gustaphe:
Failing CI notwithstanding, I like this a lot. I think it would be good to get the opinion of someone else (github suggests @t-bltg), because I have a quite Unitful focused view and this might in theory make handling axis guides slightly more cumbersome for someone who does not.

* If we intend to remove `ProtectedString` in 2.0, you can put `Base.depwarn` in the constructor. I think it's a good idea.
* I never liked this behavior, but inertia is a good thing (My preferred way to say "no label at all" is `guide=nothing`, which iiuc does the right thing even without changing `unitformat`). This way of doing it is acceptable (for now).
* Yeah, that makes sense.

t-bltg:
~~Please fix formatting using https://docs.juliaplots.org/stable/contributing/#Write-code,-and-format.~~

I also think it's a good idea to implement `get_guide`.
I don't have time to review this thoroughly though.
It would be great to check manually with a scenario, that all backends behave the same and don't regress with units.

t-bltg:
@Ickaser,do you think this PR is ready ?
Is it fixing https://github.com/JuliaPlots/Plots.jl/issues/5093 or not ?

Ickaser:
> @Ickaser,do you think this PR is ready ? Is it fixing #5093 or not ?

tldr; The bulk of the changes are here and decisions are made, but there are some details for me to work out still.

TODO:
- [x] I added some tests in the v1 version (#5098)  that aren't passing yet, once they are I will put them in this PR as well. 
- [x] If @gustaphe is OK (which I think he is) this PR will also remove the `P_str` macro and ProtectedString API, which means adding a `depwarn` to the v1 PR and making a note in whatever 2.0 update guide exists.
- [x] Put the UnitfulExt docs changes into this PR for v2. 
- [ ] Clarify the Latexify dependency (esp. if we can drop UnitfulLatexify, though that may wait for a future PR in Unitful)
- [x] Change the `unitformat=:none` and `unitformat=false` to put no unit on axis label

Ickaser:
I actually hadn't yet put a couple of the last changes from #5098 into this PR 😅 that probably wasn't obvious though

(commits https://github.com/JuliaPlots/Plots.jl/pull/5098/commits/21ecaadec4528e1f33031745f11d6f6dad7183ff, https://github.com/JuliaPlots/Plots.jl/pull/5098/commits/0f88e9edf8ff8d2c2ef94f44ddcfd257f126d3fa, https://github.com/JuliaPlots/Plots.jl/pull/5098/commits/7875bc440982e0042e6eba58b03021d96b5a8ab6 )

I can open another, smaller PR with those changes.

(That said, those last things were all to do with Latexify. Notably, the remaining problem for #5093 is that to load `UnitfulExt`, Unitful and UnitfulLatexify (two distinct packages) both need to be imported. UnitfulLatexify is only needed for the PGFPlotsX backend, which has Latexify as a dependency, so my hope is that by the time v2 is ready for release, we can have UnitfulLatexify converted to an extension that will automatically be loaded.)

BeastyBlacksmith:
Sorry, if I pulled the trigger too fast here, but indeed, the rest can go in a separate PR and it is not like the release is around the corner 😉

Issue #4947, opened 2024-06-19
[BUG] Plot annotations in awkward location


orswan:
## Details
When plotting images and adding annotations, the location of the annotation can end up in weird places with respect to the image. Here's a basic example: 
```
using Plots, Images
imgs = [zeros(Gray{Float64},64,128) for i=1:4]
plotParams = Dict(
    :ann => (:bottomleft,:auto),
    :annotationcolor => :blue,
    :size => (800,400),
    :layout => grid(1,4),
    :grid => false)
p = plot([plot(i',aspect_ratio=1,axis=false,size=(800,400)) for i in imgs]...; plotParams...)
```
Output: 
![annotation_test_2024-6-19](https://github.com/JuliaPlots/Plots.jl/assets/17443294/56112dc2-25be-4410-ba74-de6454904b7b)

### Backends

This bug occurs on ( insert `x` below )

Backend      | yes | no  | untested
-------------|-----|-----|---------
gr (default) |   x  |     |
pythonplot   |     |     |x
plotlyjs     |     |     |x
pgfplotsx    |     |     |x
unicodeplots |     |     |x
inspectdr    |     |     |x
gaston       |     |     |x

### Versions

Plots.jl version: v1.40.4
Backend  version (`]st -m <backend(s)>`): v0.72.8

Output of `versioninfo()`:
```
Julia Version 1.10.3
Commit 0b4590a550 (2024-04-30 10:59 UTC)
Build Info:
  Official https://julialang.org/ release
Platform Info:
  OS: Windows (x86_64-w64-mingw32)
  CPU: 12 × 12th Gen Intel(R) Core(TM) i5-12450H
  WORD_SIZE: 64
  LIBM: libopenlibm
  LLVM: libLLVM-15.0.7 (ORCJIT, alderlake)
Threads: 1 default, 0 interactive, 1 GC (on 12 virtual cores)
```

BeastyBlacksmith:
Fixed by https://github.com/JuliaPlots/Plots.jl/pull/5098

Issue #5035, opened 2024-11-28
[BUG] DimensionError: K and NaN are not dimensionally compatible.


Eben60:
could be related to https://github.com/JuliaPlots/Plots.jl/issues/4750

## Details
Errors on trying to plot `Vector` of `Vector`s, wherein data are `Unitful`, but some are `missing`. 

The error message mentions `NaN`s, however there were no `NaN`s in the data. 

### MWE

```
using DataFrames, Plots, Unitful
gr() # or other backend

y1 = fill(32.0u"°C", 5)
y2 = vcat(fill(missing, 2), fill(22.0u"°C", 3))
df = DataFrame((; y1, y2))
tcl = eachcol(df) |> collect
m2 = Matrix(df)

p0 = plot([y1, y2]); display(p0); # OK
p1 = plot(m2); display(p1); # OK.  As my data coming from a DataFrame, this is a usable alternative for me.
p2 = plot([tcl[1], tcl[2]]); display(p2); # OK, but similar code errored in my actual use case
p2a = plot(tcl[1]); plot!(tcl[2]); display(p2a); # OK, but similar code errored in my actual use case
p3 = plot(tcl); # errors
```
### Error stack
```
ERROR: DimensionError: K and NaN are not dimensionally compatible.
Stacktrace:
  [1] convert(::Type{Quantity{Float64, 𝚯, Unitful.FreeUnits{(K,), 𝚯, nothing}}}, x::Float64)
    @ Unitful ~/.julia/packages/Unitful/GYzMo/src/conversion.jl:106
  [2] setindex!(A::Vector{Quantity{Float64, 𝚯, Unitful.FreeUnits{(K,), 𝚯, nothing}}}, x::Float64, i::Int64)
    @ Base ./array.jl:976
  [3] macro expansion
    @ ./broadcast.jl:968 [inlined]
  [4] macro expansion
    @ ./simdloop.jl:77 [inlined]
  [5] copyto!
    @ ./broadcast.jl:967 [inlined]
  [6] copyto!
    @ ./broadcast.jl:920 [inlined]
  [7] materialize!
    @ ./broadcast.jl:878 [inlined]
  [8] materialize!
    @ ./broadcast.jl:875 [inlined]
  [9] broadcast!(f::RecipesPipeline.var"#13#14", dest::Vector{Quantity{…}}, As::Vector{Union{…}})
    @ Base.Broadcast ./broadcast.jl:844
 [10] _prepare_series_data
[...]

Issue #4793, opened 2023-07-23
Unexpected behaviour when combining an empty 3D plot with units and a contour plot [BUG]


fhagemann:
When I try to set units of the axes BEFORE plotting a heatmap, I can do so by running
```julia
using Plots, Unitful
x = (1:10)u"m"
y = (1:10)u"m"
z = ustrip.(x .* y')
plot(u"mm", u"mm")
heatmap!(x, y, z)
```
![image](https://github.com/JuliaPlots/Plots.jl/assets/30291312/717ddbf7-9185-4854-b084-1bda11796a14)

However, if z also has units, the same syntax results in a `DimensionError` (as expected):
```julia
using Plots, Unitful
x = (1:10)u"m"
y = (1:10)u"m"
z = x .* y'
plot(u"mm", u"mm")
plot!(x, y, z)
```
```DimensionError:  and m^2 are not dimensionally compatible.```

For me, a very intuitive way to fix this would be to add a third unit to the first plot command
```julia
using Plots, Unitful
x = (1:10)u"m"
y = (1:10)u"m"
z = x .* y'
plot(u"mm", u"mm", u"mm^2")
plot!(x, y, z)
```
This, however, results in a plot that looks like a 3D plot behind the contour plot
## gr
![image](https://github.com/JuliaPlots/Plots.jl/assets/30291312/4be25261-d6b9-4b02-ae1a-563c71449b22)
## pythonplot
![image](https://github.com/JuliaPlots/Plots.jl/assets/30291312/ae7d522a-de8e-4f43-a994-6018047f827f)
## plotlyjs
![image](https://github.com/JuliaPlots/Plots.jl/assets/30291312/2fbc282b-a848-4670-b7f3-b386a6ab1287)
## pgfplotsx
-- No plot --

I am aware of the keywords `xunit` and `yunit` which allow to obtain the desired plot
```julia
using Plots, Unitful
x = (1:10)u"m"
y = (1:10)u"m"
z = x .* y'
plot(x, y, z, xunit = u"mm", yunit = u"mm", zunit = u"mm^2")
```
![image](https://github.com/JuliaPlots/Plots.jl/assets/30291312/6dfa0fc5-6c1e-4ef8-8477-a6d204da13bc)

but is there a way to make also the above-listed syntax work?

### Backends

This bug occurs on ( insert `x` below )

Backend      | yes | no  | untested
-------------|-----|-----|---------
gr (default) |  x  |     |
[...]

PR #5098, opened 2025-06-06
Add `unit` axis attribute, for Plots v1

Ickaser:
## Description

Essentially identical to #5095 , but made against the `master` branch so that I'm not waiting for Plots v2--folder structures are different enough that it made sense to open as a separate PR.

@gustaphe and I discussed some in #5095; it would be helpful to get feedback from @t-bltg or @BeastyBlacksmith or someone else who is not as focused on Unitful usage.

A lot of the noise in the diff is a result of switching to `Plots.get_guide(axis)` instead of `axis[:guide]`, because that new function allows us to more elegantly handle differing unit formats in the guide strings; in any case where the axis has no units, the result of `get_guide(axis)` is `axis[:guide]` so not a big change in behavior.

Other changes involve moving some of the machinery out of the UnitfulExt so that it is accessible within Plots for axis handling.

## Attribution

## Things to consider

### DIFF
--- a/ext/UnitfulExt.jl
+++ b/ext/UnitfulExt.jl
@@ -35,16 +35,17 @@ function fixaxis!(attr, x, axisletter)
     err = Symbol(axisletter, :error)       # xerror, yerror, zerror
     axisunit = Symbol(axisletter, :unit)   # xunit, yunit, zunit
     axis = Symbol(axisletter, :axis)       # xaxis, yaxis, zaxis
-    u = pop!(attr, axisunit, _unit(eltype(x)))  # get the unit
-    # if the subplot already exists with data, get its unit
+    u = get!(attr, axisunit, _unit(eltype(x)))  # get the unit
+    # if the subplot already exists with data, use that unit instead
     sp = get(attr, :subplot, 1)
     if sp ≤ length(attr[:plot_object]) && attr[:plot_object].n > 0
-        label = attr[:plot_object][sp][axis][:guide]
-        u = getaxisunit(label)
-        get!(attr, axislabel, label)  # if label was not given as an argument, reuse
+        spu = getaxisunit(attr[:plot_object][sp][axis])
+        if !isnothing(spu)
+            u = spu
+        end
+        attr[axisunit] = u  # update the unit in the attributes
     end
     # fix the attributes: labels, lims, ticks, marker/line stuff, etc.
-    append_unit_if_needed!(attr, axislabel, u)
     ustripattribute!(attr, err, u)
     if axisletter === :y
         ustripattribute!(attr, :ribbon, u)
@@ -159,8 +160,12 @@ function fixmarkercolor!(attr)
     ustripattribute!(attr, :clims, u)
     u == NoUnits || append_unit_if_needed!(attr, :colorbar_title, u)
 end
+function fixlinecolor!(attr)
+    u = ustripattribute!(attr, :line_z)
+    ustripattribute!(attr, :clims, u)
+    u == NoUnits || append_unit_if_needed!(attr, :colorbar_title, u)
+end
 fixmarkersize!(attr) = ustripattribute!(attr, :markersize)
-fixlinecolor!(attr) = ustripattribute!(attr, :line_z)
 
 # strip unit from attribute[key]
 ustripattribute!(attr, key) =
@@ -188,36 +193,25 @@ end
 Label string containing unit information
 =======================================#
 
-abstract type AbstractProtectedString <: AbstractString end
-struct ProtectedString{S} <: AbstractProtectedString
-    content::S
-end
-struct UnitfulString{S,U} <: AbstractProtectedString
+const APS = Plots.AbstractProtectedString
+struct UnitfulString{S,U} <: APS
     content::S
     unit::U
 end
-# Minimum required AbstractString interface to work with Plots
-const S = AbstractProtectedString
-Base.iterate(n::S) = iterate(n.content)
-Base.iterate(n::S, i::Integer) = iterate(n.content, i)
-Base.codeunit(n::S) = codeunit(n.content)
-Base.ncodeunits(n::S) = ncodeunits(n.content)
-Base.isvalid(n::S, i::Integer) = isvalid(n.content, i)
-Base.pointer(n::S) = pointer(n.content)
-Base.pointer(n::S, i::Integer) = pointer(n.content, i)
-
-Plots.protectedstring(s) = ProtectedString(s)
 
 #=====================================
 Append unit to labels when appropriate
+This is needed for colorbars, mostly,
+since axes have their own handling
 =====================================#
 
 append_unit_if_needed!(attr, key, u) =
     append_unit_if_needed!(attr, key, get(attr, key, nothing), u)
 # dispatch on the type of `label`
-append_unit_if_needed!(attr, key, label::ProtectedString, u) = nothing
+append_unit_if_needed!(attr, key, label::Plots.ProtectedString, u) = nothing
 append_unit_if_needed!(attr, key, label::UnitfulString, u) = nothing
 function append_unit_if_needed!(attr, key, label::Nothing, u)
+    @info "append unit to nothing" key label u
     attr[key] = if attr[:plot_object].backend == Plots.PGFPlotsXBackend()
         UnitfulString(LaTeXString(latexify(u)), u)
     else
@@ -225,11 +219,12 @@ function append_unit_if_needed!(attr, key, label::Nothing, u)
     end
 end
 function append_unit_if_needed!(attr, key, label::S, u) where {S<:AbstractString}
+    @info "append unit to label" key label u
     isempty(label) && return attr[key] = UnitfulString(label, u)
     if attr[:plot_object].backend == Plots.PGFPlotsXBackend()
         attr[key] = UnitfulString(
             LaTeXString(
-                format_unit_label(
+                Plots.format_unit_label(
                     label,
                     latexify(u),
                     get(attr, Symbol(get(attr, :letter, ""), :unitformat), :round),
@@ -240,7 +235,7 @@ function append_unit_if_needed!(attr, key, label::S, u) where {S<:AbstractString
     else
         attr[key] = UnitfulString(
             S(
-                format_unit_label(
+                Plots.format_unit_label(
                     label,
                     u,
                     get(attr, Symbol(get(attr, :letter, ""), :unitformat), :round),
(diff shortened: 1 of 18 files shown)


t-bltg:
CI is failing on this one.

Ickaser:
This PR isn't quite ready yet, because I think it deserves a few more tests for the UnitfulExt before merging, plus some additions to the docs. But @t-bltg the CI failures seem to be either 1) a resurface of a PlotlyJS>Blink>Electron issue https://github.com/JuliaPlots/Plots.jl/pull/5075, which is randomly hitting only some of the Ubuntu test cases or 2) something to do with the 1.12 prerelease

t-bltg:
> 1) a resurface of a PlotlyJS>Blink>Electron issue https://github.com/JuliaPlots/Plots.jl/pull/5075

https://github.com/JuliaPlots/Plots.jl/pull/5100

> 2) something to do with the 1.12 prerelease

Ignore those (they are development aids), we focus on julia `release` and `lts`.

Ickaser:
Progress:

- [x] Tests added and passing
- [x] ProtectedString has a `Base.depwarn`
- [x] Added to the UnitfulExt docs 
- [x] Double check that UnitfulExt docs are correct and have everything new

Ickaser:
@gustaphe, I realized today that I was assuming the `:none` unitformat should put no unit in the axis label, where the current behavior is more like "no special symbols on the unit". Changing `unitformat=:none` from "no symbols with unit" to "no unit" is breaking, so think I will give it a new name like `unitformat=:nounit` at least for v1. But since we are making breaking changes in v2, what do you think of changing the behavior of `unitformat=:none` to print no unit at all?

Edit: `unitformat=false` follows the same code path, and I think I would also expect that to print no unit, so maybe change that at the same time.

t-bltg:
@Ickaser can you post mwe code and resulting images in your PR description, so we can see in a glance what changes ?

Ickaser:
> @Ickaser can you post mwe code and resulting images in your PR description, so we can see in a glance what changes ?

Sure, I've added that to the PR description and linked to the relevant issues.

gustaphe:
> I was assuming the `:none` unitformat should put no unit in the axis label

Not an unfair assumption, that would probably have been a better design choice.

Ickaser:
I couldn't resist trying to solve everything with this PR, so now...
- I reworked the marker_z and line_z handling, so that values on the colorbar have dimensional correctness enforced; `zunit` and `zunitformat` also get used to generate a colorbar title. This improves the situation for #4793, although the three-argument `plot(u"m", u"m", u"m")` still generates a 3D plot rather than allowing a heatmap to get put on top.
- Fixes #4947.
- I checked #5035, and although it isn't solved here, in my testing it *does* work in v2 (i.e. #5095). Not sure why exactly.

The docs changes associated with this PR are in https://github.com/JuliaPlots/PlotDocs.jl/pull/360.

With that, I _think_ this PR (for v1) is ready (although review is appreciated; I can look over it again myself in a couple days with fresh eyes if no one else can). #5095 (for v2) I still need to work out how to deal with the UnitfulLatexify dependency, but I believe it is up to date with this otherwise.

(Sorry for scattering comments and conversation across two PRs at once!)

t-bltg:
> The docs changes associated with this PR are in JuliaPlots/PlotDocs.jl#360.

The docs now live here in the Plots monorepo, https://github.com/JuliaPlots/Plots.jl/tree/v2/docs, can you reopen the PR here ?

@BeastyBlacksmith in order not to confuse contributors (since the warning in the README.md is barely noticable), can you please archive the following :
- https://github.com/JuliaPlots/PlotDocs.jl;
- https://github.com/JuliaPlots/PlotThemes.jl ?

Should we also archive these too :
- https://github.com/JuliaPlots/GraphRecipes.jl;
- https://github.com/JuliaPlots/StatsPlots.jl ?

BeastyBlacksmith:
We do need all of these for 1.x releases or don't we?

t-bltg:
> We do need all of these for 1.x releases or don't we?

You are right.

ext/UnitfulExt.jl
```diff
@@ -22,38 +22,38 @@ Main recipe
...
     if sp ≤ length(attr[:plot_object]) && attr[:plot_object].n > 0
-        label = attr[:plot_object][sp][axis][:guide]
-        u = getaxisunit(label)
-        get!(attr, axislabel, label)  # if label was not given as an argument, reuse
+        spu = getaxisunit(attr[:plot_object][sp][axis])
+        if !isnothing(spu)
+            u = spu
+        else # Subplot exists but doesn't have a unit yet
+            u = get!(attr, axisunit, _unit(eltype(x)))  # get the unit
+        end
```
t-bltg, line 43:
```suggestion
        spu = getaxisunit(attr[:plot_object][sp][axis])
        u = if isnothing(spu)  # subplot exists but doesn't have a unit yet
            get!(attr, axisunit, _unit(eltype(x)))  # get the unit
        else
            spu 
        end
```

Ickaser:
> > We do need all of these for 1.x releases or don't we?
> 
> You are right.

I think I did it right then--for 1.x, the docs changes are in https://github.com/JuliaPlots/PlotDocs.jl/pull/360, and for 2.0(#5095), the docs changes are in the single PR to the monorepo.

ext/UnitfulExt.jl
```diff
@@ -154,13 +154,32 @@ function fixaspectratio!(attr, u, axisletter)
...
+    # Then to an existing subplot's colorbar title
+    elseif sp ≤ length(attr[:plot_object]) && attr[:plot_object].n > 0
+        cbar_title = get(attr[:plot_object][sp], :colorbar_title, nothing)
+        spu = (cbar_title isa UnitfulString ? cbar_title.unit : nothing)
+        if !isnothing(spu)
+            u = spu
+            ustripattribute!(attr, key, u)
+        else
+            u = ustripattribute!(attr, key)
+        end
```
t-bltg, line 172:
```suggestion
        spu = cbar_title isa UnitfulString ? cbar_title.unit : nothing
        u = if isnothing(spu)
            ustripattribute!(attr, key)
        else
            ustripattribute!(attr, key, spu)
        end
```

src/axes.jl
```diff
@@ -111,6 +120,57 @@ end
...
+        return axis[:guide]
+    else
+        ustr = if Plots.backend_name() ≡ :pgfplotsx
+            Latexify.latexify(axis[:unit])
+        else
+            string(axis[:unit])
+        end
+        if isempty(axis[:guide])
+            return ustr
+        end
```
t-bltg, line 137:
```suggestion
        isempty(axis[:guide]) && return ustr
```

t-bltg (commented):
LGTM

gustaphe:
I think this is generally good, but I'm hunting down a regression. The second argument to `format_unit_label(l, u)` is not supposed to be a string (if the axis unit is not, I guess). It matters, because `latexify` has a two-argument method intended specifically for formatting labels:
```julia
plot(rand(5)*u"m"; yguide="x", unitformat=latexify)
``` 
should have the yguide `$x\;/\;\mathrm{m}$` (`== latexify("x", u"m")`), but with this PR it errors. I will suggest a few lines of edits that will fix this.

src/axes.jl
```diff
@@ -111,6 +120,55 @@ end
...
+        ustr = if Plots.backend_name() ≡ :pgfplotsx
+            Latexify.latexify(axis[:unit])
+        else
+            string(axis[:unit])
+        end
+        isempty(axis[:guide]) && return ustr
+        return format_unit_label(
+            axis[:guide],
+            ustr,
+            axis[:unitformat])
```
gustaphe, line 139:
```suggestion
        isempty(axis[:guide]) && return string(axis[:unit])
        return format_unit_label(
            axis[:guide],
            axis[:unit],
            axis[:unitformat])
```

gustaphe: This together with somehow setting the default unitformat for pgfplotsx to `(l, u)->"\$$l\\;\\left($(Latexify.latexify(u; env=:raw))\\right)\$"` solves my problem. I had success putting `default(unitformat=(l, u)->"\$$l\\;\\left($(Latexify.latexify(u; env=:raw))\\right)\$")` at line 1608 of `src/backends.jl` (inside `_initialize_backend(pkg::PGFPLotsXBackend)`), but I don't know if that's legal.

Ickaser:
> I think this is generally good, but I'm hunting down a regression. The second argument to `format_unit_label(l, u)` is not supposed to be a string (if the axis unit is not, I guess). It matters, because `latexify` has a two-argument method intended specifically for formatting labels:
> 
> ```julia
> plot(rand(5)*u"m"; yguide="x", unitformat=latexify)
> ```
> 
> should have the yguide `$x\;/\;\mathrm{m}$` (`== latexify("x", u"m")`), but with this PR it errors. I will suggest a few lines of edits that will fix this.

Sure enough, in `Plots.get_guide` (src/axes.jl, line 130 or thereabouts) I have the unit getting preemptively converted to a string, so that if we are in PGFPlotsX the unit is in LaTeX-friendly format. So a fix can start there

src/axes.jl
Ickaser: I would like to find a solution that doesn't involve setting a default in the backend, if I can?... So far the best idea I am coming up with is that PGFPlotsX should have its own function (or set of methods) for `format_unit_label(l, u)`, that puts `\\left` and `\\right` on all the characters.

That, or for all the cases *except* `unitformat <: Function`, `latexify` gets called on the unit before getting passed to `format_unit_label`. That way if `unitformat=latexify` gets passed, latexify gets to see the unit object itself, but if any of the defaults are used then things still work.

Increasingly I find myself of the opinion that in v2 it might be worth dropping most of the methods of `format_unit_label`--unless I'm using `:round` or `:square`, I almost always end up writing a function myself, and I think it's not too complicated an operation for most users. (Probably not more complicated than understanding how many symbols they want to pass for the current methods, at least.)

Ickaser: I went for the `unitformat isa Function` check and am letting tests run now: if this works in your cases, I might like this solution best since it doesn't involve setting any defaults, but if the resulting code is unclear I am open to other suggestions (like your original one).

Ickaser:
The latest CI failures are either 1) failure to precompile Gtk.jl on Windows (seems random: affected 1.10 and 1.11, not 1.6, and worked the last time CI ran) and 2) downstream GraphRecipes failing on a Julia type tree (which fails on 1.11 and 1.12 across branches, it seems).

I added one (1) test with latexify to the test suite, and it seems to work now, so if that's enough (and latexify solution looks good to @gustaphe) I think this is good to go.

src/axes.jl
```diff
@@ -127,15 +127,17 @@ function get_guide(axis::Axis)
         axis[:unitformat] == :nounit 
         return axis[:guide]
     else
-        ustr = if Plots.backend_name() ≡ :pgfplotsx
+        unit = if axis[:unitformat] isa Function
+            unit = axis[:unit]
```
gustaphe, line 131: One of these `unit = ` is superfluous. I would personally prefer keeping the second one and adding one at line 133 but it appears `= if` is pretty common in the rest of the code base so it's up to you :D

gustaphe:
Otherwise LGTM. Good work!

Issue #4777, opened 2023-06-14
[BUG] Failure to plot `TimeType`s as series


omus:
## Details

Plotting time types as a series results a failure:

```julia
julia> using Plots, Dates

julia> gr()
Plots.GRBackend()

julia> plot([1, 2], [1, 1])

julia> plot([[1, 2]], [[1, 1]])

julia> plot([DateTime(2001), DateTime(2002)], [1, 1])

julia> plot([[DateTime(2001), DateTime(2002)]], [[1, 1]])
ERROR: Cannot convert DateTime to series data for plotting
Stacktrace:
  [1] error(s::String)
    @ Base ./error.jl:35
  [2] _prepare_series_data(x::DateTime)
    @ RecipesPipeline ~/.julia/packages/RecipesPipeline/BGM3l/src/series.jl:8
  [3] _series_data_vector(x::DateTime, plotattributes::Dict{Symbol, Any})
    @ RecipesPipeline ~/.julia/packages/RecipesPipeline/BGM3l/src/series.jl:36
  [4] (::RecipesPipeline.var"#21#24"{Dict{Symbol, Any}})(vi::DateTime)
    @ RecipesPipeline ./none:0
  [5] iterate
    @ ./generator.jl:47 [inlined]
  [6] _series_data_vector(v::Vector{DateTime}, plotattributes::Dict{Symbol, Any})
    @ RecipesPipeline ~/.julia/packages/RecipesPipeline/BGM3l/src/series.jl:51
  [7] #21
    @ ./none:0 [inlined]
  [8] iterate(::Base.Generator{Vector{Vector{DateTime}}, RecipesPipeline.var"#21#24"{Dict{Symbol, Any}}})
    @ Base ./generator.jl:47
  [9] _series_data_vector(v::Vector{Vector{DateTime}}, plotattributes::Dict{Symbol, Any})
    @ RecipesPipeline ~/.julia/packages/RecipesPipeline/BGM3l/src/series.jl:51
 [10] macro expansion
    @ ~/.julia/packages/RecipesPipeline/BGM3l/src/series.jl:128 [inlined]
 [11] apply_recipe(plotattributes::AbstractDict{Symbol, Any}, #unused#::Type{RecipesPipeline.SliceIt}, x::Any, y::Any, z::Any)
    @ RecipesPipeline ~/.julia/packages/RecipesBase/BRe07/src/RecipesBase.jl:300
[...]

PR #5101, opened 2025-06-09
Add specialised type handling for vector of vectors

InfiniteChai:
## Description

Just getting myself familiar with the `Plots.jl` library and so thought I would try and fix some of the outstanding bugs to help me with that. 

This looks to fix https://github.com/JuliaPlots/Plots.jl/issues/4777 by introducing a specialisation for vector of vectors, applying recipes in a similar fashion to the `AbstractArray` handling.

Added unit testing to verify behaviour within the bug works. Have also tested manually that consistent behaviour applies for array type recipes (https://docs.juliaplots.org/stable/gallery/gr/generated/gr-ref045/).

### DIFF
--- a/RecipesPipeline/src/type_recipe.jl
+++ b/RecipesPipeline/src/type_recipe.jl
@@ -54,6 +54,45 @@ function _apply_type_recipe(plotattributes, v::AbstractArray, letter)
     w
 end
 
+# Specialisation to apply type recipes on a vector of vectors. The type recipe can either 
+# apply to the vector of elements or the elements themselves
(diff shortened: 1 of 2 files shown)


InfiniteChai:
Looking into test failures at the moment

t-bltg:
Please be aware that the `master` branchs corresponds to Plots 1.X and the current dev effort is oriented towards Plots 2.X in the `v2` branch.
Hence, contribution to `v2` is preferred, and backporting to `master` should follow if needed.

InfiniteChai:
Thanks @t-bltg - I saw that was happening, but wasn't sure whether it was at a stage to go straight onto v2. I'll regenerate the PR for v2 instead then.

Issue #5096, opened 2025-05-29
[BUG] Duplicate annotation related to `plot_title`


BioTurboNick:
## Details
If you simultaneously set `xlims`, `ylims`, and `plot_title`, and then later call `annotate!`, you can get duplicate annotations in an offset position.

I'm guessing it could be related to, potentially, that `plot_title` shouldn't really be provided when creating a basic plot, but only later when modifying a multi-subplot plot? `title` doesn't do it.

```
scatter([4], [4], plot_title = "x", xlims = (0, 10), ylims = (0, 10))
annotate!(4, 4, "4")
```

![Image](https://github.com/user-attachments/assets/20263cd1-3ce7-4bda-8ac6-eef80a201744)

### Backends

This bug occurs on ( insert `x` below )

Backend      | yes | no  | untested
-------------|-----|-----|---------
gr (default) |   x  |     |
pythonplot   |     |     | x
plotlyjs     |     |     | x
pgfplotsx    |     |     | x
unicodeplots |     |     | x
inspectdr    |     |     | x
gaston       |     |     | x

### Versions

Plots.jl version: v1.40.13
Backend  version (`]st -m <backend(s)>`): GR v0.73.16
Output of `versioninfo()`:
```
Julia Version 1.11.5
Commit 760b2e5b739 (2025-04-14 06:53 UTC)
Build Info:
  Official https://julialang.org/ release
Platform Info:
  OS: Linux (x86_64-linux-gnu)
  CPU: 8 × Intel(R) Xeon(R) Platinum 8488C
  WORD_SIZE: 64
  LLVM: libLLVM-16.0.6 (ORCJIT, sapphirerapids)
Threads: 4 default, 0 interactive, 2 GC (on 8 virtual cores)
Environment:
  JULIA_PKG_USE_CLI_GIT = true
  JULIA_REVISE_WORKER_ONLY = 1
```

InfiniteChai:
So this happens because `plot_title` actually creates its own sub-plot. As such when you apply `annotate!` to the plot it gets applied to every sub-plot including the title, hence the behaviour you're seeing.

What's interesting is that we get different behaviour when applying annotations directly to the initial scatter. In that case the application of the plot attribute updates doesn't carry over to the title sub-plot and we get a more expected behaviour.

```julia
scatter([4], [4], plot_title = "x", xlims = (0, 10), ylims = (0, 10), annotation=(4,4,"4"))
```

[...]

PR #5104, opened 2025-06-09
skip annotations for plot title subplot

InfiniteChai:
## Description

When applying annotations to a plot which has used `plot_title` the behaviour is inconsistent. Applying it as part of the initial plot will not apply the annotation to the title sub-plot but doing it separately will.

```
scatter([4], [4], plot_title = "x", xlims = (0, 10), ylims = (0, 10), annotation=(4,4,"4"))
```

versus

```
scatter([4], [4], plot_title = "x", xlims = (0, 10), ylims = (0, 10))
annotate!(4, 4, "4")
```

It seems sensible to never have annotations applied to the plot title sub-plot, so updating to ignore these property updates and get consistent behaviour. This fixes https://github.com/JuliaPlots/Plots.jl/issues/5096

### DIFF
--- a/PlotsBase/src/Plots.jl
+++ b/PlotsBase/src/Plots.jl
@@ -247,6 +247,8 @@ function _update_subplot_attrs(
 
     # grab those args which apply to this subplot
     for k ∈ keys(_subplot_defaults)
+        # We don't apply annotations to the plot-title sub-plot
+        k != :annotations || get(plt.attr, :plot_titleindex, 0) != subplot_index || continue
         PlotsBase.slice_arg!(plotattributes_in, sp.attr, k, subplot_index, remove_pair)
     end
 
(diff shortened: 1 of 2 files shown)


PlotsBase/src/Plots.jl
```diff
@@ -247,6 +247,8 @@ function _update_subplot_attrs(
 
     # grab those args which apply to this subplot
     for k ∈ keys(_subplot_defaults)
+        # We don't apply annotations to the plot-title sub-plot
+        k != :annotations || get(plt.attr, :plot_titleindex, 0) != subplot_index || continue
```
t-bltg, line 251: I have a hard reading this, aren't surrounding parens needed ?

t-bltg: Awesome, apart from this LGTM.

Issue #5051, opened 2025-03-09
[BUG] `mesh3d` fails for a simple mesh


pgeipi10:
```julia
using Plots, TriplotRecipes

p = Float64[
    -1.0   1.0  1.0  -1.0   0.0  0.0  -1.0  0.0  1.0
    -1.0  -1.0  1.0   1.0  -1.0  0.0   0.0  1.0  0.0
]

t = Int32[
    1  2  4  5  3  4  2  8
    5  6  7  6  8  6  9  6
    7  5  6  7  9  8  6  9
]

x, y = p[1,:], p[2,:]
z = x.^2 + y.^2 .- 1

display(plot(x, y, z, seriestype=:mesh3d, faces=t'.- 1, color=:lightblue))
```

Causes error:
```
ERROR: ArgumentError: Unsupported `:connections` type Nothing for seriestype=mesh3d
```

Sometimes the error does not appear, but then the plot is not correct

InfiniteChai:
So a few things

- The argument for `mesh3d` is `connections` to define the triangles, not `faces`. 
- Julia uses 1-based indexing so no need to subtract one from your coordinates (which for some reason plots, though not sensible).
- The coordinate indexing needs to be `Int`, not `Int32` (haven't tracked where that limitation is encoded). Just removing explicit typing works fine (or if you have it as `Int32` from elsewhere, `Int.(t)` will also work).
- `connections` needs to be a `AVec{AVec{Int}}` rather than `Matrix{Int}` so wrapping with `eachcol` gets it working

With that in mind, the following seems to work cleanly (adding in some alpha so it can be seen more clearly)

```julia
using Plots

p = [
    -1.0   1.0  1.0  -1.0   0.0  0.0  -1.0  0.0  1.0
    -1.0  -1.0  1.0   1.0  -1.0  0.0   0.0  1.0  0.0
]

t = [
    1  2  4  5  3  4  2  8
    5  6  7  6  8  6  9  6
    7  5  6  7  9  8  6  9
]

x, y = p[1,:], p[2,:]
z = x.^2 + y.^2

mesh3d(x,y,z,connections=eachcol(t), fillcolor=:blue, linecolor=:black, fillalpha=0.2)
```

![Image](https://github.com/user-attachments/assets/1b74fd6d-8a2d-4dda-813b-a53dbe362b81)

PR #5107, opened 2025-06-10
Generalise mesh3d within GR backend to support Integer types

InfiniteChai:
## Description

Small fix - noticed while looking at https://github.com/JuliaPlots/Plots.jl/issues/5051 that the GR backend only supports the system signed integer type `Int` for the `connections` within a `mesh3d`.

Seems an unnecessary constraint, so broadening to support any Integer type. The underlying `polygonmesh3d` converts it into a `Vector{Int32}` for the ccall so no need to do any conversion higher up.

Tested locally that cases with various signed and unsigned integers work as expected


t-bltg (approved):

t-bltg:
Thanks !

PR #5132, opened 2025-07-03
use `runic` formatter

t-bltg:


BeastyBlacksmith:
What was wrong with the old formatter?

Also please give people a few days to react before merging things like this.

t-bltg:
> What was wrong with the old formatter?

`Runic` is more consistent with `julia` standard for beginners.

> Also please give people a few days to react before merging things like this.

I'm fixing all the things here, for example `master` was completely broken since `1.40.13`for the docs.

Since literally no one is contributing to `Plots` these days, and since it is a bit of a nightmare to maintain `master` and `v2`, I won't accept spending time waiting for reviews.

Of course anyone interested to chime in is welcome.

t-bltg:
Also, https://github.com/JuliaPlots/Plots.jl/pull/5098 was merged too soon, bc no one built the docs, so I had to fix those.

t-bltg:
> bc no one built the docs, so I had to fix those

For that purpose, I've added an extra condition for the `docs.yml` github action : one can request to build the `docs` in a PR if one applies the labels `documentation` to the PR. This way, we can check when a PR is ready for review if it breaks the docs or not.

BeastyBlacksmith:
Don't get me wrong, I see you are doing a lot of work and I appreciate that. Still, this is collaborative project and pushing decisions without giving other people time to react and no rationale for the decision is kind of rude.

Again, you are doing great work.

Issue #5120, opened 2025-06-26
[BUG] plotly: order of plotly and require.js loading effects display


jverzani:
## Details

On a mac this set of commands does not produce a display:

```julia
using Plots
plotly()
plot(sin, 0, 2pi)
```

However, *if* I hand edit the generated HTML file and move the line

```
   <script src="https://requirejs.org/docs/release/2.3.7/minified/require.js" onload="plots_jl_plotly_init()"></script>
```

after the closing `</body>` tag, the plot displays. I think the order of loading `plotly.js` and then `require.js` matters here. I checked with Chrome, Safari, and Opera.

### Backends

This bug occurs on ( insert `x` below )

Backend      | yes | no  | untested
-------------|-----|-----|---------
gr (default) |     | x    |
pythonplot   |     |  x   |
plotlyjs     |     |   x  |
plotly   | x |  |
pgfplotsx    |     |  x   |
unicodeplots |     | x    |
inspectdr    |     |   x  |
gaston       |     |    x |

### Versions

Plots.jl version:
Backend  version (`]st -m <backend(s)>`):
```
[a03496cd] PlotlyBase v0.8.21
[f2990250] PlotlyKaleido v2.3.0
[91a5bcdd] Plots v1.40.14
```

Output of `versioninfo()`:

```
Julia Version 1.11.5
Commit 760b2e5b739 (2025-04-14 06:53 UTC)
Build Info:
  Official https://julialang.org/ release
Platform Info:
  OS: macOS (arm64-apple-darwin24.0.0)
  CPU: 10 × Apple M4
  WORD_SIZE: 64
  LLVM: libLLVM-16.0.6 (ORCJIT, apple-m1)
Threads: 1 default, 0 interactive, 1 GC (on 4 virtual cores)
```

jverzani:
This comes up as well when using `quarto`. There when there are two or more plots per page, this command becomes an issue:

```
<script src="https://requirejs.org/docs/release/2.3.7/minified/require.js" onload="plots_jl_plotly_init()">
```

(https://github.com/JuliaPlots/Plots.jl/blob/v2/PlotsBase/src/plotly.jl#L1284)

As each plot gets wrapped in a javascript function with the same name, so only one plot can  be rendered per page even after sorting out the loading order of the libraries. This was not the case in `v1.40.13`.

PR #5135, opened 2025-07-03
fix empty `plotly` plots

t-bltg:
Fix https://github.com/JuliaPlots/Plots.jl/issues/5120.

### DIFF
--- a/docs/make.jl
+++ b/docs/make.jl
@@ -26,7 +26,7 @@ eval(PlotsBase.WEAKDEPS)
 const SRC_DIR = joinpath(@__DIR__, "src")
 const WORK_DIR = joinpath(@__DIR__, "work")
 const GEN_DIR = joinpath(WORK_DIR, "generated")
-const DEV_BRANCH = ("master", "v2")[2]  # transition to v2
+const BRANCH = ("master", "v2")[2]  # transition to v2
 
 const ATTRIBUTE_SEARCH = Dict{String, Any}()  # search terms
 
@@ -70,7 +70,7 @@ end
 # ----------------------------------------------------------------------
 
 edit_url(args...) =
-    "https://github.com/JuliaPlots/Plots.jl/blob/$DEV_BRANCH/docs/" * if length(args) == 0
+    "https://github.com/JuliaPlots/Plots.jl/blob/$BRANCH/docs/" * if length(args) == 0
     "make.jl"
 else
     joinpath("src", args...)
@@ -655,7 +655,7 @@ function main(args)
         needs_rng_fix[pkg] = generate_cards(joinpath(@__DIR__, "gallery"), be, slice)
         let (path, cb, assets) = makedemos(
                 joinpath("gallery", string(be));
-                root = @__DIR__, src = joinpath(work, "gallery"), edit_branch = DEV_BRANCH
+                root = @__DIR__, src = joinpath(work, "gallery"), edit_branch = BRANCH
             )
             push!(gallery, string(pkg) => joinpath("gallery", path))
             push!(gallery_callbacks, cb)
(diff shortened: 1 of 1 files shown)


t-bltg:
I've manually checked that the 'no plot' issue is resolved.
Don't know how to write an automated test for that though.

@jverzani, do you have an example (mwe) for the 'multiple plots on one page' issue ?
Anyway, it should be fixed now by adding the (unique) uuid based tag to the function name.

BeastyBlacksmith (approved):

t-bltg:
Also on `linux` at least, the following

```julia
using Plots
plotly()
display(plot(sin, 0, 2pi))
```
writes a file in `/tmp`, which cannot be accessed by `snap` based `chromium` and `firefox`  for example (also container such as docker maybe ?), which is a pain.

So this PR adds an env variable `PLOTSBASE_TMPDIR` preempting `TMP` or `TMPDIR` - if defined - for `Base.tempdir` to control the folder where the resulting `html` file is written, and hence allowing `xdg-open` to correctly display the resulting plot in a browser.

(sorry for the dual review request)

BeastyBlacksmith:
> So this PR adds an env variable PLOTSBASE_TMPDIR preempting TMP or TMPDIR - if defined - for Base.tempdir to control the folder where the resulting html file is written, and hence allowing xdg-open to correctly display the resulting plot in a browser.

Okay, but we should have some documentation on the env vars we provide (either in the module descriptions or in the official docs)

t-bltg:
> Okay, but we should have some documentation on the env vars we provide (either in the module descriptions or in the official docs)

Thanks, done in https://github.com/JuliaPlots/Plots.jl/pull/5135/commits/b8a43d979c28d5580dcf267fbc1389e851db4ae3.

docs/src/basics.md
```diff
@@ -62,3 +62,11 @@ plot(1:10, Any[rand(10), sin])               # 2 series: rand(10) and map(sin,x)
...
 
 ---
+
+### Environment variables
+
+A few environment variables control `PlotsBase` internals:
+- `PLOTSBASE_DEFAULT_BACKEND`: default backend, preempting the one set through the `Preferences` based mecanism;
+- `PLOTSBASE_TMPDIR`: temporay files prefix for (html) files (some browser are denied access to temporary directorie such as `/tmp`);
+- `PLOTSBASE_HOST_DEPENDENCY_LOCAL`: use a local `plotly` ressource instead of a cloud hosted one.
+
```
BeastyBlacksmith, line 72:
```suggestion

### Environment variables

A few environment variables control `PlotsBase` internals:
- `PLOTSBASE_DEFAULT_BACKEND`: default backend, preempting the one set through the `Preferences` based mechanism;
- `PLOTSBASE_TMPDIR`: temporay files prefix for (html) files (some browser are denied access to temporary directorie such as `/tmp`);
- `PLOTSBASE_HOST_DEPENDENCY_LOCAL`: use a local `plotly` resource instead of a cloud hosted one.

#### For testing purposes
- `PLOTSBASE_TEST_PACKAGES`: list of backends to run the tests for
- `PLOTSBASE_PLOTLYJS_UNSAFE_ELECTRON`: ?
- `PLOTSBASE_IMG_TOL`: tolerance when comparing images during tests
- `PLOTSBASE_REFERENCE_DIR`: where to find the reference images

#### For building documentation
- `PLOTS_DOCS_DEV`: ?
- `PLOTDOCS_PACKAGES`: ?
- `PLOTDOCS_EXAMPLES`: ?
- `PLOTDOCS_ANSICOLOR`: ?

### Special julia variables

- `PLOTSBASE_DEFAULTS`: A way to set default values for plotattributes (e.g. in your `startup.jl` file). Should be a dictionary.
```
we do have a few more, I put a ? where I couldn't see at a glance what they were for

jverzani:
> I've manually checked that the 'no plot' issue is resolved. Don't know how to write an automated test for that though.
> 
> @jverzani, do you have an example (mwe) for the 'multiple plots on one page' issue ? Anyway, it should be fixed now by adding the (unique) uuid based tag to the function name.

Thanks, your fix is exactly what I would have suggested for this issue.

PR #5127, opened 2025-07-02
fix

t-bltg:
https://github.com/JuliaPlots/Plots.jl/pull/5098 was insufficiently tested.

PR #5143, opened 2025-07-04
Unitful and Latexify handling for v2

Ickaser:
## Attribution

## Things to consider

### DIFF
--- a/PlotsBase/src/Axes.jl
+++ b/PlotsBase/src/Axes.jl
@@ -347,22 +347,22 @@ function PlotsBase.attr!(axis::Axis, args...; kw...)
     return axis
 end
 
-function get_guide(axis::Axis)
-    if isnothing(axis[:unit]) || axis[:unitformat] ≡ :nounit
-        return axis[:guide]
+get_guide(axis::Axis)::String = if isnothing(axis[:guide]) 
+    ""
+elseif isnothing(axis[:unit]) || axis[:unitformat] ≡ :nounit
+    axis[:guide]
+else
+    unit = if axis[:unitformat] isa Function
+        axis[:unit]
+    elseif PlotsBase.backend_name() ≡ :pgfplotsx
+        pgext = Base.get_extension(PlotsBase, :PGFPlotsXExt)
+        isnothing(pgext) && error("PGFPlotsX extension not loaded. Problems")
+        pgext.Latexify.latexify(axis[:unit])
     else
-        ustr = if PlotsBase.backend_name() ≡ :pgfplotsx
-            pgext = Base.get_extension(PlotsBase, :PGFPlotsXExt)
-            isnothing(pgext) && error("PGFPlotsX extension not loaded. Problems")
-            pgext.Latexify.latexify(axis[:unit])
-        else
-            string(axis[:unit])
-        end
-        if isempty(axis[:guide])
-            return ustr
-        end
-        return format_unit_label(axis[:guide], ustr, axis[:unitformat])
+        string(axis[:unit])
     end
+    isempty(axis[:guide]) && return unit
+    format_unit_label(axis[:guide], unit, axis[:unitformat])
 end
 
 # Keyword options for unit formats
--- a/PlotsBase/test/test_unitful.jl
+++ b/PlotsBase/test/test_unitful.jl
@@ -292,6 +292,16 @@ end
         @test contourf(x, y, z) isa PlotsBase.Plot
     end
 
+     @testset "latexify as unitformat" begin
+        y = rand(10) * u"m^-1"
+        @test yguide(plot(y, ylabel = "hello", unitformat = latexify)) == "\$hello\\;\\left/\\;\\mathrm{m}^{-1}\\right.\$"
+
+        uf = (l, u) -> l * " (" * latexify(u) * ")"
+        @test yguide(plot(y, ylabel = "hello", unitformat = uf)) == "hello (\$\\mathrm{m}^{-1}\$)"
+        Latexify.set_default(labelformat = :square)
+        @test yguide(plot(y, ylabel = "hello", unitformat = latexify)) == "\$hello\\;\\left[\\;\\mathrm{m}^{-1}\\right]\$"
+    end
+
     @testset "colorbar title" begin
         x, y = (1:0.01:2) * m, (1:0.02:2) * s
         z = x' ./ y


Ickaser:
As is, this should address only things that were covered in #5098 and #5127.

To fully resolve #5093, https://github.com/PainterQubits/Unitful.jl/pull/795 and https://github.com/jw3126/UnitfulRecipes.jl/pull/81 will need to merge so that the Unitful extension can drop its `UnitfulLatexify` dependency and someone who loads Plots and Unitful (but doesn't need Latexify) will still have the extension load. I think that is perhaps best handled as another PR since it will involve touching a bunch of Project.toml files.

t-bltg:
FYI, I've changed the default formatter for the `Plots` repo, so now you have to use:
```
$ julia -e 'using Runic; exit(Runic.main(ARGS))' -- --inplace .
```

Ickaser:
The test failure here is because I didn't keep good track of what was new in https://github.com/PainterQubits/Unitful.jl/pull/795. When I get a chance I will change the test to reflect UnitfulLatexify's old behavior, unless that other one gets merged and released really fast.

Ickaser:
As of https://github.com/JuliaPlots/Plots.jl/pull/5143/commits/dacf8228d09a62548d738646c7b490cbef8197d9 (that last commit), this now fully depends on https://github.com/JuliaPhysics/Unitful.jl/pull/795. That makes it parallel to #5174, because I decided I may as well try to fully resolve the problem instead of spreading it across several PRs. If that seems bad then this could be split into 2 PRs, one without and one with the full removal of Latexify from UnitfulExt.

A key thing to note here is that Latexify is only required for the PGFPlotsX backend. To enable loading of the Unitful extension without needing to load Latexify (which seems like a reasonable expectation to me), I did the same thing as was done in #5098 in the `get_guide` function: in the places where `Latexify.latexify` would be called (which are inside branches with `PlotsBase.backend_name() == :pfgplotsx`, I get the `PGFPlotsX` extension and access its copy of `Latexify.latexify`. That feels like a kludge, but I think it works. 
- This would break if somehow PGFPlotsX stopped using Latexify. 
- One alternative: we could define a regular Unitful extension, then define another extension that depends on Unitful as well as Latexify that adds this logic. But it would mean that `if` statements get replaced with dispatches defined in different extensions
- I have no idea if there are severe performance implications for calling `Base.get_extension(PlotsBase, :PGFPlotsX)` every time a plot is generated

PlotsBase/Project.toml
```diff
@@ -72,7 +71,7 @@ PlotlyJSExt = "PlotlyJS"
 PlotlyKaleidoExt = "PlotlyKaleido"
 PythonPlotExt = "PythonPlot"
 UnicodePlotsExt = "UnicodePlots"
-UnitfulExt = ["Latexify", "LaTeXStrings", "UnitfulLatexify", "Unitful"]
+UnitfulExt = ["Unitful"]
```
Ickaser, line 74:
```suggestion
UnitfulExt = "Unitful"
```

BeastyBlacksmith:
I am currently failing to get the `PGFPlotsXExt` to get activated. Do you also have that problem or is it just me?

Ickaser:
> I am currently failing to get the `PGFPlotsXExt` to get activated. Do you also have that problem or is it just me?

That's not just you, and in fact a similar problem with the Plots -> UnitfulExt extension is what prompted this whole crusade of mine. 

In the Project.toml, this is how you declare multiple dependencies for an extension, i.e. if the extension module imports `Colors` then that needs to be declared here:
https://github.com/JuliaPlots/Plots.jl/blob/e271a8ea228c0f0977e8bdd54489fb7290021e04/PlotsBase/Project.toml#L70
and for the extension to load, all of those packages need to be loaded already.
In Plots v1, all of those packages are already dependencies of Plots, so with `using Plots, PGFPlotsX` you have everything loaded and the extension can activate. But as part of the efforts to slim down Plots.jl I think some of those have been dropped as deps for `PlotsBase` so they need to be manually loaded.

These extra dependencies have been manually listed here:
https://github.com/JuliaPlots/Plots.jl/blob/e271a8ea228c0f0977e8bdd54489fb7290021e04/PlotsBase/src/PlotsBase.jl#L10-L18
And for testing and documentation, they are loaded here:
https://github.com/JuliaPlots/Plots.jl/blob/e271a8ea228c0f0977e8bdd54489fb7290021e04/PlotsBase/test/runtests.jl#L17
https://github.com/JuliaPlots/Plots.jl/blob/e271a8ea228c0f0977e8bdd54489fb7290021e04/docs/make.jl#L24

I was able to whittle this down so that `UnitfulExt` only needs Unitful, and in places where it needs to be compatible with PGFPlotsX it uses the PGFPlotsX reference to LaTeXStrings and Latexify. Repeating that process for PGFPlotsX would be harder, though, unless we can make all of these packages dependencies of PGFPlotsX itself. then that would resolve things from the perspective of PlotsBase users.

BeastyBlacksmith:
Well, in that case I'd suggest to promote `"Colors", "Contour", "Latexify", "LaTeXStrings"` back to normal dependencies.

Since the backend activation shouldn't be more than
`using PlotsBase; import PGFPlotsX; pgfplotsx()`

@t-bltg what is your opinion?

t-bltg:
I think this is the wrong move here.

Plots dependencies have to be kept minimal for maintainability, and keeping TTFX minimal (which includes initial loading).

If a backend needs something but it isn't used elsewhere in PlotsBase, we should develop a wrapper such as PlotsPGFPlotsX which would load and export all the PGFPlotsX dependencies needed for this backend.

This is why I removed them for Plots 2.0. If you want a single import for convenience, then a wrapper is the solution.

BeastyBlacksmith:
Ideally package extensions would be allowed to have dependencies on their own, but I don't think this will come anytime soon.
In general, I agree that minimizing dependencies is a good thing to do, however, not at all cost.

From the user side, besides TTFX and functionality we also need to account for ease of use and consistency.
So, I am not a fan of saying "We made switching backends more cumbersome and also for some backends you need to jump through some extra hoops".

The import time of those packages is not nothing but also not too excessive:
```julia
julia> @time_imports using Colors
      3.5 ms  Statistics
     28.1 ms  FixedPointNumbers
      0.8 ms  Reexport
               ┌ 0.0 ms ColorTypes.__init__() 
    115.9 ms  ColorTypes 81.51% compilation time
     39.5 ms  Colors

julia> @time_imports using Contour
      2.8 ms  Contour

julia> @time_imports using Latexify
      2.0 ms  LaTeXStrings
     34.1 ms  MacroTools
     12.6 ms  Format
     22.3 ms  Preferences
      1.5 ms  JLLWrappers
               ┌ 41.0 ms JpegTurbo_jll.__init__() 20.77% compilation time
     42.3 ms  JpegTurbo_jll 20.09% compilation time
               ┌ 0.1 ms Zlib_jll.__init__() 
      6.9 ms  Zlib_jll
               ┌ 94.0 ms Ghostscript_jll.__init__() 
     95.8 ms  Ghostscript_jll
     15.7 ms  OrderedCollections
     43.5 ms  Latexify
```
compared to `PlotsBase`:
```julia
julia> @time_imports using PlotsBase
   320.9 ms  PlotsBase 2.17% compilation time

julia> versioninfo()
Julia Version 1.11.6
Commit 9615af0f269 (2025-07-09 12:58 UTC)
Build Info:
  Official https://julialang.org/ release
Platform Info:
  OS: Linux (x86_64-linux-gnu)
  CPU: 16 × 11th Gen Intel(R) Core(TM) i9-11900K @ 3.50GHz
  WORD_SIZE: 64
  LLVM: libLLVM-16.0.6 (ORCJIT, rocketlake)
Threads: 1 default, 0 interactive, 1 GC (on 16 virtual cores)
Environment:
  JULIA_VSCODE_REPL = 1
  JULIA_EDITOR = code
  ```
  
  Otherwise it would only make sense to me to define wrapper packages for all backends irrespective if they need extra dependencies or not, such that it is at least consistent and you can tell people to run
  ```julia
  using Plots; import Plots<Backend>; <backend>()
  ```
  For all backends. Though I am not yet convinced that managing those wrapper packages is causing less maintenance burden.

Ickaser:
> Ideally package extensions would be allowed to have dependencies on their own, but I don't think this will come anytime soon. In general, I agree that minimizing dependencies is a good thing to do, however, not at all cost.
> 
> From the user side, besides TTFX and functionality we also need to account for ease of use and consistency. So, I am not a fan of saying "We made switching backends more cumbersome and also for some backends you need to jump through some extra hoops".
> 
> Otherwise it would only make sense to me to define wrapper packages for all backends irrespective if they need extra dependencies or not, such that it is at least consistent and you can tell people to run
> 
> ```julia
> using Plots; import Plots<Backend>; <backend>()
> ```
> 
> For all backends. Though I am not yet convinced that managing those wrapper packages is causing less maintenance burden.

I think I concur with this assessment for backends in general. Though it may be unfortunate to have e.g. the latency of loading Latexify baked into PlotsBase, I think the inconvenience (and confusion) of needing to manually import these extra dependencies only for the PGFPlotsX backend is a worse problem than TTFX. Wrapper packages sound like a lot of work to manage, too, particularly since a consistent workflow would mandate using wrappers for every backend.

In fact, if we add Latexify back as a dependency of PlotsBase, then that eliminates a kludge here where I access Latexify through the PGFPlotsX extension (in a branch that only gets called when PGFPlotsX is the active backend).

BeastyBlacksmith:
> In fact, if we add Latexify back as a dependency of PlotsBase, then that eliminates a kludge here where I access Latexify through the PGFPlotsX extension (in a branch that only gets called when PGFPlotsX is the active backend).

Do add them as dependencies for now and if we find a viable alternative we can still change that later on.

Ickaser:
> > In fact, if we add Latexify back as a dependency of PlotsBase, then that eliminates a kludge here where I access Latexify through the PGFPlotsX extension (in a branch that only gets called when PGFPlotsX is the active backend).
> 
> Do add them as dependencies for now and if we find a viable alternative we can still change that later on.

Done, I added `Colors` and `Contour` as a separate commit to make it easier to revert if necessary but as this PR stands the PGFPlotsXExt should only need PGFPlotX loaded to trigger.

Ickaser:
To get tests passing, I had to tell Aqua testing to ignore Colors, Contour, LaTeXStrings, and Latexify as stale deps (meaning they are not imported in PlotsBase itself). LaTeXStrings and Latexify are definitely pretty essential to PGFPlotsXExt working right, but it might in future be possible to drop Colors and Contour entirely as dependencies. They show up only in the following spots:

https://github.com/JuliaPlots/Plots.jl/blob/50ea05e40d5fa176f2d969816ad7f9f52998bff0/PlotsBase/ext/PGFPlotsXExt.jl#L305-L317

https://github.com/JuliaPlots/Plots.jl/blob/50ea05e40d5fa176f2d969816ad7f9f52998bff0/PlotsBase/ext/PGFPlotsXExt.jl#L769-L779

Ickaser:
With https://github.com/JuliaPlots/Plots.jl/pull/5143/commits/6d9e8907b2feda6effafc76fc55da3b401f39864, I tried making these extra packages regular dependencies of PlotsBase without being actually imported, and removed the `eval(Expr(:(import Colors)))` trick that was being used to import them for docs and tests. Everything does load correctly, and loading PlotsBase does still incur the load time cost of these extra packages, but as discussed above I think that is an acceptable cost.

If there aren't other questions, I think I am good with this PR.

t-bltg:
> The import time of those packages is not nothing but also not too excessive:

The single measurement does not guarantee that this will always be the case.
Non `PGFPlotsX` users do not want to be loading unused dependencies.

I repeat, it's the wrong direction to add these dependencies back : it adds potential `Pkg` resolution issues (version compatibility), and increases the load time.

> `import Plots<Backend>; using Plots; <backend>()`

I think this solution is acceptable.

> Though I am not yet convinced that managing those wrapper packages is causing less maintenance burden.

Maintenance will be trivial if these `Plots<Backend>` are registered in the `Plots` monorepo. We don't expect them to change often.

Ickaser:
> > `import Plots<Backend>; using Plots; <backend>()`
> 
> I think this solution is acceptable.
> 
> > Though I am not yet convinced that managing those wrapper packages is causing less maintenance burden.
> 
> Maintenance will be trivial if these `Plots<Backend>` are registered in the `Plots` monorepo. We don't expect them to change often.

I do understand the reasoning here about compat and TTFX. But elsewhere this pattern is used (e.g. https://github.com/SciML/Optimization.jl/tree/master/lib/) it produces a lot of churn in registered packages and versions and compat, which I think has to be factored in as a maintenance burden. Proliferating a lot of package names like `PlotsPGFPlotsX` and `PlotsPlotlyBase` alongside `PlotsBase` feels awkward, but even more than that it is potentially confusing for the newcomers who are very likely to try out Plots early on in their time with Julia.

If we are able to remove `Colors` and `Contour`, plus lighten up the `Latexify` by removing Ghostscript, would that be enough to satisfy you @t-bltg that these particular packages are not too onerous as dependencies? What evidence would be enough to convince you?

t-bltg:
> But elsewhere this pattern is used

Yeah it isn't ideal, but it used somewhere else and does seem to work. I really think dependencies related to a **single** backend have nothing to do in `PlotsBase`.

> If we are able to remove Colors and Contour, plus lighten up the Latexify by removing Ghostscript

Probably yes. Also, we could ask to add new features as hard dependencies directly to `PGFPlotsX`, thus eliminating the maintenance here in `Plots`.

BeastyBlacksmith:
The question how we will deal with such dependencies is somewhat orthogonal to this PR and can be discussed further in https://github.com/JuliaPlots/Plots.jl/discussions/5187. Though it is a good showcase of the pros and cons.

For now it is important that the backends can be activated by a single import. So I am merging as is.

PR #5147, opened 2025-07-06
render more `html` and `svg` in docs

t-bltg:
Storing `png`s is heavy, so prefer using `svg` and `html`.


ChrisRackauckas:
> Storing pngs is heavy, so prefer using svg and html.

Oh NO! Other way around!

Is this what is causing the massive breakage for the docs everywhere? I have started seeing many places where I cannot even clone the docs anymore, with the gh-pages being 5gb, 10gb, etc. and it's all due to svgs. Are documenter builds now building SVGs? If so, we need to tell people to start clearing out their git cache via https://sciml.github.io/OrgMaintenanceScripts.jl/dev/documentation_cleanup/ because this is really starting to crash all of the servers!

t-bltg:
> Oh NO! Other way around!

Nope, it depends on your data, but fine I'll revert this as it shouldn't have landed into the backport PR https://github.com/JuliaPlots/Plots.jl/pull/5148 as it would be breaking for docs plots and web plots.

PR #5174, opened 2025-08-28
Drop UnitfulLatexify as a dependency for v1

Ickaser:
## Description

This depends on https://github.com/JuliaPhysics/Unitful.jl/pull/795 , which makes UnitfulLatexify into a LatexifyExt for Unitful. Once that is merged and released, Plots can drop UnitfulLatexify as a dependency in both v1 and v2.

Once the Unitful release is out, the compat bound here will have to be set to match. I am not sure yet if it will be 1.25 or 1.26.

Arguably, dropping support for UnitfulLatexify altogether (which this is doing) could be considered a breaking change for some users. But it's not breaking the Plots API, exactly, and we're not breaking API for Unitful or Latexify either, so the breaking change is in UnitfulLatexify, which will have a deprecation notice. (And the fix should be pretty simple for the handful of people who run into it.)

I want this change in v1 and not just v2 because then it can be used in the Unitful docs, which use the Latexify extension and demonstrate how the Latexify extension can be used with Plots.

## Attribution

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -38,7 +38,6 @@ StatsBase = "2913bbd2-ae8a-5f71-8c99-4fb6c76f3a91"
 TOML = "fa267f1f-6049-4f14-aa54-33bafae1ed76"
 UUIDs = "cf7118a7-6976-5b1a-9a39-7adc72f591a4"
 UnicodeFun = "1cfade01-22cf-5700-b092-accc4b62d6e1"
-UnitfulLatexify = "45397f5d-5981-4c77-b2b3-fc36d6e9b728"
 Unzip = "41fe7b60-77ed-43a1-b4f0-825fd5a5650d"
 
 [weakdeps]
@@ -53,7 +52,7 @@ FileIOExt = "FileIO"
 GeometryBasicsExt = "GeometryBasics"
 IJuliaExt = "IJulia"
 ImageInTerminalExt = "ImageInTerminal"
-UnitfulExt = ["Unitful", "UnitfulLatexify"]
+UnitfulExt = "Unitful"
 
 [compat]
 Aqua = "0.8"
@@ -93,7 +92,6 @@ StatsBase = "0.33 - 0.34"
 UnicodeFun = "0.4"
 UnicodePlots = "3.4"
 Unitful = "1.25"
-UnitfulLatexify = "2"
 Unzip = "0.1 - 0.2"
 julia = "1.10"
 
--- a/ext/UnitfulExt.jl
+++ b/ext/UnitfulExt.jl
@@ -8,9 +8,6 @@ import RecipesBase
 @ext_imp_use :import Unitful Quantity unit ustrip Unitful dimension Units NoUnits LogScaled logunit MixedUnits Level Gain uconvert
 import LaTeXStrings: LaTeXString
 import Latexify: latexify
-if !isdefined(Base, :get_extension)
-    import UnitfulLatexify # On Julia <1.9, this glue package loads the extension.
-end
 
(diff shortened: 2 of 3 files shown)


giordano:
Bump

Ickaser:
The test failures occur for 1.6 (where the Latexify extension does not load, so of course latexify tests fail), and downstream in GraphRecipes.jl where the `julia_type_tree` plot doesn't match an old version. The GraphRecipes one maybe deserves its own issue and PR, but is it alright if I set the Unitful+Latexify tests to only run on Julia 1.9+?

sostock:
Is there any functionality lost on Julia 1.6 or are only the tests failing? If the functionality is lost, should it be considered breaking?

One way to keep all functionality on Julia 1.6 would be to change UnitfulLatexify v2 to contain a copy of the extension on Julia < 1.9. Only on Julia ≥ 1.9 would the package be empty. Plots would then depend on UnitfulLatexify v2. It’s not particularly nice (especially since one has to continue to maintain UnitfulLatexify parallel to the Unitful extension), but I think it should work.

Ickaser:
> Is there any functionality lost on Julia 1.6 or are only the tests failing? If the functionality is lost, should it be considered breaking?
> 
> One way to keep all functionality on Julia 1.6 would be to change UnitfulLatexify v2 to contain a copy of the extension on Julia < 1.9. Only on Julia ≥ 1.9 would the package be empty. Plots would then depend on UnitfulLatexify v2. It’s not particularly nice (especially since one has to continue to maintain UnitfulLatexify parallel to the Unitful extension), but I think it should work.

The functionality is lost, unfortunately. Making UnitfulLatexify somehow load the extension even on less than 1.6 could resolve this, although my gut says it makes compat more confusing. Or including a similar hack in the Plots `UnitfulExt` could resolve it just for Plots.

Ickaser:
Some local testing on my part suggests that with a potentially fragile hack to load the extension in https://github.com/gustaphe/UnitfulLatexify.jl/pull/22, putting UnitfulLatexify back in as a nearly-empty package works. At least, the tests that are failing here pass when I load this PR and that PR in Julia 1.6, WSL Ubuntu. The most recent test failure here is because UnitfulLatexify v2.0 isn't merged and registered yet.

BeastyBlacksmith:
I think it would also be okay to drop Julia 1.6 support at this point of time

Ickaser:
> I think it would also be okay to drop Julia 1.6 support at this point of time

~OK. I'll make that change in this PR, then, if everything else looks alright?~ I put that in #5184 in case it needs to be its own conversation.

Project.toml
```diff
@@ -53,7 +53,7 @@ FileIOExt = "FileIO"
 GeometryBasicsExt = "GeometryBasics"
 IJuliaExt = "IJulia"
 ImageInTerminalExt = "ImageInTerminal"
-UnitfulExt = "Unitful"
+UnitfulExt = ["Unitful", "UnitfulLatexify"]
```
Ickaser, line 56:
```suggestion
UnitfulExt = "Unitful"
```

Project.toml
```diff
@@ -92,7 +92,8 @@ Statistics = "1"
 StatsBase = "0.33 - 0.34"
 UnicodeFun = "0.4"
 UnicodePlots = "3.4"
-UnitfulLatexify = "1"
+Unitful = "1.25"
+UnitfulLatexify = "2"
```
Ickaser, line 96:
```suggestion

```

Ickaser:
Remaining test failures here now that #5184 is merged and this is rebased are just due to the `julia_type_tree` test image of `GraphRecipes`, so unrelated to this PR.

Issue #4781, opened 2023-06-20
[BUG] Bar charts with vector attributes (e.g. multiple colors) broken in Plotly since v1.10.6


niclasmattsson:
This is partially the same issue as #4129, but I'm creating a new issue with a more descriptive title, more info on the problem including some version bisecting and a workaround that may be helpful. There may also be a separate problem related to #3623, https://github.com/JuliaPlots/StatsPlots.jl/issues/442 and the PR [#3751](https://github.com/JuliaPlots/Plots.jl/pull/3751) - these issues may have been prematurely closed as fixed.

## Details

This is the part that is the same as #4129. This particular error was introduced sometime between Plots v1.22.5 and Plots v1.23.6.

```
julia> using Plots

julia> gr()
Plots.GRBackend()

julia> bar(rand(3), color=[:red, :green, :blue])  # works

julia> plotly()
Plots.PlotlyBackend()

julia> bar(rand(3), color=[:red, :green, :blue])
Error showing value of type Plots.Plot{Plots.PlotlyBackend}:
ERROR: ArgumentError: Iterates of the argument to Flatten are not known to have constant length
Stacktrace:
  [1] flatten_length(f::Base.Iterators.Flatten{Vector{Base.Generator{UnitRange{Int64}, Plots.var"#6#10"}}}, T::Type)
    @ Base.Iterators .\iterators.jl:1185
  [2] length(f::Base.Iterators.Flatten{Vector{Base.Generator{UnitRange{Int64}, Plots.var"#6#10"}}})
    @ Base.Iterators .\iterators.jl:1187
  [3] plotly_series_shapes(plt::Plots.Plot{Plots.PlotlyBackend}, series::Plots.Series, clims::Tuple{Float64, Float64})
    @ Plots C:\Users\niclas\.julia\packages\Plots\PomtQ\src\backends\plotly.jl:762
  [4] plotly_series(plt::Plots.Plot{Plots.PlotlyBackend}, series::Plots.Series)
    @ Plots C:\Users\niclas\.julia\packages\Plots\PomtQ\src\backends\plotly.jl:558
  [5] plotly_series(plt::Plots.Plot{Plots.PlotlyBackend})
    @ Plots C:\Users\niclas\.julia\packages\Plots\PomtQ\src\backends\plotly.jl:1035
  [6] plotly_series_json
    @ C:\Users\niclas\.julia\packages\Plots\PomtQ\src\backends\plotly.jl:1041 [inlined]
[...]

Issue #4129, opened 2022-03-02
[BUG] fill_z not working for histograms in pyplot


David96:
## Details
Adding fill_z to a histogram doesn't do anything for the pyplot backend.

### Backends

This bug occurs on ( insert `x` below )

Backend      | yes | no  | untested
-------------|-----|-----|---------
gr (default) |     |  x  |
pyplot       |  x  |     | 
plotlyjs     |     |     |x
pgfplotsx    |     |     |x
unicodeplots |     |     |x
inspectdr    |     |     |x
gaston       |     |     |x

### Versions

Plots.jl version: v1.25.8
Backend  version (`]st -m <backend(s)>`):
- PyPlot: [d330b81b] PyPlot v2.10.0 `https://github.com/JuliaPy/PyPlot.jl.git#master`
Output of `versioninfo()`:
```
Julia Version 1.7.2
Commit bf53498635 (2022-02-06 15:21 UTC)
Platform Info:
  OS: Linux (x86_64-pc-linux-gnu)
  CPU: AMD Ryzen 7 3700U with Radeon Vega Mobile Gfx
  WORD_SIZE: 64
  LIBM: libopenlibm
  LLVM: libLLVM-12.0.1 (ORCJIT, znver1)
Environment:
  JULIA_NUM_THREADS = 8
```

t-bltg:
I've added a note about this here:
https://github.com/JuliaPlots/Plots.jl/blob/6e8c0de257b3dee2ede56491b6ed086be70c598f/src/backends/pythonplot.jl#L463-L465

Same root issue as in the https://docs.juliaplots.org/stable/gallery/pythonplot/generated/pythonplot-ref056/#pythonplot_ref056.

It's because the bars are drawn using 4 paths instead of a single one.

PR #5175, opened 2025-08-29
Fix handling of vector attributes in plotly code

BeastyBlacksmith:
## Description
Fix https://github.com/JuliaPlots/Plots.jl/issues/4781
Fix https://github.com/JuliaPlots/Plots.jl/issues/4129

The plotly code tried to compute the length of an iterator (which has unknown length). We solved that by collecting the iterator first.

However, the way the code durrently draws the bars, the filling of the bars does not work, since each line of the bar is drawn as a seperate plot with no area.

## Attribution

## Things to consider

### DIFF
--- a/src/backends/plotly.jl
+++ b/src/backends/plotly.jl
@@ -1,13 +1,13 @@
 # https://plot.ly/javascript/getting-started
 
 _plotly_framestyle(style::Symbol) =
-    if style in (:box, :axes, :zerolines, :grid, :none)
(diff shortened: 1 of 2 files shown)


BeastyBlacksmith:
702a8d7 changes the handling of vector attributes (yet again) to map each segment to the corresponding entry in the vector attributes instead of duplicating the entries.

still need to check whether this breaks something

PR #5176, opened 2025-09-05
Add `stem` plots for sampled data systems

jonniedie:
## Description
Discretely sampled timeseries data is usually best visualized as a `stem` plot. It is helpful especially in cases where single-point impulse functions are viewed, since it can be hard to see them on scatter plots (see below)
image

## Attribution

## Things to consider

### DIFF
--- a/.zenodo.json
+++ b/.zenodo.json
@@ -814,6 +814,11 @@
             "affiliation": "@JuliaComputing",
             "name": "Παναγιώτης Γεωργακόπουλος",
             "type": "Other"
+        },
+        {
+            "affiliation": "@ZiplineTeam",
+            "name": "Jonnie Diegelman",
+            "type": "Other"
         }
     ],
     "upload_type": "software"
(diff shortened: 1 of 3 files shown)


briederer:
Out of curiosity, how is this different from the 'sticks' linestyle?

jonniedie:
> Out of curiosity, how is this different from the 'sticks' linestyle?

I guess nothing, but this is the first time I've heard of the `sticks` linestyle after trying really hard to find something that would do this. For example, below is what the documentation shows for `linestyle`. And none of the backends I've tried so far support `sticks`. Besides this, `linestyle` doesn't seem like the right place for this, as the other `linestyle`s just change how the line is styled, not the shape/path of the line itself.

I'm not necessarily against calling it `sticks`, but `stem` is more in line with pyplot and MATLAB, so it would be nice to at least have a `stem` alias to make it more discoverable for people coming from those backgrounds.

image

BeastyBlacksmith:
> And none of the backends I've tried so far support `sticks`.

What exactly did you try?
At least on GR and PGFPlotsX
```julia
sticks(1:10, marker = true)
```
works

PR #5184, opened 2025-09-18
Up minimum compat for v1 to 1.10 (current LTS), test "min" and "lts" on CI

Ickaser:
## Description

On the current master, Plots has a Julia compat of 1.6. 
However, hxtensions have been used in Plots for a while; it seems like the burden of loading them on old Julia hasn't been too high. But for  #5174, the lack of extensions for dependencies of extensions starts to make maintenance of new code pretty difficult. A possible compat bound for getting the extensions working would be 1.9, but since 1.10 has been the LTS [for almost a year now](https://julialang.org/blog/2024/10/julia-1.11-highlights/), I propose that gets used as a lower bound for testing and compat on new versions of Plots.

This PR updates that in Project.toml, the CI pipeline (where "min", referring to minimum Julia compat, and "lts" are both set as versions). I also drop the block in `Plots.__init()__` where extensions were loaded for old versions of Julia.

## Attribution

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -29,8 +29,8 @@ jobs:
       fail-fast: false
       matrix:
         version:
-          - '1.6' # LTS (minimal declared julia compat in `Project.toml`)
-          - 'lts'
+          - 'min' # minimum declared julia compat in `Project.toml`
+          - 'lts' # LTS 
           - '1' # latest stable
         experimental:
           - false
--- a/Project.toml
+++ b/Project.toml
@@ -94,7 +94,7 @@ UnicodeFun = "0.4"
 UnicodePlots = "3.4"
 UnitfulLatexify = "1"
 Unzip = "0.1 - 0.2"
-julia = "1.6"
+julia = "1.10"
 
 [extras]
 Aqua = "4c88cf16-eb10-579e-8560-4a9242c79595"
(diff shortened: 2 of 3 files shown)


Ickaser:
@t-bltg I imagine you will want to weigh in on this

t-bltg:
> @t-bltg I imagine you will want to weigh in on this

I think too that it's reasonable to ask to for julia `1.10` lts as a minimal version for a future `Plots` version of `1.41`, but I'll let @BeastyBlacksmith make the final choice.

BeastyBlacksmith:
Can we get rid of `Requires.jl` then, or is it used for something else still?

Ickaser:
The other remaining usage of Requires is for some backend loading behavior. I didn't investigate that too closely, especially after seeing that v2 does away with that altogether (and already has a minimum julia 1.10 compat) in favor of extensions for backends. I could maybe take another look and see what exactly it's doing

Ickaser:
This macro is used elsewhere in the file for backend loading:

https://github.com/JuliaPlots/Plots.jl/blob/8d8965d40dd565472abe4ca9b26fa8411470b1a3/src/backends.jl#L96-L103

t-bltg:
Yeah in `v2` backend code loading is done through extensions and this might no be easy to backport `v2`changes into `v1`.

BeastyBlacksmith:
Alright, thanks for checking

PR #5188, opened 2025-10-06
Remove reference to `RecipesBase.debug`

penelopeysm:
## Description

`RecipesBase.debug()` was removed in https://github.com/JuliaPlots/Plots.jl/pull/4459 but the docs still refers to it.

## Attribution

## Things to consider

### DIFF
--- a/docs/src/recipes.md
+++ b/docs/src/recipes.md
@@ -435,14 +435,6 @@ This can be put anywhere in the code and will appear on the call `?my_plotfunc`.
 
 ### Troubleshooting
 
-It can sometimes be helpful when debugging recipes to see the order of dispatch inside the `apply_recipe` calls.  Turn on debugging info with:
-
-```julia
-RecipesBase.debug()
-```
-
-You can also pass a `Bool` to the `debug` method to turn it on/off.
-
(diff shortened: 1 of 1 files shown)


penelopeysm:
PS: The pull request template links to `zenodo.json` on an old commit of the repository. That should probably also be fixed. [I am listed](https://github.com/penelopeysm/Plots.jl/blob/fb25611402e6363c4049ec1999abf432e4af4d4e/.zenodo.json#L781-L783) on the v2 branch :)

I don't mind lumping that into this PR but I'm not sure if the v2 branch is the appropriate version to use, or if you'd rather use v1.

BeastyBlacksmith:
> I don't mind lumping that into this PR but I'm not sure if the v2 branch is the appropriate version to use, or if you'd rather use v1.

Ideally PRs against v2 point to the v2 branch and PRs against master to the master version. Don't know if this is possible, otherwise it doesn't really matter to which version it points

penelopeysm:
I don't think it's cleanly possible, but I added both links with some clarifying text.

BeastyBlacksmith:
Thanks 🙂

PR #5192, opened 2025-10-23
Clarify order of bounds in ribbon example

ghost:
## Description

Small modification to the ribbons example, since the original description was confusing.

## Attribution
- [x] I am listed in the appropriate version of `.zenodo.json` for [PRs against v2](https://github.com/JuliaPlots/Plots.jl/blob/v2/.zenodo.json) or [PRs against master](https://github.com/JuliaPlots/Plots.jl/blob/master/.zenodo.json) (see https://github.com/JuliaPlots/Plots.jl/issues/3503)

## Things to consider

### DIFF
--- a/PlotsBase/src/examples.jl
+++ b/PlotsBase/src/examples.jl
@@ -600,7 +600,7 @@ const _examples = PlotExample[
         "Ribbons",
         """
         Ribbons can be added to lines via the `ribbon` keyword;
-        you can pass a tuple of arrays (upper and lower bounds),
+        you can pass a tuple of arrays (lower and upper bounds, in this order),
         a single Array (for symmetric ribbons), a Function, or a number.
         """,
         quote


BeastyBlacksmith:
Note, that this will only show up in the dev Version of the docs until v2 is out. If you'd like it to show up earlier, you'd need to open a second PR against master

ghost:
Well I'm not sure how to do this so v2 it will be XD. BTW, do we have an estimated timeline for the release of v2 ?

t-bltg:
> BTW, do we have an estimated timeline for the release of v2 ?

No, although https://discourse.julialang.org/t/can-we-have-pre-release-versions-in-the-registry-v2/129887 would help so that we could make a pre-release available for the users to test `v2` (which is fully operational atm).

Issue #5213, opened 2026-01-05
Colors for mesh3d in pgfplotsx backend


darnstrom:
## Details
The following code results in a red triangle, even though I pass the color `:blue`:
```julia
using Plots
pgfplotsx()
x = [0, 1, 2, 0]
y = [0, 0, 1, 2]
z = [0, 2, 0, 1]
i = [0, 0, 0, 1]
j = [1, 2, 3, 2]
k = [2, 3, 1, 3]
mesh3d(x, y, z; connections = (i, j, k), color=:blue)
```

It works as expected with the `gr()` and `plotlyjs()`.

I can get the expected behaviour (a blue triangle) by using `extra_kwargs` as follows:
```julia
mesh3d(x, y, z; connections = (i, j, k), extra_kwargs= Dict(:series => Dict("fill" => "blue")))
```
One fix would therefore be to pass `color` to "fill" (and maybe also "draw") in the pgfplotsx backend. I would be happy to contribute, but I don't know where to start.

### Backends

This bug occurs on ( insert `x` below )

Backend      | yes | no  | untested
-------------|-----|-----|---------
gr (default) |     |  x   |
unicodeplots |     |     |
pythonplot   |     |     |
pgfplotsx    |   x  |     |
plotlyjs     |     |   x  |
plotly       |     |     |
gaston       |     |     |

### Versions

Plots.jl version: 1.41.3
Backend  version: 1.6.2
Output of `versioninfo()`:
```
Julia Version 1.12.1
Commit ba1e628ee49 (2025-10-17 13:02 UTC)
Build Info:
  Official https://julialang.org release
Platform Info:
  OS: Linux (x86_64-linux-gnu)
  CPU: 16 × AMD Ryzen 7 8840HS w/ Radeon 780M Graphics
  WORD_SIZE: 64
  LLVM: libLLVM-18.1.7 (ORCJIT, znver3)
  GC: Built with stock GC
```

BeastyBlacksmith:
The mesh3d handling is located here:

https://github.com/JuliaPlots/Plots.jl/blob/2127a0d86af755256609f0d4e4178ffab521a538/src/backends/pgfplotsx.jl#L523

and the canonical attribute to corresponding to `color` is called `seriescolor`, so probably adding a `"fill" => opt[:seriescolor]` there might do the trick.

darnstrom:
Thanks for the pointer @BeastyBlacksmith!
[...]

PR #5214, opened 2026-01-06
Handle colors for mesh3d in pgfplotsx backend

darnstrom:
## Description
This PR adds support for colors in `mesh3d` in the `pgfplotsx` backend.
The example in #5213 now correctly creates a blue triangle (similar to the GR and PlotlyJS backend.)
I have also tested it with several calls to `mesh3d`, and the result is the same as for GR and PlotlyJS.

## Attribution
- [x] I am listed in the appropriate version of `.zenodo.json`

## Things to consider


BeastyBlacksmith:
Great, would you mind opening a PR against v2 also?

There the backend code is located here: https://github.com/JuliaPlots/Plots.jl/blob/a58d0ffdfacc1dc86b8f870e2082727df34f3c63/PlotsBase/ext/PGFPlotsXExt.jl#L723

Issue #4816, opened 2023-08-30
[BUG] Assertion error `total_plotarea_vertical > 0mm`. Plots remains unusable afterwards


yha:
## Details

Sometimes plot display fails with one of the assertions at 
https://github.com/JuliaPlots/Plots.jl/blob/4ad4a04a176ba55e69a1263acee8bd0f3034f09a/src/layouts.jl#L342C1-L343C42
and afterwards it's impossible to display plots in the same session.

This bug is weirdly history-dependent, but I've managed to found this reproducer (note that the plot needs to be displayed twice to trigger the bug)

```
julia> using Plots

julia> function badplot()
           plt1 = heatmap(rand(4,4); title = "title")
           plt2 = plot(fill(NaN,4), fill(NaN,4))
           annotate!([NaN, NaN], [NaN, NaN], Plots.text.(["a", "b"], 12))

           plot(plt1, plt2)
       end
badplot (generic function with 1 method)

julia> badplot()

julia> badplot()
Error showing value of type Plots.Plot{Plots.GRBackend}:
ERROR: AssertionError: total_plotarea_vertical > 0mm
Stacktrace:
  [1] update_child_bboxes!(layout::Plots.GridLayout, minimum_perimeter::Vector{Measures.AbsoluteLength})
    @ Plots C:\Users\sternlab\.julia\packages\Plots\3BCH5\src\layouts.jl:343
  [2] update_child_bboxes!(layout::Plots.GridLayout)
    @ Plots C:\Users\sternlab\.julia\packages\Plots\3BCH5\src\layouts.jl:315
  [3] prepare_output(plt::Plots.Plot{Plots.GRBackend})
    @ Plots C:\Users\sternlab\.julia\packages\Plots\3BCH5\src\plot.jl:251
  [4] display(#unused#::Plots.PlotsDisplay, plt::Plots.Plot{Plots.GRBackend})
    @ Plots C:\Users\sternlab\.julia\packages\Plots\3BCH5\src\output.jl:168
  [5] display(x::Any)
    @ Base.Multimedia .\multimedia.jl:340
  [6] #invokelatest#2
    @ .\essentials.jl:816 [inlined]
  [7] invokelatest
    @ .\essentials.jl:813 [inlined]
  [8] print_response(errio::IO, response::Any, show_value::Bool, have_color::Bool, specialdisplay::Union{Nothing, AbstractDisplay})
[...]

PR #5736, opened 2026-04-24
Fix GR text-state corruption causing persistent layout assertion failures

AJ0070:
Fixes a GR backend state-corruption bug behind issue #4816 where plotting with non-finite text coordinates/extents could poison subsequent plots in the same session and trigger layout assertions (`total_plotarea_vertical > 0mm` / `total_plotarea_horizontal > 0mm`).

### Root cause
When GR was asked to draw or measure text at invalid coordinates (for example, `annotate!` with `NaN` positions, or labels/ticks after invalid log-scale inputs), GR could return non-finite text extents (`inqtext` / `inqtextext`). Those values propagated into subplot padding and layout calculations, causing assertion failures. Because GR state persisted, later unrelated plots also failed until restart.

### Fix
In `GRExt.jl`:
- `gr_text(...)`: skip text draw when `(x, y)` is non-finite.
- `gr_inqtext(...)`: sanitize non-finite returned text bounds and fall back to finite zero extents.

This prevents non-finite values from entering layout padding math while preserving normal behavior for valid inputs.

### Testing
Added regression test in `test_misc.jl` covering:
- Non-finite annotation reproducer (including repeated call and history dependence)
- Invalid negative log-scale case followed by a valid labeled log-scale plot

Strengthened test path to renderer-level checks (`show(..., MIME("image/png"), ...)`) so GR drawing and measuring is exercised directly.

Additional GR stress verification in a clean environment (repeated bad → good render cycles, including `twinx`) completed without reintroducing assertion failures or session poisoning.

### Attribution
- [ ] I am listed in the appropriate version of `.zenodo.json` for PRs against v2 or PRs against master (see https://github.com/JuliaPlots/Plots.jl/issues/3503)

### Things to consider

### DIFF
--- a/PlotsBase/ext/GRExt.jl
+++ b/PlotsBase/ext/GRExt.jl
@@ -444,20 +444,30 @@ end
 
 gr_inqtext(x, y, s) = gr_inqtext(x, y, string(s))
 gr_inqtext(x, y, s::AbstractString) =
-if (occursin('\\', s) || occursin(r"10\^{|2\^{|e\^{", s)) &&
-        match(r".*\$[^\$]+?\$.*", String(s)) ≡ nothing
-    GR.inqtextext(x, y, s)
-else
-    GR.inqtext(x, y, s)
+begin
+    coords =
+        if (occursin('\\', s) || occursin(r"10\^{|2\^{|e\^{", s)) &&
+                match(r".*\$[^\$]+?\$.*", String(s)) ≡ nothing
+            GR.inqtextext(x, y, s)
+        else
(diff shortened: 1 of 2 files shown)


AJ0070:
The code block mentioned in the issue description works fine locally but there is failure on CI. I am trying to figure it out. It would be great if someone could help out.

BeastyBlacksmith:
Thank you for looking into this. The remaining failures seem to be due to PlotlyJS failing to precompile on windows and thus unrelated.

AJ0070:
Got it, thanks!

PR #5746, opened 2026-05-12
Add GR colorbar border width

hunterbastian:
## Summary

- add a `colorbar_borderwidth` subplot attribute for GR colorbars
- wire the attribute into GR colorbar axis drawing instead of always using width `1`
- add aliases/docs plus coverage for keyword acceptance and GR SVG rendering

Refs #3174. This is limited to the remaining GR colorbar border-thickness slice.

## Tests

- `git diff --check`
- not run locally: Julia is not installed in this Codex environment, so execution coverage should come from CI


BeastyBlacksmith:

PR #5747, opened 2026-05-12
Add GR colorbar border color

hunterbastian:
## Summary
- add a `colorbar_bordercolor` subplot attribute with `cbordercolor` aliases
- resolve `:match` through `foreground_color_subplot` and normalize explicit color values with the other subplot colors
- wire the GR colorbar border/tick-axis line to the new attribute and cover it in alias/output tests

Refs #3174

## Tests
- `git diff --check`
- Not run locally: Julia is not installed in this environment, so the Julia test suite is left to CI.


BeastyBlacksmith:

PR #5750, opened 2026-05-17
Add GR colorbar width and height attributes

nexiumbiz-debug:
## Summary
- add GR support for `colorbar_width` and `colorbar_height` subplot attributes
- keep `:auto` as the default so existing GR colorbars keep their current dimensions
- add aliases, argument docs, and a GR SVG smoke test for the new attributes

Refs #3174.

## Tests
- `git diff --check`
- Not run: Julia is not installed on this Windows environment (`julia` is not on PATH)


nexiumbiz-debug:
CI is fully green now: codespell, runic, the full Julia LTS/current/pre matrix on Ubuntu/Windows/macOS, and both Codecov patch/project checks all passed. This is still scoped to the GR colorbar width/height attributes plus aliases, docs, and the SVG smoke test.

BeastyBlacksmith:

Issue #3887, opened 2021-10-19
[BUG] aspect_ratio = :equal introduces undesired white space between subplots and titles


stevenaleung:
## Details
Setting `aspect_ratio = :equal` introduces undesired white space between subplots and titles. The white space can be manually removed using `top_margin`, but this can be cumbersome when making many plots

Example:
```
random_values = randn(10,10);

h1 = heatmap(random_values)
heatmap!(title="Plot 1", xlabel="X", ylabel="Y")

h2 = heatmap(random_values)
heatmap!(title="Plot 2", xlabel="X", ylabel="Y")

p = plot(h1, h2, layout=(1,2))
```
<img src="https://user-images.githubusercontent.com/4917419/137826663-70f28f72-81d8-4d95-a0a6-133129077479.png" height=50% width=50%>

```
plot!(aspect_ratio=:equal, xlims=(0.5,10.5), ylims=(0.5,10.5))
```
<img src="https://user-images.githubusercontent.com/4917419/137826950-7be3b2d8-41ef-4533-8a26-6f8d6aeed3bc.png" height=50% width=50%>

In both cases, the value of `p[1][:top_margin]` is the same (1.0mm)

I can set a negative `top_margin` to reduce the white space, but needing to manually determine a new `top_margin` for every new plot I make quickly becomes unwieldy (from my experience so far, I haven't found a `top_margin` value that works for all plots)

The desired behavior would be to have the title text shift such that it remains just above the subplot

<img src="https://user-images.githubusercontent.com/4917419/137826976-b9b57f65-e1aa-4a14-a1c9-14470588fc12.png" height=50% width=50%>

### Backends

This bug occurs on ( insert `x` below )

Backend      | yes | no  | untested
-------------|-----|-----|---------
gr (default) | x    |     |
pyplot       |     |     | x
plotlyjs     |     |     | x
pgfplotsx    |     |     | x
unicodeplots |     |     | x
inspectdr    |     |     | x
gaston       |     |     | x

### Versions

Plots.jl version: v1.13.2
Backend  version (`]st -m <backend(s)>`): GR v0.57.4
Output of `versioninfo()`:
```
Julia Version 1.6.0
Commit f9720dc2eb (2021-03-24 12:55 UTC)
Platform Info:
  OS: macOS (x86_64-apple-darwin19.6.0)
  CPU: Intel(R) Core(TM) i7-9750H CPU @ 2.60GHz
  WORD_SIZE: 64
[...]

PR #5741, opened 2026-05-04
Add GR backend support for colorbar_ticks

AJ0070:
This enables users to customize tick positions and labels on GR colorbars (both vertical and horizontal).

#### Core Implementation
Updated `gr_draw_colorbar()` in `GRExt.jl` to:

- Call `get_colorbar_ticks(sp)` to obtain optimized tick positions and labels
- Render tick marks manually using GR polyline operations
- Render tick labels with proper font formatting (respects `colorbar_tickfont*` attributes)
- Support log-scale colorbars via `colorbar_scale`
- Handle both vertical and horizontal colorbar orientations

#### Tests
Added 3 test cases to `test_backends.jl`:

- Vertical colorbar with tick positions only: `colorbar_ticks=[0.2, 0.5, 0.8]`
- Vertical colorbar with tick positions and labels: `colorbar_ticks=([0.2, 0.5, 0.8], ["low", "mid", "high"])`
- Horizontal colorbar with ticks: `colorbar=:top`, `colorbar_ticks=[0.3, 0.6, 0.9]`

#### Changelog
Updated `NEWS.md` with feature description.

### Attribution

- [ ] I am listed in the appropriate version of `.zenodo.json` for PRs against v2 or PRs against master (see [#3503](https://github.com/JuliaPlots/Plots.jl/issues/3503))

### Checklist

- [x] Does it work on log scales?
- [x] Does it work in layouts?
- [x] Does it work in recipes?
- [x] Does it work with multiple series in one call?
- [x] PR includes or updates tests?
- [x] PR includes or updates documentation?

closes #3560

PR #5756, opened 2026-05-29
Control gap between title and plot

harishjayaraj10:
## Description
When `aspect_ratio=:equal` is used, the plot viewport (`vp_plt`) is shrunk centered inside the subplot bounding box (`vp_sp`). The title was anchored to `vp_sp.ymax` with `valign=:top`, which caused three issues:     

1. **Floating gap** — the title stayed at the image top while the plot moved down, creating unwanted whitespace (issue #3887). 
2. **Title flush with border** — `top_margin` had no visible effect because the title was already at the edge.  
3. **No gap control** — users had no way to set the distance between the plot and the title independently of `top_margin`.

This change:                                                                                                                                                                                                                                                  
   - Anchors the title to the **plot viewport** (`vp_plt.ymax`) instead of the outer bbox, with `valign=:bottom`, so the title tracks the plot area.                                                                                                                                                                                       
   - Introduces a new `title_gap` attribute (default `1mm`) that controls the space between the plot top edge and the title bottom.                                                                                                                                                                                                       
   - Makes `top_margin` control the space **above** the title naturally.                                                                                                                                                                                         
   - Adds `title_gap` to the generic `title_padding()` so other backends can benefit from the same attribute.                   

## Attribution
- [ ] I am listed in the appropriate version of `.zenodo.json` for [PRs against v2](https://github.com/JuliaPlots/Plots.jl/blob/v2/.zenodo.json)    

## Things to consider
- [X] Does it work on log scales? *(Title positioning is independent of axis scale)*   
- [X] Does it work in layouts? *(Tested with `layout=(1,2)` — see #3887 reproducer)*    
- [X] Does it work in recipes? *(Title attributes flow through recipes unchanged)*     
- [X] Does it work with multiple series in one call? *(Title is per-subplot, unaffected by series count)*  


harishjayaraj10:
The reference image tests are expected to fail because this PR changes title positioning. The saved baseline images were generated with the old vp_sp.ymax anchor behavior. Please run the reference image update workflow when reviewing.

BeastyBlacksmith:
You can just run the tests and add the updated reference images in the assets folder

harishjayaraj10:
@BeastyBlacksmith Done — regenerated the GR reference images into the assets folder.

Note: temporarily skipped example 67 ("Heatmap colorbar ticks"). GR 0.73.25 changed `axis()` to take a `String` instead of a `Char` (it was `axis(::Char)` in 0.73.24), so the colorbar code's `GR.axis('X')`/`GR.axis('Y')` now errors with `no method matching axis(::Char; ...)`. #5741 passed because it ran against 0.73.24. Skipped to keep this PR scoped to the title change.

BeastyBlacksmith:
>  temporarily skipped example 67

would you mind changing the calls to `GR.axis` to accomondate for that and bump the GR compat entry in Project.toml?

assets/PlotsBase/gr/0.1.0/ref063.png
BeastyBlacksmith, line 1: can we avoid that the title collides with the ticks here? Ideally it would be above the ticklabel.

harishjayaraj10: Sure. I will update this.

harishjayaraj10:
> would you mind changing the calls to `GR.axis` to accomondate for that and bump the GR compat entry in Project.toml?

I will raise a followup PR for this so the change is cleaner for future reference. Today or tomorrow. Would that be okay?

BeastyBlacksmith:
Sure

assets/PlotsBase/gr/0.1.0/ref063.png
harishjayaraj10: The new (method and) image ref063.png is cleaner and avoids the overlap. Kindly check.

PlotsBase/src/Commons/attrs.jl
```diff
@@ -428,6 +428,7 @@ const _subplot_defaults = KW(
     :titlefontvalign => :vcenter,
     :titlefontrotation => 0.0,
     :titlefontcolor => :match,
+    :title_gap => 1mm,
```
BeastyBlacksmith, line 431: please also add documentation for that

harishjayaraj10: Done.

BeastyBlacksmith:
Nice work, thank you !

PR #5755, opened 2026-05-25
Add colorbar styling controls across GR, Plotly, PGFPlotsX, and PythonPlot

partyplatter08-lab:
/claim #3174

## Summary
- add colorbar tick line, border, width, and height controls across GR, Plotly/PlotlyJS, PGFPlotsX, and PythonPlot
- wire colorbar tick/title font settings and explicit colorbar ticks into Plotly and PythonPlot colorbar rendering
- emit PGFPlotsX colorbar style options for tick lines, borders, and dimensions
- update the Heatmap colorbar ticks example and GR reference image, with backend tests covering the generated options

## Validation
- `git diff --check origin/v2...HEAD`
- Runic formatting check across all touched Julia files
- targeted GR reference comparison for example 67: `Difference: 0.0`
- `JULIA_PKGEVAL=true CI=true PLOTSBASE_TEST_PACKAGES=GR,UnicodePlots,PythonPlot GKS_WSTYPE=nul MPLBACKEND=agg julia --project=PlotsBase/test PlotsBase/test/runtests.jl`
- `CI=true GKS_WSTYPE=nul MPLBACKEND=agg julia --project=PlotsBase/test PlotsBase/test/test_plotly.jl`
- `CI=true GKS_WSTYPE=nul MPLBACKEND=agg julia --project=PlotsBase/test -e 'import PGFPlotsX; include("PlotsBase/test/test_pgfplotsx.jl")'`
- rendered styled colorbar smoke outputs locally for GR and PythonPlot

Prepared with Codex assistance.

PR #5771, opened 2026-07-29
Add colorbar border and size attributes

AJ0070:
Closes the remaining items of #3174 (border color, border thickness, bar width, bar height).

### New subplot attributes

| attribute | default | meaning |
|---|---|---|
| `colorbar_border_color` | `:match` | frame color, setting it alone draws the frame |
| `colorbar_border_width` | `:auto` | frame thickness, `0` disables |
| `colorbar_width` | `:auto` | bar thickness, as a fraction of the plot area width |
| `colorbar_height` | `:auto` | bar length, as a fraction of the plot area height |

`:auto` keeps each backend's own default, so nothing renders differently unless asked.
Resolution is shared through `Colorbars.colorbar_border(sp)`.

Supported by `gr`, `plotly`/`plotlyjs`, `pgfplotsx` and `pythonplot`.
The `plotly`/`plotlyjs` colorbar was previously unstyled and now also honors `colorbar_ticks`
and the colorbar title / tick fonts.

### Also fixed

`GR.drawaxis` lost its axis-spec argument in GR 0.73.25, which made `colorbar_ticks` throw
(reference example 67 errored). The axis object carries its own `X`/`Y` spec, so the call is
now `GR.drawaxis(axis)`.

## Attribution

## Things to consider
- [x] Does it work on log scales? Tested with `colorbar_scale = :log10`
- [x] Does it work in layouts? The new example is a 2x2 layout, also covered by tests
- [x] Does it work in recipes? Tested through a `@userplot` setting the attributes
- [x] Does it work with multiple series in one call? Tested with several `marker_z` series sharing one colorbar
- [x] PR includes or updates tests? Backend tests for `gr`/`pythonplot`, `plotly` and `pgfplotsx`; GR reference images 67/67 pass; default rendering verified byte-identical before/after on all `gr` and `pythonplot` examples
- [x] PR includes or updates documentation? `arg_desc` entries (attribute tables and `plotattr`), new example 68 "Colorbar border and size" with one panel per attribute, and a `NEWS.md` entry


AJ0070:
@BeastyBlacksmith

AJ0070:
@BeastyBlacksmith gentle ping!

AJ0070:
@BeastyBlacksmith

BeastyBlacksmith:
You might want to check out the script here (haven't run it on this PR): https://github.com/JuliaPlots/Plots.jl/pull/5755#pullrequestreview-4410258115

AJ0070:
Thanks, that was useful. Ran it on all four backends and fixed what it turned up.

`colorbar_tickcolor`, `colorbar_ticklinewidth` and a magic `colorbar_tickfont` didn't exist yet, and `colorbar_bordercolor` / `colorbar_borderlinewidth` weren't recognised spellings, so most of the styling in the script was silently doing nothing. All of it works now. One thing worth knowing: an unknown attribute with a Symbol value isn't ignored, it lands in `extra_kwargs` and gets written straight into the tex, which is why panel 1 didn't render at all on pgfplotsx.

On the three you pointed out:

- `colorbar_formatter` works on GR now. Same gap was in plotly and pgfplotsx, fixed there too, and pgfplotsx was dropping custom tick labels entirely.
- log scale colours match GR on plotly, pgfplotsx and pythonplot.
- pythonplot ticks point outward, and tick label rotation is applied.

Example 68 in the gallery covers the attributes and the reference images pass. Happy to post the script's output per backend if you want to see it.

PlotsBase/src/examples.jl
```diff
@@ -1284,19 +1284,21 @@ const _examples = PlotExample[
...
                 cbar_heatmap(title = "border color", colorbar_border_color = :red),
                 cbar_heatmap(title = "border width", colorbar_border_width = 3),
+                cbar_heatmap(title = "tick color", colorbar_tickcolor = :red),
+                cbar_heatmap(title = "tick width", colorbar_ticklinewidth = 3),
                 cbar_heatmap(title = "bar width", colorbar_width = 0.2),
                 cbar_heatmap(title = "bar height", colorbar_height = 0.5),
-                layout = (2, 2),
-                size = (900, 700),
+                layout = (2, 3),
+                size = (1_000, 700),
```
BeastyBlacksmith:
```suggestion
                size = (1_200, 700),
````
should look better that way

**Comment by BeastyBlacksmith** [comment]
> Happy to post the script's output per backend if you want to see it.

That would be helpful

File: PlotsBase/src/examples.jl (hunk 1 above)
Comment (AJ0070) [reply PlotsBase/src/examples.jl]: Done, and it does look better. Reference image regenerated.

**Comment by AJ0070** [comment]
Here is your script on the current branch, one section per backend. Same three plots each: the styled pair, `colorbar_scale = :log10`, and `colorbar_formatter = x -> "test"`.

<details open><summary><b>gr</b></summary>





</details>

<details><summary><b>pythonplot</b></summary>





</details>

<details><summary><b>pgfplotsx</b></summary>





</details>

<details><summary><b>plotly</b></summary>





</details>

<details><summary><b>plotlyjs</b></summary>





</details>

The log scale plots are the interesting comparison: all five now compress the dark end the same way, where before only gr did.

**Comment by AJ0070** [comment]
@BeastyBlacksmith the only red job is `julia lts - windows-latest`, and it's the StatsPlots `interaction` test rather than anything colorbar related. Knockout downloads `knockout.js` at build time and that file is missing from the cached depot, so WebIO throws `Invalid dependency (must be a url, file, or AssetRegistry path)`.

The version split explains itself: on julia 1 the `Pkg.instantiate(; workspace = true)` branch pulls the test-only deps in and their build runs, but on lts (1.10) they only arrive via `Pkg.test`, and Pkg won't rebuild a package that's already in the depot. So it stays broken until the cache is refreshed instead of clearing itself. Same test passes on `julia 1 - windows-latest` in this run.

Want me to send a separate PR for it? Refreshing the cache key is the one line fix, or making lts fetch the test deps the way 1.12 does would address the cause. Happy to leave it alone too.

**Comment by BeastyBlacksmith** [comment]
would be great if could add that fix to this PR

**Comment by AJ0070** [comment]
@BeastyBlacksmith Pushed it as a47d0b94.

`Knockout` and `WebIO` download their JS at build time instead of shipping it, and `Pkg.instantiate` only builds packages it has just downloaded, so once a cached depot carries them without those files nothing repairs it. Forcing the build from a throwaway environment in the dev step fixes it regardless of what the cache restored:

```julia
Pkg.activate(; temp = true)
Pkg.add(["Knockout", "WebIO"])
Pkg.build(["Knockout", "WebIO"])
```

Checked locally by deleting `assets/knockout.js` from the depot: `instantiate` leaves it missing, the forced build brings it back.

AJ0070:
@BeastyBlacksmith the Knockout failure is gone, but both Windows jobs now hit a different one: `display(pl)` in the PlotlyJS testset throws `ECONNREFUSED` from Blink, which spawns electron and then can't reach its control port within its retry budget.

Looks like a pre-existing flake rather than anything from this PR. The same error hit #5776's branch at 09:25 UTC today, roughly six hours before the CI commit here existed and on a branch that doesn't have it, and only 3 of ~50 Windows jobs across the last 25 runs show it. Both Windows jobs on this branch also passed that same testset at 9282a6d3, with electron starting fine.

Could you re-run the two Windows jobs? Being honest about the one thing that nags: both jobs with the CI commit failed and both without it passed, though I can't see any mechanism connecting a `Pkg.build` in a temp env to electron's startup. If it comes back on a re-run I'd wrap the `display(pl)` at test_backends.jl:53 in a try on Windows, since it sits outside any `@test` and asserts nothing.

AJ0070:
The re-run got past electron on lts, but `julia 1 - windows` has now failed three times in that same `PlotlyJS` testset, each time for a different reason at `display(pl)`:

- `IOError: connect: connection refused (ECONNREFUSED)`, electron not coming up inside Blink's retry budget
- `cannot bind to port; may already be in use or access denied`, Blink's Mux server losing the race for its random port
- once it didn't even get that far, CondaPkg couldn't take its lock at `PlotsBase\test\.CondaPkg\lock` and PythonPlot failed to precompile (the lts job logged the same lock waits and got through, so that one is a race too)

Pushed f894954a, which makes that `display` warn instead of error on windows only. Nothing is asserted on its result, so the two real assertions in the testset still run, and linux and macos keep exercising the real path. Verified locally by forcing the guarded branch: it catches the same `ECONNREFUSED` and the testset passes 2/2.

Happy to drop it if you'd rather keep windows strict, but the same failure has shown up on branches without any of my changes, so it looks like runner flakiness rather than something worth chasing here.

AJ0070:
@BeastyBlacksmith is that approach okay with you, or would you rather I take it out?

If you'd prefer not to special case windows, the alternatives are dropping `display(pl)` from that testset altogether, since it asserts nothing either way, or leaving windows strict and just re-running when it trips. Any of the three is fine by me, it's your call which noise you'd rather live with.

BeastyBlacksmith:
Well. The point of the display is to make sure that the display does not error. So I am against hiding it, but its also unrelated to this PR, thus we can just reset the branch.
I am also curios, whether the temporary environment is really necessary or if its enough to just force the build by calling `Pkg.build(["Knockout", "WebIO"])`.

AJ0070:
Fair enough, reset the branch back to badd160b so the `display` is strict again.

On the temp environment: I tried plain `Pkg.build(["Knockout", "WebIO"])` first and it errors, because those packages are never in that environment.

```
Pkg.Types.PkgError("The following package names could not be resolved:
 * Knockout (not found in project or manifest)
 * WebIO (not found in project or manifest)")
```

`Knockout` comes in through `Interact`, which is a `StatsPlots` test dep, and `StatsPlots` isn't a member of the root `[workspace]`, so it never lands in that manifest on any julia version. It only exists inside the env `Pkg.test("StatsPlots")` builds. On lts even `WebIO` is missing, since plain `instantiate()` doesn't fetch test-only deps there either.

With them present as indirect deps the plain call does work and restores the asset, so the `Pkg.add` is only there to make them resolvable without adding test-only deps to the root manifest. Happy to write it differently if you have a preference.

AJ0070:
@BeastyBlacksmith just ping me if you need any assistance

BeastyBlacksmith:
I thought there is no need for v2 to support 1.10, since probably at the time of release its unlikely to still be lts, so I changed it to `min`, but that seems to try to pull even older julia versions.
Probably, there is some julia compat missing.

AJ0070:
Yes, and it's the root `Project.toml`. `setup-julia` resolves `min` from the project at the repository root, and that one only has `[workspace]`, so there was no bound to work from and it went for the oldest julia it knows:

```
##[error]Encountered error: Could not find x86_64/0.1.2 binaries
```

Your `4f9f3654` set `julia = "1.12"` on all seven package projects, just not on the root. Pushed 7c4c95f2 which adds it there too.

Also: adding `StatsPlots` to the workspace is the piece I was missing, that's why `Interact` and `Knockout` never showed up in that manifest. `Pkg.build(["Interact"])` reaches them fine, Pkg builds dependencies transitively.

One thing to be aware of: with `min` at 1.12 and `1` also on 1.12.x, those two rows test the same version until 1.13 is out.

PlotsBase/ext/GRExt.jl
```diff
@@ -670,10 +685,37 @@ end
 # in case someone wants to modify these hardcoded factors
 const gr_cbar_width = Ref(0.03)
 const gr_cbar_offsets = Ref((0.02, 0.07))
+# ndc space kept on the right of the colorbar for its tick labels and title
```
BeastyBlacksmith: What means `ndc`?

PlotsBase/ext/GRExt.jl
```diff
@@ -105,6 +105,21 @@ const _gr_attrs = PlotsBase.merge_with_base_supported(
...
+        :colorbar_formatter,
+        :colorbar_tickcolor,
+        :colorbar_ticklinewidth,
+        :colorbar_tickfont,
+        :colorbar_tickfontfamily,
+        :colorbar_tickfontsize,
+        :colorbar_tickfonthalign,
+        :colorbar_tickfontvalign,
+        :colorbar_tickfontrotation,
+        :colorbar_tickfontcolor,
```
BeastyBlacksmith, line 118: I'd like the canonical keywords to be all underscores, e.g. `colorbar_tick_font_color` and then add one underscore aliases

BeastyBlacksmith (approved):
A few minor comments, but in general I am on board with this

PlotsBase/ext/GRExt.jl
AJ0070: Normalized device coordinates, GR's `0..1` square over the canvas. Reworded both comments to say it in full.

PlotsBase/ext/GRExt.jl
AJ0070: Done in 04bc1464 for the ones this PR adds: `colorbar_tick_color` and `colorbar_tick_line_width` are now the canonical names, with `colorbar_tickcolor` / `colorbar_ticklinewidth` kept as aliases, plus `colorbar_tick_font` for the magic `colorbar_tickfont`. `colorbar_border_color`, `colorbar_border_width`, `colorbar_width` and `colorbar_height` were already fully underscored.

I left the pre-existing font families alone: `colorbar_tickfontcolor`, `colorbar_titlefontsize` and friends would become `colorbar_tick_font_color`, `colorbar_title_font_size` and so on. That's doable, and the magic expansion already supports it (`legend_font` gets the trailing underscore treatment in `process_font_attr!`), but it means renaming the defaults, the `_match_map` entries and every backend's usage, which felt out of scope for a colorbar styling PR. Happy to do it here if you want it in one go, or as a follow up.

BeastyBlacksmith: Nah, that would need to go in a separate PR, but wouldn't mind, if you want to go for that.

AJ0070: Sounds good, I'll do the font families as a follow-up PR once this lands.

AJ0070:
Follow-up with the font family rename is up: #5783

AJ0070:
@BeastyBlacksmith submitted invoice on opencollective

PR #5757, opened 2026-06-05
Harish/update gr version

harishjayaraj10:
## Description
- Update GR.axis calls to accept String instead of Char to match the `GR 0.73.25` API. 
- Bump GR compat entry accordingly and re-enable the previously skipped example 67 in reference image tests.  

## Attribution
- [X] I am listed in the appropriate version of `.zenodo.json` for [#5756]

## Things to consider
- [ ] Does it work on log scales? (Not relevant to the changes in this branch)
- [ ] Does it work in layouts? (Not relevant to the changes in this branch)
- [ ] Does it work in recipes? (Not relevant to the changes in this branch)
- [ ] Does it work with multiple series in one call? (Not relevant to the changes in this branch)
- [X] PR includes or updates tests?
- [ ] PR includes or updates documentation?

PR #5772, opened 2026-07-29
Update axis ticks and expand limits for integer values

### LINKED ISSUE #5738: Integer number of ticks not working
## Details

A [previous issue](https://github.com/JuliaPlots/Plots.jl/issues/2155#issuecomment-524849857) showed that one could conveniently specify an integer number of `ticks` like this:
`plot(sin, -π, π, xticks = 5, yticks = 7)`

However, that feature no longer seems to work (at least not on macOS with GR).

### Backends

gr (default) |   x  |     |
(others untested)

### Versions

Plots.jl version:
v1.41.6
GR v0.73.24
Output of `versioninfo()`:
```
Julia Version 1.12.6
Commit 15346901f00 (2026-04-09 19:20 UTC)
Build Info:
  Official https://julialang.org release
Platform Info:
  OS: macOS (arm64-apple-darwin24.0.0)
  CPU: 12 × Apple M3 Pro
  WORD_SIZE: 64
  LLVM: libLLVM-18.1.7 (ORCJIT, apple-m3)
  GC: Built with stock GC
Threads: 6 default, 1 interactive, 6 GC (on 6 virtual cores)
Environment:
[lots of paths...]
```

harishjayaraj10:
## Description

Fixes #5738 

`xticks` = 5 computes 5 ticks but some fall outside the axis limits and get clipped, so fewer show up:

```julia
julia> p = plot(sin, -π, π, xticks = 5);
julia> PlotsBase.get_ticks(p[1], :x)      # ([-4.0, -2.0, 0.0, 2.0, 4.0], ...)
julia> PlotsBase.axis_limits(p[1], :x)    # (-3.33, 3.33) → only 3 drawn
```

`optimize_ticks` returns the wider span it needs (`strict_span` = false) as its 2nd/3rd value & the code dropped it. It was assigned back to `axis[:lims]` until #3346 commented the line out during the colorbar refactor (I believe it is not intentional).

`axis_limits` now expands to that span, and `get_ticks` lays ticks out over the unexpanded limits (`expand_int_ticks = false`), so they always fit. Otherwise the two are mutually dependent & disagree on log scales. Both paths share `optimize_int_ticks` so they can't drift apart again. 

_**Skipped when explicit lims are given, on polar and on discrete axes.**_

Also fixes swapped parameters in `get_ticks(ticks::Int, dvals, cvals, ...)`, which returned tick positions and labels the wrong way round on discrete axes.

Note: an exact count of round ticks can widen limits noticeably (`xticks = 8` on sin gives (-10, 4)). `colorbar_ticks::Int` has the same clipping issue but needs a different fix, since expanding clims would change the colors.

## Attribution
- [x] I am listed in the appropriate version of `.zenodo.json` for [#5756 & #5757]

### DIFF
--- a/PlotsBase/src/Axes.jl
+++ b/PlotsBase/src/Axes.jl
@@ -111,24 +111,17 @@ function get_axis(sp::Subplot, letter::Symbol)
 end
 
 """
-Expand `(amin, amax)` to the span that `ticks = n::Int` needs, so that all `n` requested
-ticks are within the axis limits instead of being clipped by the backend.
-
-`get_ticks` lays out the ticks over the limits returned here, so the expansion is iterated
-until it is a fixed point of `optimize_int_ticks` - the span it returns for its own output.
-Widening is monotonic, hence no data can be hidden and the iteration terminates.
+Expand `(amin, amax)` so that all `n` ticks requested by `ticks = n::Int` fit inside the axis
+limits. Ticks are laid out over the *unexpanded* limits (`get_ticks` passes
+`expand_int_ticks = false`), hence they always fall inside the span returned here.
 """
-function expand_limits_to_ticks(amin, amax, n::Int, scale::Symbol, max_passes = 5)
+function expand_limits_to_ticks(amin, amax, n::Int, scale::Symbol)
     sf, invsf, _ = scale_inverse_scale_func(scale)
     smin, smax = sf(amin), sf(amax)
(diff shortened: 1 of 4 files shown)


harishjayaraj10:
CI failures are the pre-existing PythonPlot/matplotlib libraqm breakage.

harishjayaraj10:
@BeastyBlacksmith A gentle ping for review.

BeastyBlacksmith:
Thanks 🙂

PR #5773, opened 2026-07-29
pin matplotlib below 3.10 to avoid the libraqm/harfbuzz symbol clash

harishjayaraj10:
## Description
Fixes CI on all platforms, which is currently red on v2 before any test runs (julia 1/lts on macOS, Windows and Ubuntu all fail identically).

## Attribution
- [x] I am listed in the appropriate version of `.zenodo.json` for [#5756 & #5757]


AJ0070:
I ran into the same two breakages while working on #5771 and can confirm this is the right fix. I had independently landed the identical `GR.drawaxis(axis)` and `PlotThemes/test/Project.toml` changes before finding this PR. On the `libraqm` side: the conda-forge build is the problem rather than matplotlib itself, since installing matplotlib 3.11.1 with `pip` into a venv and pointing `JULIA_PYTHONCALL_EXE` at it works fine, while the `CondaPkg` env reproduces the crash exactly.

A few data points on the last red job (`julia 1 - windows-latest`), which may help unstick this:

- It fails with `ArgumentError: Package Blink [ad839575-38b3-5650-b840-f874b8c74a25] is required but does not seem to be installed`, while precompiling `PlotlyJS`. `Blink v0.12.9` *is* listed in the same job's `Pkg.status` output, so it resolves but never gets installed. The added `Pkg.instantiate()` does not appear to fetch it on Windows, and `Pkg.instantiate(; workspace = true)` might be what is wanted there.
- It looks Windows/Blink specific rather than something this PR causes. On a branch off current `v2` *without* these changes, `julia 1 - windows-latest` also dies on Blink, but with a different error: `InitError: AssertionError: isfile(f)` inside `Blink.resource`, i.e. Blink installed but a resource file missing. That same job passed on 2026-08-12 and failed on 2026-08-13 with no relevant change in between, so Blink on Windows seems independently flaky right now.
- This run is from 2026-07-30 and `v2` has moved since (#5772, #5774, #5778). A rebase and re-run would give current signal before chasing the Windows job. Note that #5778 landed the same `PlotThemes/test/Project.toml` hunk, so that one can be dropped on rebase.

For reference, `v2` CI today without this PR: linux and macos die in `PythonPlot.__init__` on the `libraqm` symbol before a single test runs, and `julia lts - windows-latest` errors in the GR reference tests on `drawaxis(::Char, ::GR.GRAxis)` (65 passed, 1 errored). Both are fixed here, so this seems worth landing even with the Windows job outstanding.

harishjayaraj10:
@AJ0070 That fixed it. Thanks for the inputs. 

@BeastyBlacksmith A gentle ping...

Issue #3419, opened 2021-04-05
[BUG] aspect_ratio=:equal not works in 3d plot


JinraeKim:
## Details
I don't know how to set `aspect_ratio=:equal` for 3d plot (like `axis_equal` in `MATLAB`).
It seems that only `x` and `y` axes' aspect ratio are equal.

- example code
```julia
using Plots


function test()
    N = 10
    xs = range(0, stop=10, length=N)
    ys = range(0, stop=20, length=N)
    zs = range(0, stop=50, length=N)
    p = plot(; aspect_ratio=1.0)
    plot!(xs, ys, zs)
end
test()
```
- result
<img width="689" alt="image" src="https://user-images.githubusercontent.com/43136096/113574808-cc5e8600-9657-11eb-8843-b937679ae075.png">

- Notes: from [the original post](https://discourse.julialang.org/t/plots-jl-aspect-ratio-1-0-for-3d-plot/58607).

### Backends
This bug occurs on ( insert `x` below )

Backend      | yes | no  | untested
-------------|-----|-----|---------
gr (default) |  x  |     |
pyplot       |     |     | x
plotly       |     |     | x
plotlyjs     |     |     | x
pgfplotsx    |     |     | x
inspectdr    |     |     | x

### Versions

Plots.jl version: v1.11.0
Backend  version (`]st -m`): don't know
Output of `versioninfo()`:
```
Julia Version 1.6.0
Commit f9720dc2eb (2021-03-24 12:55 UTC)
Platform Info:
  OS: Linux (x86_64-pc-linux-gnu)
  CPU: AMD Ryzen 9 5900X 12-Core Processor
  WORD_SIZE: 64
  LIBM: libopenlibm
  LLVM: libLLVM-11.0.1 (ORCJIT, generic)

```

antoniosgeme:
Heads up, this got closed automatically by a `Fixes #3419` line I put in the commit message
for #5774, and that overreached. Sorry about that.

What #5774 actually does is stop `aspect_ratio` from corrupting 3D limits. The guard was
`RecipesPipeline.is3d(:sp)`, passing a symbol, which always resolved to the
`is3d(st) = false` fallback, so the 2D aspect correction was running on 3D subplots and
silently widening x/y. Real bug, but not this one.

`aspect_ratio = :equal` still does nothing useful in 3D, which is what this issue is about.
#4148 is the PR that would actually close it.

[...]

PR #5774, opened 2026-08-03
fix `aspect_ratio` corrupting 3D axis limits

antoniosgeme:
## Description

`axis_limits` passes the symbol `:sp` rather than the subplot:

```julia
!(aspect_ratio ≡ :none || RecipesPipeline.is3d(:sp))
```

`is3d(::Symbol)` resolves to `is3d(Val{:sp})` and hits the `is3d(st) = false` fallback, so
the guard never fires for 3D. The 2D aspect correction then runs on 3D subplots, widening
their x/y limits by a factor computed from the plot area's pixel width and height:

```julia
julia> using PlotsBase; import GR; gr()

julia> x, y = range(0, 10, length = 10), range(0, 1, length = 10);

julia> z = [xi * yi for xi in x, yi in y];

julia> pl = surface(x, y, z, aspect_ratio = :equal); PlotsBase.prepare_output(pl);

julia> PlotsBase.ylims(pl)
(-2.750533194953867, 3.750533194953867)
```

The y data spans 0 to 1. Nothing downstream uses these limits to set 3D box proportions,
so the widening shows up only as wrong ranges and ticks. I think this is what's behind
#3419, where x and y come out equalized against each other and z doesn't.

`is3d(sp::Subplot)` is defined in `Commons`, so `is3d(sp)` looks like what was meant.

This doesn't make `aspect_ratio` do anything useful in 3D, it just stops it doing damage.
#4148 is the PR for the former and this shouldn't conflict with it.

The test needs `prepare_output`, as `plotarea` is empty until the plot is laid out. It
fails without the one character change and passes with it. Full `test_axes.jl` passes
locally, runic clean.

Same typo is on `master` at `Plots/src/axes.jl:758`. Happy to open that one too.

## Attribution
- [x] I am listed in the appropriate version of `.zenodo.json`

Left out to keep the diff small, can add it here if you want.

## Things to consider

Checked each of these against a 3D subplot with `aspect_ratio` in `(:none, :equal, 1)`,
comparing limits to the same plot with no `aspect_ratio` set:

- `zscale = :log10`: limits identical, the log path is upstream of this block
- two 3D subplots in a `(1,2)` layout, which have different `plotarea`s and so used to
  drift apart from each other: both now match their references
- `aspect_ratio --> 1` set from inside a recipe rather than as a kwarg: unchanged
- multi series 3D call: unchanged

Also checked the other direction, that 2D is untouched. `plot(x, y, aspect_ratio = :equal)`
still widens ylims from `(-0.03, 1.03)` to `(-3.021, 4.021)` exactly as before.

No docs change: `aspect_ratio` is documented for the 2D case only, and that behaviour is
unchanged.

### DIFF
--- a/PlotsBase/src/Axes.jl
+++ b/PlotsBase/src/Axes.jl
@@ -163,7 +163,7 @@ function Commons.axis_limits(
             !has_user_lims &&
                 consider_aspect &&
                 letter in (:x, :y) &&
-                !(aspect_ratio ≡ :none || RecipesPipeline.is3d(:sp))
+                !(aspect_ratio ≡ :none || RecipesPipeline.is3d(sp))
         )
         aspect_ratio = aspect_ratio isa Number ? aspect_ratio : 1
         area = PlotsBase.plotarea(sp)
--- a/PlotsBase/test/test_axes.jl
+++ b/PlotsBase/test/test_axes.jl
@@ -141,6 +141,22 @@ end
 @testset "3D Axis" begin
     ql = quiver([1, 2], [2, 1], [3, 4], quiver = ([1, -1], [0, 0], [1, -0.5]), arrow = true)
     @test ql[1][:projection] == "3d"
+
+    @testset "#3419 - `aspect_ratio` must not alter 3D limits" begin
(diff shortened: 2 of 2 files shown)


BeastyBlacksmith:
Thanks 🙂

Issue #4532, opened 2022-11-21
[FR] change `minorticks` semantics


t-bltg:
Change `minorticks` to be the number of ticks between major ticks not the number of intervals.
The keyword `minorticks` is inconsistent with its description.

See https://github.com/JuliaPlots/Plots.jl/pull/4528 and https://github.com/JuliaPlots/Plots.jl/issues/4508.

PR #5783, opened 2026-08-20
Use underscored names for the colorbar font attributes

AJ0070:
## Description

Follow-up to #5771, as discussed there.

The colorbar font families were the only ones still spelled compactly, while every other
font family in Plots uses the fully underscored form. This makes the underscored spelling
canonical and keeps the previous names as aliases, so `plotattr` and the attribute tables
document one name per attribute instead of two competing ones.

| before | now |
|---|---|
| `colorbar_titlefont` | `colorbar_title_font` |
| `colorbar_tickfont` | `colorbar_tick_font` |
| `colorbar_fontfamily` | `colorbar_font_family` |
| `colorbar_<title\|tick>font<family\|size\|halign\|valign\|rotation\|color>` | `colorbar_<title\|tick>_font_<...>` |

`process_font_attr!` now appends the trailing underscore for `colorbar_title_font` and
`colorbar_tick_font` as it already did for `legend_font`, which is what makes the magic
attributes expand into the new names.

No behaviour change: the old spellings, their non-underscored forms and the `cb`/`cbar`/
`colorkey` prefixes all still resolve.

## Attribution

## Things to consider
- [x] Does it work on log scales? Rename only, the existing `colorbar_scale = :log10` tests still pass
- [x] Does it work in layouts? Covered by the existing colorbar testset
- [x] Does it work in recipes? Covered by the existing `@userplot` colorbar test
- [x] Does it work with multiple series in one call? Covered by the existing `marker_z` colorbar test
- [x] PR includes or updates tests? The colorbar testset now checks both spellings, magic and per attribute; GR reference images 68/68 pass
- [x] PR includes or updates documentation? `arg_desc` entries renamed (attribute tables and `plotattr`) and a `NEWS.md` entry


AJ0070:
@BeastyBlacksmith

AJ0070:
@BeastyBlacksmith Are there any other priority issues that you’d like me to take up?

BeastyBlacksmith:
Uh, I got plenty of those 😅 . For starters anything that is on the 2.0 Milestone, but also regarding the canonicalization of keyword arguments: What you've done here should be done to all the current keywords, however this will break all the recipes relying on keyword names, so in https://github.com/JuliaPlots/Plots.jl/pull/4783 I started to introduce syntax for getting an attribute in a way that would not break under an internal renaming, but I never finished it. So that would also be a thing.

Could also have a video chat or something if you'd like to talk about things to do.

AJ0070:
@BeastyBlacksmith plenty to go at then. I'd rather keep it in the issues and PRs if that's alright, easier for me to fit around other things and it leaves a written trail.

I read #4783. The `canonical_key` hook plus a read spelling in recipe bodies looks like the right shape. Two things before I put work into it:

1. It targets `master` and predates the `PlotsBase` split, so it conflicts and the files have all moved. Port to `v2` as a fresh PR, or rebase in place? Also `getattr` is already taken on `v2` as an internal, `getattr(plt::Plot, ::Int = 1)` in `PlotsBase/src/plot.jl`.
2. `$attr` and `<--` are both implemented, with `$` commented as the recommended one. Keep both, or pick one? `$` collides with normal interpolation in a recipe body, which is the part that worries me.

On doing this to the rest of the keywords: the families don't agree with each other today. Legend went to `legend_font_pointsize` / `legend_font_family`, but `_magic_subplot_attrs` exposes `:title_font` while the expansions are still `titlefontsize` / `titlefonthalign`, and annotations have `annotationfontsize` sitting next to `annotationhalign` with no `font` segment. If you give me the target shape for those, I can apply it to the rest.

Agreed the read path has to come first though. A recipe doing `get(plotattributes, :markersize, 8)` would silently get the fallback after a rename rather than throw, which is worse than a plain error.

Meanwhile, is #4532 free? The `minorticks` FIXME is still sitting in `Ticks.jl:45` and it looks self contained.

BeastyBlacksmith:
#4532 is also free and indeed a much smaller adventure. I'd consider anything free, really.

On the other topic:
1. I'd just open a new PR targeting `v2`. AFAIK the current `getattr` is not used anywhere and can just be rebranded.
2. I'd keep both. Initially I only had `<--`, but using that in anything that is not an assignment looks kind of awkward. Like (`xrotation --> <-- yrotation`). What do you mean with normal interpolation? String interpolation? I think, it should be possible to implement, that e.g. `"My $($title)"` works. But I am not married to the syntax here, if you have something better in mind, im open to that.

AJ0070:
@BeastyBlacksmith correction first: plain string interpolation does not collide. `"$var"` parses to `Expr(:string, :var)`, the parser eats the `$`, so `e.head === :$` never sees it. I ran your `process_recipe_body!` over your own RecipesBase test body and `plotattributes[:hello] = "$var"` comes out untouched, so that test is sound.

And your nested idea already works as written, no implementation needed: `"My $($title)"` parses to `(:string, "My ", (:$, :title))`, and your rewriter turns it into `"My $(plotattributes[canonical_key(:title)])"`.

What I should have pointed at is `$` that survives parsing, i.e. inside quoted code. That one is not hypothetical, it is in this repo. `RecipesPipeline/src/group.jl:95` and `:116`, in the recipe that backs every `plot(...; group = g)`:

```julia
:($key) := split_attribute(plt, key, val, groupby.group_indices[i])
```

That is the dynamic-attribute-key idiom, `$key` interpolated at runtime from the loop variable. With the PR the re-walk descends `:(=)` then `:ref` then `:quote`, finds `Expr(:$, :key)` and rewrites it to

```julia
plotattributes[:(plotattributes[canonical_key(:key)])] = split_attribute(...)
```

which throws `MethodError: Cannot convert an object of type Expr to an object of type Symbol`. Every grouped plot. I checked whether keeping the old `e.head ≢ :call` guard saves it, and it does not, that path never crosses a `:call`.

So I think `$` has to be quote-aware: track quoting depth, only rewrite at depth 0. That kills this and the `ex = :(y = $val)` class together, and you keep `f($attr)`.

Three smaller ones from testing the branch:

- `return z` in a recipe body is a `MethodError: no method matching process_recipe_body!(::Symbol)` at expansion time, since the new unconditional recursion runs on `expr.args[i]` after the `:return` branch replaced it with a non-Expr. Same for bare `return` and `return :auto`. `return (a, b)` survives, which is why it would go green: the only in-tree recipe `return` is a tuple.
- On the write path `k = canonical_key(k)` runs before the `isa(k, Symbol)` QuoteNode wrap, so `c --> :red` canonicalises but `:c --> :red`, the spelling your test uses, does not. It also runs at expansion time inside RecipesBase, so it depends on Plots being loaded when the recipe package precompiles. Splicing `$RecipesBase.canonical_key($k)` the way `is_key_supported` does would fix both.
- `$(:markershape)` gives `plotattributes[:(:markershape)]` and a runtime KeyError. Easy to write by accident when `<--` takes a quoted symbol.

On `getattr`: no need to rebrand, `Tuple{Plot,Int}` and `Tuple{Plot,Symbol}` have empty intersection so they dispatch apart, `detect_ambiguities` is clean, and your branch already ships both without touching `plot.jl`. It is used once, `PlotsBase.getattr(pl)` at `test_layouts.jl:112`. Unexported either way, so delete or keep, your call.

Taking #4532 now.

Issue #4900, opened 2024-03-05
[FR] Retrieve Plot Attributes


nickbrady:
It would be great if Plots had ways to retrieve plot attributes, not just set them. For instance it would be useful to setup conditional control protocols depending on what theme or background color is being employed.

i.e.
```julia
display(Plots.theme)
> :dark
```

BeastyBlacksmith:
I startet implementing this in https://github.com/JuliaPlots/Plots.jl/pull/4783, but put it on hold because of CI issues.

PR #5786, opened 2026-08-21
Let recipes read attributes without hard-coding canonical names

AJ0070:
Port of the recipe-facing half of #4783 to `v2`, with the quoting problem from #5785 solved.

## What it adds

`RecipesBase.canonical_key`, an extension hook next to `is_key_supported`, and two spellings inside a `@recipe` body that read an attribute through it:

```julia
lw = $linewidth            # unary
war <-- :markershape       # binary, in assignment position
:zrotation --> $xrotation  # also works on the right of an arrow, and as a call argument
```

PlotsBase implements the hook as `get(_keyAliases, k, k)`. The call is spliced into the generated code instead of being resolved during macro expansion, so it does not depend on which packages happen to be loaded when a recipe package precompiles.

## Quoting

`$` already means interpolation inside quoted code, so the walk carries the quoting state:

- `quoted` is set on entering `Expr(:quote)` (both `:( ... )` and `quote ... end`) and never clears. Inside a quote every `$` is Julia's own, at every nesting level, so there is no depth at which one can be reclaimed.
- `dollars` is cleared inside macro calls other than `@series`, because macros quote implicitly. `@eval f(::$T) = 1` contains no `Expr(:quote)` at all, its `Expr(:$, :T)` sits at structural depth 0, so a plain depth counter does not see it. Making `$` opt-in per macro fails closed: an unknown macro can never have its interpolation corrupted.
- `arrows` is the old `e.head ≢ :call` guard kept as a flag. We now descend into calls so `something($markercolor, :red)` works, while `Dict(1 => 2)` is still left alone.

This is what preserves `RecipesPipeline/src/group.jl:95` and `:116`, the grouping idiom where the key is interpolated at run time:

```julia
:($key) := split_attribute(plt, key, val, groupby.group_indices[i])
```

Rewriting that turns the key into an `Expr`, and every `plot(...; group = g)` dies with `MethodError: Cannot convert an object of type Expr to an object of type Symbol`.

A total fallback method makes non-`Expr` arguments a no-op, which is also what lets the `:return` arm hand back a `Symbol` safely.

## Verification

- New `attribute reads` testset in `RecipesBase/test/runtests.jl`: 23 assertions covering the reads, alias resolution, quoted code left alone (`:($key) :=`, `:(y = $val)`, `quote`, `@eval`, `@btime`, string interpolation, backticks), all five `return` payload shapes, and a computed key erroring at expansion time. Existing testsets unchanged.
- Ran the rewriter over all 156 recipe bodies in the repo (both `@recipe` and PlotsBase's `@attributes`): output identical to today's everywhere, and `Meta.lower` accepts every rewritten form.
- Grouped plots exercised directly against `gr` (`group =` on plot, scatter and histogram, so both `split_attribute` and `groupedvec2mat`).
- GR reference images 68/68 match.

## Known limits

- `$attr` inside a macro other than `@series`, say `@assert $markershape ≡ :auto`, passes through and fails later at `syntax: "$" expression outside quote`. This cannot be turned into a friendlier error, because a surviving unquoted `$` under a macrocall is exactly what the legitimate `@eval` case looks like. The workaround is a temporary above the macro call.
- `$` on the left of an arrow (`$key --> 5`) hits the same thing, since the key is deliberately never walked. Dynamic writes keep the existing `:($key) := v` spelling.

## Not included

`getattr` for `Plot` / `Subplot` / `Axis` / `Series` (the user-facing half of #4783, and #4900) is left for a follow-up, along with deleting the unused `getattr(plt::Plot, ::Int = 1)`.

## Attribution

## Things to consider
- [x] Does it work on log scales? Not applicable, this is macro expansion, no data path
- [x] Does it work in layouts? GR reference images 68/68 cover them
- [x] Does it work in recipes? That is the whole change, and all 156 in-tree recipe bodies expand identically
- [x] Does it work with multiple series in one call? Grouped plots and `@series` blocks tested
- [x] PR includes or updates tests? 23 new assertions in `RecipesBase/test/runtests.jl`
- [x] PR includes or updates documentation? `NEWS.md` entry; the syntax should also get a docs section once the spelling is settled


RecipesBase/Project.toml
```diff
@@ -1,7 +1,7 @@
 name = "RecipesBase"
 uuid = "3cdcf5f2-1ef4-517c-9805-6587b60abb01"
 author = ["Tom Breloff (@tbreloff)"]
-version = "1.3.4"
+version = "1.4.0"
```
BeastyBlacksmith:
```suggestion
version = "2.0"
```
since old recipes will be broken on Plots v2, this prevents that you can use old recipes with Plots v2.

AJ0070: I read this as bumping RecipesBase to 2.0 and moving the pins with it (the suggestion as written would replace the `version =` line itself).

I tried it and the workspace stops resolving: `RecursiveArrayTools` and `SciMLBase` pin `RecipesBase = "1.3.4"`, and they come in through `docs` via `OrdinaryDiffEq`, so CI would be red on something outside the repo.

Wider picture: 496 packages in General depend on RecipesBase, and 4 of them have compat that would admit a 2.0. The most common entry is a bare `"1"`, 244 of those. So the gate isn't really "old recipes can't be used with Plots v2", it's "Plots v2 can't coexist with any of the other 492 until each one merges a compat PR", including SciMLBase, Symbolics, MLJBase, CategoricalArrays, Measurements and TimeZones. Most of them only ever *write* attributes with `-->`, which aliases still handle, so they'd be updating for a break that doesn't affect them.

Worth separating two things: this PR is additive (all 156 recipe bodies in the repo expand identically, and `$attr` in unquoted code used to be a syntax error, so nothing changes meaning). The break you want to gate is the rename itself, which hasn't happened yet.

So my suggestion is to put the 2.0 bump on the PR that does the renames, where the break actually is, and keep this one at 1.4. If you'd rather have the signal out now I'll do it, but it needs compat PRs upstream to land first or CI stays red. Your call, happy either way.

BeastyBlacksmith: > The break you want to gate is the rename itself, which hasn't happened yet.

That is true, but the renames solely happen in Plots and not RecipesBase. My thinking here is that if someone wants to use the new syntax, they need to opt-in. 

Plots v1 can itself declare compat with RecipesBase v2 since it will still use the old names (just need to define the `canonical_keys` function).
The migration path would then be to Release RecipesBase v2 first, make PRs throughout the ecosystem and only release Plots v2 once the majority is on board.

AJ0070: That migration path makes sense, and opt-in is the right framing. Bumped to 2.0 in 7ec898c8, and moved the in-workspace pins (`PlotsBase`, `RecipesPipeline`, `GraphRecipes`, `StatsPlots`) with it.

Heads up that CI on this PR will be red until the ecosystem side happens, since the workspace itself can't resolve. Exactly seven external packages block it, all reachable from `docs`:

| package | pin |
|---|---|
| SciMLBase | `1.3.4` |
| RecursiveArrayTools | `1.3.4` |
| CategoricalArrays | `1.1` |
| IntervalSets | `1` |
| Polynomials | `0.7, 0.8, 1` |
| Rotations | `1` |
| TimeZones | `0.7, 0.8, 1` |

Each needs a one-line `RecipesBase = "1, 2"`. Happy to open those seven, that unblocks this repo specifically and is a decent start on the wider sweep. Say the word and I'll start with SciMLBase and RecursiveArrayTools, since those two are what actually pull the rest in here.

One thing worth confirming before the release: nothing in this PR changes the meaning of existing code, so Plots v1 should be able to widen to `RecipesBase = "1, 2"` without any source change. `canonical_key(key) = key` is the fallback in RecipesBase, so v1 keeps working on identity until it wants alias resolution, and `process_recipe_body!` keeps its signature. If you want I can put that compat PR up against `master` too, so v1 is ready before v2 releases.

BeastyBlacksmith: Well for CI we can split the docs in its own workspace and just not build the documentation until those compat PRs are done

AJ0070: Tried it. Splitting docs out takes it from seven blockers to five, but not to green, because the rest come in through the test environments rather than docs:

- `Rotations` via `ImageTransformations`, `IntervalSets` via `AxisArrays`, `Polynomials` via `ImageBinarization`, all from the `Images` stack in `PlotsBase/test` that the reference image tests use
- `CategoricalArrays` and `TimeZones` via `RData`, i.e. RDatasets in the StatsPlots tests

Every subproject lists its `test` env under `[workspace]`, so those share the root manifest and the compat applies.

So those five need `RecipesBase = "1, 2"` regardless of what happens to docs. I can open the five now, and SciMLBase plus RecursiveArrayTools for when docs comes back.

Do you want the docs split in this PR too, or separately once the compat PRs are in?

AJ0070:
The seven compat changes are prepared and ready to open, waiting on a 2.0 tag since the registry only has 1.3.4 today and the maintainers would have nothing to resolve against:

SciMLBase, RecursiveArrayTools, CategoricalArrays, IntervalSets, Polynomials, Rotations, TimeZones. One line each, existing ranges kept, no version bumps (CompatHelper convention, which they all take for compat-only changes).

I checked what each actually reads rather than widening blindly, since the whole point of the major bump is that the author has looked:

- SciMLBase and RecursiveArrayTools: `get(plotattributes, :xscale, :identity)`
- IntervalSets: `:background_color`, `:linewidth`, `:markersize`
- Polynomials: `:xlims`, `:ylims`
- TimeZones: `:xguide`
- Rotations: only `delete!` of its own kwargs (`:origin`, `:boxsize`, `:axissize`)
- CategoricalArrays: no attribute access at all, it is a type recipe

All of those are already canonical names, so the renames would not have touched any of these seven. Worth knowing if it shapes how wide you want the sweep to go: the packages that get forced to update are mostly ones the break would never have reached.

Tag it and I will open all seven.

AJ0070:
@BeastyBlacksmith four things waiting on you when you get a chance:

- A RecipesBase 2.0 tag. #5786 is red until then, and I have the seven compat PRs ready to fire (SciMLBase, RecursiveArrayTools, CategoricalArrays, IntervalSets, Polynomials, Rotations, TimeZones).
- The docs split you suggested: in this PR, or separately once the compat PRs are in? It only gets us from seven blockers to five either way, the rest come in through the test envs.
- The naming shape from #5785: is it `title_font_size` or `title_font_pointsize`, and do annotations get a `font` segment? That decides the rest of the rename.
- Now that #5794 is in, `xlims = :round` and `:symmetric` overlap with the modifiers of the same name. Fold them in, or leave both?

BeastyBlacksmith:
I think in [this discourse post](https://discourse.julialang.org/t/rfc-using-a-major-version-bump-for-syntax-opt-in/139080), I came to the conclusion that using a 2.0 release as an opt-in doesn't provide a large benefit.

So lets keep it as a minor version bump.

> Now that https://github.com/JuliaPlots/Plots.jl/pull/5794 is in, xlims = :round and :symmetric overlap with the modifiers of the same name. Fold them in, or leave both?

I don't get this. Can you rephrase that for me?

> The naming shape from https://github.com/JuliaPlots/Plots.jl/discussions/5785: is it title_font_size or title_font_pointsize, and do annotations get a font segment? That decides the rest of the rename.

I think we can canonicalize to `title_font_size`. I think you already can pass `Font` objects as an annotation, though thats a bit unfortunate as an API because you can't use that in recipes, so it should get font attributes as keywords

AJ0070:
Reverted to the minor bump, so #5786 is 1.4 again and the workspace resolves. That also means the seven compat PRs are unnecessary, which is the better outcome.

Canonicalizing to `title_font_size`, noted. And agreed on annotations taking font attributes as keywords rather than only a `Font` object.

On the modifiers, let me try again. #5794 gave `limits_modifiers` three modifiers: `widen`, `round`, `symmetric`. But `:round` and `:symmetric` were already accepted as *limits*, and still are:

```julia
plot(1:5, xlims = :round)                        # old spelling, still works
plot(1:5, xlimits_modifiers = (round = true,))   # new spelling, same result
```

So there are two ways to say the same thing. They are also not quite equivalent: `xlims = :symmetric` runs before the degenerate span fixup, the modifier runs after, so `plot([3,3], ylims = :symmetric)` and `ylimits_modifiers = (symmetric = true,)` disagree.

Do you want `xlims` to go back to being only limits (a tuple or `:auto`), with `:round` and `:symmetric` deprecated in favour of the modifiers? Or leave both spellings?

BeastyBlacksmith:
>  so plot([3,3], ylims = :symmetric) and ylimits_modifiers = (symmetric = true,) disagree

What is the difference?

AJ0070:
`y = [3, 3]`:

```
ylims = :symmetric                        (-3.18, 3.18)
ylimits_modifiers = (symmetric = true,)   (-4.0,  4.0)
```

`lims = :symmetric` runs before the `amax ≤ amin` fixup, the modifier after, so it symmetrizes `(3, 4)`. Otherwise they agree, an explicit chain just turns off the `:auto` widen.

BeastyBlacksmith:
only accept the modifier

PR #5787, opened 2026-08-22
Make `minorticks` count ticks rather than intervals

### LINKED ISSUE #4532: [FR] change `minorticks` semantics

AJ0070:
Closes #4532.

`minorticks = n` drew `n - 1` ticks, because the number was used as the count of intervals a major gap gets cut into. #4528 hit this and corrected the docstring to match the code rather than the other way round, leaving `# FIXME in 2.0` markers in `Ticks.jl` and `test_axes.jl`. This does it the other way round.

The whole conversion lives in `num_minor_intervals`, so `n` ticks becomes `n + 1` intervals and every consumer (GR, PGFPlotsX, PythonPlot, Gaston, and the `get_minor_ticks` path) follows without changes. `0` still means no minor ticks, and `:auto` / `true` are unchanged, so nothing renders differently unless an explicit count was passed. Log axes still get their 9 intervals per decade on `:auto`.

Two things did carry explicit counts and were adjusted so they look the same as before: the `dao` theme (5 to 4) and `ggplot2` (2 to 1). Example 40 passes `minorticks = 2` and now draws two ticks per gap instead of one, so `ref040.png` is regenerated; that one is a real visual change and it is the point of the fix.

The testset in `test_axes.jl` collapses to the identity, which is the readable version of what the FIXME was complaining about.

## Attribution

## Things to consider
- [x] Does it work on log scales? `:auto` still gives 9 intervals per decade, and an explicit count converts the same way
- [x] Does it work in layouts? GR reference images cover them, 68/68
- [x] Does it work in recipes? No recipe surface, the change is in tick generation
- [x] Does it work with multiple series in one call? Ticks are per axis, unaffected
- [x] PR includes or updates tests? `test_axes.jl` rewritten for the new semantics, 264 assertions pass
- [x] PR includes or updates documentation? `arg_desc` back to "Number of minor ticks between major ticks", plus a NEWS entry under breaking changes


t-bltg (approved):
I must admit I forgot the logic behind using intervals, but the more  consistent it gets, the better.

Haven't checked for correctness (I don't have time for this), but I trust you for checking all the corner cases.

Pending CI for merging.

AJ0070:
On log axes `n` is per decade rather than per gap between majors: `n = 1` puts one tick at 5.5, `n = 3` gives 3.25 / 5.5 / 7.75, and `:auto` still gives the 2 through 9. Linear axes give `n` between consecutive majors.

Separately, `get_minor_ticks` returns duplicates on log scales and includes the major positions. `plot([1, 1000], yscale = :log10, minorticks = :auto)` gives the 2..9 run twice, plus 1.0 and 10.0. The `:auto` log path is identical before and after this PR, so it is pre-existing rather than something here. Want an issue for it?

t-bltg:
Can you post a before / after image for log-scales ?

AJ0070:
`yscale = :log10`, minor gridlines on, counting the minor ticks inside the first decade.

before


after


| `yminorticks` | before | after |
|---|---|---|
| `:auto` | 8 | 8 |
| `1` | 0 | 1 |
| `3` | 2 | 3 |
| `8` | 7 | 8 |

`:auto` is untouched, which is why the reference images still match apart from example 40. The explicit counts are the ones that move, and `1` is the clearest case: it used to draw nothing at all.

t-bltg:
> Want an issue for it?

Yes, we should document what is wrong and what needs to be fixed aside this PR.

AJ0070:
Opened #5789 with the reproducer and the two causes in `get_minor_ticks`. Short version: `sub` is computed from the first two majors and then applied to the phantom-padded pairs, which only span one decade, so those over-run into the next pair; and for `e > 1` the range starts exactly on the decade boundary, which is a major tick. Only bites when majors are more than one decade apart, which is why `:auto` on a normal decade axis looks right.

PR #5788, opened 2026-08-22
Accept a single position in `hline` and `vline`

### LINKED ISSUE #2129: support scalars for hline! and vline!

AJ0070:
Closes #2129.

`hline(0.73)` errored with "No user recipe defined for Float64". An integer was worse than an error: it fell through to `@recipe f(n::Integer)` in RecipesPipeline, which adds `n` empty series, so `vline(100)` drew a hundred lines (reported on the thread in Nov 2024).

This is the scoped version AshtonSBradley suggested on the issue: `Real` methods on the four shorthands only, so a lone position wraps itself in a vector before it can reach the generic single-argument recipes. `plot(n)` is untouched and still adds `n` empty series, so the Lorenz example in the docs that daschw flagged keeps working.

One thing to rule on before merging. This changes what an integer means for these two functions specifically: `vline(3)` used to add three empty series and now draws one line at `x = 3`. I think that is the bug rather than a feature, and it is what the last comment on the issue reports, but it is a behaviour change and it is your call. If you would rather keep integers out of it, dispatching on `AbstractFloat` instead of `Real` fixes the reported case and leaves `vline(3)` alone, at the cost of `hline(1)` and `hline(1.0)` doing different things.

The broader change yha and AshtonSBradley discussed, dropping the integer to empty series recipe entirely, is not here: that one is genuinely breaking and needs its own decision.

## Attribution

## Things to consider
- [x] Does it work on log scales? The position is passed through unchanged, the recipes are the same ones a vector goes through
- [x] Does it work in layouts? `hline!(plt, v)` with an explicit plot has its own method and is tested
- [x] Does it work in recipes? No recipe surface, these are shorthand methods
- [x] Does it work with multiple series in one call? Vectors are unchanged, a scalar is just a one element vector
- [x] PR includes or updates tests? New testset in `test_shorthands.jl` checking scalar and one element vector agree for `Float64`, `Int`, `Rational` and `Float32`, both bang forms, and that `plot(3)` still gives three series
- [x] PR includes or updates documentation? Docstrings for `hline` and `vline` mention the scalar form, plus a NEWS entry

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -17,6 +17,8 @@
   The previous spellings (`colorbar_titlefont`, `colorbar_tickfontsize`, `colorbar_fontfamily`, ...) still work when setting attributes
 
 ### Fixed
+- `hline`, `vline` and their `!` forms accept a single position, so `hline(0.73)` no longer errors and
+  `vline(100)` draws one line instead of a hundred empty series (#2129)
 - GR backend: `colorbar_ticks` no longer errors with `GR ≥ 0.73.25` (`GR.drawaxis` signature change)
 - GR backend: `colorbar_formatter` is now applied (it was ignored unless `colorbar_ticks` was given explicitly)
 - PGFPlotsX backend: the colorbar now shows custom tick labels, from either `colorbar_ticks` pairs or `colorbar_formatter`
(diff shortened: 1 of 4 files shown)


t-bltg (approved):
Thanks for fixing this ! Pending CI for merging.

PR #5791, opened 2026-08-27
Handle a lone position wherever the series type is set

AJ0070:
Follows up https://github.com/JuliaPlots/Plots.jl/pull/4904#discussion_r3853923402: #5788 only fixed the `vline(...)` entry point, so a recipe that sets the series type walked straight past it.

On merged `v2`:

| | `plot(x; seriestype = :vline)` |
|---|---|
| `0.73` | `Cannot convert Float64 to series data for plotting` |
| `3` | three empty series |

Same for a recipe doing `seriestype := :vline; return 0.73`.

`_series_data_vector` already receives `plotattributes`, so it knows the series type. This adds a `takes_positions` trait next to `is3d` and uses it in the two scalar methods, which covers the shorthands, the direct `plot(x; seriestype = ...)` call and the recipe case in one place. The six methods #5788 added to `shorthands.jl` are redundant now and are removed. `hspan` and `vspan` come along for free, which #5788 had not covered.

Unchanged: `plot(3)` still gives three empty series (its series type is `:path`), `plot(0.73)` still errors, and vectors behave exactly as before.

One behaviour change worth pointing at: `plot(3; seriestype = :vline)` draws one line instead of three empty series. That is what `vline(3)` already does after #5788, so this makes the two agree.

## Attribution

## Things to consider
- [x] Does it work on log scales? The position is passed through unchanged, same path a vector takes
- [x] Does it work in layouts? GR reference images 68/68
- [x] Does it work in recipes? That is the case this fixes, and there are tests for both a float and an integer returned from a recipe
- [x] Does it work with multiple series in one call? Vectors are untouched, a scalar is just a one element vector
- [x] PR includes or updates tests? New `takes_positions` testset in `RecipesPipeline/test/runtests.jl` (31 assertions, including that `:path` keeps the empty-series and error behaviour), and `test_shorthands.jl` extended to cover the seriestype path and two recipes, 87 assertions total
- [x] PR includes or updates documentation? NEWS entry updated; the `hline` / `vline` docstrings already mention the scalar form


RecipesPipeline/src/utils.jl
```diff
@@ -156,6 +156,20 @@ end
...
+number is a one element series rather than a request for that many empty series.
+"""
+takes_positions(st) = false
+for st in (:hline, :vline, :hspan, :vspan)
+    @eval takes_positions(::Type{Val{Symbol($(string(st)))}}) = true
+end
+takes_positions(st::Symbol) = takes_positions(Val{st})
+takes_positions(plotattributes::AbstractDict) =
+    takes_positions(get(plotattributes, :seriestype, :path))
+
```
BeastyBlacksmith: lets just set the default here and move the Plots specific handling of series to PlotsBase

AJ0070: Done in 5d080c35. Trait default and forwarders stay in RecipesPipeline, the four series types are declared in `recipes.jl` next to their recipes, same as `is_surface` for `:hexbin`.

PR #5794, opened 2026-09-01
Replace `widen` with a composable `limits_modifiers` axis attribute

### LINKED ISSUE #3556: [FR] Rework axislimits modifiers
Actually what I really would like is a new axis attribute lets say `limits_modifiers` where I can mix and match like
```julia
julia> plot(1:5) |> Plots.xlims # still widen by default
(0.88, 5.12)

julia> plot(1:5, xlimits_modifiers = :round) |> Plots.xlims # in this case same as `limits_modifiers = false/nothing/:none`
(1.0, 5.0)

julia> plot(1:5, xlimits_modifiers = (:widen, :round) |> Plots.xlims
(0.0, 6.0)

julia> plot(1:5, xlimits_modifiers = (:symmetric, :widen) |> Plots.xlims
(-5.12, 5.12)
```
But I'm not sure we can do that non-breaking.

_Originally posted by @BeastyBlacksmith in https://github.com/JuliaPlots/Plots.jl/issues/3339#issuecomment-856134778_

AJ0070:
Closes #3556.

`limits_modifiers` replaces `widen`. It takes a named tuple of modifiers applied left to right, so the setting rides on the modifier and order is whatever you wrote:

```julia
plot(1:5)                                                       # (0.88, 5.12), the default `:auto`
plot(1:5, xlimits_modifiers = (widen = false,))                 # (1, 5)
plot(1:5, xlimits_modifiers = (widen = true, round = true))     # (0, 6)
plot(1:5, xlimits_modifiers = (symmetric = true, widen = true)) # (-5.3, 5.3)
```

`:widen`, `:round` and `:symmetric` are the modifiers; a bare name is shorthand for switching one on, so `:round` and `(:widen, :round)` work too. `:none` applies nothing. The default `:auto` behaves as `widen = :auto` did, so nothing renders differently unless you ask for it. `widen` is removed rather than deprecated, per the discussion; master keeps it.

Two things to flag:

- Your last example in the issue annotates `(symmetric, widen)` as `(-5.12, 5.12)`, but left to right that is `(-5.3, 5.3)`; `(-5.12, 5.12)` is widen then symmetric. Both orders are tested side by side.
- `xlims = :round` and `xlims = :symmetric` are untouched. They are not quite the same as the modifiers of the same name (`:symmetric` as a limit runs before the degenerate span fixup), so folding them in is worth its own PR.

## Attribution

## Things to consider
- [x] Does it work on log scales? `:widen` and `:round` go through the existing scale aware `scale_lims` and `round_limits`
- [x] Does it work in layouts? Reference images 68/68
- [x] Does it work in recipes? `spy`, `bar` and the shape recipes were ported
- [x] Does it work with multiple series in one call? The `:auto` heuristic still scans every series on the axis
- [x] PR includes or updates tests? New `limits modifiers` testset in `test_axes.jl`, plus the existing `widen` call sites ported across four test files
- [x] PR includes or updates documentation? `arg_desc` entry for `limits_modifiers`, `lims` updated to point at it, NEWS under breaking changes

### DIFF
--- a/PlotsBase/src/Axes.jl
+++ b/PlotsBase/src/Axes.jl
@@ -246,19 +246,15 @@ function auto_limits_modifiers(axis::Axis)
 end
 
 """
-    limits_modifiers(axis)
-    limits_modifiers(axis, spec)
+    normalize_limits_modifiers(letter, spec)
 
-Resolve `limits_modifiers` into the named tuple of modifiers `axis_limits` folds over, left to
-right. Resolving an already resolved chain returns it unchanged.
+Validate `spec` and put it in the canonical form: a named tuple of the modifiers that are
+switched on, in the order they should be applied, or `:auto`. Warns about, and drops,
+anything invalid. `preprocess_attributes!` runs this once so `axis_limits` does not have to.
 """
-limits_modifiers(axis::Axis) = limits_modifiers(axis, axis[:limits_modifiers])
-
 # `(symmetric = true, widen = 1.2)`: named tuples keep their order, so the chain reads left
-# to right, and a modifier that takes a setting carries it as its value. Resolving drops the
-# invalid and the switched off, so resolving a resolved chain is a no-op.
-function limits_modifiers(axis::Axis, spec::NamedTuple)
-    letter = axis[:letter]
+# to right, and a modifier that takes a setting carries it as its value
+function normalize_limits_modifiers(letter, spec::NamedTuple)
     keep = filter(keys(spec)) do m
         valid_limits_modifier(letter, m) &&
             valid_limits_setting(letter, m, spec[m]) &&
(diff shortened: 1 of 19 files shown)


BeastyBlacksmith:
A few points:
 - `plot(1:5)                                          # (0.88, 5.12)` that is already a widened xlimit. The tight limits would be (1, 5).
 - Its probably better to use a named tuple or dictionary as input. Such that `plot(1:5, xlimits_modifier = (widen = false,))` would give you the unwidened result
 - The deprecation of the keywords should go on master and on v2 we can just remove it

AJ0070:
All three done in 4cf42d22.

Named tuple it is, and it is better: the setting rides on the modifier, so `(widen = 1.2,)` replaces the `:widen => 1.2` pair I had, and `(widen = false,)` switches one off inside a chain. Named tuples keep their order, so left to right falls out. It also drops the parenthesization footgun I flagged, since there is no longer a `=>` to collide with `-->`.

```julia
plot(1:5, xlimits_modifiers = (widen = false,))                 # (1, 5)
plot(1:5, xlimits_modifiers = (widen = true, round = true))     # (0, 6)
plot(1:5, xlimits_modifiers = (symmetric = true, widen = 1.2))  # (-6, 6)
```

A bare name still works as shorthand for switching one on, so `:round` and `(:widen, :round)` are accepted. Say if you would rather only have the named tuple.

`widen` is now simply gone from v2, no shim. I have not touched master; tell me if you want the deprecation there as a separate PR.

On the first point: yes, `(0.88, 5.12)` is the widened result. That line was showing the default is unchanged, tight would be `(1, 5)`. Reworded in the description.

PlotsBase/src/Axes.jl
```diff
@@ -216,25 +222,79 @@ function Commons.axis_limits(
...
+
+Resolve `limits_modifiers` into the tuple of modifiers `axis_limits` folds over, left to
+right. Resolving an already resolved tuple returns it unchanged.
+"""
+limits_modifiers(axis::Axis) = limits_modifiers(axis, axis[:limits_modifiers])
+
+# `(symmetric = true, widen = 1.2)`: named tuples keep their order, so the chain reads left
+# to right, and a modifier that takes a setting carries it as its value
+limits_modifiers(axis::Axis, spec::NamedTuple) =
+    limits_modifiers(axis, Tuple(k => v for (k, v) in pairs(spec)))
```
BeastyBlacksmith: what is the reason to convert named tuples to tuples of pairs and not just use them like they are?

AJ0070: No good reason, it was so filtering out the invalid and the switched off entries was a one liner, and so the `(:widen, :round)` shorthand could share the path. Keeping the named tuple is nicer and it turns out just as short, since `NamedTuple{keep}(spec)` takes a runtime tuple of keys:

```julia
function limits_modifiers(axis::Axis, spec::NamedTuple)
    letter = axis[:letter]
    keep = filter(keys(spec)) do m
        valid_limits_modifier(letter, m) &&
            valid_limits_setting(letter, m, spec[m]) &&
            spec[m] ≢ false
    end
    return NamedTuple{keep}(spec)
end
```

Switched in 66c5965a. The resolved chain is now a named tuple too, so the third positional argument of `axis_limits` takes exactly what the attribute takes: `axis_limits(sp, :x, (widen = false,))`. Shorthand goes the other way now, `(:widen, :round)` builds the named tuple.

Same numbers everywhere, and `axis_limits` is 768 bytes and about 6.5 us per call, which is where it was.

PlotsBase/src/Commons/aliases.jl
```diff
@@ -234,6 +234,7 @@ add_aliases(:fillalpha, :fa, :falpha, :fα, :fillopacity, :fopacity)
 # axes attributes
 add_axes_aliases(:guide, :label, :lab, :l; generic = false)
 add_axes_aliases(:lims, :lim, :limit, :limits, :range)
+add_axes_aliases(:limits_modifiers, :limits_modifier, :lims_modifiers, :lims_modifier)
```
BeastyBlacksmith:
```suggestion
add_axes_aliases(:limits_modifiers, :limits_modifier, :lims_modifiers, :lims_modifier, :lim_mod, :lims_mod)
```

AJ0070: Added, all six resolve including the lettered forms.

BeastyBlacksmith (approved):
Apart from making the formatter happy this looks good to me

AJ0070:
The runic failure is not from this PR: Runic 1.10 reindents generators, and it flags five files here that this branch does not touch. Fixed separately in #5795, this goes green once that lands.

AJ0070:
@BeastyBlacksmith this should be good to go!

AJ0070:
@BeastyBlacksmith this should be ready to merge!

PlotsBase/src/Axes.jl
```diff
@@ -216,25 +222,78 @@ function Commons.axis_limits(
...
+limits_modifiers(axis::Axis) = limits_modifiers(axis, axis[:limits_modifiers])
+
+# `(symmetric = true, widen = 1.2)`: named tuples keep their order, so the chain reads left
+# to right, and a modifier that takes a setting carries it as its value. Resolving drops the
+# invalid and the switched off, so resolving a resolved chain is a no-op.
+function limits_modifiers(axis::Axis, spec::NamedTuple)
+    letter = axis[:letter]
+    keep = filter(keys(spec)) do m
+        valid_limits_modifier(letter, m) &&
+            valid_limits_setting(letter, m, spec[m]) &&
```
BeastyBlacksmith, line 260: One last thing, the validation logic should be moved to the preprocessing in `preprocess_attributes!`, so it only needs to happen once

AJ0070: Moved to `preprocess_attributes!` in c287b1dd, `axis_limits` now only resolves `:auto`. Also rebased on the runic fix. About 4x faster per call.

PR #5795, opened 2026-09-03
Reformat for Runic 1.10

AJ0070:
Runic 1.10 indents generator expressions differently, so the `format` job fails on v2 and on every PR against it. This is `runic --inplace` on the five affected files, nothing else: `git diff -w` is empty.


AJ0070:
@BeastyBlacksmith runic is passing!

PR #5796, opened 2026-09-09
Fix push! and append! on plot series

### LINKED ISSUE #4825: [FR] `append!` cannot accept collections (as would be expected from the name)

AJ0070:
Closes #4825.

Three things were wrong with `push!` / `append!` on a plot, all in the same few lines:

```julia
pl = plot(1:3);   append!(pl.series_list[1], [4,5,6])   # silently did nothing
pl = plot([1.0]); push!(pl, 4)                          # BoundsError, the 4 was read as a series index
pl = plot([1.0]); append!(pl, 1, [4,5,6])               # MethodError
```

`extend_series_data!` never stored the result back, so extending a range (which is immutable, so it is extended by `vcat` rather than in place) discarded the new points. The indexed methods matched with zero data arguments, so `push!(pl, 4)` meant "push nothing to series 4"; they now take at least one datum. And `append!` splatted a vector into `extend_series!`, so a three element vector was read as x, y, z; it now takes a collection per coordinate.

Existing behaviour is unchanged: the tests only used floats, which sidestepped the integer case.


AJ0070:
Aqua caught a real one: the `Real...` and `AbstractVector...` methods both matched `append!(plt)` with no arguments. Both take a datum now, and `test_ambiguities` passes.

PlotsBase/src/Plots.jl
```diff
@@ -82,10 +82,22 @@ Base.iterate(plt::Plot) = iterate(plt.subplots)
 # -------------------------------------------------------
 # push/append for one series
 
+# the indexed methods take at least one datum, else `push!(plt, 4)` would read the `4` as a
```
BeastyBlacksmith:
```suggestion
# the indexed methods take at least one argument, else `push!(plt, 4)` would read the `4` as a
```
Should be updated in other places accordingly

AJ0070: Applied, and made the `append!` comment two lines down read the same way so the "also" still lines up.

PlotsBase/test/test_utils.jl
```diff
@@ -96,6 +96,41 @@
...
+    @testset "push! / append!" begin
+        # github.com/JuliaPlots/Plots.jl/issues/4825
+        y(pl, i = 1) = pl.series_list[i][:y]
+
+        # a range is immutable, so extending it has to replace it rather than drop the result
+        pl = plot(1:3)
+        append!(pl.series_list[1], [4, 5, 6])
+        @test y(pl) == 1:6
+
+        # an integer datum used to be read as the series index
```
BeastyBlacksmith:
```suggestion
        # an integer argument used to be read as the series index
```

AJ0070: Applied, and used the same wording in the `NEWS.md` entry.

PR #5797, opened 2026-09-10
Take limits only in `lims`, rounding and symmetry are modifiers

AJ0070:
Follows #5794 and https://github.com/JuliaPlots/Plots.jl/pull/5786#issuecomment-5614054743.

`lims` now takes limits only, `:auto` or a two element tuple or vector. `:round` and `:symmetric` moved to `limits_modifiers`, so there is one spelling each:

```julia
plot([1.05, 2.0, 2.95], ylimits_modifiers = (round = true,))   # was ylims = :round
plot(1:5, xlimits_modifiers = (symmetric = true,))             # was xlims = :symmetric
```

Passing the old values warns and falls back to the data limits.

This also fixes the ordering difference: `lims = :symmetric` ran before the degenerate span fixup and the modifier after, so `plot([3,3], ylims = :symmetric)` gave `(-3.18, 3.18)` against the modifier's `(-4, 4)`. There is only one path now.

Note the modifier does not widen unless asked, where `lims = :symmetric` left `:auto` on. `(symmetric = true, widen = true)` is the old behaviour.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -30,6 +30,9 @@
   as with the other backends, and `colorbar_scale` now maps the colors, not just the tick labels
 
 ## Breaking changes
+- `lims` takes limits only: `:auto`, or a two-element tuple or vector. `xlims = :round` and
+  `xlims = :symmetric` are now `xlimits_modifiers = (round = true,)` and `(symmetric = true,)`,
+  which also compose with `:widen` (#3556)
 - The `widen` axis attribute is replaced by `limits_modifiers`, which composes. It takes a named tuple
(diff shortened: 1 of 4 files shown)


BeastyBlacksmith:
It still widens by default, correct?

AJ0070:
Yes, `limits_modifiers = :auto` is still the default and still widens the clipping prone series types, so `plot(1:3)` is unchanged at `(0.94, 3.06)`.

What does not widen is an explicit chain: `xlimits_modifiers = (symmetric = true,)` replaces the auto chain, so `plot(1:5, ...)` gives `(-5.0, 5.0)` where `xlims = :symmetric` gave `(-5.3, 5.3)`. Adding `widen = true` restores it.

PR #5802, opened 2026-09-16
Use underscored names for every font attribute

AJ0070:
## Description

Follows #5783 and https://github.com/JuliaPlots/Plots.jl/pull/5786#issuecomment-5604418047.

Three spellings were canonical at once: `titlefontsize`, `legend_font_pointsize` and, since #5783, `colorbar_title_font_size`. Every font attribute is now `<name>_font_<property>`:

```julia
titlefontsize       -> title_font_size
xtickfontcolor      -> xtick_font_color
guidefontfamily     -> guide_font_family
plot_titlefontsize  -> plot_title_font_size
fontfamily          -> font_family
legend_font_pointsize -> legend_font_size
```

The old names stay as aliases on the input side, reading one back throws as it already did for the legend and colorbar fonts.

Annotations gain the missing `font` segment, so `annotationhalign` is `annotation_font_halign`, and pick up `annotation_font` as a magic attribute. `_annotationfont` already built a `Font` out of all six, only the names said otherwise.

Two things fall out of this:

- `title_font`, `plot_title_font`, `tick_font` and `guide_font` were listed as magic attributes but nothing ever popped them, so `plot(1:3, title_font = 20)` was silently ignored while `titlefont = 20` worked. Both spellings work now.
- `legend_titlefont` was documented but was not an attribute or an alias, so it was ignored too. It is an alias of `legend_title_font` now, and the duplicate `_arg_desc` entry is gone.

The `pointsize` half comes from `Font`, whose field `@add_attributes` splices into the attribute name, so that macro maps `pointsize` to `size`. Both TODOs in `process_font_attr!` that marked the transition are gone with it.

Accessor names are left alone: `titlefont(sp)`, `tickfont(ax)` and `guidefont(ax)` stay compact like `legendfont` and `colorbartickfont` beside them. Plotly's `tickfont` and `titlefont` JSON keys are untouched.

Also fixes `:legend_title_font_valigm` in the GR supported list.

## Attribution

## Things to consider
- [x] Does it work on log scales? Not applicable, attribute names only
- [x] Does it work in layouts? `test_layouts` passes, and the themes still apply their fonts
- [x] Does it work in recipes? `corrplot` and `cornerplot` set `title_font`, `test_recipes` passes
- [x] Does it work with multiple series in one call? Yes, nothing touches the series path
- [x] PR includes or updates tests? New `Fonts/Names` testset in `test_components.jl`
- [x] PR includes or updates documentation? `NEWS.md`, `arg_desc.jl`, `attributes.md` and `layouts.md`


t-bltg (commented):
I'm always in favor of consistency.

t-bltg:
Side question, is it normal that we still have some remaining such as `xtickfontcolor`, `xtickfontfamily` and `xtickfontsize`, (and `ytick...`) ?

AJ0070:
Those are aliases, not attributes. `add_axes_aliases(:tick_font_size, :tickfontsize)` fans the old spelling out over `x`, `y` and `z`, so `xtickfontcolor` resolves to `xtick_font_color` the same way `tickfontsize` resolves to `tick_font_size`, and all six properties are covered for each letter.

No canonical name is spelled compactly any more:

```julia
julia> C = PlotsBase.Commons;

julia> filter(k -> occursin(r"font[a-z]|pointsize", string(k)),
           union(C._all_attrs, C._all_magic_attrs, C._lettered_all_axis_attrs))
Symbol[]
```

In the tree the only places those strings appear are the `NEWS.md` example and the alias tests in `test_components.jl`. If you saw them somewhere else, point me at it and I will take a look.

t-bltg:
Oh ok, sorry I was also confused by a previous local `docs` build which matched a simple grep command.

